Issue #78, opened 2019-08-20
Documentation and support for representation and respect of human identities


Firstyear:
As an identity management software implementation, we have a very specific goal - store and provide access to people's identity information. However, identities are complex. From social and cultural perspectives they are all unique, and deserve respect. It's important to consider that a project like this will be used in places where people may not have a choice to use this. IE you are hired in a business and this is their IDM platform. We must consider our impact on people who are not directly able to choose this platform - but these people still deserve our respect.

In order to help guide our development and documentation, we should consider strategies and methods to guide us to always uphold the rights of people and represent their identities correctly. 

While today I have personally made many steps to achieve this, having a more formal or well defined guide would help support others, support other projects, and to allow discussion of these ideas.

Having this defined helps us as a project because it allows us to bring in a set of social standards to conversations, besides just technical or behavioral. For example. A feature is requested for firstname/lastname schema in the server. Today this is technically valid - they are just attributes. It's okay by the code of conduct as that is limited to interactions on the development resources. This means we need something extra to say "I think there are people we will exclude in this feature".

This issue is intended to host discussion of how we as developers can guide a project to always consider and respect the rights of others.

Firstyear:
I'm changing this issues milestone - not because it's not important, but because I think it will be some time before it's "closed" or rather, finished. I have merged #84 which is an initial skeleton. We now have a starting point, and I want that collaboration and discussion to continue.

Firstyear:
I think that we haven't had movement on this, but our work has shown what we want to achieve. If we find new use cases as we go, we can add them.

PR #84, opened 2019-09-11
Add initial skeleton of the statement of rights

Firstyear:
Implements #78, documentation on rights and ethics for consumers of our software.

- [ - ] cargo fmt has been run
- [ - ] cargo test has been run and passes

### DIFF
--- a/ethics/EXAMPLES.md
+++ b/ethics/EXAMPLES.md
@@ -1,0 +1,48 @@
+
+# Examples of situations for consideration
+
+## Ability to be forgotten
+
+### Deletion is delete not flagging
+
+When an account is deleted it must be truly deleted, not just flagged for future delete. Note
+that some functionality like the recycle bin, we must keep the account details, but a recycle
+bin purge does truly delete the account.
+
+## Self determination and autonomy
+
+### Self name change
+
+People should be able to change their own name at anytime. Consider divorce, leaving abusive partners
+or other personal decisions around why a name change is relevant.
+
+This is why names are self-service writeable at any time.
+
+## Free from harmful discrimination of any kind
+
+### Cultural and Social awareness of name formats
+
+All name fields should be case sensitive utf8 with no max or min length limit. This is
+because names can take many forms such as.
+
+* firstname middlename lastname
+* firstname lastname
+* firstname firstname lastname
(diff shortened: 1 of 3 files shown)


Firstyear:
@excitedleigh @ryankurte It would be great to have you review and potential refactor or expand this. It's a bit of a rough draft, but I think it's the kind of idea I want to go for in this document.

ethics/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+## Rights of people
+
+All people using this software should expect to have the right to:
+
+* Ability to be forgotten
+* Self determination and autonomy
+* Free from harmful discrimination of any kind
+* Informed consent and Privacy of their data
+* To use and access this software regardless of ability
```
ryankurte: how about something like "to access and understand data held and shared on their behalf", intended to support access to their account audit trail with logins / information access / whatever else is created through their interaction with the service, in a supporting GDPR or NZ Privacy Act kinda way?

ethics/EXAMPLES.md
```diff
@@ -0,0 +1,48 @@
...
+People should be able to change their own name at anytime. Consider divorce, leaving abusive partners
+or other personal decisions around why a name change is relevant.
+
+This is why names are self-service writeable at any time.
+
+## Free from harmful discrimination of any kind
+
+### Cultural and Social awareness of name formats
+
+All name fields should be case sensitive utf8 with no max or min length limit. This is
```
ryankurte: naive question, do you _need_ more than one name field (like, is there a benefit to the first name / lastname / whatever arrangement vs. just having one name field)? just that some people that will want it?

and, on the basis that all these fields are customisable anyway (afaik from talking to you about this, might be incorrect), is there a way that you could recommend but also gently influence this (with a default configuration or something)?

Firstyear: Ahhh name fields are displayName/legalName. They have different privacy settings by default in the project. But I should make taht clearer (there is plenty of documentation about why firstname/lastname is not usable).

JJJollyjim: One of my biggest frustrations with software is the name/preferredName structure, which guides people into always using the wrong one. I definitely like your approach.

PR #92, opened 2019-09-18
Run tests and check rustfmt as Github actions

JJJollyjim:
- [x] cargo fmt has been run

### DIFF
--- a/.github/workflows/formatting.yml
+++ b/.github/workflows/formatting.yml
@@ -1,0 +1,12 @@
+name: Check formatting
+
+on: [push]
+
+jobs:
+  rustfmt:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v1
+    - name: Check formatting
+      run: cargo fmt -- --check
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,0 +1,19 @@
+name: Run tests
+
+on: [push]
+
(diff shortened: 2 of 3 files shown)


JJJollyjim:
Ahh, I've just realized Github actions is still in beta, and it looks like you don't have it. It will likely not work on this repo.

JJJollyjim:
Here's what it looks like on my repo, for this commit: https://github.com/JJJollyjim/kanidm/runs/227228813

The runner only has two cores, so build takes around 5m and tests take around 10.

Let me know if you want me to port this to a different CI provider: I just thought I would add something quick, and hadn't used Actions yet

Firstyear:
Honestly, I have no strong feelings on which CI provider is better or worse than any other. It's great to have something added for certain as the project becomes "more than just me". 

The tests should get faster once I add indexing - there was a pretty major perf regression when I added the default IDM entries because we now have schema + acp + groups which is 85 entries, and we do full table scans. So in the future the CI should be quicker.

The main concern would be socket opening in the proto tests, but if they all passed for you then I guess it's all good? I'm happy to merge this, but what changes do I need to make to my account to add github-actions? Or should we wait until is available publicly?

JJJollyjim:
Closing this in favour of a CircleCI configuration, since there is a waitlist on Github Actions, and the release date of November has been taken off the site (so other contributors may not have access for a while)

PR #97, opened 2019-09-19
Add CircleCI configuration

JJJollyjim:
Got the time below 2 minutes (mostly by running the tests with `--release`, plus caching stuff that CircleCI offers which GitHub doesn't)

You'll need to log in to CircleCI, and I think allow the use of external Orbs, which definitely means something not ominous

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,46 @@
+version: 2.1
+orbs:
+  rust: glotrade/rust@0.1.3
+
+
+commands:
+  cargo-install:
+    description: "Install a binary from crates.io, caching the built binary"
+    parameters:
+      crate:
+        description: "The crate to install"
+        type: string
+    steps:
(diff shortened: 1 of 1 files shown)


Firstyear:
I think that adding --release isn't really a good idea with tests, so I'd rather not do this. I know that the time is probably important, but as mentioned, the current performance of tests is due to lack of indexing that I'm (literally) working on at the moment to fix. So I'd rather not use --release in CI.

When you say "I need to login" what do I need to do? Is there some documentation you can point me too? Thanks!

Firstyear:
Hey @JJJollyjim did you want to remove the --release or are you happy for me to take over the branch and finish this PR myself? Thanks again!

Firstyear:
Someone contributed a different method of this, so closing this PR. Thanks for your time @JJJollyjim :)

Issue #20, opened 2019-02-02
Investigate and change from sqlite to sled


Firstyear:
The sled db project plans to add mvcc as an option. We require this style of transactions so if they support mvcc, we should change from sqlite to sled in our idl layer.

Firstyear:
After some discussion with the author of SLED, it may be a long time before SLED supports the transaction functions we need, so this may not be possible for some time.

NiryaAestus:
Currently no CI for Windows either.

Firstyear:
Good to know. I'm currently not in a rush to change, so we'll see if we need to change from sqlite to something else, or if it's possible to support two idl layers. I think it won't be too dificult to support differnt idl providers in the future.

Firstyear:
I think it's not going to be possible as sled transaction model is too-different from what we required today.

PR #98, opened 2019-09-19
WIP: Replace openssl with rustls+ring in the server

JJJollyjim:
Ring is a gradual rewrite of parts of openssl in rust, and rustls is a pure-rust SSL library built on top of it.

Combined with #20, this would remove all native dependencies... almost.

Reqwest is depending on native-tls, so on unix platforms we'd still be using openssl, as a transitive dependency. Reqwest has a feature to depend on rustls, but we run into an issue:

Reqwest depends on rustls 0.16, but actix-web 0.7 depends on rustls 0.13. Because of the assembly code in ring, the versions cannot coexist. This means that for now, the server at least is completely openssl free, but the client (and the Cargo.lock) is not.

To unify the dependencies (and probably for other reasons...) we'd probably need to update to actix-web 1.0, a major breaking change (inexplicably, it no longer uses actix!)

- [x] cargo fmt has been run

### DIFF
--- a/kanidmd/Cargo.toml
+++ b/kanidmd/Cargo.toml
@@ -50,7 +50,6 @@ time = "0.1"
 
 concread = "0.1"
 
-openssl = "0.10"
 rustls = "0.14.0"
 ring = "0.13.5"
 
--- a/kanidmd/Cargo.toml
+++ b/kanidmd/Cargo.toml
@@ -52,6 +52,7 @@ concread = "0.1"
 
 openssl = "0.10"
 rustls = "0.14.0"
+ring = "0.13.5"
 
 rpassword = "0.4"
 num_cpus = "1.10"
--- a/kanidmd/Cargo.toml
+++ b/kanidmd/Cargo.toml
@@ -20,7 +20,7 @@ path = "src/server/main.rs"
 kanidm_proto = { path = "../kanidm_proto" }
 
 actix = "0.7"
-actix-web = { version = "0.7", features = ["ssl"] }
+actix-web = { version = "0.7", features = ["rust-tls"] }
 
 bytes = "0.4"
 log = "0.4"
@@ -51,6 +51,7 @@ time = "0.1"
 concread = "0.1"
 
 openssl = "0.10"
+rustls = "0.14.0"
 
 rpassword = "0.4"
 num_cpus = "1.10"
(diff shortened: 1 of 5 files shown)


Firstyear:
This PR puts me in a difficult place. It was a careful decision that led me to use OpenSSL in the first place. 

I really respect pure rust crypto libs. I really do. But the issue is two fold. The first is certification. If this is to be used by a business (like a major opensource company that I work for), then we need to use openssl or nss. We can not use libraries like ring because they have not gone through the same level of assessment. The second part is that the api offered by ring and others is not complete enough to support webauthn-rs - they don't allow rebuilding of keys from x, y coords that is needed for the ecdsa routines. This means OpenSSL will *always* have to be included because of the webauthn-rs depenedency, so ring is now just another extra lib that we "didn't need" as we already require OpenSSL. 

The update to actix 1.0 is something that I really want to do, but I have delayed due to the really poor documentation and attitude of the actix project, and the generally difficult nature of rust futures (but this has improved). I have been hoping that actix would improve and honestly I haven't looked lately, but there are signs it's getting better. I considered that perhaps rewriting away from actix may also be a future path too, given await is now in rust, so a lower level tokio+hyper interface could be considered. But certainly it needs more thought and design.

So for now, I would "reject" (softly) this PR. For this I would change it to be:

* actix-web 1.0 has to be a seperate PR, and there is plenty of my own mistakes and clean up that needs to happen here too. This has to be the first step before we talk about cryptography libs.
* OpenSSL must be the default choice for people with the project.
* Ring is a feature that is optional choice for the webserver TLS, but OpenSSL will always be required due to Webauthn.

So I hope this helps explain my position. I really do appreciate your time and input on this.

Conan-Kudo:
Please don't use RusTLS/Ring. The author's terrible attitude to people makes it an extremely unpleasant project to interact with, and the author's penchant for yanking old releases to break the world is even more unpleasant to deal with.

NiryaAestus:
From the windows discussion. Supporting OpenSSL seems better. (Also proper dynamic linking support)

I also found that if OpenSSH is enabled Windows already has LibreSSL installed. Which is maintained by the reputable OpenSSH/OpenBSD team. 
If it's an alternate option to LibreSSL it would be good for supporting: Dragonfly BSD, OpenBSD and Void Linux, and apparently Windows.

NiryaAestus:
nvm [openssl](https://docs.rs/openssl/0.10.24/openssl/) create already supports libressl transparently.

Firstyear:
Okay, in this case I'm going to close it. I think it would be valuable to open an actix 1.0 refactor issue though, or at least an investigation about other possible actor / web frameworks we could use for the api. Thanks!

Issue #126, opened 2019-10-30
ssh key distribution features


Firstyear:
Add support for SSH key distribution to clients. Initially this would be uncaching.

PR #146, opened 2019-11-14
126 ssh key features

Firstyear:
Implements #126 ssh key distribution features. This adds an improved SSH key validator, and the related REST endpoints for SSH keys management due to being a non-trivial type for parsing - we could consider a str -> json instead for future to avoid the custom routines perhaps. This also adds a tool for openssh to consume by the AuthorizedKeysCommand in sshd_config, which gets and lists keys as such:

```
kanidm_ssh_authorizedkeys admin -H https://localhost:8080 -D anonymous -C ../insecure/ca.pem
ssh-ed25519 AAAC3NzaC1lZDI1NTE5AAAAIAeGW1P6Pc2rPq0XqbRaDKBcXZUPRklo0L1EyR30CwoP william@amethyst
```

### DIFF
--- a/kanidm_client/src/lib.rs
+++ b/kanidm_client/src/lib.rs
@@ -435,11 +435,13 @@ impl KanidmClient {
         self.perform_post_request(format!("/v1/account/{}/_ssh_pubkeys", id).as_str(), sk)
     }
 
+    /*
     pub fn idm_account_rename_ssh_pubkey(&self, id: &str, oldtag: &str, newtag: &str) -> Result<(), ClientError> {
         self.perform_put_request(format!("/v1/account/{}/_ssh_pubkeys/{}", id, oldtag).as_str(), newtag.to_string())
     }
+    */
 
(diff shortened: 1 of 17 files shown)


kanidm_client/src/lib.rs
```diff
@@ -68,7 +68,10 @@ impl KanidmClient {
     }
 
     fn build_reqwest(ca: &Option<reqwest::Certificate>) -> Result<reqwest::Client, reqwest::Error> {
-        let client_builder = reqwest::Client::builder().cookie_store(true);
+        let client_builder = reqwest::Client::builder()
+            .cookie_store(true)
+            // .danger_accept_invalid_hostnames(true)
+            .danger_accept_invalid_certs(true);
```
Firstyear: Probably should remove this, I think that this was necesitated by a breaking change in reqwest or hyper where the add_root_certificate seems to no longer be respected :(

Issue #88, opened 2019-09-15
acp_enable should be may, and filter on reload should change


Firstyear:
Due to the way migrations work, we enforce certain attribute states, but we *don't* migrate or change attributes not listed in the migration. This poses a cchallenge on default access controls, where someone may want to disable the acp, but the acp_enable boolean is currently must, and we set it to true at migration time. 

We should change acp_enable to may, and then change the reload filter to andnot(acp == false) which would mean that acp not present == true. Then we remove the acp_enable flag from the acp migrations, which would let people alter this value as intended.

PR #152, opened 2019-11-29
acp_enable is optional and non-presence implies true

Firstyear:
Implements #88, acp_enable becomes optional. This is important with regard to migrations. When we migrate, if an attribute is present in the migration we always set the value in the case of a single-value attribute (multivalues interleave). This means an admin who may set acp_enable to false on default shipped access controls would have them forcefully re-activated on each server restart. This change makes acp_enable optional, and removes it from the migration. If *NOT* present on an ACP, it's considered "true", so that when an admin overrides this value to false, we will permanently respect that choice.

### DIFF
--- a/kanidmd/src/lib/constants.rs
+++ b/kanidmd/src/lib/constants.rs
@@ -473,7 +473,6 @@ pub static JSON_IDM_ACP_XX_V1: &'static str = r#"{
         "name": ["idm_acp_xx"],
         "uuid": ["00000000-0000-0000-0000-ffffff0000XX"],
         "description": ["Builtin IDM Control for xx"],
-        "acp_enable": ["true"],
         "acp_receiver": [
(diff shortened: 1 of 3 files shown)


kanidmd/src/lib/constants.rs
```diff
@@ -473,7 +473,6 @@ pub static JSON_IDM_ACP_XX_V1: &'static str = r#"{
         "name": ["idm_acp_xx"],
         "uuid": ["00000000-0000-0000-0000-ffffff0000XX"],
         "description": ["Builtin IDM Control for xx"],
-        "acp_enable": ["true"],
         "acp_receiver": [
```
Perri-Boulton: I think this should just be delted.

Firstyear: Thanks for commenting! In a diff the '-' here means a removed line, and a '+' means a line was added. Similar the reason for the colour coding.

But good work to connect that (indeed) this attribute should be removed from the template :)

PR #166, opened 2020-01-13
Move a bit the code to 2018 edition. Fix most of the clippy errors.

aplanas:
I would like to move the codebase into the 2018 edition.

This PR is:

* Dropping some of the `extern crate`
* Fixing some typos in Cargo.toml
* Fixing most of the clippy errors / warnings

Note that there are still two clippy errors related with complexity. I am not confident to address them with my current rust knowledge, but I will try in a future PR (if you do not fix them before).

- [X] cargo fmt has been run

### DIFF
--- a/kanidmd/src/lib/actors/v1_write.rs
+++ b/kanidmd/src/lib/actors/v1_write.rs
@@ -32,12 +32,12 @@ pub struct CreateMessage {
 
 impl CreateMessage {
     pub fn new(uat: Option<UserAuthToken>, req: CreateRequest) -> Self {
-        CreateMessage { uat: uat, req: req }
+        CreateMessage { uat, req }
     }
 
     pub fn new_entry(uat: Option<UserAuthToken>, req: ProtoEntry) -> Self {
         CreateMessage {
-            uat: uat,
+            uat,
             req: CreateRequest { entries: vec![req] },
         }
     }
@@ -54,7 +54,7 @@ pub struct DeleteMessage {
 
 impl DeleteMessage {
     pub fn new(uat: Option<UserAuthToken>, req: DeleteRequest) -> Self {
-        DeleteMessage { uat: uat, req: req }
+        DeleteMessage { uat, req }
     }
 }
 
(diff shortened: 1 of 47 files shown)


Firstyear:
This is an awesome change @aplanas, I'm reviewing it now. :) Thanks so much!

kanidmd/src/lib/actors/v1_write.rs
```diff
@@ -765,7 +758,7 @@ impl Handler<PurgeTombstoneEvent> for QueryServerWriteV1 {
 
     fn handle(&mut self, msg: PurgeTombstoneEvent, _: &mut Self::Context) -> Self::Result {
```
Firstyear: It looks like you've removed the result here?

kanidmd/src/lib/actors/v1_write.rs
```diff
@@ -786,7 +778,7 @@ impl Handler<PurgeRecycledEvent> for QueryServerWriteV1 {
 
     fn handle(&mut self, msg: PurgeRecycledEvent, _: &mut Self::Context) -> Self::Result {
         let mut audit = AuditScope::new("purge recycled");
-        let res = audit_segment!(&mut audit, || {
+        audit_segment!(&mut audit, || {
```
Firstyear: And here?

Firstyear:
Beside those two comments about the returns this change is awesome - huge amount of work, really happy with it. You've made a lot of good improvements to the apis and efficiency, so I'll be happy to merge it shortly :D Thanks so much!

kanidmd/src/lib/actors/v1_write.rs
aplanas: Good one. Here clippy complains about that we assign to `res` the unit value. `type Result = ()` is defined to this trait, so the only value that can be returned is `()` for `handle` IIUC.

kanidmd/src/lib/actors/v1_write.rs
aplanas: Same. Only `()` can be returned.

Firstyear:
Okay, I'll merge and investigate those two functions separately. Thanks so much!

kanidmd/src/lib/actors/v1_write.rs
Firstyear: Okay I looked and I can see it now, good spot. Thanks!

Issue #129, opened 2019-10-31
Add pam/nsswitch clients


Firstyear:
For linux/unix auth we will require pam/nsswitch clients to be available

Issue #153, opened 2019-11-29
Posix group and account administration lifecycle


Firstyear:
Add tooling for managing the posix account extensions on account and their lifecycle.

PR #179, opened 2020-02-10
129 pam nsswitch stage 1 daemon

Firstyear:
Implements #129, pam and nsswitch daemon capability. This is stage 1, which adds a localhost unix domain socket resolver, a ssh key client, support to the server for generating unix tokens, an async client lib, and client handles for adding posix extensions to accounts and groups.


Firstyear:
It's pretty likely I'll also do #153 in the same issue here.

Firstyear:
Right, done!

Firstyear:
Implements #129, pam and nsswitch daemon capability. This is stage 1, which adds a localhost unix domain socket resolver, a ssh key client, support to the server for generating unix tokens, an async client lib, and client handles for adding posix extensions to accounts and groups.

Issue #108, opened 2019-09-26
Extend testing of default entries


Firstyear:
We have some testing of the default entries, but having more around things like asserting who can do what, a test of the bootstrap process, and even things like ensuring anonymous can't self write would be great. While I trust the server is currently implementing all of this correctly, having more tests of some of these assumptions from the default entries would be great!

pando85:
Hi @Firstyear, I'm working on this and I'm trying to understand all well while I'm learning about Rust.
All comments are going to help me so if you want to review my code and make some comments, I will appreciate it.
I started testing users read/write capabilities and I don't know how to modify self `name` and self `displayname`. [This](https://github.com/pando85/kanidm/blob/fec793d7af246a7bb6dc3d1cc8e0c4ab6f63d3f7/kanidm_client/tests/proto_v1_test.rs#L22) is the commit were I'm already working on.
Thanks and I'm glad to read your code. I like it. Great job.

Firstyear:
Hey there, thanks for looking into this! I'm glad you like the code, and happy to help.

First piece of advice, is that maybe for this test it's a good idea to break it out to another file, and import the "run_test" fn into the new test case.

With the writes, there are a few paths to consider, that create, modify and delete of entries are all seperate in the access control system, and that modifies are able to define present/delete controls, as well as extra controls on which classes may be defined on an entry. So as you probably know, the proto test is using the front end http api, which requires the client library here https://github.com/kanidm/kanidm/blob/master/kanidm_client/src/lib.rs so if there are missing api end points, you may need to add them to the client (but I think I have defined them all already?). 

The first way to do a modify is via the raw modify call here: https://github.com/kanidm/kanidm/blob/master/kanidm_client/src/lib.rs#L494 this takes a filter of what entries to affect, and a list of "modifications" that define states you want present. So an example of changing name would be in pseudo code:

```
rsclient.modify(
    filter!(eq("name", "testaccount")),
    ModifyList::new_list(vec![
        Modify::Purged("name"),
        Modify::Present("name", "testaccount_new_name")
    ])
);
```

Keep in mind the "raw" apis, are really that, raw. They represent the internal DB model of how to alter and affect entries, so that's what it's a little bit unpolished - every rest api end point translates to this model underneath, but the rest end points exist to help define higher level operations.

Anyway, the other option is to look at this api end point:

https://github.com/kanidm/kanidm/blob/master/kanidm_client/src/lib.rs#L612

[...]

PR #213, opened 2020-04-30
Extend testing of default entries

pando85:
Add default entries test to apply behaviours according to
`designs/default_idm_layout.rst`.

Add expected behaviours for:
  - Users
  - Account managers
  - Group managers
  - Admins
  - People Managers
  - Anonymous clients
  - Radius servers

Also, refactor `kanidmd_client` tests to separate into different files
and fix some documentation typos

Resolves: #108

- [x] cargo fmt has been run

### DIFF
--- a/kanidm_client/tests/default_entries.rs
+++ b/kanidm_client/tests/default_entries.rs
@@ -1,0 +1,622 @@
+#![deny(warnings)]
+use std::collections::HashSet;
+
+use kanidm_client::KanidmClient;
+use kanidm_proto::v1::{Filter, Modify, ModifyList};
+
+mod common;
+use crate::common::{run_test, ADMIN_TEST_PASSWORD};
+
+static USER_READABLE_ATTRS: [&str; 9] = [
+    "name",
+    "spn",
+    "displayname",
+    "class",
+    "memberof",
+    "uuid",
+    "gidnumber",
+    "loginshell",
+    "ssh_publickey",
+];
+static SELF_WRITEABLE_ATTRS: [&str; 7] = [
+    "name",
+    "displayname",
+    "legalname",
+    "radius_secret",
+    "primary_credential",
+    "ssh_publickey",
+    "unix_password",
+];
+static DEFAULT_HP_GROUP_NAMES: [&str; 22] = [
+    "idm_admins",
+    "system_admins",
+    "idm_people_manage_priv",
+    "idm_people_account_password_import_priv",
+    "idm_people_extend_priv",
+    "idm_people_write_priv",
+    "idm_people_read_priv",
+    "idm_group_manage_priv",
+    "idm_group_write_priv",
+    "idm_account_manage_priv",
+    "idm_account_write_priv",
+    "idm_account_read_priv",
+    "idm_radius_servers",
+    "idm_hp_account_manage_priv",
+    "idm_hp_account_write_priv",
+    "idm_hp_account_read_priv",
+    "idm_schema_manage_priv",
+    "idm_hp_group_manage_priv",
+    "idm_hp_group_write_priv",
+    "idm_acp_manage_priv",
+    "domain_admins",
+    "idm_high_privilege",
+];
+static DEFAULT_NOT_HP_GROUP_NAMES: [&str; 2] =
+    ["idm_account_unix_extend_priv", "idm_group_unix_extend_priv"];
+
+fn create_user(rsclient: &KanidmClient, id: &str, group_name: &str) -> () {
+    rsclient.idm_account_create(id, "Deeeeemo").unwrap();
+
+    // Create group and add to user to test read attr: member_of
+    match rsclient.idm_group_get(&group_name).unwrap() {
+        Some(_) => (),
+        None => rsclient.idm_group_create(&group_name).unwrap(),
+    }
+
+    rsclient
+        .idm_group_add_members(&group_name, vec![id])
+        .unwrap();
+}
+
+fn is_attr_writable(rsclient: &KanidmClient, id: &str, attr: &str) -> Option<bool> {
+    println!("writing to attribute: {}", attr);
+    match attr {
+        "radius_secret" => Some(
+            rsclient
+                .idm_account_radius_credential_regenerate(id)
+                .is_ok(),
+        ),
+        "primary_credential" => Some(
+            rsclient
+                .idm_account_primary_credential_set_password(id, "dsadjasiodqwjk12asdl")
+                .is_ok(),
+        ),
+        "ssh_publickey" => Some(
+            rsclient
+                .idm_account_post_ssh_pubkey(
+                    id,
+                    "k1",
+                    "ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIAeGW1P6Pc2rPq0XqbRaDKBcXZUPRklo0\
+                L1EyR30CwoP william@amethyst",
+                )
+                .is_ok(),
+        ),
+        "unix_password" => Some(
+            rsclient
+                .idm_account_unix_cred_put(id, "dsadjasiodqwjk12asdl")
+                .is_ok(),
+        ),
+        entry => {
+            let new_value = match entry {
+                "acp_receiver" => "{\"Eq\":[\"memberof\",\"00000000-0000-0000-0000-000000000011\"]}".to_string(),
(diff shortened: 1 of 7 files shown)


pando85:
I have a dude because I notice that I duplicate some documentation present in `designs/default_idm_layout.rst` and put near tests implementations. I don't know if there is a better approach.

Also, any comment is good received, I'm starting in Rust and I would appreciate any advice.

kanidm_client/tests/default_entries.rs
```diff
@@ -0,0 +1,622 @@
+#![deny(warnings)]
+use std::collections::HashSet;
+
+use kanidm_client::KanidmClient;
+use kanidm_proto::v1::{Filter, Modify, ModifyList};
+
+mod common;
+use crate::common::{run_test, ADMIN_TEST_PASSWORD};
+
+static USER_READABLE_ATTRS: [&str; 9] = [
```
Firstyear, line 10: I wonder if these should be in #cfg(test) in constants near the definitions of the access controls so that they are updated in sync?

kanidm_client/tests/default_entries.rs
```diff
@@ -0,0 +1,622 @@
...
+static SELF_WRITEABLE_ATTRS: [&str; 7] = [
+    "name",
+    "displayname",
+    "legalname",
+    "radius_secret",
+    "primary_credential",
+    "ssh_publickey",
+    "unix_password",
+];
+static DEFAULT_HP_GROUP_NAMES: [&str; 22] = [
```
Firstyear, line 30: I think you don't need the size spec (22) here because you have a statically defined array it's implied?

kanidm_client/tests/default_entries.rs
```diff
@@ -0,0 +1,622 @@
...
+    rsclient.idm_account_person_extend(id).unwrap();
+
+    ["ssh_publickey", "legalname", "mail"]
+        .iter()
+        .for_each(|attr| {
+            assert!(is_attr_writable(&rsclient, id, attr).unwrap());
+        });
+
+    // Write radius credentials
+    if id != "anonymous" {
```
Firstyear, line 135: Could it be worth testing that anonymous gets an error in this path since anon should never have a radius cred?

kanidm_client/tests/default_entries.rs
```diff
@@ -0,0 +1,622 @@
...
+
+// read schema entries.
+// TODO: write schema entries
+#[test]
+fn test_default_entries_rbac_admins_schema_entries() {
+    run_test(|rsclient: KanidmClient| {
+        rsclient
+            .auth_simple_password("admin", ADMIN_TEST_PASSWORD)
+            .unwrap();
+        let default_classnames: HashSet<String> = [
```
Firstyear, line 391: You could beild this set from the actual schema rather than hardcoding? But maybe that needs schema apis to be added ....

Firstyear:
Hey there, I've added a few comments, but otherwise this looks really good, great stuff!

kanidm_client/tests/default_entries.rs
pando85: I think I cannot do in other way. Example [here](https://stackoverflow.com/questions/23810032/how-to-specify-const-array-in-global-scope-in-rust).

kanidm_client/tests/default_entries.rs
pando85: This function is used to add all possible attributes to an account, if that account is anonymous we cannot add radius credentials but who it is a helper function to populate accounts with attributes we cannot fail.

kanidm_client/tests/default_entries.rs
pando85: Where? I don't know where you're referring sorry :S

kanidm_client/tests/default_entries.rs
pando85: I don't know how to get all pub static from constants module.

kanidm_client/tests/default_entries.rs
Firstyear: if you move the USER_READABLE_ATTRS to kanidmd/src/lib/constants/, then mark the module as `#[cfg(test)]`, so that way we can have USER_READABLE_ATTRS defined near the access controls?

kanidm_client/tests/default_entries.rs
Firstyear: Ohh wow, okay, ignore my comment then!

kanidm_client/tests/default_entries.rs
Firstyear: My point is more that if the account is anonymous we *expect* to see the failure, so we have the function return a result or similar if we expect the addition to succeed OR fail. This way we are both testing that attributes can be added by who-we-expect but also that they can't be added by accounts we don't expect.

kanidm_client/tests/default_entries.rs
Firstyear: Ahhh that's not what I mean. Schema is loaded into the database as entries, so there is a rest api here for it:

https://github.com/kanidm/kanidm/blob/master/kanidmd/src/lib/core/mod.rs#L1500

So you could have the client call that api, then you can create the "all possible classnames" from the live running schema of the server. Does that make sense?

pando85: I test against that response. See line L418. Am I missing something?

kanidm_client/tests/default_entries.rs
pando85: Ok, but I how could I populate all attributes to `anonymous` without using that function? I will need to refactor in a way that I can't see. Could you make an example please?

kanidm_client/tests/default_entries.rs
Firstyear: OHHHH I see. Okay, I missed this. All good then!

kanidm_client/tests/default_entries.rs
Firstyear: Well, rsclient has the associated auth session. So you could have  rsclient do an auth_anonymous, then you have your test just call create_user_with_all_attrs, and you expect a failure instead, so create_user could return a result, and then your test expects on that instead? 

I'm not saying you populate attributes to anonymous, I'm saying you authenticate AS anonymous and attempt to populat the attributes on an account.

kanidm_client/tests/default_entries.rs
```diff
@@ -0,0 +1,622 @@
...
+}
+
+// Account Managers
+// read and write to accounts, including write credentials but NOT private data (see people manager)
+// ability to lock and unlock accounts, excluding high access members.
+#[test]
+fn test_default_entries_rbac_account_managers() {
+    run_test(|mut rsclient: KanidmClient| {
+        rsclient
+            .auth_simple_password("admin", ADMIN_TEST_PASSWORD)
```
Firstyear, line 268: Gah I just realised. You don't want to do this as admin. You need to bootstrap a test-admin that you add to each priv group one at a time, so that you guarantee you are only testing that one expect groups privileges rather than admin who comes with a set of rights already.

kanidm_client/tests/default_entries.rs
pando85: Line 591. Anonymous tests are at the end.

Firstyear: ARRGHH how did I miss this! Okay ignore me :)

kanidm_client/tests/default_entries.rs
pando85: mmmm I use admin just too bootstrap my test environment. In that case I create a account_manager and a test entry to check all behaviours against them.

kanidm_client/tests/default_entries.rs
pando85: WIP

pando85:
Rebase to pass tests

kanidm_client/tests/default_entries.rs
pando85: This is not clear to me, what could we do here?

Firstyear: No, we can still use the admin for bootstrapping stuff, just the test needs to be checked as account_manager which I see you are doing now, so I think it's okay.

Firstyear:
Okay, I think this is good to go unless you have any other questions?

pando85:
For me is ok too, thanks!

PR #214, opened 2020-04-30
Fix code format and sqlite errors in CI

pando85:
This will fix CI and tests.

To avoid this happen again in the future I will set a branch protection for master and avoid to push commits directly or merge branches without passing tests.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -12,13 +12,11 @@ jobs:
         uses: actions/checkout@v2
 
       - name: Install stable toolchain
-        uses: actions-rs/toolchain@v1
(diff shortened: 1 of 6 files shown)


pando85:
From [https://lib.rs/crates/libsqlite3-sys](https://lib.rs/crates/libsqlite3-sys)
> If you use the bundled feature, libsqlite3-sys will use the cc crate to compile SQLite from source and link against that. This source is embedded in the libsqlite3-sys crate and is currently SQLite 3.30.1 (as of rusqlite 0.21.0 / libsqlite3-sys 0.17.0). This is probably the simplest solution to any build problems.

pando85:
Fix in #215

PR #215, opened 2020-05-01
Fix sqlite fails in CI tests and fmt error

pando85:
Add `libsqlite3-sys` crate to avoid sqlite3 missmatching version.

From [https://lib.rs/crates/libsqlite3-sys](https://lib.rs/crates/libsqlite3-sys)
> If you use the bundled feature, libsqlite3-sys will use the cc crate to compile SQLite from source and link against that. This source is embedded in the libsqlite3-sys crate and is currently SQLite 3.30.1 (as of rusqlite 0.21.0 / libsqlite3-sys 0.17.0). This is probably the simplest solution to any build problems.

Remove sqlite OS packages dependencies for build.

Also, we fix a format error to allow CI tests to pass again.

- [X] cargo fmt has been run

### DIFF
--- a/kanidm_unix_int/Cargo.toml
+++ b/kanidm_unix_int/Cargo.toml
@@ -58,11 +58,15 @@ serde_derive = "1.0"
 serde_cbor = "0.11"
 structopt = { version = "0.3", default-features = false }
 
+libsqlite3-sys = { version = "0.17" }
 rusqlite = { version = "0.21" }
 r2d2 = "0.8"
 r2d2_sqlite = "0.14"
 
 reqwest = { version = "0.10" }
+
+[features]
+default = [ "libsqlite3-sys/bundled" ]
 
 [dev-dependencies]
 kanidm = { path = "../kanidmd", version = "0.1" }
--- a/kanidmd/Cargo.toml
+++ b/kanidmd/Cargo.toml
@@ -51,6 +51,7 @@ serde_cbor = "0.11"
 serde_json = "1.0"
 serde_derive = "1.0"
 
+libsqlite3-sys = { version = "0.17" }
 rusqlite = { version = "0.21", features = ["backup"] }
 r2d2 = "0.8"
 r2d2_sqlite = "0.14"
@@ -71,6 +72,9 @@ idlset = { version = "0.1" , features = ["use_smallvec"] }
 
 zxcvbn = "2.0"
 base64 = "0.12"
+
+[features]
+default = [ "libsqlite3-sys/bundled" ]
 
 [dev-dependencies]
 criterion = "0.3"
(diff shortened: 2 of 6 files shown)


kanidm_unix_int/Cargo.toml
```diff
@@ -58,6 +58,7 @@ serde_derive = "1.0"
 serde_cbor = "0.11"
 structopt = { version = "0.3", default-features = false }
 
+libsqlite3-sys = { version = "0.17", features = ["bundled"] }
```
Firstyear, line 61: I'm a bit worried about this, because some distros that end up using this may not like this. Is there a way for us to have a feature flag for if we use bundled libsqlite?

Firstyear:
Happy with the other changes, but I'm a bit worried about the sqlite one. I'll think about it a bit, because I do want to think about future supportability of the project on distros, and sqlite seems like something we should be able to pull in from the distro. 

Saying this we do already have to vendor many libs, so I think yeah, thought needed from me :) 

Again, great work as always :D

pando85:
Ok, I understand that distros will want to use that from its system, it could be configurable. This is the shortest way to fix the issue and this open the way to [static binaries](https://github.com/kanidm/kanidm/commit/c07fad5e0fed93ef1c4c179a7ac50b198664f2f5) and lighter images that will be nice for kanidmd but could not be interested for kanidm_unix_int.

I think we must go with a configurable version without static link for this package to build against all distros that we want to support and add tests for that distros.

kanidm_unix_int/Cargo.toml
ryankurte: Something like
```
[features]
...
default = [ "libsqlite3-sys/bundled" ]
```

should do the job, then you can opt-out with --no-default-features

Firstyear:
@pando85 What do you think about using the suggestion that @ryankurte made for `default = [ "libsqlite3-sys/bundled" ]` ?

pando85:
@Firstyear great. Included! I think that it is the most versatile solution.

kanidmd/Cargo.toml
```diff
@@ -51,6 +51,7 @@ serde_cbor = "0.11"
 serde_json = "1.0"
 serde_derive = "1.0"
 
+libsqlite3-sys = { version = "0.17", features = ["bundled"] }
```
Firstyear, line 54: I think you forgot to update this one too? Otherwise it looks great!

pando85: Is Kanidmd going to be distributed as package or only as a container? I think that containers are the standard for this kind of services and if is only a container we don't need to support both compilations, but as you prefer.

Firstyear: It may be packaged, and already a package exists for suse, but I haven't really advertised it much. So I think we should be consistent in our configuration here.

pando85: Greats. Here is it.

PR #219, opened 2020-05-06
Create light images with static binaries

pando85:
- Change base builder image to `ekidd/rust-musl-builder` for create static
binaries.
- Add two flavours: alpine and stand alone binary.
- Add makefile commands for build, test and push kanidmd images and
  refactor previous code.
- Add missing things to `.dockerignore`.
- Refactor CI to use Makefile.

Implements # .

- [x] cargo fmt has been run

This is a propose to create safest and lightest images.
We could include a SUSE flavour too if you need. I think that alpine and scratch are enough but I understand your situation.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,18 +1,52 @@
+.PHONY: help build/kanidmd build/radiusd test/kanidmd push/kanidmd push/radiusd vendor-prep doc install-tools prep
 
-docker-kanidmd:
-	docker build -f kanidmd/Dockerfile -t kanidm/server:latest .
+IMAGE_BASE ?= kanidm
+IMAGE_VERSION ?= latest
 
-docker-radiusd:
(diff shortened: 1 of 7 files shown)


pando85:
This builds could be improved with some template to avoid code repetition and we could include more images or just static binaries apart of `kanidmd`.

Makefile
```diff
@@ -1,18 +1,52 @@
...
+push/radiusd:
+	@docker push $(IMAGE_BASE)/radius:$(IMAGE_VERSION)
 
 vendor-prep:
 	cargo vendor
 	tar -czf vendor.tar.gz vendor
 
-doc-local:
+doc: ## build doc local
+doc:
```
Firstyear, line 45: Is this double line a mistake maybe?

pando85: No, test it with `make help`. It is for dependencies and documentation the first line.

Firstyear: Mind blown. Okay, that's awesome.

PR #220, opened 2020-05-06
Merge docker image github actions into unique

pando85:
Use matrix and Makefile to build both images in the same workflow.

Also fix a typo bug with kanidmd image build on CI.


pando85:
Remove static binaries (muslc) and back to SUSE image because performance problems.

PR #221, opened 2020-05-07
216 - add initial /status endpoint

Firstyear:
This is a stub implementation of 216, adding a `/status` endpoint. Importantly this actually contacts a status thread, which means that the main thread pools aren't affected by this. The idea of the thread pool is that later we can have this thread aggregate statistics and information from the other operation threads that can be displayed in the status.


pando85:
Great stuff! It could be interesting that our metrics have [this format](https://openmetrics.io/) to be compatible with modern metrics collection solutions.

PR #223, opened 2020-05-08
Fix format error, fix CI

pando85:


Firstyear:
ONE DAY we will solve this stupid format thing ....

PR #228, opened 2020-05-11
CI update all kanidm component version and tag

### LINKED ISSUE #225: Github auto-tag of releases
According to @pando85 it's possible to have git auto-tag git commits based on changes to Cargo.toml.

It would be good to have this, but also we will need a way to keep the verison of all cargo.toml's in sync.

pando85:
Add make entry to have all cargo versions updated from VERSION file.
Add `tag_version` github action to update all cargo.toml and tag it and
push it.

In future cargo will support managing all dependencies from main
Cargo.toml. Ref: https://github.com/rust-lang/cargo/issues/3931

Resolve #225


VERSION
```diff
@@ -0,0 +1 @@
+0.1.2
```
Firstyear, line 1: Can we make the version 1.0.0-rc10 ? Is that even valid in rust? I know actix-web has 3.0.0-alpha.2, so I'd be happy to tweak it too. This way it lines up with what I've been tagging in git already and the versions in build.opensuse.org.

Firstyear:
Beside the one version thing, looks awesome!

pando85:
Update:
>    Remove cargo fmt check from CI because we have problems with different
    versions and fails continually. Instead, we use cargo-husky to ensure
    that all people run it automatically in its local, but this is not
    obligatory.
    Remove all versions between dependencies in internal packages.
    Update to version 1.0.0-rc10, according to packages published.

kanidmd/Cargo.toml
```diff
@@ -26,7 +26,7 @@ path = "src/server/main.rs"
 
 
 [dependencies]
-kanidm_proto = { path = "../kanidm_proto", version = "0.1" }
+kanidm_proto = { path = "../kanidm_proto" }
```
Firstyear, line 29: The version lines need to stay, else we can't publish to crates.io.

pando85:
Revert:
> Remove all versions between dependencies in internal packages.

PR #231, opened 2020-05-14
In progress logging design

Firstyear:
Work in progress logging design - it'll probably change a bit as I start to implement the feature.

### DIFF
--- a/designs/logging.rst
+++ b/designs/logging.rst
@@ -37,10 +37,10 @@ performance of any operation, and make informed decisions about tuning (such as 
 or threads), and developers should be able to identify code paths that are under pressure
 and could be targets for improvement.
 
-* Containers
+* Containers/Systemd
 
-Logs should be access via an API, and support some querying or extraction that can be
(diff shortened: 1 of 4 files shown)


designs/logging.rst
```diff
@@ -0,0 +1,170 @@
...
+* Performance
+
+Administrators and Developers should be able to analyse fine grained information about the
+performance of any operation, and make informed decisions about tuning (such as caches or
+or threads), and developers should be able to identify code paths that are under pressure
+and could be targets for improvement.
+
+* Containers
+
+Logs should be access via an API, and support some querying or extraction that can be
```
pando85, line 42: I think that usually working with containers is enough to write in stdout/stderr. Admins must ensure to take logs and put in a place to be reviewed and accessible (via API or whatever).

pando85:
Good document.

designs/logging.rst
Firstyear: That's a good point. I think I will make it stdout/stderr only to start with just because it will work with systemd as well that way. So I'll updated this part, and then merge. Thanks!

PR #236, opened 2020-05-20
Phase 2, logging refactor

Firstyear:
Implements 99, the second phase of the logging refactor. This adds tagged log attributes, and improves the format of what we collect so that it's easier to identify and trace specific areas of the server during development and administration.

### DIFF
--- a/kanidmd/src/lib/audit.rs
+++ b/kanidmd/src/lib/audit.rs
@@ -10,15 +10,46 @@ use chrono::DateTime;
 use uuid::adapter::HyphenatedRef;
 use uuid::Uuid;
 
-#[macro_export]
+#[derive(Debug, Serialize, Deserialize)]
+pub enum LogTag {
+    AdminError,
+    AdminWarning,
+    AdminInfo,
+    RequestError,
+    Security,
+    SecurityAccess,
+    Filter,
+    FilterError,
+    FilterWarning,
+    Trace,
+}
+
+impl fmt::Display for LogTag {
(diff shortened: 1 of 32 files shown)


Firstyear:
Also worth noting that due to these log changes, a few tweaks to indexing and query execution is in this PR, and it improves the performance of tests by 25% 😳

kanidmd/src/lib/audit.rs
```diff
@@ -10,31 +10,172 @@ use chrono::DateTime;
...
+    Filter,
+    FilterError,
+    FilterWarning,
+    Trace,
+}
+
+impl fmt::Display for LogTag {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match self {
+            LogTag::AdminError => write!(f, "admin::error 🚨"),
```
pando85, line 30: I dunno if this icons will help or will make some difficulties showing logs and parsing them.

Firstyear: Parsing there are still unique string tags associated, and I found the icons helpful? None of this is set in stone though, but for me when looking at the full log it helped things to standout. I'm attaching two screenshots as an example. Notice how easy it is to spot the access control lines? 

Screen Shot 2020-05-22 at 16 28 53

Screen Shot 2020-05-22 at 16 29 13

pando85: I find them pretty cool, helps with visualization and with that format I think that they are easy to parse. So I love them.

PR #246, opened 2020-06-09
199 ldap gateway

### LINKED ISSUE #199: Add an LDAP gateway
Kanidm uses a http interface, which many legacy applications won't use or support. we could add a read only ldap gateway to the project to support these legacy applications and their ability to connect to kanidm, which would make us immediately more available in many use cases. 

A major question is how would a user authenticate and which password would they connect with through ldap as ldap lacks proper chal-resp auth that we use on the main auth paths.

Firstyear:
Fixes #199 - adds an LDAP gateway to the server. It supports TLS if configured for the webserver, using the same parameters. It is a read only interface, only supporting bind via the configured posix password. 

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1499,9 +1499,11 @@ dependencies = [
  "crossbeam",
  "env_logger",
  "futures",
+ "futures-util",
  "idlset",
  "kanidm_proto",
  "lazy_static",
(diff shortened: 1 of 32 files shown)


Firstyear:
This is failing as I need to update ldap3_server

Issue #103, opened 2019-09-24
Add name2uuid/uuid2name index


Firstyear:
Due to the high amount of value resolution we do on ingress/egress of messages, we should have an index to resolve this faster than loading entries.

Issue #245, opened 2020-06-09
Add limited ldap compat attributes 


Firstyear:
Some attributes we should have for compat to support legacy IDM services, to facilitate migration for applications like nsUniqueId, EntryUUID, and others.

PR #247, opened 2020-06-10
245 ldap compat

Firstyear:
Partially implements #245. This improves filter and entry attribute name rewrite, and adds the nsunique schema. I'm thinking about the best way to proceed here and we want to have compatability for in the future. 


Firstyear:
Going to delay review on this, I deployed a trial version and encountered some issues, so this needs some extra work.

Firstyear:
Fixes #103. I added the name2spn/uuid2name index because it's capable of handling the rdn of entries. Also added tests for ldap bind expectations around valid dns, stabilised what rdn/dn is returned on every entry, and added the index to the arc cache.

PR #289, opened 2020-07-16
Adopt semantic versioning

phozzy:
Fixes #

- [ ] cargo fmt has been run

Hi @Firstyear probably this goes out of procedures, but I saw that you tried to implement automatic versioning using github actions. I think you disabled it since you are not satisfied with its current state now.
So I dare to propose a workflow that I use in my "coding dojo" project. It uses https://github.com/marketplace/actions/github-tag-bump action, that is also based on a quite famous project: https://github.com/fsaintjacques/semver-tool .

### DIFF
--- a/.github/workflows/semver.yml
+++ b/.github/workflows/semver.yml
@@ -1,0 +1,18 @@
+name: Bump semver
+on:
+  push:
+    branches:
+    - master
+jobs:
+  release_build:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        fetch-depth: '0'
+    - name: Bump version and push tag
+      uses: anothrNick/github-tag-action@1.22.0
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        WITH_V: true
+        DEFAULT_BUMP: 'patch'


phozzy:
The only one thing that I don't understand is that Cargo does not support versioning from git tag, or does it?

Firstyear:
Hey @phozzy great to have you submit this PR! It doesn't go out of procedures at all, happy to have your input.

We tried to make it so that there was a version file, and that when changed that it would update the git tag and the Cargo.toml version strings too. (I think, it was actually @pando85 who did that work!). I think it mostly worked but it was the github side of things and permissions I think we hit a problem with - probably because I'm not very knowledgeable about github bits. @pando85 has been busy recently, but I wonder if they have some thoughts on what we can do here?

I think you are right though, that Cargo does not support getting the version from git tag, but it's a cool idea .... Perhaps we could use something like this to achieve it? 

https://crates.io/crates/cargo-release

I'm happy for it to be a partially manual process too by the way, since it's up to me to do it for now :)

Firstyear:
As an extra follow up, perhaps it's a good idea to test cargo-release on another project first, like https://github.com/kanidm/concread which is "less public" and if we make a mistake it's less of a drama etc.

phozzy:
Okay! Then I'll close this PR for now, and will test it on concread project.

PR #316, opened 2020-09-03
314 improve async

### LINKED ISSUE #314: Rewrite away from actix to native tokio/async
Actix has a number of limitations especially around stream handlers and certain types of queueing. We should rewrite away from actix-actors and use native tokio spawned tasks instead. 

This will require

* async capable concread
* changes to queuing and event passing.
* Attaching the delayed event queue to a spawned task

Firstyear:
Fixes #314 - this completely removes actix and actix-web from the codebase, replacing it with tokio and http-rs/tide. Due to a current temporary limit in tokio parts with openssl/libressl, rustls is used for the webserver, but I'll change this back once that issue is resolved. For now there are likely some other clippy issues, but the next step now is that I can *finally* run cargo outdated and update this and the other kanidm/* deps to be up to date due to no longer being held back on versions by actix. So following this, I need to finish clippy warnings, and run cargo outdated and cargo audit. 


kanidmd/server.toml
```diff
@@ -5,4 +5,4 @@ db_fs_type = "zfs"
 tls_ca = "../insecure/ca.pem"
 tls_cert = "../insecure/cert.pem"
 tls_key = "../insecure/key.pem"
-log_level = "perffull"
+log_level = "verbose"
```
pando85, line 8: Is this the default configuration?

Firstyear: The server.toml here is meant for testing really, so maybe I should set this to the maximum level instead?

pando85: Yep, it makes sense.

PR #319, opened 2020-09-10
259 reduce clones

### LINKED ISSUE #259: Reduce clones in idl arc
We have to create a new cache key every ready. We should probably find how to borrow to an & cachekey properly to reduce copies on lookups.

Firstyear:
Fixes #259. Thanks to the linked worked example, we can reduce a problematic set of clones during filter metadata injection. Previously we had to create a new hashset every time we went to resolve index metadata in queries, but with this change we can now just use the copy-on-write hashset instead. This will improve cache access, reduces clones, and more. In a cargo test run this takes out nearly 15% of the execution time (on my system reducing the test time by nearly 35 seconds).

https://github.com/sunshowers/borrow-complex-key-example/blob/master/src/lib.rs 


Firstyear:
Actually, I think I just worked out how to implement https://github.com/kanidm/concread/issues/42 which means I can extend this to the arcache too.

PR #321, opened 2020-09-15
250 cookie to auth bearer

### LINKED ISSUE #250: Convert cookie to auth bearer header
We should use an auth bearer header instead of cookies. They don't have limits on their size and have better protections against cross site theft than cookies. This will also let us stash and use them from the CLI and other tools for session persistence.

Firstyear:
Fixes #250, replacing cookies with auth-bearer tokens. This is done using fernet with randomised keys each startup. The reason for this is that in the future the size of the auth token may exceed cookie limits, so we must be able to understand and process auth bearer. Additionaly, this lets us store the tokens for say the kanidm cli as reqwest today can't persist a cookie jar. 

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1038,6 +1038,18 @@ dependencies = [
 ]
 
 [[package]]
+name = "fernet"
+version = "0.1.3"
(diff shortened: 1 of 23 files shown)


Firstyear:
Okay, now this is ready to go :)

PR #334, opened 2020-12-05
Fix/ci missing libudev

pando85:
Fixes CI with authenticator library.

- [X] cargo fmt has been run

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -8,7 +8,7 @@ jobs:
 
 
   build:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     steps:
     - uses: actions/checkout@v2
 
@@ -16,7 +16,6 @@ jobs:
       run: |
         sudo apt install -y \
           libpam0g-dev \
-          libssl1.0-dev \
           libudev-dev
 
     - name: Build
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -16,7 +16,8 @@ jobs:
       run: |
         sudo apt install -y \
           libpam0g-dev \
-          libssl1.0-dev
+          libssl1.0-dev \
+          libudev-dev
 
     - name: Build
       run: cargo build --verbose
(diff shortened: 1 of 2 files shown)


.github/workflows/rust.yml
```diff
@@ -8,15 +8,15 @@ jobs:
...
+    runs-on: ubuntu-20.04
     steps:
     - uses: actions/checkout@v2
 
     - name: Install dependencies
       run: |
         sudo apt install -y \
           libpam0g-dev \
-          libssl1.0-dev
+          libudev-dev
```
Firstyear, line 19: We should consider leaving libssl1.0-dev in place, as a future change to maek some downstream packaging easier I've had to remove bundling by default, now you have to request them. So we probably also need to set `--features=libsqlite3-sys/bundled` in the cargo build line, or add libsqlite dev as well I think

kanidm_client/tests/common.rs
```diff
@@ -25,7 +33,12 @@ pub fn run_test(test_fn: fn(KanidmClient) -> ()) {
     let (mut ready_tx, mut ready_rx) = mpsc::channel(1);
     let (mut finish_tx, mut finish_rx) = mpsc::channel(1);
 
-    let port = PORT_ALLOC.fetch_add(1, Ordering::SeqCst);
+    let port = loop {
```
Firstyear, line 36: I think we should bound this loop to a set number of attempts so that we don't wait forever, maybe 5 or 10 attempts?

Firstyear (changes requested):
Looks really good mate, as always really appreciate your work and time on this. Only a couple of small comments :)

kanidm_client/tests/common.rs
pando85: Ok, add it because I don't see clearly how to do it without adding to much complexity for that non critical code.

Firstyear: All good, I'll add the change myself :)

Firstyear:
Thank you so much @pando85 for always helping so much!

PR #348, opened 2021-02-01
202 totp cli enrollment

### LINKED ISSUE #202: Add cli tools for adding/removing and managing TOTP
While the server supports TOTP, the cli tools have no yet gained the ability to add/remove or otherwise manage these tokens. This should be added.

Firstyear:
Fixes #202 - This adds support for enrolling and removing totp on the cli, as well as a rebuilt work flow for login to allow dynamic prompting of what credetials are required. 

### DIFF
--- a/kanidm_proto/src/v1.rs
+++ b/kanidm_proto/src/v1.rs
@@ -265,7 +265,7 @@ pub struct UnixUserToken {
 impl fmt::Display for UnixUserToken {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         writeln!(f, "---")?;
-        writeln!(f, "spn: {}", self.name)?;
+        writeln!(f, "spn: {}", self.spn)?;
         writeln!(f, "name: {}", self.name)?;
         writeln!(f, "displayname: {}", self.displayname)?;
         writeln!(f, "uuid: {}", self.uuid)?;
--- a/kanidm_proto/src/v1.rs
+++ b/kanidm_proto/src/v1.rs
@@ -605,13 +605,17 @@ impl TOTPSecret {
             .replace("%3A", "")
             .replace(" ", "%20");
         let label = format!("{}:{}", issuer, accountname);
-        let secret = base32::encode(base32::Alphabet::RFC4648 { padding: false }, &self.secret);
         let algo = self.algo.to_string();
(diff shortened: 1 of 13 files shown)


kanidm_proto/src/v1.rs
```diff
@@ -463,6 +463,17 @@ impl PartialEq for AuthMech {
     }
 }
 
+impl fmt::Display for AuthMech {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            AuthMech::Anonymous => write!(f, "Anonymous (no credentials)"),
```
JuxhinDB, line 469: Not sure if this makes sense, but if you want to make things easy for multi-lingual support, you might want to convert these to indexable codes that a client/UI can map and render themselves.

JuxhinDB: Actually given that this is CLI support, this probably does not matter.

Firstyear: It's a good idea to think about multi-lingual yes, but as you say, I think that's going to be more the focus of the WebUI more than anything.

kanidm_tools/src/cli/account.rs
```diff
@@ -282,6 +294,90 @@ impl AccountOpt {
...
+                    ) {
+                        Ok(v) => v,
+                        Err(e) => {
+                            eprintln!("Error Starting Registration -> {:?}", e);
+                            return;
+                        }
+                    };
+
+                    // display the tok.
+                    eprintln!("You should scan the follow QR code with your OTP App.");
```
Charcol, line 311: "Scan the following QR code with your OTP app."

kanidm_tools/src/cli/account.rs
```diff
@@ -282,6 +294,90 @@ impl AccountOpt {
...
+                        }
+                    };
+                    let image = code
+                        .render::<unicode::Dense1x2>()
+                        .dark_color(unicode::Dense1x2::Light)
+                        .light_color(unicode::Dense1x2::Dark)
+                        .build();
+                    eprintln!("{}", image);
+
+                    eprintln!("Alternately, you can manually enter the following OTP details:");
```
Charcol, line 327: Suggest: "Alternatively"

PR #368, opened 2021-03-07
Fix typo in kanidm_unixd socket path

Flakebi:
The path is /var/run/kanidm-unixd/sock, not /var/run/kanidm.sock.

- [x] cargo fmt has been run ← this fails in `kanidmd/src/lib/macros.rs` at line 18 btw (on nightly)


Firstyear:
Sadly due to distro's not having new enough rust versions we are stuck on 1.45 for a while which is why fmt on nightly / clippy may be different :( 

Otherwise, thanks for the fix! Much appreciated!

PR #370, opened 2021-03-09
Add credential display command

### LINKED ISSUE #364: Credential status display
We need a way to show the current status of credentials, and what they contain including labels. This probably needs some thought about how to handle multi-factor, multiple credentials, and other api keys etc. 

cc @vDorst for raising this.

Firstyear:
Fixes #364 - this adds a credential display command to the cli, and the api so it can be used later. cc @vDorst


Firstyear:
Looks like a permissions thing. I'll double check the access controls and what I have there, but no matching attributes is what you get when there are no credentials to display. I might change that message to make it clearer about what's going on. Good find though!

vDorst:
Great! 

Changing the `--` to ` *` is way better!

Question: `uuid` in the status overview should it match the `uuid` of the queried user?

I see no more issues at the moment.

Firstyear:
It's the uuid of the credential. There is some underlying support for multiple credentials per account and the uuid allows us to select which one we are working on directly. So I would be happy to hide the UUID if it's confusing. At the very least though I'll merge this now since it works for you. Again, thanks so much for your time to test and give constructive feedback!

PR #387, opened 2021-03-22
379 change loginshell

### LINKED ISSUE #379: Cannot change loginshell via kanidm account posix set
I tried changing the posix shell of a user by executing `kanidm` again, but it doesn’t work.

Steps to reproduce:
```bash
kanidm account create testuser testuser
kanidm account posix set --shell /bin/sh testuser
kanidm account posix set --shell /bin/bash testuser
# Fails with Error -> Http(400, Some(SchemaViolation(InvalidAttributeSyntax("loginshell"))), "5b62715c-023a-450f-a4e0-8e684cdaafcf")
```

I ended up deleting the loginshell with raw commands: `kanidm raw modify '{"eq":["name","testuser"]}' mod.json`
mod.json:
```json
[
    { "purged": "loginshell" }
]
```

Firstyear:
Fixes #379 this fixes the generated internal modify to correct purge the previous login shell state to prevent attribute errors. This also adds a test to ensure the behaviour works. 

cc @Flakebi 

### DIFF
--- a/kanidmd/src/lib/actors/v1_write.rs
+++ b/kanidmd/src/lib/actors/v1_write.rs
@@ -1085,8 +1085,14 @@ impl QueryServerWriteV1 {
             "class".into(),
             Value::new_class("posixaccount"),
         )))
+        .chain(iter::once(gidnumber.as_ref().map(|_| {
+            Modify::Purged("gidnumber".into())
+        })))
         .chain(iter::once(gidnumber.map(|n| {
             Modify::Present("gidnumber".into(), Value::new_uint32(n))
+        })))
+        .chain(iter::once(shell.as_ref().map(|_| {
+            Modify::Purged("loginshell".into())
         })))
         .chain(iter::once(shell.map(|s| {
             Modify::Present("loginshell".into(), Value::new_iutf8(s.as_str()))


vDorst (approved):
My default shell is set in the config file to `/bin/fish`

```
# kanidm account posix set --shell /bin/bash test_user -D idm_admin
# su test_user
Password:
bash-5.0$ exit

# kanidm account posix set --shell /bin/fish test_user -D idm_admin
# su test_user
Password:
Welcome to fish, the friendly interactive shell
Type help for instructions on how to use fish
```

vDorst:
I wonder if we need a fail-safe option?

When the shell don't exists, the user can't login.
I think that is fine for a local user.
But in a network with different computers and distro's, the chance that all the computers has that custom shell that you using, is minimal.
Maybe switch back to the configured default found in `/etc/kanidm/unixd`? 

```
# kanidm account posix set --shell /bin/shell_dont_exists test_user -D idm_admin
# su test_user
Password:
su: failed to execute /bin/shell_dont_exists: No such file or directory
```

Firstyear:
I agree we need the fail safe here, but I think it's a fair bit of work to add it. because we need to check the file exists, and probably need a way to cache that so we aren't doing an fs operation for each nsswitch call. So I think it needs a new issue.

PR #389, opened 2021-03-22
Fix Dockerfile to check return values.

OttoHollmann:
Fixes #, check docker build return values.
Now it's possible, if part of build command fails, docker build will continue anyway. And if there are some artifacts in `/usr/src/kanidm/target/release/` they will be reused instead of newly build binaries  (which I suppose is reason of invalid docker images).

- [ - ] cargo fmt has been run
- [ - ] cargo test has been run and passes


Firstyear:
Looks good! thanks!

PR #391, opened 2021-03-24
Base web UI

### LINKED ISSUE #388: Broken docker images
Seems like docker images and docker build is broken (at least for me).

```bash
$ docker pull kanidm/server:latest
$ docker run --rm -it kanidm/server:latest /bin/bash
b20736e6b4f8:/ # /sbin/kanidmd -h
Illegal instruction (core dumped)
```
And it's same for `:devel`

If I try to build own image using `make build/kanidmd`,

on SLES or Debian I get:
```
error: failed to get `base32` as a dependency of package `kanidm_proto v1.1.0-alpha.3 (/usr/src/kanidm/kanidm_proto)`

Caused by:
  failed to initialize index git repository

Caused by:
  failed to resolve path '/root/.cargo/registry/index/github.com-1ecc6299db9ec823/.git/': Operation not permitted; class=Os (2)
ls: cannot access '/usr/src/kanidm/target/release/': No such file or directory
The command '/bin/sh -c if [ "$(uname -m)" == "x86_64" ]; 		then export RUSTFLAGS='-Ctarget-cpu=haswell'; 	fi; 	if [ "$(uname -m)" == "aarch64" ]; 		then export RUSTFLAGS=''; 	fi; 	if [ "${SCCACHE_REDIS}" != "" ]; 		then 			export CC="/usr/bin/sccache /usr/bin/clang" && 			export RUSTC_WRAPPER=sccache && 			sccache --start-server; 		else 			export CC="/usr/bin/clang"; 	fi; 	export RUSTC_BOOTSTRAP=1 && 	echo $RUSTC_BOOTSTRAP && 	echo $RUSTC_WRAPPER && 	echo $RUSTFLAGS && 	echo $CC && 	cargo build 		--features=simd_support,libsqlite3-sys/bundled 		--release; 	if [ "${SCCACHE_REDIS}" != "" ]; 		then sccache -s; 	fi; 	ls -al /usr/src/kanidm/target/release/' returned a non-zero code: 2
make: *** [Makefile:28: build/kanidmd] Error 2
```

But if I run build command on Tumbleweed, it reports different error, (on another run, it will print SIGILL error for different package) e.g.:
```
error: failed to run custom build command for `syn v1.0.64`

Caused by:
  process didn't exit successfully: `/usr/src/kanidm/target/release/build/syn-27d5e4bee826a3c8/build-script-build` (signal: 4, SIGILL: illegal instruction)
warning: build failed, waiting for other jobs to finish...
error: build failed
```
and then build continues (see #389  ).

Which looks really strange, because makefile runs only `docker build` and rest of build is done inside container, so it should be independent on host.

Firstyear:
Fixes #388 - This adds support for compilation profiles, which aids in docker image compatability. These are required so we have profiles to know where static resources are found. This introduces the basic build tooling for constructing wasm outputs and serving these from kanidmd.

TODO: Document in the developer guide how to build these parts, and improve the kanidm book with the new docker image name suffixes (server:x86_64_v3_latest) 


Firstyear:
This adds documentation to the developer readme for the webui parts, and to the main book about the cpu optimised images.

kanidm_book/src/installing_the_server.md
```diff
@@ -10,18 +10,26 @@ You can fetch these with:
...
 There are a number of configuration steps you must perform before you can run this container.
 
 ## System Requirements
 
 ### CPU
 
-You must have a CPU that is from 2013 or newer (Haswell, Ryzen). This is due to the usage of
-SIMD for a number of internal components.
+If you are using the cpu-optimised version, you must have a CPU that is from 2013 or newer
+(Haswell, Ryzen). This is due to the usage of SIMD for a number of internal components.
```
vDorst, line 27: Based on [Intel Wiki](https://en.wikipedia.org/wiki/List_of_Intel_Pentium_processors#Comet_Lake_based_Pentiums) , It is safe to say that all Intel Celeron and Pentium-branded cpu are not supported. Even the current one that are released this year/even this month.

Maybe add which SIMD instructions are important. I think important one is `AVX2`.

Better highlight these points make it easier to making the right choice between the two versions.

Firstyear: I'll list the exact flags that are used :)

PR #399, opened 2021-04-01
Feature Proposal: Interpersonal Identity Verification

### LINKED ISSUE #337: Feature request: verify identity functionality

For example:

QnnOkabayashi:
Fixes #337 

- [ ] cargo fmt has been run

Added proposal for a feature that would allow users to verify identities of company representatives


inter-ident-verif/README.md
```diff
@@ -0,0 +1,62 @@
+# Interpersonal Identity Verification
+### Inspired by GitHub user [Charcoal](https://github.com/kanidm/kanidm/issues/337)
```
Firstyear, line 2: Charcol, not Charcoal :)

inter-ident-verif/README.md
```diff
@@ -0,0 +1,62 @@
+# Interpersonal Identity Verification
```
Firstyear, line 1: The file should be in designs/inter-ident-verify.md not inter-ident-verify/readme.md :)

inter-ident-verif/README.md
```diff
@@ -0,0 +1,62 @@
...
+___
+## Proposed solution
+Many companies have various branches that operate different parts of the company. For example, if Apple Tech Support were to call you, it likely wouldn't be Tim Cook on the other end. Instead, it would be some branch of Apple that put on a tightly vetted list of Apple-trusted callers. If a caller calls a user and the user needs to verify that branch is a trusted Apple caller, they need a way to verify this with only the knowledge that the caller is supposedly an Apple-verified source. This is where Kanidm comes into play.
+
+All users (individuals, companies, company branches, etc.) of this verification system will have an existing RSA key linked to their Kanidm credentials on the servers. This key will not change (unless for very good reason), and the private key is only known by the exact account without exception. Since these keys are linked to credentials, we can safely find the authentic public key of any user we want, without real time communication between servers around the world.
```
Firstyear, line 14: This is good to see you consider global replication here!

inter-ident-verif/README.md
```diff
@@ -0,0 +1,62 @@
...
+1. User generates random, unique message, and stores a copy for later.
+2. User encrypts message with their private key, ensuring that the end target (Apple) must acknowledge them when modifying the result.
+3. User encrypts result with Apple public key, preventing the caller from decrypting to get the original message.
+4. User sends result to Apple Tech Support.
+
+### Apple Tech Support
+5. Apple Tech Support encrypts message with Apple Tech Support private key, ensuring that Apple knows it's from them.
+6. Apple Tech Support sends result to Apple.
+
+### Apple
```
Firstyear, line 33: It's worth defining all the parties involved in this operation, and where kanidm sits in this. At the moment I can only assume "Apple" is the kanidm site.

It's probably also worth not saying Apple but "corp".

inter-ident-verif/README.md
```diff
@@ -0,0 +1,62 @@
...
+An important aspect of this process is that users are able to read their hashes over the phone. To increase the readability of these hashes, I propose a modified number representation system similar to hexadecimal where each digit has 64 variants, `{a-z, A-Z, 0-9, *, #}`, allowing 6-bits of information to be stored per digit. With only 6 digits in this number system, we can represent up to 68,719,476,736, covering all 32-bit numbers. With 11 digits, we can represent up to 7.38e19, covering all 64-bit numbers. For shorter messages, this should be enough to read these values over voice.
+___
+## User interface
+When the user receives a call, they ask for the callers root company. They then navigate and log into the Kanidm web interface on their smartphone, search up the root company, and click a button that does steps 1-3, and displays the result to the user in the number system specified above. This code is read over the phone, where 5-15 are performed remotely for the user. Once the caller reads the resulting message back to the user, they will enter this back into the web interface, which performs steps 17-19, and tells the user if the verification was successful.
```
Firstyear, line 61: This assumes the user has an account on the kanidm instance at the company, yes? 

It's not clear here how the "read numbers" fit in with steps 17-19 or steps 5-15. 

Additionally, it is worth asking how this defends from replay attacks? I assume the message in this case is actually a nonce?

Firstyear (changes requested):

designs/inter-ident-verify.md
```diff
@@ -0,0 +1,78 @@
...
+* X, local to A
+* Y, local to S and U
+
+All users (individuals, companies, company branches, etc.) of this verification system will have an existing RSA key linked to their Kanidm credentials on their local servers. This key will not change (unless for very good reason), and the private key is only known by the exact account without exception. Since these keys are linked to credentials, we can safely find the authentic public key of any user we want, without real time communication between servers around the world. Additionally, all trusted callers for A will also have a copy of their information (name, public key) linked to A's credentials on server X. This is acceptable since there are limited trusted callers for A. Furthermore, server Y will store a copy of A's information (name, public key) locally. This is acceptable, since there are limited large corporations. It's also important to note that A doesn't need to store any information about any particular U on server X.
+
+In summary:
+* X stores A, which holds name and public key of S
+* Y stores U and S, and also name and public key of A
```
QnnOkabayashi, line 34: Specified Kanidm sites as requested

designs/inter-ident-verify.md
```diff
@@ -0,0 +1,78 @@
...
+
+If A wanted to contact U, the CEO wouldn't call them directly. Instead, they would reach out through S, which lives in a tightly vetted whitelist of callers approved by A. S calls U, and U must verify that S is a trusted branch of A, with only the knowledge that S supposedly is trusted by A. U doesn't care about the exact identity of the caller, only that they are trusted by A, since no scam company could make A's tightly vetted whitelist of trusted callers. Since they're not contacting A directly, there are four key connections that must be verified:
+1. U needs to know the S is trusted by A.
+2. A needs to verify the S is trusted.
+3. U must verify their request for verification was directly verified by A and nobody else.
+4. U needs to know the response received is not a replay attack, and that the response received is unique to their request for verification.
+
+This can be accomplished by sending a nonce encrypted message to A, and then sending it back to U with a different encryption pattern. If the message received is the same as the message sent, then all identities are verified. Since the message sent is always unique, replay attacks are meaningless.
```
QnnOkabayashi, line 23: Address the issue of replay attacks

inter-ident-verif/README.md
QnnOkabayashi: I don't know how to respond to comments with quoted code so I made responses below detailing how I addressed things you brought up. Thanks for the feedback!

designs/inter-ident-verify.md
```diff
@@ -0,0 +1,78 @@
...
+___
+## User interface
+When the user receives a call, they ask for the callers root company. They then navigate and log into the Kanidm web interface on their smartphone, running on server Y, search up the root company, and click a button that does steps 1-3, and displays the result to the user in the number system specified above. This hash is displayed to the user, where they can transmit the code directly over the phone since it should be relatively short. The caller enters these numbers into their end, where steps 5-15 are performed remotely from the user. Once the caller reads the resulting hash back to the user, they will enter this back into the web interface, which performs steps 17-19, and tells the user if the verification was successful. This allows the user to verify that the caller on the other end of the phone is the company that the corporation sees and validates.
```
QnnOkabayashi, line 75: Specified how reading the numbers is related to steps. Does this make sense / how can I clarify here?

designs/inter-ident-verify.md
```diff
@@ -0,0 +1,78 @@
...
+___
+## Proposed solution
+Many companies have various branches that operate different parts of the company. For example, take the following parties:
+* U, the individual user
+* A, the big name corporation the user knows of
+* S, a branch of A that handles tech support, and is trusted by A
+
+If A wanted to contact U, the CEO wouldn't call them directly. Instead, they would reach out through S, which lives in a tightly vetted whitelist of callers approved by A. S calls U, and U must verify that S is a trusted branch of A, with only the knowledge that S supposedly is trusted by A. U doesn't care about the exact identity of the caller, only that they are trusted by A, since no scam company could make A's tightly vetted whitelist of trusted callers. Since they're not contacting A directly, there are four key connections that must be verified:
+1. U needs to know the S is trusted by A.
```
Firstyear, line 18: I think these letter abreviations make it less clear. User, Support, Corp works?

designs/inter-ident-verify.md
```diff
@@ -0,0 +1,78 @@
...
+
+### Kanidm Server setup
+In this situation, there are two Kanidm instances which can be remote from each other:
+* X, local to A
+* Y, local to S and U
+
+All users (individuals, companies, company branches, etc.) of this verification system will have an existing RSA key linked to their Kanidm credentials on their local servers. This key will not change (unless for very good reason), and the private key is only known by the exact account without exception. Since these keys are linked to credentials, we can safely find the authentic public key of any user we want, without real time communication between servers around the world. Additionally, all trusted callers for A will also have a copy of their information (name, public key) linked to A's credentials on server X. This is acceptable since there are limited trusted callers for A. Furthermore, server Y will store a copy of A's information (name, public key) locally. This is acceptable, since there are limited large corporations. It's also important to note that A doesn't need to store any information about any particular U on server X.
```
Firstyear, line 30: When you say "on their local server" do you mean, "stored as an attribute in their own account entry on the kanidm server"? Or is the key stored outside of the kanidm instance somehow?

Also please use ECDSA or ED25519 instead of RSA :)

designs/inter-ident-verify.md
```diff
@@ -0,0 +1,78 @@
...
+In summary:
+* X stores A, which holds name and public key of S
+* Y stores U and S, and also name and public key of A
+
+### Methodology
+#### Individual user (U)
+1. U generates random, unique message, and stores a copy for later.
+2. U encrypts message with their private key, ensuring that the end target (A) must acknowledge them when modifying the result.
+3. U encrypts result with A public key from server Y, preventing S from decrypting to get the original message.
+4. U sends result to S via phone call.
```
Firstyear, line 41: This procedure seems interesting, but I think you also should give a subsequent version which shows a fully worked example of what is generated, what is read out, etc.

designs/inter-ident-verify.md
Firstyear: I think you need to do a second, stepped through worked example.

Firstyear (changes requested):
I think you are actually solving a different problem here maybe. I think that the intent is say ... if I from software engineering call someone in finance, and we want to verify each other's as individuals within a single corp. 

What you have here though, is a bit closer to say ... you have a support person, who needs to be verified, so there needs to be a seperate "support group" verification credential that they can access through membership f that group. Which is similar, but different.

For example, it's common that to do this kind of thing, you need each party to know the identity of the other, and it may not be appropriate fyor a support staff member to hand out their username or identity. So you want the user to verify the support *group* and the member of that group can verify the user. Because then you are verifying their legitimacy to the membership.

But I think that should be a second stage. For now, I think you may benefit to rework the example for "user 1 speaking to user 2", and then doing group membership verification is a stage 2 (but probably has a lot of overlap).

QnnOkabayashi:
> I think you are actually solving a different problem here maybe. I think that the intent is say ... if I from software engineering call someone in finance, and we want to verify each other's as individuals within a single corp.
> 
> What you have here though, is a bit closer to say ... you have a support person, who needs to be verified, so there needs to be a seperate "support group" verification credential that they can access through membership f that group. Which is similar, but different.
> 
> For example, it's common that to do this kind of thing, you need each party to know the identity of the other, and it may not be appropriate fyor a support staff member to hand out their username or identity. So you want the user to verify the support _group_ and the member of that group can verify the user. Because then you are verifying their legitimacy to the membership.
> 
> But I think that should be a second stage. For now, I think you may benefit to rework the example for "user 1 speaking to user 2", and then doing group membership verification is a stage 2 (but probably has a lot of overlap).

QnnOkabayashi:
Thanks for the feedback! I have some brief questions:

> I think you are actually solving a different problem here maybe. I think that the intent is say ... if I from software engineering call someone in finance, and we want to verify each other's as individuals within a single corp.
>
> What you have here though, is a bit closer to say ... you have a support person, who needs to be verified, so there needs to be a seperate "support group" verification credential that they can access through membership f that group. Which is similar, but different.

I agree with the second part, but what do you mean in the first part? Do you mean I should shift my approach to be applicable for intra-corporation communication instead of between a support group and an outside customer?

> For example, it's common that to do this kind of thing, you need each party to know the identity of the other, and it may not be appropriate fyor a support staff member to hand out their username or identity. So you want the user to verify the support _group_ and the member of that group can verify the user. Because then you are verifying their legitimacy to the membership.

This like what I'm currently proposing, correct?

> But I think that should be a second stage. For now, I think you may benefit to rework the example for "user 1 speaking to user 2", and then doing group membership verification is a stage 2 (but probably has a lot of overlap).

Right now, I'm envisioning three scenarios:
1. User 1 speaking to user 2
2. Support group speaking to user
3. SWE branch speaking to finance

Are you essentially suggesting that scenario 1 is a good MVP, and that scenarios 2/3 would be good extensions?

Again, thank you for the feedback!

Firstyear:
> 
> I agree with the second part, but what do you mean in the first part? Do you mean I should shift my approach to be applicable for intra-corporation communication instead of between a support group and an outside customer?

Yep - initially we need intra-corp communication verification (to prevent classes of social engineering attacks). The external communication/verification can be looked at later :) 

> 
> > For example, it's common that to do this kind of thing, you need each party to know the identity of the other, and it may not be appropriate fyor a support staff member to hand out their username or identity. So you want the user to verify the support _group_ and the member of that group can verify the user. Because then you are verifying their legitimacy to the membership.
> 
> This is like what I'm currently proposing, correct?

I didn't think it was completely clear in the proposal is all :) You never mentioned that it was a group membership, and being a member of that group is what you were validating. 

> 
> > But I think that should be a second stage. For now, I think you may benefit to rework the example for "user 1 speaking to user 2", and then doing group membership verification is a stage 2 (but probably has a lot of overlap).
> 
> Right now, I'm envisioning three scenarios:
> 
> 1. User 1 speaking to user 2
> 2. Support group speaking to user
> 3. SWE branch speaking to finance
> 
> Are you essentially suggesting that scenario 1 is a good MVP, and that scenarios 2/3 would be good extensions?
> 
> Again, thank you for the feedback!

Yep, that's what I'm suggesting! 

I think the most important step next is to write out the user experience scenarios and how those look - the technical design really comes second and interacts from that.

QnnOkabayashi:
> > I agree with the second part, but what do you mean in the first part? Do you mean I should shift my approach to be applicable for intra-corporation communication instead of between a support group and an outside customer?
> 
> Yep - initially we need intra-corp communication verification (to prevent classes of social engineering attacks). The external communication/verification can be looked at later :)

I'm curious about the motivation to prioritize intra-corporation communication. Wouldn't most corporations already have valid forms on intra-corporation communication already? I would imagine that more of this communication occurs over email / official corp methods of messaging.

On the other hand, I can envision scenarios where corps calling external users would be very beneficial over secure channels like email. Examples could be suspicious credit card behavior that should be addressed immediately, contacting an incapacitated persons emergency contact, and so on. The user is especially vulnerable in both of these scenarios, and I could definitely see as targets for social engineering scams. 

Obviously you know Kanidm's place more than I, but I want to get this cleared up before I focus on one of them.

Firstyear:
> > > I agree with the second part, but what do you mean in the first part? Do you mean I should shift my approach to be applicable for intra-corporation communication instead of between a support group and an outside customer?
> > 
> > 
> > Yep - initially we need intra-corp communication verification (to prevent classes of social engineering attacks). The external communication/verification can be looked at later :)
> 
> I'm curious about the motivation to prioritize intra-corporation communication. Wouldn't most corporations already have valid forms on intra-corporation communication already? I would imagine that more of this communication occurs over email / official corp methods of messaging.

I think @Charcol could weigh in much more here - the reality is that a lot of attacks today like phishing, spear phishing, whaling, and more all still happen. So while you may have intra-corporation messaging, you still need a way to perform out of band verification that the communication is legitimate. So this is actually the most important initial use case for this.  

> 
> On the other hand, I can envision scenarios where corps calling external users would be very beneficial over secure channels like email. Examples could be suspicious credit card behavior that should be addressed immediately, contacting an incapacitated persons emergency contact, and so on. The user is especially vulnerable in both of these scenarios, and I could definitely see as targets for social engineering scams.
> 
> Obviously you know Kanidm's place more than I, but I want to get this cleared up before I focus on one of them.

External communications adds extra challenges that build on intra corp verification, so start with intra corp :)

JuxhinDB:
If I can pitch in briefly on one small point -- 

> I'm curious about the motivation to prioritize intra-corporation communication. Wouldn't most corporations already have valid forms on intra-corporation communication already? I would imagine that more of this communication occurs over email / official corp methods of messaging.

A feature like this would generally get baked into a policy & process around critical stakeholders in a business that are deemed high risk (i.e. C-level, high access users, insider lists for publicly traded companies etc.). The value here is that certain processes (e.g. procurement in excess of >$100k) can mandate intra-corporation verification through kanidm to prevent against social engineering attacks and non-repudiation. Hope that helps.

Firstyear:
@Charcol Do you mind reviewing this PR as is, and giving your insight too? It would really help @QnnOkabayashi I'm sure :)

designs/inter-ident-verify.md
```diff
@@ -3,76 +3,96 @@
...
+An important aspect of this process is that users are able to read their hashes over the phone. To increase the readability of these hashes, I propose using English sentences to encode hashes. Since Ed25519 hashes create 512-bit signatures, we would need 36 words if we had a word bank with 20,000 words. If we followed a basic sentence structure to make these envoding more natural to read, where each sentence could be made up of 5 words, we could encode a hashing in ~7 simple sentences, which is incredible for the security we are ensuring.
+
+TODO: this requires 20,000 words for each object in the sentence, so not quite feasible?
```
Firstyear, line 76: There is an easier way I think here. I think you could have a TOTP half-key on each users account, and when User A wants to verify user B, you concat their secrets together to get a full TOTP key that is a unique combination between the two individuals. Then you can simply display the TOTP and have the person read it.

For challenge response, You could have it so that User A + User B key is one direction, and User B  + User A key is the other since both will yield unique TOTPs. Then it's a 6 digit number that is continually changing, and is combined from User A + User B, without either user ever having the content of the secret exposed (Kani itself would arbitrate that you may read or see only your key muxed with someone elses).

This defends from say User C being able to impersonate User A, because Kani would only allow them to mix User C's key with the half key from A *or* B. They will never be able to see the factor that A and B both provide.

PR #402, opened 2021-04-07
Create mfa_backup_code.rst

### LINKED ISSUE #163: Account recovery (backup) mfa codes
We should support having backup mfa codes so that single-device accounts can be restored.

victorcwai:
Fixes #163, draft design for mfa backup code

- [ ] cargo fmt has been run


designs/mfa_backup_code.rst
```diff
@@ -0,0 +1,81 @@
...
+A user can only have one Backup code, if user wants a new Backup code:
+
+- if the existing Backup code is still valid, cannot create new Backup code
+- if no existing/valid Backup code, create/replace with a new Backup code
+
+The login session should only last for 5 mins with a Claim that allows users to change login credentials.
+
+Code Generation
+------------------------
+The Backup code could be generated by TOTP/HOTP, or a cryptographically secure pseudorandom number generator.
```
Firstyear, line 29: look at kanidmd/src/lib/utils.rs readable_password_from_random :)

designs/mfa_backup_code.rst
```diff
@@ -0,0 +1,81 @@
...
+
+Based on the Microsoft temporary access pass:
+
+Add a new class/object: Backup code policy
+
+- Each Kanidm has a Backup code policy
+- Define which users/groups can use the Backup code
+   - User can only login if they are included in the Backup code policy
+- Define parameters like max/min lifetime, one-time use and length
+- Only ``idm_hp_account_manage_priv`` can modify it?
```
Firstyear, line 43: Or self modification if they are in the correct policy group. I think this policy actually could already be done pretty easily via a membership to a backup_code_allowed group and the existing access profile infrastructure.

designs/mfa_backup_code.rst
```diff
@@ -0,0 +1,81 @@
...
+Overview
+------------------------
+
+If the user has a Backup code, then they can replace the TOTP/WebAuthn challenge with it during login. For example, TOTP + password auth will be Backup code + password instead.
+
+To implement this, we can add Backup code as a new AuthCredential, so that it can be used in the auth process.
+
+To validate Backup code, we simply check if the input matches the Backup code saved in database, which should be hashed with salt (similar to how password is validated).
+
+A user can only have one Backup code, if user wants a new Backup code:
```
Firstyear, line 20: I'd check how github does this - since the backup code is "once use" and get's scratched off (deleted) after use, then we also need:

* An async message to remove that backup code from the account after usage
* Multiple backup codes.

Github for example issues batches of 5 or 10 at a time that can be printed. I think we should do similar.

designs/mfa_backup_code.rst
```diff
@@ -0,0 +1,81 @@
...
+Add a new class/object: Backup code
+
+- Issue and save the Backup code in the database with predefined parameters of lifetime, one-time use and length
+- ``idm_hp_account_manage_priv``, ``idm_account_manage_priv`` and normal users themselves can create/read/delete it?
+
+Workflow
+================
+1. User request Backup code from high-priv users (``idm_hp_account_manage_priv`` or ``idm_account_manage_priv``) via phone/email/etc.
+2. User receive Backup code from high-priv users via phone/email/etc.
+3. User login with the Backup code
```
Firstyear, line 69: This sounds more like a password reset flow? I think this may be out of scope ....

designs/mfa_backup_code.rst
```diff
@@ -0,0 +1,81 @@
...
+
+Or
+
+1. User can generate Backup code themselves if they are already authenticated (for later use)
```
Firstyear, line 73: Yes, the work flow here should only be users self requesting I think. In other cases I think that a full credential reset on account lockout may be better ....

Firstyear (changes requested):

designs/mfa_backup_code.rst
victorcwai: I see your point. I suppose this should be separated into a "password reset" flow.

designs/mfa_backup_code.rst
victorcwai: Github issues a batch of 16 codes. When a code is used, it is no longer valid (i.e. you can only use the remaining 15 codes).
When user generates a new batch of backup codes, the previous codes will no longer be valid.

victorcwai: One question though: Why do we need a separate async message? Couldn't we just do it in the same transaction?

Firstyear: Yep, that's the thought I had is to have a batch of codes and you use one at a time. 

Well, the thing about the auth process is that currently while it's in a single transaction lock, that transaction is NOT a writeable transaction, and the intent is to make it so that authentications can proceed in parallel.

Because the server can also operate as a single thread, and there are a limited set of db tickets, that also prevents opening nested transactions too.

So the correct way here to perform the "write" to invalidate the code is to send a message to the async task queue - you can see how this is done for "on login pw upgrades". A similar process could be used here, but when the recovery code is used and is found correct, we immediately send the async task to deactivate it. Theoretically, this means there is a small window between "use of the backup code" and the async action being processed to actually cause the invalidation, but this window is short. So I think it's an acceptable compromise here.

designs/mfa_backup_code.rst
Firstyear: Well, password reset or an account setup flow I think. Because any password reset, those passwords/tokens need a time limit on their usage and they are a single factor, where these backup mfa codes can't have one since the user has to be able to use them from anypoint from now to the future, and they are to "replace" a lost TOTP/Webauthn token so they are still part of an MFA setup.

designs/mfa_backup_code.rst
victorcwai: Yep I was wondering what if another agent login with the same code while the async action is being processed. 
Thanks for clearing that up!

designs/mfa_backup_code.rst
```diff
@@ -17,65 +17,35 @@ To implement this, we can add Backup code as a new AuthCredential, so that it ca
...
 Code Generation
 ------------------------
-The Backup code could be generated by TOTP/HOTP, or a cryptographically secure pseudorandom number generator. 
-Attackers should not be able to guess it even if they know our source code.
-
-Implementation 1
-------------------------
 
-Based on the Microsoft temporary access pass:
+The Backup code could be generated by TOTP/HOTP, or a cryptographically secure pseudorandom number generator. Attackers should not be able to guess it even if they know our source code.
```
Firstyear, line 25: I'd remove TOTP/HOTP here, as I don't think these are viable options here.

designs/mfa_backup_code.rst
```diff
@@ -17,65 +17,35 @@ To implement this, we can add Backup code as a new AuthCredential, so that it ca
...
-- Only ``idm_hp_account_manage_priv`` can modify it?
+Implementation
+------------------------
 
 Add a new class/object: Backup code
 
-- Issue and save the Backup code in the database, it must adhere to Backup code policy
-- ``idm_account_manage_priv`` can create/read/delete it for normal users, but not themselves?
-- ``idm_hp_account_manage_priv`` can create/read/delete it for ``idm_account_manage_priv`` users and normal users?
+A user can generate one batch of Backup codes at a time, which will be saved in the database
```
Firstyear, line 34: 'saved in the database in a non-reversible format (KDF)'

designs/mfa_backup_code.rst
```diff
@@ -17,65 +17,35 @@ To implement this, we can add Backup code as a new AuthCredential, so that it ca
...
+   - Each code in the batch can be used only once
+   - If the user generates a new batch of Backup codes, the last batch of codes become invalid.
 
-Workflow
+To invalidate the Backup code after usage, we should send an async message to the async task queue. Since the auth process is currently in a single transaction lock that is NOT writeable, using an async action allows the auth process to proceed in parallel and prevent opening nested transactions. **Note** that there is a small window between "use of the backup code" and the async action being processed to actually cause the invalidation, but this window is short enough that it's an acceptable compromise.
```
Firstyear, line 38: We also need to comment on how we need rate limiting/bruteforce protection against these codes similar to what exists for passwords.

designs/mfa_backup_code.rst
```diff
@@ -17,65 +17,35 @@ To implement this, we can add Backup code as a new AuthCredential, so that it ca
...
-3. User login with the Backup code
 
-Implementation 2
-------------------------
-
-Add a new class/object: Backup code
-
-- Issue and save the Backup code in the database with predefined parameters of lifetime, one-time use and length
-- ``idm_hp_account_manage_priv``, ``idm_account_manage_priv`` and normal users themselves can create/read/delete it?
+We can decide which users/groups can use the Backup code via a membership to a ``backup_code_allowed`` group and making use of the existing access profile infrastructure.
```
Firstyear, line 43: `idm_account_mfa_backup_code_self_priv` I think is the better name for this. For other users to admin backup codes for others it should be `idm_account_mfa_backup_code_manage_priv`. See kanidmd/src/lib/constants/entries.rs for examples of existing groups for privileges.

Firstyear (changes requested):

Firstyear (approved):

Firstyear:
Looks great!

PR #419, opened 2021-04-24
62 idm qs cleanup

### LINKED ISSUE #62: Seperate IDM migrations from QS
The server *could* be built in such a way as to allow a "generic" DB to be built (with some work to split out plugins). To help, the IDM specific migrations should be moved to the IDMServer component, rather than in the QS component, to properly seperate these intents.

Firstyear:
Fixes #62 - this cleans up how we call the qs stack with the IDM components, and means that we can properly have these layered, allowing things like the password badlist set to be stored in COW on the IDMServer (Rather than the query server). 

### DIFF
--- a/kanidmd/src/lib/access.rs
+++ b/kanidmd/src/lib/access.rs
@@ -26,17 +26,15 @@ use std::cell::Cell;
 use std::ops::DerefMut;
 use uuid::Uuid;
 
-use crate::audit::AuditScope;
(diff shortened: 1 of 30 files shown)


Firstyear:
cc @victorcwai this will make it easier for you to put the password badlist cow cell onto the IDMServer for it to be using in IdmAuthTransactions rather than in the query server. It exposes a changed_uuid set from the query server that you can check during fn commit in write :)

PR #420, opened 2021-04-24
Making clippy happy

yaleman:
Fixes a load of `cargo clippy` complaints. 

Replaces the use of the string username `"admin"` in tests with a single common variable - same as the password is for same tests.

- [x] cargo fmt has been run

All the above tests were run on rust 1.49.0 across:

 - openSUSE Tumbleweed
 - openSUSE Leap 15.2
 - Ubuntu Bionic 
 - macOS Big Sur on M1
 - Debian Buster

### DIFF
--- a/kanidmd/src/lib/lib.rs
+++ b/kanidmd/src/lib/lib.rs
@@ -3,7 +3,7 @@
 #![deny(clippy::unwrap_used)]
 #![deny(clippy::expect_used)]
 // #![deny(clippy::panic)]
-#![allow(clippy::unreachable)]
+#![deny(clippy::unreachable)]
 #![deny(clippy::await_holding_lock)]
 #![deny(clippy::needless_pass_by_value)]
 #![deny(clippy::trivially_copy_pass_by_ref)]
--- a/kanidmd/src/lib/lib.rs
+++ b/kanidmd/src/lib/lib.rs
@@ -3,7 +3,7 @@
 #![deny(clippy::unwrap_used)]
 #![deny(clippy::expect_used)]
 // #![deny(clippy::panic)]
-#![deny(clippy::unreachable)]
+#![allow(clippy::unreachable)]
 #![deny(clippy::await_holding_lock)]
 #![deny(clippy::needless_pass_by_value)]
 #![deny(clippy::trivially_copy_pass_by_ref)]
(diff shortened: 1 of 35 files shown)


yaleman:
🔥 #420 blazeit

yaleman:
... test failed? That's rude :(

Firstyear:
Test failure is a known random issue in tests.

kanidmd/src/lib/lib.rs
```diff
@@ -3,7 +3,7 @@
 #![deny(clippy::unwrap_used)]
 #![deny(clippy::expect_used)]
 // #![deny(clippy::panic)]
-#![deny(clippy::unreachable)]
+#![allow(clippy::unreachable)]
```
Firstyear, line 6: We should deny unreachhable an annotate individual locations where we need to allow it I think.

kanidmd_web_ui/src/login.rs
```diff
@@ -20,6 +20,7 @@ extern "C" {
 pub struct LoginApp {
     link: ComponentLink<Self>,
     inputvalue: String,
+    #[allow(dead_code)]
```
Firstyear, line 23: Shouldn't be needed, lstorage should be used in the latest commits.

kanidmd_web_ui/src/login.rs
```diff
@@ -262,7 +263,7 @@ impl Component for LoginApp {
     type Properties = ();
 
     fn create(_: Self::Properties, link: ComponentLink<Self>) -> Self {
-        ConsoleService::log(format!("create").as_str());
+        ConsoleService::log(&"create".to_string());
```
Firstyear, line 266: Should just be "create" no need to & and to_string here.

Firstyear:
@yaleman Do you mind if I merge #419 first? I think it may need you to rebase though. Alternately if you do these little fix ups I'll merge this first, then I'll rebase 419 to this change. Which ever is easier :)

kanidmd_web_ui/src/login.rs
yaleman: done

yaleman:
@Firstyear go ahead with #419  - happy to rebase :)

kanidmd_web_ui/src/login.rs
yaleman: done

PR #425, opened 2021-04-28
397 Caching password badlist

### LINKED ISSUE #397: Caching for password badlist
As per discussion in #159 and #394, implement a cache for password badlist in memory since it will be frequently accessed.
Then we can pass the cached map of badlist by reference rather than by value.

victorcwai:
Fixes #397, keep password badlist in memory

- [x] cargo fmt has been run

TODO: write tests for concurrent read/write lock?


kanidmd/src/lib/idm/server.rs
```diff
@@ -39,10 +39,13 @@ use async_std::task;
...
 
-use concread::bptree::{BptreeMap, BptreeMapWriteTxn};
 use concread::hashmap::HashMap;
+use concread::{
+    bptree::{BptreeMap, BptreeMapWriteTxn},
+    CowCell,
+};
 use rand::prelude::*;
-use std::time::Duration;
+use std::{collections::BTreeSet, sync::Arc, time::Duration};
```
Firstyear, line 48: I think this should be a hashset since this will be accessed very frequently with fast lookups :)

kanidmd/src/lib/idm/server.rs
```diff
@@ -67,6 +70,7 @@ pub struct IdmServer {
     async_tx: Sender<DelayedAction>,
     // Our webauthn verifier/config
     webauthn: Webauthn<WebauthnDomainConfig>,
+    pw_badlist_cache: Arc<CowCell<BTreeSet<Value>>>,
```
Firstyear, line 73: This should be HashSet<String> I think since during any password operation we have the password cleartext as a string first, and that's more compact in the HashSet.

kanidmd/src/lib/idm/server.rs
```diff
@@ -84,6 +88,7 @@ pub struct IdmServerAuthTransaction<'a> {
     // For flagging eventual actions.
     async_tx: Sender<DelayedAction>,
     webauthn: &'a Webauthn<WebauthnDomainConfig>,
+    pw_badlist_cache: Arc<CowCell<BTreeSet<Value>>>,
```
Firstyear, line 91: Shouldn't this be a CowCellReadTxn, rather than an arc<cowcell<...>> here?

kanidmd/src/lib/idm/server.rs
```diff
@@ -898,14 +903,8 @@ impl<'a> IdmServerProxyWriteTransaction<'a> {
...
         let lc_password = PartialValue::new_iutf8(cleartext);
-        let badlist_entry = self
-            .qs_write
-            .internal_search_uuid(au, &UUID_SYSTEM_CONFIG)
-            .map_err(|e| {
-                ladmin_error!(au, "Failed to retrieve system configuration {:?}", e);
-                e
-            })?;
-        if badlist_entry.attribute_value_pres("badlist_password", &lc_password) {
+        let pw_badlist_cache = &*self.pw_badlist_cache.read();
```
Firstyear, line 906: If you change the type to a HashSet<string> you can do the check pw_badlist_cache.contains(cleartext) before the conversion to lc_password :)

kanidmd/src/lib/idm/server.rs
```diff
@@ -1522,11 +1521,30 @@ impl<'a> IdmServerProxyWriteTransaction<'a> {
...
                 self.mfareg_sessions.commit();
                 self.qs_write.commit(au)
             }
         )
     }
+
+    fn reload_password_badlist(&self, au: &mut AuditScope) -> Result<(), OperationError> {
+        match self.qs_write.get_password_badlist(au) {
+            Ok(badlist_entry) => {
+                let mut pw_badlist_cache_write_txn = self.pw_badlist_cache.write();
```
Firstyear, line 1540: I don't think this is quitecorrect. You should take the .write() in the start of the ProxyWrite transaction, so you can just directly call into current value. Then you commit it before qs_write.commit.

Firstyear (changes requested):
Looks like your on the right track here. I think you need to change how you approach these transactions a bit. You need to take the read/write transactions during the start of the proxywrite and proxy read types. 

Plenty of comments inline, but generally good start!

Firstyear:
I also had a think about this this morning, how does this load the badlist cache on restart of the server? I think we may have to fix up initialise_helper in queryServer to be split into a few parts and driven by the IdmServer. I can be responsible for that though, since it was my past mistake :)

kanidmd/src/lib/idm/server.rs
victorcwai: My thinking was we should get the write lock (.write()) only when the changed_uuid contains UUID_SYSTEM_CONFIG.
But I suppose it doesn't really matter if we are using CowCell and there's only 1 writer?

Firstyear: The main reason to "always take it" in the lock form of the transaction is that if a reader was to "mistakenly" be coded to attempt to take the write during a read, there is a possibility of deadlock.

So it's better here given there is only a single writer, to take the write lock at the start of the write, to ensure we really do have access to it, and similar, we take the read lock at the start of a read, so we know we can never accidentaly deadlock during an operation. 

It's more about preventing human/programmer error (we are all likely to make mistakes!) than anything.

And of course, because of the structure of the code, the lock is cheap to take anyway :)

victorcwai: Ah good, I see your point!

kanidmd/src/lib/idm/server.rs
victorcwai: Well then I suppose we should also change from BTreeSet to HashSet in authsession.rs too, right? 

Also, is there any easy way to convert from BTreeSet to HashSet? I am guessing I have to use a iterator and map from `Value` to `String` and insert the strings to a new HashSet.

Firstyear: Yeah, when you are building the cache value during the fn commit, you'll take the btreeset of values from the entry, and then iterate and map that to a hashset<String> :)

kanidmd/src/lib/idm/server.rs
```diff
@@ -902,9 +906,8 @@ impl<'a> IdmServerProxyWriteTransaction<'a> {
 
         // check a password badlist to eliminate more content
         // we check the password as "lower case" to help eliminate possibilities
-        let lc_password = PartialValue::new_iutf8(cleartext);
-        let pw_badlist_cache = &*self.pw_badlist_cache.read();
-        if pw_badlist_cache.contains(&lc_password) {
+        // also, when pw_badlist_cache is read from DB, it is read as Value (iutf8 lowercase)
```
Firstyear, line 909: Good catch! 👍

Firstyear (approved):
Okay this looks really good! Choice is up to you if you want to make one more (extension) change to this. 

We can merge it as it is then do a second change to make the badlist a reference to the authsession operations, or we can do the reference change now.

I think that we should remove pw_badlist_set from AuthSession, and just pass it as a reference to fn validate_creds each call instead. This means we'll be able to prevent the use of .clone() on the HashSet. What do you think? Do you want to do that in this PR, or do you want to merge this and do it in a second one? 

Regardless, great work, it looks really good!

Firstyear:
Great work! Thank you!

PR #432, opened 2021-05-06
Add ability to pick a server role

### LINKED ISSUE #426: Config flag to not build/load/require the WebUI?
Maybe just a feature flag on-build? `kanidmd` won't run without all the web goodies being in a particular path, which is an ergo issue during development :(

Definitely a low-priority request :)

Firstyear:
Fixes #426 adds a configuration element for "role" which can define what features we expose from the server. 

### DIFF
--- a/kanidmd/src/lib/core/https.rs
+++ b/kanidmd/src/lib/core/https.rs
@@ -13,7 +13,7 @@ use crate::actors::v1_write::{
     InternalRegenerateRadiusMessage, InternalSshKeyCreateMessage, ModifyMessage,
     PurgeAttributeMessage, RemoveAttributeValueMessage, ReviveRecycledMessage, SetAttributeMessage,
 };
-use crate::config::{TlsConfiguration, ServerRole};
+use crate::config::{ServerRole, TlsConfiguration};
 use crate::event::AuthResult;
 use crate::filter::{Filter, FilterInvalid};
 use crate::idm::AuthState;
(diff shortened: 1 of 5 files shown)


kanidmd/src/lib/core/https.rs
```diff
@@ -1310,17 +1311,22 @@ pub fn create_https_server(
...
-        .map_err(|e| {
-            error!(
-                "Failed to serve pkg dir {} -> {:?}",
-                env!("KANIDM_WEB_UI_PKG_PATH"),
-                e
-            );
-        })?;
+
+    // If we are no-ui, we remove this.
+    if !matches!(role, ServerRole::WriteReplicaNoUI) {
```
yaleman, line 1316: Can this be replaced with a ServerRole method? It might be premature opt but also... maybe not?

```
- if !matches!(role, ServerRole::WriteReplicaNoUI)
+ if role.should_show_ui() {
```

```
impl ServerRole {
     fn should_show_ui(&self) -> bool {
         !matches(self, WriteReplicaNoUI)
     }
 }
 ```

yaleman (commented):
Looks good, I just have a ... thought about how to do the go/nogo.

kanidmd/src/lib/core/https.rs
Firstyear: I think it's premature opt for now. When we need it, we can cross that bridge I think :)

PR #435, opened 2021-05-06
277 radius not acc pw

### LINKED ISSUE #277: Check radius pw is not main acct pw

Firstyear:
Fixes #277 this prevents the radius password, or variations of it, being used as the main account or unix credential.

### DIFF
--- a/kanidmd/src/lib/idm/server.rs
+++ b/kanidmd/src/lib/idm/server.rs
@@ -947,13 +947,54 @@ impl<'a> IdmServerProxyWriteTransaction<'a> {
         pce: &PasswordChangeEvent,
     ) -> Result<(), OperationError> {
         let account = self.target_to_account(au, &pce.target)?;
+
+        // Deny the change if the account is anonymous!
+        if account.is_anonymous() {
+            return Err(OperationError::SystemProtectedObject);
+        }
+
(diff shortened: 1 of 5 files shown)


kanidmd/src/lib/idm/server.rs
```diff
@@ -955,12 +955,16 @@ impl<'a> IdmServerProxyWriteTransaction<'a> {
         // I don't think so - because we should only be showing how STRONG the pw is ...
```
yaleman, line 955: *I* think so... enumerating passwords is possible if that's a thing :)

yaleman (changes requested):
Yeah, if the system's going to compare against an existing password and provide feedback to the user that it's invalid (even if it doesn't say that) then ... it's possible to infer a current password, without rate limits. Maybe put *that* check after the "can the user make the change" permission check?

Firstyear:
Done, it's possible now to do a partial mod application which checks everything "up until" applying to the DB. This means access, schema, plugins are run, then we check the quality, then we finish the application. This avoids the attack method you have noted here.

yaleman (approved):
Sweet, all checks out from my reading of it. Also - comments explaining the flow are much appreciated as someone Outside Williams Head trying to understand it. :)

PR #437, opened 2021-05-09
20210509 cleanup

Firstyear:
This cleans up some un-needed structures, starts to replace au with audit as a variable name, and resolves all clippy errors from 1.47.


yaleman (approved):
I like the rename of au to audit because... that was confusing previously. Cleanups good.

The test fail's just that OTP thing...

PR #440, opened 2021-05-15
Adding an example config file

yaleman:
Adds an example config file, mainly because when I'm packaging kani I want a default one ^_^

- [x] book chapter updated

### DIFF
--- a/examples/config
+++ b/examples/config
@@ -1,0 +1,4 @@
+# this should be at /etc/kanidm/config, and configures the kanidm command line tool
+uri = "https://idm.example.com"
+verify_ca = true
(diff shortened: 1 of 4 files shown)


examples/config
```diff
@@ -0,0 +1,4 @@
+# this should be at /etc/kanidm/config, and configures the kanidm command line tool
```
Firstyear, line 1: Or ~/.config/kanidm

Firstyear (commented):
Beside that one change, looks excellent :D

examples/config
yaleman: Done 😄

PR #442, opened 2021-05-18
This allows TOTP to accept an OTP that is one step behind

Firstyear:
This allows a TOTP that is one step behind to be accepted. The reason is that if someone is "just before" the edge of the totp time step window, and due to network latency or similar, they may be incorrectly rejected. It also makes the system a bit more forgiving with totp apps or password managers where you may not be able to see how long the current time window is remaining valid for. Due to the rate limiting on TOTP allowing on 3 attempts, this means each TOTP is valid for two time windows, allowing 6 attempts within 60 seconds. The odds of a correct guess are 1 in 166,667, from 1 in 333,333 in the current configuration. 

### DIFF
--- a/kanidmd/src/lib/credential/totp.rs
+++ b/kanidmd/src/lib/credential/totp.rs
@@ -166,11 +166,14 @@ impl Totp {
     }
 
     pub fn verify(&self, chal: u32, time: &Duration) -> bool {
+        let secs = time.as_secs();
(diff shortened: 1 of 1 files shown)


Firstyear:
@victorcwai Do you mind reviewing this for me when you have some time :)

victorcwai (approved):
Looks good to me :)

PR #449, opened 2021-05-21
I might have become clippy this time

yaleman:
Assuages #my_need_to_garden

Note the change to `kanidmd/src/lib/value.rs`, I'm pretty sure it's OK?

- [x] cargo fmt has been run


Firstyear:
I really should rename service princple name to security principle name since that's a better description of it. I was too busy listening to someone on the samba team at the time, and not busy enough checking my words were correct. So if you want you could do that now, or we can leave it for another pr.

yaleman:
> I really should rename service princple name to security principle name since that's a better description of it. I was too busy listening to someone on the samba team at the time, and not busy enough checking my words were correct. So if you want you could do that now, or we can leave it for another pr.

I'll do it in that spot, sure.

Firstyear (approved):

Issue #450, opened 2021-05-21
LDAP connections - log client IP on connect


yaleman:
Similar to #191 it'd be good to log which IP's connecting so we can identify it.

yaleman:
Looks like we *should* be able to identify it from here: https://github.com/kanidm/kanidm/blob/122966978555677499e5f1417e04cc5c7e7b2061/kanidmd/src/lib/core/ldaps.rs#L35

Or ... I'm thinking moving it earlier, into the acceptors - as early as we can in the connection - https://github.com/kanidm/kanidm/blob/122966978555677499e5f1417e04cc5c7e7b2061/kanidmd/src/lib/core/ldaps.rs#L101-L120

Firstyear:
i'm thinking we should move the creation of the operations auditscope to as "early as possible" in the operation here so we can collect this kind of metadata. I want to make similar changes in the https stack.

PR #452, opened 2021-05-22
Trying to log client IP for LDAP clients

yaleman:
Relates to #450 

Moved creation of `eventid` out into the acceptor which means there's a single eventid to tie the connection together for troubleshooting, instead of starting it individually after the connection is stood up. 

- [x] cargo fmt has been run

### DIFF
--- a/kanidmd/src/lib/core/ldaps.rs
+++ b/kanidmd/src/lib/core/ldaps.rs
@@ -21,18 +21,21 @@ use uuid::Uuid;
 
 struct LdapSession {
     uat: Option<LdapBoundToken>,
+    eventid: Uuid,
 }
 
 impl LdapSession {
-    fn new() -> Self {
+    fn new(eventid: &uuid::Uuid) -> Self {
         LdapSession {
             // We start un-authenticated
             uat: None,
+            eventid: *eventid,
         }
     }
 }
 
 async fn client_process<W: AsyncWrite + Unpin, R: AsyncRead + Unpin>(
+    eventid: Uuid,
     mut r: FramedRead<R, LdapCodec>,
     mut w: FramedWrite<W, LdapCodec>,
     _paddr: net::SocketAddr,
@@ -40,12 +43,11 @@ async fn client_process<W: AsyncWrite + Unpin, R: AsyncRead + Unpin>(
 ) {
     // This is a connected client session. we need to associate some state to the
     // session
-    let mut session = LdapSession::new();
+    let mut session = LdapSession::new(&eventid);
     // Now that we have the session we begin an event loop to process input OR
     // we return.
     while let Some(Ok(protomsg)) = r.next().await {
-        // Start the event
-        let eventid = Uuid::new_v4();
+        debug!("Processing LDAP eventid={:?}", session.eventid);
         let uat = session.uat.clone();
         let qs_result = qe_r_ref.handle_ldaprequest(eventid, protomsg, uat).await;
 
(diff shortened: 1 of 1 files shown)


kanidmd/src/lib/core/ldaps.rs
```diff
@@ -21,31 +21,33 @@ use uuid::Uuid;
...
+    eventid: Uuid,
 }
 
 impl LdapSession {
-    fn new() -> Self {
+    fn new(eventid: &uuid::Uuid) -> Self {
         LdapSession {
             // We start un-authenticated
             uat: None,
+            eventid: *eventid,
```
Firstyear, line 32: It would make more sense to name this session ID I think

kanidmd/src/lib/core/ldaps.rs
```diff
@@ -21,31 +21,33 @@ use uuid::Uuid;
 
 struct LdapSession {
     uat: Option<LdapBoundToken>,
+    eventid: Uuid,
 }
 
 impl LdapSession {
-    fn new() -> Self {
+    fn new(eventid: &uuid::Uuid) -> Self {
```
Firstyear, line 28: Clippy will be upset here because you are passing by reference a type that is copy ;)

Firstyear (changes requested):
I think there may be a better way here. I think you could pass in `log: Sender<AuditScope>,` to `create_ldap_server` , then you can make a small auditscope in the "accept" that sends the auditscope containing the session id + ip. Then you can just refer to the session id in future ldap events.

But the alternate is that we could put the connection IP address into the ldap session, then we adjust kanidmd/src/lib/event.rs EventOrigin to have an Entry and the current IP address / remote address of the client. We can then reuse this with https I think because we'll need to handle that (as well as X-FORWARDED-FOR).

kanidmd/src/lib/core/ldaps.rs
Firstyear: Or connection_id

Firstyear: I think I'm wondering if it's a good idea to have the eventid "stay the same" over multiple ldap operations because of how we do the batching of logs in the auditscope. I'm not sure - I can see how combinidng it to one is a good idea, but I can also see how it could be confusing too, because they are meant to be unique per-operation ids

yaleman: In context, connectionId would make more sense. Sessions tend to indicate some kind of long-lived thing in my head. (k)operationId's something that would be relevant for the individual operations that happen within the connection's lifetime.

(This'll be better articulated in the logging thesis I'm mostly-done writing)

kanidmd/src/lib/core/ldaps.rs
yaleman: Clippy was fine with it!

Firstyear: *gasp*

yaleman:
> I think there may be a better way here. I think you could pass in `log: Sender<AuditScope>,` to `create_ldap_server` , then you can make a small auditscope in the "accept" that sends the auditscope containing the session id + ip. Then you can just refer to the session id in future ldap events.
> 
> But the alternate is that we could put the connection IP address into the ldap session, then we adjust kanidmd/src/lib/event.rs EventOrigin to have an Entry and the current IP address / remote address of the client. We can then reuse this with https I think because we'll need to handle that (as well as X-FORWARDED-FOR).

Both approaches have their merits, I can't follow the code deep enough to figure out those work though...  might have to leave that to someone with a better understanding of the code. 🦀 

Happy to pull this PR in the meantime - it'd already turned from a "just output the client IP" to "so I moved things around a bunch 🥳"

PR #459, opened 2021-05-27
Add ldap vattr start

### LINKED ISSUE #439: Alias/virtual fields for LDAP searches
My use case is for LDAP, there's some fields which are queried by various clients which cause filter errors and failed queries. As discussed elsewhere this is likely to be needed for mappings on things like SAML too, so ... maybe best to be something that can be configured outside the source code, either at a config-file level or in the DB?

| queried field | kanidm field | other notes
| --- | --- | --- |
| [entrydn](https://ldapwiki.com/wiki/EntryDN) | normalised spn, maybe? | "a normalized form of the entry's DN" ([RFC5020](https://datatracker.ietf.org/doc/html/rfc5020)) - Splunk looks for this |
| email/emailAddress | mail | https://ldapwiki.com/wiki/1.2.840.113549.1.9.1 |
| firstname | ? | Not sure how to map this in a way that meets kanidm's human-centric goals |
| comment | ? | possibly just a blank response? |
| keys | ssh_publickey? | |

Firstyear:
Fixes #439 - this adds the ability to have field aliases. Currently these are hardcoded. In the future, we could be able to use the system schema to extract/cache what these maps are, as the current design allows the source of the mappings to be changed from the ldap_vattr_map() function to something else. 


kanidmd/src/lib/entry.rs
```diff
@@ -1462,28 +1462,72 @@ impl Entry<EntryReduced, EntryCommitted> {
...
+        let attr_map = attr_map?;
+        // Stage 2 - transform and get all our attr - names out that we need to return.
+        //                  ldap a, kani a
+        let attr_names: Vec<(&str, &str)> = if all_attrs {
+            // Join the set of attr keys, and our requested attrs.
+            self.attrs
+                .keys()
+                .map(|k| (k.as_str(), k.as_str()))
+                .chain(
+                    l_attrs
```
Firstyear, line 1500: This means that every search you always get keys AND ssh_publickeys back. Should we not generate these virtual attrs in a default search?  And only use them when requested?

yaleman (commented):
The rest looks sane an explicable from a development/extension perspective too.

kanidmd/src/lib/entry.rs
Firstyear: In the update I think I have a reasonable compromise. If you req '*' you just get kanidm attrs, if you req '+' you get all kanidm attrs + all mapped attributes so you can see what mappings exist. And then you can request "what you like".

Firstyear:
@yaleman Did you want to test this with your application that needed the compat mappings? Or want me to merge and we'll fix any missing maps later in seperate tickets? It should be easier now to add this.

yaleman:
I'll give it a quick test and see if there's any major weirdness :)

yaleman:
Splunk queries groups and memberships OK, I haven't worked out how to set `mail` for my user so I can't test auth - but that's definitely a big improvement :) Thanks!

Firstyear:
> Splunk queries groups and memberships OK, I haven't worked out how to set `mail` for my user so I can't test auth - but that's definitely a big improvement :) Thanks!

For now there is no way to set mail, because I haven't written a mail syntax parser for storing it correctly in the server. I also wanted to make sure primary/non-primary mail is handled too, and that may involve ldap tags. So for now, you could map spn to mail instead as a hack? Can you configure that in splunk? Or would we need to do that here?

yaleman:
> > Splunk queries groups and memberships OK, I haven't worked out how to set `mail` for my user so I can't test auth - but that's definitely a big improvement :) Thanks!
> 
> For now there is no way to set mail, because I haven't written a mail syntax parser for storing it correctly in the server. I also wanted to make sure primary/non-primary mail is handled too, and that may involve ldap tags. So for now, you could map spn to mail instead as a hack? Can you configure that in splunk? Or would we need to do that here?

I *could* map the fields but that'd break a whole load of other things due to it being used for sending notifications etc. I'll avoid it for now. :)

Issue #414, opened 2021-04-16
automatic clearing of stale credentials from kanidm_tokens


yaleman:
`~/.cache/kanidm_tokens` gets full of all the used tokens, is there an auto-cleanup task, or could there be one made?

Two reasons:

 1. security - it's not good to list what accounts have been used on this machine by this account - `user_a` has access to `idm_admin` - rad, target them!
 2. it's neater. uWu.

Firstyear:
Yep, agreed. This needs to happen. I think to achieve this I need to make some changes to how sessions are current managed. Today they are encrypted with fernet, but that makes them opaque. I think to do this nicely, I need to swap to signing (ie ecdsa or ed25519) and then we can expose an expiry field so that the client also knows what time the token becomes invalid. That will allow them to be cleaned up.

PR #460, opened 2021-05-27
Fix unixd token expiry issue

Firstyear:
After #414 we added a second error code for session expiry to indicate to clients why they were receiving 401 error codes. Due to an over-zealous pattern match in unixd, this caused the generic handler to be taken rather than forcing the re-authentication path that is required. This fixs that so that any 401 code will always force a re-auth, and has a better generic handler. 


Firstyear:
Sorry for the sudden merge, it's affecting some systems and I want to do the build asap.

Issue #120, opened 2019-10-11
Have a split admin read and admin write, affected by claims (on both)


Firstyear:
With AzureAD as inspiration we should have:

* admin_read as a claim inherited privilege
* once you take admin_read then you can elevate to admin_write if needed.

Firstyear:
I think this can be closed since we already have this via the privilege levels now, and other claim based access schemes are a future goal.

PR #462, opened 2021-05-30
64 120 session claims

### LINKED ISSUE #64: Support session claims
Claims are an important modern security concept, allowing ephemeral, or scoped privileges for each session of authentication or token. Examples are that you may have a claim for interactive logins that allows changing your own displayname, but your "email" token does not have this claim so can not change your name. Another example is the github style "sudo" mode which is a time limited claim allowing modification of passwords or 2fa tokens. 

This means claims should:

* Be inherited from groups to "allow" access of the claim
* The claim is "associated" to a credential, IE interactive pw + mfa, or a api password type.
* The claim is added to the userAuthToken
* The claim is processed into the current Entry during an operation for auth filters to be applied.
* The claim associated on the credential must be indexed/searchable to allow memberof style removal if you are removed from the group.

Firstyear:
Fixes #64 #120 this adds the ability to extract claims and add extra detail into our user auth tokens.

I'm not sure I like the claims-in-access controls right now. Right now as these claims could be session-dependent we don't have a way to "signal back" why an operation failed. For example, if the user has multiple credentials,, they would get an error that they can't perform a modification but no reason to *why*. So I think that for now, I want to leave most of this infrastructure in place but just not extract claims into the local entry as it creates ambiguity. 

Examples: 

Today one of the claims is authlevel_strong/authclass_mfa. But these would be better as credential policies for groups/members instead. Another is which type of credential that was used, but instead we could use a sudo-like feature in the server instead. Another could be "what's your source IP" but that could be a rule in the credential/session that defines that it can only be used from within some IP range, or similar. Alternately that could be an actual acp rule.

So for now,  Ithink claims will probably end up as a oauth-only thing, and we won't use them internally. 


victorcwai:
So currently the uat (with expiry time) contains an `AuthType`, which could be used for "claims" in the future. Is that right?

> Right now as these claims could be session-dependent we don't have a way to "signal back" why an operation failed. 

Not sure I understand this part, how come we can't signal back when an operation failed?

> So I think that for now, I want to leave most of this infrastructure in place but just not extract claims into the local entry as it creates ambiguity.

By local entry, do you mean the entry at the client side?

Firstyear:
> So currently the uat (with expiry time) contains an `AuthType`, which could be used for "claims" in the future. Is that right?

Well, claims are "pieces of information in the token" that the token holder can use to make decisions about or inspect. I did consider using these "inside the server" for making access decisions, but for the reasons below, I think it's not a good idea - instead we can just have claims in the token so clients can make their own decisions and present information about the session etc. 

> 
> > Right now as these claims could be session-dependent we don't have a way to "signal back" why an operation failed.
> 
> Not sure I understand this part, how come we can't signal back when an operation failed?
> 
> > So I think that for now, I want to leave most of this infrastructure in place but just not extract claims into the local entry as it creates ambiguity.
> 
> By local entry, do you mean the entry at the client side?

These two are kind of related. This involves a lot in how access controls work. When you go to perform a search or modify we need to make decisions about "who you are" and "do you have this access". Today this is done in https://github.com/kanidm/kanidm/blob/master/kanidmd/src/lib/access.rs which is a pretty intense piece of code. 

But there are at it's core two things - a filter (query) that expresses "what entries does this access control apply this operation to" and a second filter that expresses "who is granted this access". In access, these are the https://github.com/kanidm/kanidm/blob/master/kanidmd/src/lib/access.rs#L293 receiver (who inherits this access) and targetscope (what is affected by this access). 

So for receiver to work, we need a fully-inflated copy of your entry so we can apply this filter. This is why we have UserAuthToken to Entry code, where we search your entry, and then we associate your Entry to an Identity, and that Identity is referenced in the Event. This is what I mean by the "local entry", rather the Entry that is bound to the Identity of the current UserAuthToken so that we can make decisions about which access controls apply. 

Now the second part is about how we can't "signal back". Let's assume we did inject these claims to the Entry, and that allows access controls to make decisions about who does or does not receive this permission. 

In the non-claim case there are really two paths - you either match the access control (have the right) or you do not. When you do not have access, we signal a "permission denied" - we don't say "oh you could have done this IF you had X, Y or Z access controls applied". We just say "permission denied". 

Now lets assume we add in claims here. You would add some user confusion about "why can I do operation X sometimes, but not other times?". This is because the different sessions, may have different claims, and the filter has gone from matching to not matching. But remember, we can't say "oh you would have matched, but only if X, Y, Z, and that would have granted you the access". And really that's what we need here is a way to say "look, normally you *could* do this thing, but right now there is something dynamic that you need to change". The current access code really can't handle that scenario, so having something like claims in our access controls could really confuse users here. 

So for now, Im leaving the infrastructure there, because we COULD fix this in the future, but I think we need to consider other rules about auth policy etc which may be better placed. For example, why have a "claim" of authtype_mfa in an access control, when we could be better to enforce that to be a member of some group, you must fufil certain password requirements instead. This is clearer for the user ( you have the right or don't), clearer for the admin, and is easier for feedback. 

Does that help explain it better?

victorcwai (approved):
Thanks for the explanation. Ok this looks good to me.

Issue #461, opened 2021-05-29
Mail attribute handling


Firstyear:
We need to be able to set mail attributes, but we also need a way to tag which mail addr is primary, display those tags in ldap, and have correct parsing of mail addrs  https://www.mailboxvalidator.com/resources/articles/acceptable-email-address-syntax-rfc/

Firstyear:
Relates #459

yaleman:
This could be handy for validation - https://docs.rs/validator/0.13.0/validator/fn.validate_email.html

> Validates whether the given string is an email based on the [HTML5 spec](https://html.spec.whatwg.org/multipage/forms.html#valid-e-mail-address). [RFC 5322](https://tools.ietf.org/html/rfc5322) is not practical in most circumstances and allows email addresses that are unfamiliar to most users.

I agree with the comment on the RFC - truly RFC-compliant emails are horrible things and unlikely to be used except in trolling... I'm not sure how often ones outside the HTML5 spec are actually used...

Firstyear:
@yaleman Can I resolve this? We have email address handling mostly in place now.

yaleman:
Yep, I think that one's 
 sorted. I was able to set up email for OAUTH2 use OK.

PR #465, opened 2021-06-03
Add email syntax

Firstyear:
Part one of #461 - this adds the syntax to support email addresses and validation of their content, and a method to serialise to the DB that can be extended with attribute tagging in the future. Part two will address administration of these values. 

### DIFF
--- a/kanidm_client/src/asynchronous.rs
+++ b/kanidm_client/src/asynchronous.rs
@@ -835,7 +835,7 @@ impl KanidmAsyncClient {
         attr: &str,
         values: &[&str],
     ) -> Result<bool, ClientError> {
-        let m: Vec<_> = values.iter().map(|v| (*v).to_string()).collect();
+        let msg: Vec<_> = values.iter().map(|v| (*v).to_string()).collect();
         self.perform_post_request(format!("/v1/account/{}/_attr/{}", id, attr).as_str(), m)
             .await
     }
--- a/kanidm_client/src/asynchronous.rs
+++ b/kanidm_client/src/asynchronous.rs
@@ -836,7 +836,7 @@ impl KanidmAsyncClient {
         values: &[&str],
     ) -> Result<bool, ClientError> {
         let msg: Vec<_> = values.iter().map(|v| (*v).to_string()).collect();
-        self.perform_post_request(format!("/v1/account/{}/_attr/{}", id, attr).as_str(), m)
+        self.perform_post_request(format!("/v1/account/{}/_attr/{}", id, attr).as_str(), msg)
             .await
     }
 
--- a/kanidm_client/src/asynchronous.rs
+++ b/kanidm_client/src/asynchronous.rs
@@ -829,6 +829,17 @@ impl KanidmAsyncClient {
             .await
     }
 
+    pub async fn idm_account_add_attr(
+        &self,
+        id: &str,
+        attr: &str,
+        values: &[&str],
+    ) -> Result<bool, ClientError> {
+        let m: Vec<_> = values.iter().map(|v| (*v).to_string()).collect();
+        self.perform_post_request(format!("/v1/account/{}/_attr/{}", id, attr).as_str(), m)
+            .await
+    }
+
     pub async fn idm_account_set_attr(
         &self,
         id: &str,
--- a/kanidm_client/src/lib.rs
+++ b/kanidm_client/src/lib.rs
@@ -571,6 +571,15 @@ impl KanidmClient {
         tokio_block_on(self.asclient.idm_account_purge_attr(id, attr))
     }
 
+    pub fn idm_account_add_attr(
+        &self,
+        id: &str,
+        attr: &str,
+        values: &[&str],
+    ) -> Result<bool, ClientError> {
+        tokio_block_on(self.asclient.idm_account_add_attr(id, attr, values))
+    }
+
     pub fn idm_account_set_attr(
         &self,
         id: &str,
--- a/kanidm_client/tests/default_entries.rs
+++ b/kanidm_client/tests/default_entries.rs
@@ -99,7 +99,7 @@ fn is_attr_writable(rsclient: &KanidmClient, id: &str, attr: &str) -> Option<boo
         entry => {
             let new_value = match entry {
                 "mail" => format!("{}@example.com", id),
-                "acp_receiver" => "{\"eq\":[\"memberof\",\"00000000-0000-0000-0000-000000000011\"]}".to_string(),
+                "acp_receiver" => r#"{"eq":["memberof","00000000-0000-0000-0000-000000000011"]}"#.to_string(),
                 "acp_targetscope" => "{\"and\": [{\"eq\": [\"class\",\"access_control_profile\"]}, {\"andnot\": {\"or\": [{\"eq\": [\"class\", \"tombstone\"]}, {\"eq\": [\"class\", \"recycled\"]}]}}]}".to_string(),
                  _ => id.to_string(),
             };
--- a/kanidm_client/tests/default_entries.rs
+++ b/kanidm_client/tests/default_entries.rs
@@ -98,6 +98,7 @@ fn is_attr_writable(rsclient: &KanidmClient, id: &str, attr: &str) -> Option<boo
         ),
         entry => {
             let new_value = match entry {
+                "mail" => format!("{}@example.com", id),
                 "acp_receiver" => "{\"eq\":[\"memberof\",\"00000000-0000-0000-0000-000000000011\"]}".to_string(),
                 "acp_targetscope" => "{\"and\": [{\"eq\": [\"class\",\"access_control_profile\"]}, {\"andnot\": {\"or\": [{\"eq\": [\"class\", \"tombstone\"]}, {\"eq\": [\"class\", \"recycled\"]}]}}]}".to_string(),
                  _ => id.to_string(),
--- a/kanidmd/src/lib/be/dbvalue.rs
+++ b/kanidmd/src/lib/be/dbvalue.rs
@@ -81,6 +81,11 @@ pub struct DbValueTaggedStringV1 {
 }
 
 #[derive(Serialize, Deserialize, Debug)]
+pub struct DbValueEmailAddressV1 {
+    pub d: String,
+}
+
+#[derive(Serialize, Deserialize, Debug)]
 pub enum DbValueV1 {
     U8(String),
     I8(String),
@@ -99,6 +104,7 @@ pub enum DbValueV1 {
     CI(DbCidV1),
     NU(String),
     DT(String),
+    EM(DbValueEmailAddressV1),
 }
 
 #[cfg(test)]
--- a/kanidmd/src/lib/constants/schema.rs
+++ b/kanidmd/src/lib/constants/schema.rs
@@ -561,7 +561,8 @@ pub const JSON_SCHEMA_CLASS_ACCOUNT: &str = r#"
         "ssh_publickey",
         "radius_secret",
         "account_expire",
-        "account_valid_from"
+        "account_valid_from",
+        "mail"
       ],
       "systemmust": [
         "displayname",
--- a/kanidmd/src/lib/constants/schema.rs
+++ b/kanidmd/src/lib/constants/schema.rs
@@ -56,7 +56,7 @@ pub const JSON_SCHEMA_ATTR_MAIL: &str = r#"
         "mail"
       ],
       "syntax": [
-        "UTF8STRING"
+        "EMAIL_ADDRESS"
       ],
       "uuid": [
         "00000000-0000-0000-0000-ffff00000041"
--- a/kanidmd/src/lib/schema.rs
+++ b/kanidmd/src/lib/schema.rs
@@ -200,6 +200,7 @@ impl SchemaAttribute {
             SyntaxType::Cid => v.is_cid(),
             SyntaxType::NsUniqueId => v.is_nsuniqueid(),
             SyntaxType::DateTime => v.is_datetime(),
+            SyntaxType::EmailAddress => v.is_email_address(),
         };
         if r {
             Ok(())
@@ -378,6 +379,15 @@ impl SchemaAttribute {
             SyntaxType::DateTime => ava.iter().fold(Ok(()), |acc, v| {
                 acc.and_then(|_| {
                     if v.is_datetime() {
+                        Ok(())
+                    } else {
+                        Err(SchemaError::InvalidAttributeSyntax(a.to_string()))
+                    }
+                })
+            }),
+            SyntaxType::EmailAddress => ava.iter().fold(Ok(()), |acc, v| {
+                acc.and_then(|_| {
+                    if v.is_email_address() {
                         Ok(())
                     } else {
                         Err(SchemaError::InvalidAttributeSyntax(a.to_string()))
@@ -1206,7 +1216,7 @@ impl<'a> SchemaWriteTransaction<'a> {
                     unique: false,
                     phantom: true,
                     index: vec![],
-                    syntax: SyntaxType::UTF8STRING,
+                    syntax: SyntaxType::EmailAddress,
                 },
             );
             self.attributes.insert(
@@ -1219,7 +1229,7 @@ impl<'a> SchemaWriteTransaction<'a> {
                     unique: false,
                     phantom: true,
                     index: vec![],
-                    syntax: SyntaxType::UTF8STRING,
+                    syntax: SyntaxType::EmailAddress,
                 },
             );
             self.attributes.insert(
--- a/kanidmd/src/lib/schema.rs
+++ b/kanidmd/src/lib/schema.rs
@@ -230,165 +230,30 @@ impl SchemaAttribute {
             return Err(SchemaError::InvalidAttributeSyntax(a.to_string()));
         };
         // If syntax, check the type is correct
-        match self.syntax {
-            SyntaxType::Boolean => ava.iter().fold(Ok(()), |acc, v| {
-                acc.and_then(|_| {
-                    if v.is_bool() {
-                        Ok(())
-                    } else {
-                        Err(SchemaError::InvalidAttributeSyntax(a.to_string()))
-                    }
-                })
-            }),
-            SyntaxType::SYNTAX_ID => ava.iter().fold(Ok(()), |acc, v| {
-                acc.and_then(|_| {
-                    if v.is_syntax() {
-                        Ok(())
-                    } else {
-                        Err(SchemaError::InvalidAttributeSyntax(a.to_string()))
-                    }
-                })
-            }),
-            SyntaxType::Uuid => ava.iter().fold(Ok(()), |acc, v| {
-                acc.and_then(|_| {
-                    if v.is_uuid() {
-                        Ok(())
-                    } else {
-                        Err(SchemaError::InvalidAttributeSyntax(a.to_string()))
-                    }
-                })
-            }),
-            // This is the same as a UUID, refint is a plugin
-            SyntaxType::REFERENCE_UUID => ava.iter().fold(Ok(()), |acc, v| {
-                acc.and_then(|_| {
-                    if v.is_refer() {
-                        Ok(())
-                    } else {
-                        Err(SchemaError::InvalidAttributeSyntax(a.to_string()))
-                    }
-                })
-            }),
-            SyntaxType::INDEX_ID => ava.iter().fold(Ok(()), |acc, v| {
-                acc.and_then(|_| {
-                    if v.is_index() {
-                        Ok(())
-                    } else {
-                        Err(SchemaError::InvalidAttributeSyntax(a.to_string()))
-                    }
-                })
-            }),
-            SyntaxType::Utf8StringInsensitive => ava.iter().fold(Ok(()), |acc, v| {
-                acc.and_then(|_| {
-                    if v.is_insensitive_utf8() {
-                        Ok(())
-                    } else {
-                        Err(SchemaError::InvalidAttributeSyntax(a.to_string()))
-                    }
-                })
-            }),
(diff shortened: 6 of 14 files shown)


kanidmd/src/lib/value.rs
```diff
@@ -237,6 +243,7 @@ impl fmt::Display for SyntaxType {
                 SyntaxType::Cid => "CID",
                 SyntaxType::NsUniqueId => "NSUNIQUEID",
                 SyntaxType::DateTime => "DATETIME",
+                SyntaxType::EmailAddress => "EMAIL_ADDRESS",
```
QnnOkabayashi, line 246: Since it's only writing a `&str` we should just use `f.write_str(match self { ... })` to clear up lines 226-227.

kanidmd/src/lib/value.rs
```diff
@@ -281,6 +288,7 @@ pub enum PartialValue {
     Cid(Cid),
     Nsuniqueid(String),
     DateTime(OffsetDateTime),
+    EmailAddress(String),
```
QnnOkabayashi, line 291: How come everything is a `String`? Do we not have more explicit types for things? Why not something like:

```suggestion
    EmailAddress {
        local: String,
        domain: String,
    },
```

kanidmd/src/lib/value.rs
```diff
@@ -1101,6 +1129,10 @@ impl Value {
             DbValueV1::DT(s) => PartialValue::new_datetime_s(&s)
                 .ok_or(())
                 .map(|pv| Value { pv, data: None }),
+            DbValueV1::EM(DbValueEmailAddressV1 { d }) => Ok(Value {
```
QnnOkabayashi, line 1132: What does the `DbValueV1` represent exactly? And what are `PartialValue`s? Also I think we should make make two constructors for `Value`, since the `data` field seems to be `None` most of the time. So we could have constructors like `Value::dataless` that only takes the `pv` field, and `Value::with_data` that takes both fields.

kanidmd/src/lib/value.rs
```diff
@@ -1590,6 +1628,22 @@ mod tests {
...
     }
 
+    #[test]
+    fn test_value_email_address() {
+        // https://html.spec.whatwg.org/multipage/forms.html#valid-e-mail-address
+        let val1 = Value::new_email_address_s("william@blackhats.net.au");
+        let val2 = Value::new_email_address_s("alice@idm.example.com");
+        let val3 = Value::new_email_address_s("test+mailbox@foo.com");
+        let inv1 = Value::new_email_address_s("william");
+        let inv2 = Value::new_email_address_s("test~uuid");
```
QnnOkabayashi, line 1638: What is the `_s` suffix for? Also we can use [Django's email tests](https://github.com/django/django/blob/main/tests/validators/tests.py#L48), which is what the source code of validator uses for testing.

kanidmd/src/lib/be/dbvalue.rs
```diff
@@ -80,6 +80,11 @@ pub struct DbValueTaggedStringV1 {
     pub d: String,
 }
 
+#[derive(Serialize, Deserialize, Debug)]
+pub struct DbValueEmailAddressV1 {
```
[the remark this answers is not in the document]
Firstyear: Database encoding format 1. This way if we have some horrendous issues, we make a "v2" struct, then we still have the ability to read V1 entries from non-upgraded DB's, and we re-serialised them out to v2 in the future. It's about future proofing :)

kanidm_client/tests/default_entries.rs
```diff
@@ -98,6 +98,7 @@ fn is_attr_writable(rsclient: &KanidmClient, id: &str, attr: &str) -> Option<boo
         ),
         entry => {
             let new_value = match entry {
+                "mail" => format!("{}@example.com", id),
                 "acp_receiver" => "{\"eq\":[\"memberof\",\"00000000-0000-0000-0000-000000000011\"]}".to_string(),
```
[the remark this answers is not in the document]
Firstyear: Yeah, I agree with this.

kanidmd/src/lib/be/dbvalue.rs
```diff
@@ -80,6 +80,11 @@ pub struct DbValueTaggedStringV1 {
...
+    pub d: String,
```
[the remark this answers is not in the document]
Firstyear: The intent is later to extra metadata. Remember, this type of dbvalue is "how it will be encoded into the db" so we minimise the struct field names because they are serialised and this reduces our IO in/out. So we have the "email" in its raw-serialised form, and the higher level Value type is out "rich" type that *could* have an EmailAddr type, if one existed. 

Saying this, the Value type needs a rewrite - like all code, it has some techdebt and poor choice :)

kanidm_client/tests/default_entries.rs
```diff
@@ -98,6 +98,7 @@ fn is_attr_writable(rsclient: &KanidmClient, id: &str, attr: &str) -> Option<boo
         ),
         entry => {
             let new_value = match entry {
+                "mail" => format!("{}@example.com", id),
```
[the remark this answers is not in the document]
Firstyear: This is a test case where we are changing/adapting the value of the attribute based on well ... which attribute it is. Because we send the attribute to the server as a string, the str match makes sense. it *could* be an enum, but that enum would need to encapsulate all possible attributes that the server has in it's schema - and of course that schema adapts and grows over time. Which is where the problems start.

So actually, by being strings, we are "close to" the experience tthat the current version of the api has for users, and it also promotes fragility in the tests - we want to be sensitive and fail fast, which in this situation, is exactly what happened to lead me to creating a value for this.

Also the hardcoded example.com doesn't matter - it's a test case :)

kanidmd/src/lib/constants/schema.rs
```diff
@@ -56,7 +56,7 @@ pub const JSON_SCHEMA_ATTR_MAIL: &str = r#"
         "mail"
       ],
       "syntax": [
-        "UTF8STRING"
+        "EMAIL_ADDRESS"
```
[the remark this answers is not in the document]
Firstyear: So the structure of schema is a whole thing. We store our schema in the database, so that in the future it can be replicated. Schema is something that admins can extend and customise, and having it in the database means we can make it consistent over the set of nodes in the topology.

Now of course, this has a problem. We have schema in the database - but we need schema to *query* that database in the first place.

So in schema.rs, we load an inmemory schema that is "just enough" to then process the database in an efficient manner (ie hit indexes) and then it is reloaded from the db/disk.

So these json structures are the DB schema elements, and we run them in a "migration stage" during server create and bootstrap. So these are used in the "initialise" functions in server.rs to load these into the database, and tthen we can use these for future migrations etc.

kanidmd/src/lib/schema.rs
```diff
@@ -384,6 +385,15 @@ impl SchemaAttribute {
...
             }),
+            SyntaxType::EmailAddress => ava.iter().fold(Ok(()), |acc, v| {
+                acc.and_then(|_| {
+                    if v.is_email_address() {
+                        Ok(())
+                    } else {
+                        Err(SchemaError::InvalidAttributeSyntax(a.to_string()))
+                    }
+                })
+            }),
```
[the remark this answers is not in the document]
Firstyear: William of the past is not always the best programmer, so probably, this code is trash and could be better. I would like to commit this and give it a crack. Especially if all shortcircuits.

kanidmd/src/lib/value.rs
```diff
@@ -210,6 +215,7 @@ impl SyntaxType {
             SyntaxType::Utf8StringIname => 14,
             SyntaxType::NsUniqueId => 15,
             SyntaxType::DateTime => 16,
+            SyntaxType::EmailAddress => 17,
```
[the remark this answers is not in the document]
Firstyear: Yes, there is a reason to go back and forth. Do I remember why? No, not at all. 

I think I hardcoded the usize here so that if the fields are re-arranged, we keep the values stable. More investigation needed?

kanidmd/src/lib/value.rs
Firstyear: That's an awesome suggestion :)

kanidmd/src/lib/value.rs
Firstyear: Because I am not always aware if there are better types. Value.rs does need to be rewritter because I didn't know how to make borrow as a trait work in the past between value and partialvalue. But because value is fully in memory, we can and freely change these types if we find something better in the future.

kanidmd/src/lib/value.rs
Firstyear: As before, db varue is how this is serialised down. Value is in memory.

kanidmd/src/lib/value.rs
Firstyear: takes a str ref, rather than a string. Again, something something past william, poor choices, something something.

kanidmd/src/lib/value.rs
yaleman: What's the functional benefit of the local/domain breakdown for an email address enum? They're essentially arbitrary strings with some validation - like usernames/phone numbers/physical addresses.

yaleman (approved):
Rad. I like me sum emailz.

kanidmd/src/lib/value.rs
Firstyear: That's a good point too yaleman. We do validate the content as it comes in so we know the string data is correct too. Saying this, database corruption could be a thing, but even making this a stricter type won't protect us from that ...

kanidmd/src/lib/value.rs
Firstyear: Given I've mentioned it a few times, opening https://github.com/kanidm/kanidm/issues/468 to address it.

kanidmd/src/lib/schema.rs
```diff
@@ -229,161 +230,30 @@ impl SchemaAttribute {
...
-            SyntaxType::DateTime => ava.iter().fold(Ok(()), |acc, v| {
-                acc.and_then(|_| {
-                    if v.is_datetime() {
-                        Ok(())
-                    } else {
-                        Err(SchemaError::InvalidAttributeSyntax(a.to_string()))
-                    }
-                })
-            }),
+        let valid = match self.syntax {
```
Firstyear, line 233: @QnnOkabayashi Thanks to your inspiration, this section is MUCH cleaner!

kanidmd/src/lib/value.rs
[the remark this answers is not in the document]
QnnOkabayashi: Is the `SyntaxType` used to describe a type of message being sent back and forth, I have a hard time imagining we would want to change it back and forth on the same machine. If this is the case, couldn't we use some [`serde` macros](https://serde.rs/variant-attrs.html) on the `SyntaxType` variants to tell each one to serialize to a unique `usize`? This would let us have associated `usize`s in the definition, instead of having different functions containing massive `match` statements.

Something like this:
```rust
pub enum SyntaxType {
    #[serde(rename = "0")]
    UTF8STRING,
    #[serde(rename = "1")]
    Utf8StringInsensitive,
    ...
    #[serde(rename = "16")]
    DateTime,
    #[serde(rename = "17")]
    EmailAddress,
}
```

I'm not sure if it will allow serialization to `usize`s though, or only strings.

kanidmd/src/lib/value.rs
QnnOkabayashi: My thinking was that although all email addresses are `String`s, the opposite isn't necessarily true. This leaves anyone looking at the code to assume that the `String` field is a valid email address, which I wasn't certain of until William's comment. Also, it's odd that you can unpack the `EmailAddress` variant and get a `String`, and then it can do anything that a `String` can do that an email address shouldn't be able to do, like join it with another `String`.

But if we had an explicit `EmailAddr` type, we could give it special email address functionality (not sure exactly what this would look like though) and make it generally more robust. Thus, the variant would look like:

```suggestion
    EmailAddress(EmailAddr)
```

Where `EmailAddr` is declared as
```rust
struct EmailAddr {
    local: String,
    domain: String,
}
```

This could probably be used in many other places as well. But if we tread carefully and don't use this in a lot of places, you're both right that it won't make much difference at all.

kanidmd/src/lib/value.rs
QnnOkabayashi: Ah makes sense. What are your thoughts on the two `Value` initializers I proposed?

Also, I recall you saying somewhere that we used short variable names like `d` to reduce packet size when sending information. Can't we still use explicit, longer names and use `serde` macros the serialize them into shorter variants?

kanidmd/src/lib/value.rs
```diff
@@ -210,35 +215,33 @@ impl SyntaxType {
...
-                SyntaxType::RadiusUtf8String => "RADIUS_UTF8STRING",
-                SyntaxType::SshKey => "SSHKEY",
-                SyntaxType::SecurityPrincipalName => "SECURITY_PRINCIPAL_NAME",
-                SyntaxType::UINT32 => "UINT32",
-                SyntaxType::Cid => "CID",
-                SyntaxType::NsUniqueId => "NSUNIQUEID",
-                SyntaxType::DateTime => "DATETIME",
-            }
-        )
+        f.write_str(match self {
```
QnnOkabayashi, line 225: Yes! Much better :)

kanidmd/src/lib/value.rs
```diff
@@ -1349,7 +1394,8 @@ impl Value {
             PartialValue::Utf8(s)
             | PartialValue::Iutf8(s)
             | PartialValue::Iname(s)
-            | PartialValue::Nsuniqueid(s) => vec![s.clone()],
+            | PartialValue::Nsuniqueid(s)
+            | PartialValue::EmailAddress(s) => vec![s.clone()],
             PartialValue::Refer(u) | PartialValue::Uuid(u) => {
```
QnnOkabayashi, line 1399: Did this pass `cargo fmt`?

kanidm_client/src/asynchronous.rs
```diff
@@ -829,6 +829,17 @@ impl KanidmAsyncClient {
             .await
     }
 
+    pub async fn idm_account_add_attr(
+        &self,
+        id: &str,
+        attr: &str,
+        values: &[&str],
+    ) -> Result<bool, ClientError> {
+        let m: Vec<_> = values.iter().map(|v| (*v).to_string()).collect();
```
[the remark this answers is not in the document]
Firstyear: Changed m to msg. 

Yes the Vec<_> hint is needed. Because self.perform_post_request takes `R: Serialise`, without the hint here .collect doesnt know what to collect into. I could use the turbofish syntax, but honestly, I always forget it. I think it's .collect::Vec<_>() but I think that's a bit less clear than `let msg: Vec<_> = ... `

kanidm_client/src/lib.rs
```diff
@@ -571,6 +571,15 @@ impl KanidmClient {
         tokio_block_on(self.asclient.idm_account_purge_attr(id, attr))
     }
 
+    pub fn idm_account_add_attr(
+        &self,
+        id: &str,
+        attr: &str,
+        values: &[&str],
+    ) -> Result<bool, ClientError> {
```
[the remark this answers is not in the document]
Firstyear: It represents past williams mistakes. I thought that there would be a case where you could set and it would be Ok() but change nothing (false), but I think we never indicate that so it should be cleaned up. That could be a new issue from this (cleanup kanidm_client bool/return values).

kanidm_client/tests/default_entries.rs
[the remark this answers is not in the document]
Firstyear: Tests should be fragile - they should fail fast when something - anything - changes. We don't want robust tests that self-heal, and fix things magically, or accidentally work when they shouldn't. We want them to be fragile and sensitive so they fail in slight changes so that we can understand and react to these in development.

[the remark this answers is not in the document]
Firstyear: Ohh and I mean the JSON HTTP api version here with v1

kanidmd/src/lib/schema.rs
```diff
@@ -229,161 +230,30 @@ impl SchemaAttribute {
...
+            SyntaxType::UINT32 => ava.iter().all(Value::is_uint32),
+            SyntaxType::Cid => ava.iter().all(Value::is_cid),
+            SyntaxType::NsUniqueId => ava.iter().all(Value::is_nsuniqueid),
+            SyntaxType::DateTime => ava.iter().all(Value::is_datetime),
+            SyntaxType::EmailAddress => ava.iter().all(Value::is_email_address),
+        };
+        if valid {
+            Ok(())
+        } else {
+            Err(SchemaError::InvalidAttributeSyntax(a.to_string()))
```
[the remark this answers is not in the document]
Firstyear: The issue we are stuck on 1.47 due to a requirement at SUSE. I'm working on something to fix this though.

kanidmd/src/lib/value.rs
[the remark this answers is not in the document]
Firstyear: I think this is a case for a seperate issue to investigate and clean up to be sure of what we are doing.

kanidmd/src/lib/value.rs
Firstyear: Well, remember this is also an inmemory representation. So we can change it to local/domain when and if we find a need for that. Today, we don't have that need, so what benefit does it bring? Remember a key part of this design is that we split the DB (stable, can't change format) from the in memory (flexible, can change, adapt over time), so if we find out tomorrow we need this split, we can doit immediately :) 

So for now, I think unless we have a need to do this, we should leave it. And when we do have the need for it, we can easily retro fit this in.

kanidmd/src/lib/value.rs
Firstyear: I think the value thing is actually going to come up in https://github.com/kanidm/kanidm/issues/468 so let's move that talk there. And second we can use the longer names + serde macros, but again - william of the past didn't know this existed at the time. So my advice is open a new issue and just fix it. I know it's really tempting to want to clean up everything we see immediately in one code review, but we have to also consider a focus of this issue. It can quickly go from "add mail attributes" to "rewrite everything". So I'm not opposed to this change, but please open an issue so we can focus on it appropriately there :)

kanidmd/src/lib/value.rs
Firstyear: I'm on a different version of rust 1.47 so it would have been formatted, but may have a different fmt version.

kanidm_client/src/asynchronous.rs
```diff
@@ -835,7 +835,7 @@ impl KanidmAsyncClient {
         attr: &str,
         values: &[&str],
     ) -> Result<bool, ClientError> {
-        let m: Vec<_> = values.iter().map(|v| (*v).to_string()).collect();
+        let msg: Vec<_> = values.iter().map(|v| (*v).to_string()).collect();
         self.perform_post_request(format!("/v1/account/{}/_attr/{}", id, attr).as_str(), m)
```
QnnOkabayashi:
```suggestion
        let msg: Vec<String> = values.iter().map(str::to_string).collect();
        self.perform_post_request(format!("/v1/account/{}/_attr/{}", id, attr).as_str(), msg)
```

QnnOkabayashi (changes requested):

kanidmd/src/lib/schema.rs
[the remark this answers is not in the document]
QnnOkabayashi: One last thing, would your suggestion [here](https://github.com/kanidm/kanidm/issues/468#issuecomment-855210520) for storing variants at a higher level deprecate this function?

kanidm_client/src/lib.rs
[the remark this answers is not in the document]
QnnOkabayashi: If I wanted to work on this, should I open an issue, create my own branch, and then submit a PR for it? Also, there are a bunch of changes that you made here on this branch, how can I get access to them so I don't have to deal with a lot of merge conflicts?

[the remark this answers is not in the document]
Firstyear: Well, just open the issue first, then once this is merged *then* you create your branch and work from there :)

kanidmd/src/lib/schema.rs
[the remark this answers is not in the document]
Firstyear: I'm not sure. Research would be needed. Can you comment that in #468 if this function would still be needed?

kanidmd/src/lib/value.rs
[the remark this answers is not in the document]
Firstyear: @QnnOkabayashi did you want to open this issue?

kanidmd/src/lib/value.rs
Firstyear: @QnnOkabayashi Can you open the issue fro dbvalue serde renames?

kanidm_client/src/asynchronous.rs
Firstyear: It's not letting me apply this, but it's an easy manual change I'll just push it.

Firstyear: Ahh it doesn'th ork because it's an &&str here because of iter over &str. We'd need into iter maybe. Anyway, ignoring for now, as this is a minor nit.

Firstyear:
@QnnOkabayashi At this point it's just opening some cleanup issues and a comment on #468 - is there anything else blocking this merging at this point?

QnnOkabayashi:
> @QnnOkabayashi At this point it's just opening some seperate cleanup issues and a comment on #468 - is there anything else blocking this merging at this point?

Looks good to me, go ahead and merge 👍

kanidmd/src/lib/value.rs
QnnOkabayashi: Okay, I opened it at https://github.com/kanidm/kanidm/issues/473

PR #467, opened 2021-06-03
Tracing logging

QnnOkabayashi:
Fixes #

- [x] cargo fmt has been run

Ignore `inter-ident-verify.md` for now.
`daemon.rs` is mostly refactoring.
I mostly need feedback on how to make `cache_layer_test.rs` work with the `tracing::subscriber::with_default` so I can have scoped logging working, instead of setting a global default.

I also think that we could use `tracing` for logging everywhere, since it's so flexible and can support both human-readable logging and json logging.


kanidm_unix_int/tests/cache_layer_test.rs
```diff
@@ -514,31 +514,58 @@ fn test_cache_account_pam_allowed() {
...
+            // ! TRACING BEGIN
+            let span = span!(Level::INFO, "test_cache_account_pam_allowed");
+            let _enter = span.enter();
+            // ? Seems like these subscribers will allow us to have json loggers
+            // ? and human-readable loggers at the same time, and select between
+            // ? them easily?
+
+            let subscriber = FmtSubscriber::builder()
+                .with_max_level(Level::INFO)
+                // .event_format(format::json()) // ! uncomment for JSON
```
QnnOkabayashi, line 526: I looked around in `unix_config.rs`, but I wasn't able to figure out how I can use that to configure which formatting to do here. Are the config objects passed around? Would I need to create my own field to determine the logging type or is there an existing one?

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "restructuredtext.confPath": ""
```
Firstyear, line 2: do you mind adding to .gitignore an ignore on .vscode? Is there a reason to add his to the repo?

designs/inter-ident-verify.md
```diff
@@ -0,0 +1,98 @@
+# Interpersonal Identity Verification
```
Firstyear, line 1: Do you need some help with rebasing/cherry-pick to clean up your branch?

kanidm_unix_int/src/cache.rs
```diff
@@ -786,14 +789,19 @@ impl CacheLayer {
...
                     );
+                    // error!(
+                    //     "unknown account or is not a valid posix account - eventid {}",
+                    //     opid
+                    // );
                     Ok(None)
                 }
-                er => {
-                    error!("client error -> {:?}", er);
+                err => {
```
Firstyear, line 803: Ahh william of the past, even to lazy to add two 'rr's.

kanidm_unix_int/src/daemon.rs
```diff
@@ -89,14 +90,15 @@ impl Decoder for TaskCodec {
     type Error = io::Error;
 
     fn decode(&mut self, src: &mut BytesMut) -> Result<Option<Self::Item>, Self::Error> {
-        match serde_cbor::from_slice::<TaskResponse>(&src) {
-            Ok(msg) => {
+        let maybe_task_response = serde_cbor::from_slice::<TaskResponse>(&src)
```
Firstyear, line 93: Do we need this let? Couldn't we just directly return after the ok()?

kanidm_unix_int/src/daemon.rs
```diff
@@ -338,10 +331,9 @@ async fn handle_client(
             }
             ClientRequest::Status => {
                 debug!("status check");
-                if cachelayer.test_connection().await {
-                    ClientResponse::Ok
-                } else {
-                    ClientResponse::Error
+                match cachelayer.test_connection().await {
```
Firstyear, line 334: Will clippy complain if you match on a bool?

kanidm_unix_int/src/daemon.rs
```diff
@@ -374,25 +366,19 @@ async fn main() {
...
+    });
     if cfg_path.exists() {
-        let cfg_meta = match metadata(&cfg_path) {
-            Ok(v) => v,
-            Err(e) => {
-                error!("Unable to read metadata for {} - {:?}", cfg_path_str, e);
-                std::process::exit(1);
-            }
-        };
+        let cfg_meta = metadata(&cfg_path).unwrap_or_else(|e| {
```
Firstyear, line 374: This is a neat idea, I like it.

kanidm_unix_int/src/daemon.rs
```diff
@@ -497,25 +475,26 @@ async fn main() {
...
-                        .unwrap_or_else(|| "<db_par_path_buf invalid>")
+                        .unwrap_or("<db_par_path_buf invalid>")
                 );
                 std::process::exit(1);
             }
             if i_meta.permissions().readonly() {
                 warn!("WARNING: DB folder permissions on {} indicate it may not be RW. This could cause the server start up to fail!", db_par_path_buf.to_str()
-                .unwrap_or_else(|| "<db_par_path_buf invalid>")
+                .unwrap_or("<db_par_path_buf invalid>")
+                // is there a reason we use `or_else` here when it's just a lit str?
```
Firstyear, line 485: Could just be past william didn't know. :)

kanidm_unix_int/tests/cache_layer_test.rs
```diff
@@ -514,31 +514,58 @@ fn test_cache_account_pam_allowed() {
...
+                .finish();
+            // ? This is specific to only applications-
+            // ? Ask if kanidm will ever be used as a library?
```
Firstyear, line 529: It will not :)

kanidm_unix_int/tests/cache_layer_test.rs
```diff
@@ -514,31 +514,58 @@ fn test_cache_account_pam_allowed() {
...
+            // ? I'm guessing not but double check
+
+            // Set a global subscriber (bad)
```
Firstyear, line 532: Why is this bad?

kanidm_unix_int/tests/cache_layer_test.rs
```diff
@@ -514,31 +514,58 @@ fn test_cache_account_pam_allowed() {
...
+            // tracing::subscriber::set_global_default(subscriber)
+            //     .expect("setting default subscriber failed");
+            // ! TRACING END
+
+            // ! This is what I want to do, but I get the error message:
+            // ! `await` is only allowed inside `async` functions and blocks
+            // ! only allowed inside `async` functions and blocks
+            // And if I make the block of the closure async, then there's
+            // nothing that will block on it and actually run it.
```
Firstyear, line 541: I think maybe you should move the tracing subscriber outside of this section, above `let fut = async move`. That way your setting up tracing in "sync" land, and then that can wrap the `let fut = ...` and `rt.block_on()`. That would resolve your problem I think :)

Firstyear (commented):
Really solid start mate. Appreciate the clean ups too :)

designs/inter-ident-verify.md
QnnOkabayashi: I think I know how to do it now. How should I go about removing it from this PR though?

.vscode/settings.json
QnnOkabayashi: Oops, I'll fix that now.

kanidm_unix_int/src/daemon.rs
QnnOkabayashi: I was trying to add clarity, but we could do it that way. I was also wondering if we could just do this:
```suggestion
        let maybe_task_response = serde_cbor::from_slice(&src); // type inferencer has enough information
        src.clear();
        Ok(maybe_task_response.ok())
```

I'm curious why we have so many nested `Result`'s and `Option`'s as return types, doesn't that somewhat defeat the point? Shouldn't we always just be returning a `Result`, and make more expressive `Err` types for better readability? I think the compiler would also be able to do more optimizations like this as well.

kanidm_unix_int/src/daemon.rs
QnnOkabayashi: Yeah I think I over complicated this, I'll revert.

kanidm_unix_int/tests/cache_layer_test.rs
QnnOkabayashi: It's bad because I was using it as a hack to make my code work, when I really should have been using scoped `subscriber`s. But I think we should end up using it as a safety net for default logs, probably with JSON formatting so we can do easy lookups. For most logging that we want to look at, I think we should use scoped `subscriber`s with human-readable formatting and pretty colors :)

One question, where should I initialize the global `subscriber`? I think we want to do it as early as possible, but I don't know where the entry point is.

kanidm_unix_int/tests/cache_layer_test.rs
QnnOkabayashi: This fixed the issue, thanks!

QnnOkabayashi:
Thanks for the feedback! I went through and came up with a few more questions:
* How do I remove irrelevant files from the PR?
* When I set the capture level to `DEBUG`, my `subscriber` picks up on a ton of logs like `hyper::client::pool: pooling idle connection for ("http", 127.0.0.1:28080)`.  Are we using the `tracing` crate anywhere else, or could it be picking up on logs intended for `env_logger`?
* If an `async` function call is `await`ed, is the function guaranteed to finish before continuing past the `await` call? I'm curious because the docs are very wary about using `span`s with `async`/`await` code.

yaleman (commented):
> * When I set the capture level to `DEBUG`, my `subscriber` picks up on a ton of logs like `hyper::client::pool: pooling idle connection for ("http", 127.0.0.1:28080)`.  Are we using the `tracing` crate anywhere else, or could it be picking up on logs intended for `env_logger`?

Tracing provides new handlers for the standard macros that env_logger provides (debug/info/error etc) so that'll be why - https://docs.rs/tracing/0.1.22/tracing/#for-log-users

Firstyear:
> Thanks for the feedback! I went through and came up with a few more questions:
> 
> * How do I remove irrelevant files from the PR?

You'll need to "have fun". It's likely some git branching stuff, so check git status/git log to see where you are. 

From there you need to decide what's happened:

* if you have commits that aren't meant to be there, but contain the extra files, git rebase -i can allow you to delete commits.
* If you have files all in one commit, you can unstage with git reset --soft, which will allow you to then re-add/commit the changes that you want and you can revert the others
* If you have a bit more confusion, you can checkout master as a clean branch, pull from upstream, then create a new branch and cherry-pick commits through

Remember, in most of these cases, take a backup first before you do anything, and also remember, that in git any mistake can be undo with the reflog if needed. 

It may be a good idea to check out @Charcol 's blog about gitting out of trouble: http://char.id.au/git-talks.html . Worst case we can have a quick call or something and sort it out. 

> * When I set the capture level to `DEBUG`, my `subscriber` picks up on a ton of logs like `hyper::client::pool: pooling idle connection for ("http", 127.0.0.1:28080)`.  Are we using the `tracing` crate anywhere else, or could it be picking up on logs intended for `env_logger`?

It's picking up tracing statements from other libraries (dependencies). So you may want to look at the subscriber docs or the env var if you can do something like DEBUG=kanidm or similar to filter which crates are logged. 

> * If an `async` function call is `await`ed, is the function guaranteed to finish before continuing past the `await` call? I'm curious because the docs are very wary about using `span`s with `async`/`await` code.

Yes and no.

Await means "pause execution of this stack here, and yield to the main tokio run time to find something else to do". When this task is read, the await "continues" and completes, and then the final result is returned. So lets say you have:

```
let x: Result<_, _> = foo.await;
```

In the case that foo.await pauses, x is still "undefined", but once it unpauses and completes, x will be allocated and now a result, so subsequent code can continue. It gets a little more complex wrt data dependencies but for most cases consider that:

```
PREVIOUS
let x: Result<_, _> = foo.await;
AFTER
```

Anything "AFTER" will always be after the await, anything PREVIOUS will always be previous to that.

kanidm_unix_int/src/cache.rs
```diff
@@ -731,13 +732,15 @@ impl CacheLayer {
             .await
         {
             Ok(Some(n_tok)) => {
-                debug!("online password check success.");
+                event!(Level::DEBUG, "online password check success.");
```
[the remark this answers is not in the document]
QnnOkabayashi: Nope, no benefit at all. I was thinking of changing macro calls to FQS like `tracing::debug!(...)`, but I'm not sure if that's a good idea because the `env_logger`'s macros seem to work fine for `tracing`.

kanidm_unix_int/tests/cache_layer_test.rs
QnnOkabayashi: Once I have a field and a default, how do I actually get access to this field in functions? For example, one of the tests looks like this:

```rust
#[test]
fn test_cache_account_pam_nonexist() {
    let subscriber = FmtSubscriber::builder()
        .with_max_level(Level::INFO)
        // if I wanted JSON formatting, I would say so here
        .finish();
    // other stuff here...
}
```
I guess I'm a bit confused because there is no config objects in scope. If this isn't possible, what scenarios would I have access to the config fields?

Another thing I'm concerned about is that the formatting of the subscriber varies at a type level. How can I work this in?

kanidm_unix_int/src/cache.rs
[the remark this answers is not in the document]
yaleman: Sounds like it's best to leave it with the original things in that case :)

kanidm_unix_int/src/daemon.rs
```diff
@@ -373,6 +373,7 @@ async fn main() {
     if cfg_path.exists() {
         let cfg_meta = metadata(&cfg_path).unwrap_or_else(|e| {
             error!("Unable to read metadata for {} - {:?}", cfg_path_str, e);
+            // Is is safe to call this? Docs say that no destructors will be run...
```
yaleman, line 376: At this point it's so early in the lifecycle that killing everything is fine - it hasn't opened the database or anything else. Memory will be cleaned up etc.

kanidm_unix_int/src/daemon.rs
```diff
@@ -439,7 +440,7 @@ async fn main() {
     });
 
     // Check the pb path will be okay.
-    if cfg.db_path != "" {
+    if cfg.db_path.is_empty() {
```
yaleman, line 443: At this point, you're changing this from "if cfg.db_path is set" to "if cfg.db_path is unset" - this inverts the test.

kanidm_unix_int/src/daemon.rs
```diff
@@ -542,13 +543,10 @@ async fn main() {
...
                     // Did it come from root?
-                    match socket.peer_cred() {
-                        Ok(ucred) if ucred.uid() == 0 => (), // all good!
-                        _ => {
-                            // move along.
-                            debug!("Task handler not running as root, ignoring ...");
-                            continue;
-                        }
+                    if !matches!(socket.peer_cred(), Ok(ucred) if ucred.uid() == 0) {
+                        // is this too obfuscated?
```
yaleman, line 547: It does seem pretty cognitively complex...

kanidm_unix_int/src/unix_config.rs
```diff
@@ -85,13 +94,14 @@ impl KanidmUnixdConfig {
...
                 debug!("Unabled to open config file [{:?}], skipping ...", e);
+                // I wish there was a way to use `unwrap_or_else` here
                 return Ok(self);
             }
         };
 
         let mut contents = String::new();
         f.read_to_string(&mut contents)
-            .map_err(|e| eprintln!("{:?}", e))?;
+            .map_err(|e| eprintln!("{:?}", e))?; // shouldn't this be a log?
```
yaleman, line 104: Yes, it should be.

yaleman (changes requested):
Looks good, but one of the changes to the daemon startup seem to introduce a bug

kanidm_unix_int/src/daemon.rs
Firstyear: At the least make it `!cfg.db_path.is_empty()`

kanidm_unix_int/src/daemon.rs
Firstyear: Yeah, I think this needs a clean up.

kanidm_unix_int/src/daemon.rs
Firstyear: And even if, we don't system.exit() within a db txn, so we know the ondisk content is stable, so this is fine.

QnnOkabayashi:
I'm at the point where I think I need to set up an instance to do further testing. I tried by following the instructions [here](https://github.com/kanidm/kanidm/blob/master/DEVELOPER_README.md#development-server-quickstart-for-interactive-testing), but when I got to running `cargo run -- recover_account -c ./server.toml -n admin`, I got:

```
   Compiling kanidm_tools v1.1.0-alpha.3 (/Users/quinn/Documents/Programming/rust/kanidm/kanidm_tools)
    Finished dev [unoptimized + debuginfo] target(s) in 8.48s
     Running `target/debug/kanidm recover_account -c ./server.toml -n admin`
error: Found argument 'recover_account' which wasn't expected, or isn't valid in this context

USAGE:
    kanidm <SUBCOMMAND>

For more information try --help
```
Any tips?

Firstyear:
> I'm at the point where I think I need to set up an instance to do further testing. I tried by following the instructions [here](https://github.com/kanidm/kanidm/blob/master/DEVELOPER_README.md#development-server-quickstart-for-interactive-testing), but when I got to running `cargo run -- recover_account -c ./server.toml -n admin`, I got:
> 
> ```
>    Compiling kanidm_tools v1.1.0-alpha.3 (/Users/quinn/Documents/Programming/rust/kanidm/kanidm_tools)
>     Finished dev [unoptimized + debuginfo] target(s) in 8.48s
>      Running `target/debug/kanidm recover_account -c ./server.toml -n admin`
> error: Found argument 'recover_account' which wasn't expected, or isn't valid in this context
> 
> USAGE:
>     kanidm <SUBCOMMAND>
> 
> For more information try --help
> ```
> 
> Any tips?

You need to run the server binary from in the kanidmd directory - at the work space root it runs the cli tool instead. 

Try `cd kanidmd` then try again :)

yaleman:
> You need to run the server binary from in the kanidmd directory - at the work space root it runs the cli tool instead.
> 
> Try `cd kanidmd` then try again :)

or `cargo run --bin kanidmd -- server --config /dev/null --debug verbose` etc etc ::D

Firstyear:
> > You need to run the server binary from in the kanidmd directory - at the work space root it runs the cli tool instead.
> > Try `cd kanidmd` then try again :)
> 
> or `cargo run --bin kanidmd -- server --config /dev/null --debug verbose` etc etc ::D

--config /dev/null would end in some quite comical results I'm sure :)

yaleman:
> > > You need to run the server binary from in the kanidmd directory - at the work space root it runs the cli tool instead.
> > > Try `cd kanidmd` then try again :)
> > 
> > 
> > or `cargo run --bin kanidmd -- server --config /dev/null --debug verbose` etc etc ::D
> 
> --config /dev/null would end in some quite comical results I'm sure :)

It's a good way to force it to run and output something, then die when it can't find config, which is what I was testing at the time 🤣

QnnOkabayashi:
> > I'm at the point where I think I need to set up an instance to do further testing. I tried by following the instructions [here](https://github.com/kanidm/kanidm/blob/master/DEVELOPER_README.md#development-server-quickstart-for-interactive-testing), but when I got to running `cargo run -- recover_account -c ./server.toml -n admin`, I got:
> > ```
> >    Compiling kanidm_tools v1.1.0-alpha.3 (/Users/quinn/Documents/Programming/rust/kanidm/kanidm_tools)
> >     Finished dev [unoptimized + debuginfo] target(s) in 8.48s
> >      Running `target/debug/kanidm recover_account -c ./server.toml -n admin`
> > error: Found argument 'recover_account' which wasn't expected, or isn't valid in this context
> > 
> > USAGE:
> >     kanidm <SUBCOMMAND>
> > 
> > For more information try --help
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Any tips?
> 
> You need to run the server binary from in the kanidmd directory - at the work space root it runs the cli tool instead.
> 
> Try `cd kanidmd` then try again :)

This fills my console with a ton of logs with massive objects displayed in their `Debug` implementation. What is the server trying to do on startup? Which files are generating these logs? And what are all these `Filter` objects and their giant debug msgs doing?

QnnOkabayashi:
Also, what is the function/purpose of the `lperf_trace_segment` macro like found here

https://github.com/kanidm/kanidm/blob/c2d74ced2b5a1ee3cb65378ac349eb18ed3d1aa6/kanidmd/src/lib/be/mod.rs#L523

I did a little bit of digging through the macro, and it seems like we could simplify this and remove the macro by making it a method of the `AuditScope`. Something like:
```rust
use crate::audit::LogTag;

impl AuditScope {
    fn lperf_tag_segment<T>(&mut self, id: &'static str, tag: LogTag, fun: impl Fn() -> T) -> T {
        if (self.level & tag as u32) == tag as u32 {
            use std::time::Instant;

            // start timer.
            let start = Instant::now();

            // Create a new perf event - this sets
            // us as the current active, and the parent
            // correctly.
            let pe = unsafe { self.new_perfevent(id) };

            // fun run time
            let r = fun();
            // end timer, and diff
            let end = Instant::now();
            let diff = end.duration_since(start);

            // Now we are done, we put our parent back as
            // the active.
            unsafe { self.end_perfevent(pe, diff) };

            // Return the result. Hope this works!
            r
        } else {
            fun()
        }
    }
}
```
I don't think this would be particularly helpful for logging. But I do think it would remove some of the mystery that comes with using macros, since they're difficult to reason about without looking at their source code.

Firstyear:
> This fills my console with a ton of logs with massive objects displayed in their `Debug` implementation. What is the server trying to do on startup? Which files are generating these logs? And what are all these `Filter` objects and their giant debug msgs doing?

On startup it's literally doing a full database create and migration. Filter are internal searches, and you're seeing lots of times when the server is checking "hey do I need to migrate this". This is probably because by default you may have a higher verbosity level set.

Firstyear:
> Also, what is the function/purpose of the `lperf_trace_segment` macro like found here
> 
> https://github.com/kanidm/kanidm/blob/c2d74ced2b5a1ee3cb65378ac349eb18ed3d1aa6/kanidmd/src/lib/be/mod.rs#L523
> 

It's the macro that creates a "span" to do performance traces of function execution times. 


> 
> I don't think this would be particularly helpful for logging. But I do think it would remove some of the mystery that comes with using macros, since they're difficult to reason about without looking at their source code.

macros vs functions both hide what they do without looking at their source by their very nature. But in this case, I think it's better to leave it as a macro until we know if: 

* We want to change to tracing instead
* Or if we stick with auditscope, if there is a benefit to changing this behaviour

The main reason we could consider fn's instead is that we had to remove some trace spans as async closures aren't available yet I don't think?

QnnOkabayashi:
In the process of implementing `tracing` on `kanidmd/src/lib/plugins/memberof.rs`, I was having trouble reasoning about the logging macros that are currently being used regarding the current `AuditScope` strategy for logging. I decided to do a quick-and-dirty rewrite for the macros as methods of `AuditScope`, since that would allow vscode to provide suggestions/type hints/tips to help me better understand the logic of the code.

However, I'm having a compilation error:
```
error[E0599]: no function or associated item named `new_impersonate_entry_ser` found for struct `event::CreateEvent` in the current scope
   --> kanidmd/src/lib/macros.rs:260:30
    |
260 |                 CreateEvent::new_impersonate_entry_ser(e_str, create_entries.clone())
    |                              ^^^^^^^^^^^^^^^^^^^^^^^^^ function or associated item not found in `event::CreateEvent`
    | 
   ::: kanidmd/src/lib/event.rs:517:1
    |
517 | pub struct CreateEvent {
    | ---------------------- function or associated item `new_impersonate_entry_ser` not found for this

error[E0599]: no method named `write` found for struct `server::QueryServer` in the current scope
   --> kanidmd/src/lib/macros.rs:265:31
    |
265 |             let qs_write = qs.write(duration_from_epoch_now());
    |                               ^^^^^ method not found in `server::QueryServer`
    | 
   ::: kanidmd/src/lib/server.rs:62:1
    |
62  | pub struct QueryServer {
    | ---------------------- method `write` not found for this
    |
    = help: items from traits can only be used if the trait is implemented and in scope
```
Both of these methods are direct methods of the types, and not `trait` impls, and I made sure all the proper types are `use`d. Why is this happening?

QnnOkabayashi:
I did a lot more digging around `tracing` today and found a lot of features that align nicely with what we're already doing with our `AuditScope` type.

The `lqueue` macro in `kanidmd/src/lib/audit.rs`, and other macros invoking it with specific arguments (`lfilter`, `ladmin_warning`, ...), is analogous to `tracing`s standard log macros (`trace`/`debug`/`info`/`warn`/`error`). The only difference is that it incorporates a `LogTag` in it to track the level of the log, but it also serves as a flag for where the log occurred, using the `Admin`, `Request`, `Security`, and `Filter` prefixes in `LogTag`s variant names. We can encode this information in the logs in `tracing` by using fields. Here's an example of what this would look like:
```rust
// using `AuditScope`
ladmin_error!(au, "An error occurred");

// using `tracing`
tracing::error!(tag = "admin", "An error occurred");
```
In JSON formatting, this adds a `"tag":"admin"` key-value pair to the log objects `fields` array, meaning we can easily access and filter on it. 

For consistency, we could define a custom macros to add these tags:
```rust
// a macro to expand this
admin_error!("Oh no")
// into this
tracing::error!(tag = "admin", "Oh no")
```

Similarly, the `lperf_tag_segment` macro and it's helper macros (`lperf_op_segment`, ...) is analogous to `tracing`s `span`s. The `lperf` family of macros also uses `LogTag`s, but this issue is solved the same as above. The only thing I haven't figured our yet is if the spans have built-in timers, but I don't think this would be hard to implement.

I currently see four major benefits to using `tracing`:
1. More flexibility with logging formatting (JSON, pretty, ansi colors (lol)...)
2. Exporting tasks that are not kanidm-related (aka logging) to a crate that specializes in it and is optimized for it. This also means that if we want a new feature, it likely already exists. Just less code to manage in Kanidm.
3. We don't have to pass `&mut AuditScope` references around everywhere.
4. `tracing` was built for `async` compatibility, meaning it can be used almost anywhere in Kanidm.

I'm still stuck on a few issues, though: 
* I'm having an issue with all `tracing` logs getting directed to `stdout`, so my logs are getting mixed with everything that's supposed to be printed to the console. However, other logs are going to `stderr`. Any ideas on how to fix this? I struggled when googling around for this. I'm logging everything to files because there's so much stuff, and then using cmd+f to find the logs I'm trying to generate.
```
$ cargo test --package kanidm --lib -- plugins::memberof::tests::test_create_mo_single --exact --nocapture > out.log 2> err.log
```
* When should I be concerned about running performance tests? It seems like `tracing` is more than capable of doing everything we're already doing with `AuditScope`.

Firstyear:
> Question: does the "perf" part stand for "performance"?

Yep

> 
> I currently see four major benefits to using `tracing`:
> 
> 1. More flexibility with logging formatting (JSON, pretty, ansi colors (lol)...)
> 2. Exporting tasks that are not kanidm-related (aka logging) to a crate that specializes in it and is optimized for it. This also means that if we want a new feature, it likely already exists. Just less code to manage in Kanidm.
> 3. We don't have to pass `&mut AuditScope` references around everywhere.
> 4. `tracing` was built for `async` compatibility, meaning it can be used almost anywhere in Kanidm.
> 
> I'm still stuck on a few issues, though:
> 
> * I'm having an issue with all `tracing` logs getting directed to `stdout`, so my logs are getting mixed with everything that's supposed to be printed to the console. However, other logs are going to `stderr`. Any ideas on how to fix this? I struggled when googling around for this. I'm logging everything to files because there's so much stuff, and then using cmd+f to find the logs I'm trying to generate.

You could set the auditscope to quiet which mutes all the current audit scope logging. That would let you see what you want to see. 

> 
> ```
> $ cargo test --package kanidm --lib -- plugins::memberof::tests::test_create_mo_single --exact --nocapture > out.log 2> err.log
> ```
> 
> * When should I be concerned about running performance tests? It seems like `tracing` is more than capable of doing everything we're already doing with `AuditScope`.

IMO there won't be a need for performance tests because there is a specific requirement that I have which will absolve us of the need for it.

That is, the thread that is currently writing the log event should not be the thread that displays it to stdout/stderr.

See, the current design of auditscope is such that all the messages are collected to the audit scope, then they are sent in a queue to a dedicated log-writing task/thread, and THAT does the output/writing. The majority of the performance impact of logging comes from writing it out!

The immediate question you may have is "why".

* Stdout/stderr actually contain a lock, and that adds many fine-grained serialisation points through out the code that impacts concurrency. This is why we don't want "logs" writing immediately as they come in.
* In the case of a crash (unlikely, but it may happen), we can go to the thread in question, and observe it's log state/stack for that operation to understand the events leading up to the failure. 
* We want logs to be batched - if you look at something like:

```
[09/Jun/2021:14:45:15.911692001 +1000] conn=76 op=1 RESULT err=0 tag=101 nentries=0 wtime=0.000313903 optime=0.000059015 etime=0.000371896
[09/Jun/2021:14:45:15.912023298 +1000] conn=79 op=-1 fd=68 closed error - B1
[09/Jun/2021:14:45:15.912029434 +1000] conn=78 op=-1 fd=65 closed error - B1
[09/Jun/2021:14:45:15.912489806 +1000] conn=76 op=-1 fd=66 closed error - B1
[09/Jun/2021:14:45:15.912508286 +1000] conn=91 fd=65 slot=65 SSL connection from 172.24.20.54 to 172.24.20.4
[09/Jun/2021:14:45:15.922040722 +1000] conn=92 fd=66 slot=66 SSL connection from 172.24.20.54 to 172.24.20.4
[09/Jun/2021:14:45:15.922782871 +1000] conn=82 TLS1.3 128-bit AES-GCM
[09/Jun/2021:14:45:15.922983809 +1000] conn=82 op=0 BIND dn="uid=account_12917285583703370811,ou=people,dc=example,dc=com" method=128 version=3
[09/Jun/2021:14:45:15.923169190 +1000] conn=82 op=0 RESULT err=0 tag=97 nentries=0 wtime=0.011671096 optime=0.000200500 etime=0.011870460 dn="uid=account_12917285583703370811,ou=people,dc=example,dc=com"
[09/Jun/2021:14:45:15.927894437 +1000] conn=93 fd=68 slot=68 SSL connection from 172.24.20.54 to 172.24.20.4
[09/Jun/2021:14:45:15.927926421 +1000] conn=94 fd=77 slot=77 SSL connection from 172.24.20.54 to 172.24.20.4
[09/Jun/2021:14:45:15.928484214 +1000] conn=95 fd=79 slot=79 SSL connection from 172.24.20.54 to 172.24.20.4
[09/Jun/2021:14:45:15.928600474 +1000] conn=81 TLS1.3 128-bit AES-GCM
```

Which lines here relate to each other? What events are correlated? While we are seeing an accurate picture of events over time, it becomes hard to trace "what happened in this ONE event"? 

The way kanidm does this is by batching log items into a single audit scope, we can emit them together.

```
[a39475da-2d64-494b-b670-b6e76c5f4e64 security::info 🔐] search initiator: -> User( anonymous@idm.blackhats.net.au, 00000000-0000-0000-0000-ffffffffffff )
[a39475da-2d64-494b-b670-b6e76c5f4e64 security::access 🔓] entry c1510395-c5d0-4f31-ad94-38a8d6547604 matches acs idm_all_acp_read
[a39475da-2d64-494b-b670-b6e76c5f4e64 security::access 🔓] requested filter attributes --> {"uuid"} allowed attributes   --> {"class", "displayname", "gidnumber", "loginshell", "member", "memberof", "name", "spn", "ssh_publickey", "uuid"}
[a39475da-2d64-494b-b670-b6e76c5f4e64 security::access 🔓] search attr decision --> true
[a39475da-2d64-494b-b670-b6e76c5f4e64 security::access 🔓] allowed 1 entries ✅
[a39475da-2d64-494b-b670-b6e76c5f4e64 security::access 🔓] target entry c1510395-c5d0-4f31-ad94-38a8d6547604 matches acs idm_all_acp_read
[a39475da-2d64-494b-b670-b6e76c5f4e64 security::access 🔓] requested attributes --> None allowed attributes   --> {"class", "displayname", "gidnumber", "loginshell", "member", "memberof", "name", "spn", "ssh_publickey", "uuid"}
[a39475da-2d64-494b-b670-b6e76c5f4e64 security::access 🔓] attribute set reduced on 1 entries ✅
[fc649e96-6155-450f-9617-e3a1ab2ffabf security::info 🔐] search initiator: -> User( anonymous@idm.blackhats.net.au, 00000000-0000-0000-0000-ffffffffffff )
[fc649e96-6155-450f-9617-e3a1ab2ffabf security::access 🔓] entry 886b45ce-98b2-4d79-9069-c9ff25bc0232 matches acs idm_all_acp_read
[fc649e96-6155-450f-9617-e3a1ab2ffabf security::access 🔓] requested filter attributes --> {"uuid"} allowed attributes   --> {"class", "displayname", "gidnumber", "loginshell", "member", "memberof", "name", "spn", "ssh_publickey", "uuid"}
[fc649e96-6155-450f-9617-e3a1ab2ffabf security::access 🔓] search attr decision --> true
[fc649e96-6155-450f-9617-e3a1ab2ffabf security::access 🔓] allowed 1 entries ✅
```

Here it's a lot easier to see the components that relate to each other as a human reading this event log. We know that anonymous has done two searches, and what it was they requested. We could make it clearer by resolving the target uuid's to an spn/name, but even now, we still have a solid idea of the content of each "transaction" or "event".

So getting tracing to do the same would be invaluable for human readability.

IMO that's why performance won't be a major impact, because if we approach it in this way, we are making our logs effectively async, and avoiding locking :)

QnnOkabayashi:
> You could set the auditscope to quiet which mutes all the current audit scope logging. That would let you see what you want to see.

I worded that badly, I meant to say the issue was that I can't figure out how to redirect to `stderr` instead of `stdout`. cmd+f to find my logs isn't really an issue.

> Here it's a lot easier to see the components that relate to each other as a human reading this event log. We know that anonymous has done two searches, and what it was they requested. We could make it clearer by resolving the target uuid's to an spn/name, but even now, we still have a solid idea of the content of each "transaction" or "event".
> 
> So getting tracing to do the same would be invaluable for human readability.
> 
> IMO that's why performance won't be a major impact, because if we approach it in this way, we are making our logs effectively async, and avoiding locking :)

One idea to solve this blocking issue would be to use `std::sync::mpsc`, or an alternative send/receive async library. We can think about this more later.

Also, the `tracing` crate allows us to create `Subscriber`s that only listen for logs with certain fields that you can specify at runtime. If I can find a way to make `Subscriber`s withhold logs until explicitly told to write, I can easily simulate the scoping behavior by associating a `Subscriber` with each `AuditScope`. This way, I would be able to address all the strengths that `AuditScope` already has, in addition to the benefits I listed above.

QnnOkabayashi:
I added a `LogBuffer` type (currently in a temporary spot for testing) that is able to collect all the logs and release them all at once. I currently have it tied to a `Subscriber` object, so all you have to do is add a `.with_writer(LogBuffer::default())` call during the subscriber building process, and all the logs get stored in a buffer until the `Subscriber` is dropped, and then all the logs are written.

One thing I was unsure about is that in order to do this, I had to bend over backwards to meet the `tracing_subscriber::fmt::MakeWriter` trait requirements. It basically forced me to return a new `io::Write` object each time, so I had to store the buffer in an `Arc<Mutex<_>>`. This means that each log sent to the buffer involves an `Arc::clone` call :(. If we're not planning to send the `LogBuffer` across threads while logs are actively being taken, I can do some stuff to get around this I think. I do understand that we plan do send it to another thread for the actual dumping, and that should still be fine because we're not cloning it anymore at that point. I also still need to figure out how we're sending it to another thread during the `drop` call.

Firstyear:
> I worded that badly, I meant to say the issue was that I can't figure out how to redirect to `stderr` instead of `stdout`. I got it working now with the [`with_writer`](https://docs.rs/tracing-subscriber/0.2.18/tracing_subscriber/fmt/struct.SubscriberBuilder.html#method.with_writer) method.

You could also just redirect this in shell with: `2>&1`

> I added a `LogBuffer` type (currently in a temporary spot for testing) that is able to collect all the logs and release them all at once. I currently have it tied to a `Subscriber` object, so all you have to do is add a `.with_writer(LogBuffer::default())` call during the subscriber building process, and all the logs get stored in a buffer until the `Subscriber` is dropped, and then all the logs are written.

So I assume with this logbuffer type, we would have a new log buffer per event and then it's written out at the end of the operation? 

> 
> One thing I was unsure about is that in order to do this, I had to bend over backwards to meet the `tracing_subscriber::fmt::MakeWriter` trait requirements. It basically forced me to return a new `io::Write` object each time, so I had to store the buffer in an `Arc<Mutex<_>>`. This means that each log sent to the buffer involves an `Arc::clone` call :(. If we're not planning to send the `LogBuffer` across threads while logs are actively being taken, I can do some stuff to get around this I think. I do understand that we plan do send it to another thread for the actual dumping, and that should still be fine because we're not cloning it anymore at that point. I also still need to figure out how we're sending it to another thread during the `drop` call.

I think I need to see the code for this before I can comment here. In theory, if each buffer is used in one operation at a time, then a mutex won't really matter too much since it will be uncontended. But it depends how you've modeled this log buffer. Generally IMO anytime you need to reach for a Mutex you have to wonder "what have I done wrong" in a framework like Kanidm :P

Firstyear:
Okay, one other thing. You did a test of tracing in the unix integration, and that's what this pr is to swap that over. Is there a reason we don't want to make unix int just use tracing and commit, and then focusing on tracing in a new issue/set of prs for the server core?

Firstyear:
Just noticed your changes in kanidmd - can we split these out to a new PR so we can discuss them seperately to the unix_int changes?

Firstyear:
@QnnOkabayashi Do you mind rebasing to master to resolve these conflicts so we can merge?

QnnOkabayashi:
> @QnnOkabayashi Do you mind rebasing to master to resolve these conflicts so we can merge?

I'm having some confusion rebasing to master. I pulled the changes from master onto my forks master branch. From there, I tried to rebase it off of my master and even after resolving conflicts (I just accepted all the changes from master), I was getting error messages. I was super confused because one of the functions I had had different parameters than that in master and my branch. How did this even happen?

Also, I'm not ready to merge yet because I haven't added many of the changes I wanted to add yet. I'm probably going to work on https://github.com/kanidm/kanidm/pull/497 first before I work on this.

Firstyear:
What else did you want to add to this change for unix_int that is different to 497 out of curiosity :) 

I've sent you conflict fix strategies out-of-band :)

QnnOkabayashi:
@yaleman Do you have any particular recommendations for the parameters that should be included in a JSON log? Here's an example of the format I'm currently planning on implementing:
### Span
```json
{
    "uuid": "00000000-0000-0000-0000-000000000000",
    "timestamp": "2021-07-20T18:59:30.605201+00:00",
    "level": "TRACE",
    "message": "try_from_entry_ro",
    "log-type": "span",
    "nanos-direct": 282,
    "nanos-total": 1000,
}
```

### Event
```json
{
    "uuid": "00000000-0000-0000-0000-000000000000",
    "timestamp": "2021-07-20T18:59:30.605201+00:00",
    "level": "INFO",
    "message": "Some request info...",
    "log-type": "event",
    "tag": "request.info",
    "spans": [
        "try_from_entry_ro",
        "server::internal_search",
    ]
}
```

PR #469, opened 2021-06-07
163 account recovery code

### LINKED ISSUE #163: Account recovery (backup) mfa codes

victorcwai:
Fixes #163 (WIP) 
Looking for feedback!
Draft code for using backup code to login and how to create it. I mostly follow the code for doing authentication with totp/password generation.

- [x] cargo fmt has been run

### DIFF
--- a/kanidm_proto/src/v1.rs
+++ b/kanidm_proto/src/v1.rs
@@ -382,6 +382,12 @@ impl fmt::Display for CredentialStatus {
     }
 }
 
+#[derive(Debug, Serialize, Deserialize, Clone)]
+pub struct BackupCodesView {
+    // Or use SetCredentialResponse::BackupCodes?
+    pub backup_codes: Vec<String>,
+}
+
 /* ===== low level proto types ===== */
 
 // ProtoEntry vs Entry
@@ -682,6 +688,7 @@ pub enum SetCredentialRequest {
     // Remove
     WebauthnRemove(String),
     GenerateBackupCode,
+    BackupCodeRemove,
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
--- a/kanidm_proto/src/v1.rs
+++ b/kanidm_proto/src/v1.rs
@@ -310,7 +310,7 @@ pub enum CredentialDetailType {
     GeneratedPassword,
     Webauthn(Vec<String>),
     /// totp, webauthn
-    PasswordMfa(bool, Vec<String>),
+    PasswordMfa(bool, Vec<String>, bool),
 }
 
 #[derive(Debug, Serialize, Deserialize, Clone)]
@@ -341,12 +341,17 @@ impl fmt::Display for CredentialDetail {
                     write!(f, "")
                 }
             }
-            CredentialDetailType::PasswordMfa(totp, labels) => {
+            CredentialDetailType::PasswordMfa(totp, labels, backup_code) => {
                 writeln!(f, "password: set")?;
                 if *totp {
                     writeln!(f, "totp: enabled")?;
                 } else {
                     writeln!(f, "totp: disabled")?;
+                }
+                if *backup_code {
+                    writeln!(f, "backup_code: enabled")?;
+                } else {
+                    writeln!(f, "backup_code: disabled")?;
                 }
                 if labels.is_empty() {
                     writeln!(f, "webauthn: no authenticators")
@@ -522,6 +527,7 @@ pub enum AuthCredential {
     Password(String),
     Totp(u32),
     Webauthn(PublicKeyCredential),
+    BackupCode(String),
 }
 
 impl fmt::Debug for AuthCredential {
@@ -531,6 +537,7 @@ impl fmt::Debug for AuthCredential {
             AuthCredential::Password(_) => write!(fmt, "Password(_)"),
             AuthCredential::Totp(_) => write!(fmt, "TOTP(_)"),
             AuthCredential::Webauthn(_) => write!(fmt, "Webauthn(_)"),
+            AuthCredential::BackupCode(_) => write!(fmt, "BackupCode(_)"),
         }
     }
 }
@@ -674,6 +681,7 @@ pub enum SetCredentialRequest {
     WebauthnRegister(Uuid, RegisterPublicKeyCredential),
     // Remove
     WebauthnRemove(String),
+    GenerateBackupCode,
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
--- a/kanidm_proto/src/v1.rs
+++ b/kanidm_proto/src/v1.rs
@@ -310,7 +310,7 @@ pub enum CredentialDetailType {
     GeneratedPassword,
     Webauthn(Vec<String>),
     /// totp, webauthn
-    PasswordMfa(bool, Vec<String>, bool),
+    PasswordMfa(bool, Vec<String>, usize),
 }
 
 #[derive(Debug, Serialize, Deserialize, Clone)]
@@ -348,7 +348,7 @@ impl fmt::Display for CredentialDetail {
                 } else {
                     writeln!(f, "totp: disabled")?;
                 }
-                if *backup_code {
+                if *backup_code > 0 {
                     writeln!(f, "backup_code: enabled")?;
                 } else {
                     writeln!(f, "backup_code: disabled")?;
@@ -748,6 +748,7 @@ pub enum SetCredentialResponse {
     Token(String),
     TotpCheck(Uuid, TotpSecret),
     WebauthnCreateChallenge(Uuid, CreationChallengeResponse),
+    BackupCodes(Vec<String>),
 }
 
 /* Recycle Requests area */
--- a/kanidmd/src/lib/actors/v1_write.rs
+++ b/kanidmd/src/lib/actors/v1_write.rs
@@ -643,7 +643,7 @@ impl QueryServerWriteV1 {
                         idms_prox_write
                             .generate_backup_code(&mut audit, &gbe)
                             .and_then(|r| idms_prox_write.commit(&mut audit).map(|_| r))
-                            .map(SetCredentialResponse::Token)
+                            .map(SetCredentialResponse::BackupCodes)
                     }
                 }
             }
--- a/kanidmd/src/lib/actors/v1_write.rs
+++ b/kanidmd/src/lib/actors/v1_write.rs
@@ -3,6 +3,7 @@ use std::sync::Arc;
 use tokio::sync::mpsc::UnboundedSender as Sender;
 
 use crate::idm::event::GenerateBackupCodeEvent;
+use crate::idm::event::RemoveBackupCodeEvent;
 use crate::prelude::*;
 
 use crate::event::{
@@ -644,6 +645,25 @@ impl QueryServerWriteV1 {
                             .generate_backup_code(&mut audit, &gbe)
                             .and_then(|r| idms_prox_write.commit(&mut audit).map(|_| r))
                             .map(SetCredentialResponse::BackupCodes)
+                    }
+                    SetCredentialRequest::BackupCodeRemove => {
+                        let rbe = RemoveBackupCodeEvent::from_parts(
+                            &mut audit,
+                            // &idms_prox_write.qs_write,
+                            ident,
+                            target_uuid,
+                        )
+                        .map_err(|e| {
+                            ladmin_error!(
+                                audit,
+                                "Failed to begin internal_credential_set_message: {:?}",
+                                e
+                            );
+                            e
+                        })?;
+                        idms_prox_write
+                            .remove_backup_code(&mut audit, &rbe)
+                            .and_then(|r| idms_prox_write.commit(&mut audit).map(|_| r))
                     }
                 }
             }
--- a/kanidmd/src/lib/actors/v1_write.rs
+++ b/kanidmd/src/lib/actors/v1_write.rs
@@ -2,6 +2,7 @@ use std::iter;
 use std::sync::Arc;
 use tokio::sync::mpsc::UnboundedSender as Sender;
 
+use crate::idm::event::GenerateBackupCodeEvent;
 use crate::prelude::*;
 
 use crate::event::{
@@ -623,6 +624,26 @@ impl QueryServerWriteV1 {
                         idms_prox_write
                             .remove_account_webauthn(&mut audit, &rwe)
                             .and_then(|r| idms_prox_write.commit(&mut audit).map(|_| r))
+                    }
+                    SetCredentialRequest::GenerateBackupCode => {
+                        let gbe = GenerateBackupCodeEvent::from_parts(
+                            &mut audit,
+                            // &idms_prox_write.qs_write,
+                            ident,
+                            target_uuid,
+                        )
+                        .map_err(|e| {
+                            ladmin_error!(
+                                audit,
+                                "Failed to begin internal_credential_set_message: {:?}",
+                                e
+                            );
+                            e
+                        })?;
+                        idms_prox_write
+                            .generate_backup_code(&mut audit, &gbe)
+                            .and_then(|r| idms_prox_write.commit(&mut audit).map(|_| r))
+                            .map(SetCredentialResponse::Token)
                     }
                 }
             }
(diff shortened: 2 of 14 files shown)


kanidmd/src/lib/actors/v1_write.rs
```diff
@@ -624,6 +625,26 @@ impl QueryServerWriteV1 {
...
+                                audit,
+                                "Failed to begin internal_credential_set_message: {:?}",
+                                e
+                            );
+                            e
+                        })?;
+                        idms_prox_write
+                            .generate_backup_code(&mut audit, &gbe)
+                            .and_then(|r| idms_prox_write.commit(&mut audit).map(|_| r))
+                            .map(SetCredentialResponse::Token)
```
Firstyear, line 646: This should probably use a new type like BackupCodes on the `SetCredentialResponse` enum so it's really clear what these are.

kanidmd/src/lib/core/https.rs
```diff
@@ -1278,6 +1307,11 @@ pub fn create_https_server(
     account_route
         .at("/:id/_credential/:cid/_lock")
         .get(do_nothing);
+    account_route
+        .at("/:id/_credential/backup_code")
```
Firstyear, line 1311: These should be under :cid or `.at("/:id/_credential/primary")`. This is for future multiple credential handling

kanidmd/src/lib/credential/mod.rs
```diff
@@ -211,6 +213,34 @@ impl Password {
     }
 }
 
+#[derive(Clone, Debug)]
+pub struct BackupCode {
```
Firstyear, line 217: Shouldn't this be BackupCodes (plural, append an S here).

kanidmd/src/lib/credential/mod.rs
```diff
@@ -706,6 +768,59 @@ impl Credential {
         }
     }
 
+    pub(crate) fn update_backup_code(&self, backup_code: HashSet<String>) -> Self {
+        let type_ = match &self.type_ {
+            // TODO: we should panic/reject CredentialType != PasswordMfa?
```
Firstyear, line 773: we should reject (not panic), so this would change to `Result<Self, OperationError>` to represent that errorstate

kanidmd/src/lib/credential/mod.rs
```diff
@@ -706,6 +768,59 @@ impl Credential {
...
+            CredentialType::Password(pw) | CredentialType::GeneratedPassword(pw) => {
+                debug_assert!(false);
+                CredentialType::PasswordMfa(
```
Firstyear, line 776: I don't think this is valid to do this convert, since a backup code exists to recover from lost webauthn/totp.

kanidmd/src/lib/credential/mod.rs
```diff
@@ -706,6 +768,59 @@ impl Credential {
...
+            CredentialType::PasswordMfa(pw, totp, wan, _) => CredentialType::PasswordMfa(
+                pw.clone(),
+                totp.clone(),
+                wan.clone(),
+                Some(BackupCode {
+                    code_set: backup_code,
+                }),
+            ),
+            CredentialType::Webauthn(wan) => {
+                debug_assert!(false);
```
Firstyear, line 794: Should return a Err() instead.

kanidmd/src/lib/credential/mod.rs
```diff
@@ -706,6 +768,59 @@ impl Credential {
...
+            }
+        };
+        Credential {
+            type_,
+            claims: self.claims.clone(),
+            uuid: self.uuid,
+        }
+    }
+
+    // pub(crate) fn remove_backup_code(&self) -> Self {
```
Firstyear, line 805: I think you should uncomment this, it's good :) 

If you make it so only PasswordMfa takes backup code, it's a much simpler code path. You could probabry consider returning Err() when not PMFA here though so that we respond that the transform is invalid though.

kanidmd/src/lib/credential/mod.rs
```diff
@@ -731,8 +846,8 @@ impl CredentialType {
     fn is_valid(&self) -> bool {
         match self {
             CredentialType::Password(_) | CredentialType::GeneratedPassword(_) => true,
-            CredentialType::PasswordMfa(_, m_totp, webauthn) => {
-                m_totp.is_some() || !webauthn.is_empty()
+            CredentialType::PasswordMfa(_, m_totp, webauthn, _) => {
+                m_totp.is_some() || !webauthn.is_empty() // ignore backup code (it should only be a complement for totp/webauth)
```
Firstyear, line 850: I think this check won't be needed if you change the code above to only set backup code if totp/webauthn are some. (which they have to be if you are password MFA anyway).

kanidmd/src/lib/idm/account.rs
```diff
@@ -357,6 +358,39 @@ impl Account {
         }
     }
 
+    pub(crate) fn gen_backup_code_mod(
+        &self,
+        backup_code: HashSet<String>,
```
Firstyear, line 363: plural

kanidmd/src/lib/idm/authsession.rs
```diff
@@ -301,6 +309,23 @@ impl CredHandler {
                             CredState::Denied(BAD_TOTP_MSG)
                         }
                     }
+                    (AuthCredential::BackupCode(code_chal), _, _, Some(backup_code)) => {
```
Firstyear, line 312: You need to extend and test this state machine flow in the tests at the bottom of this file. This is a security critical code path, so requires *extensive* testing. However, the logic here appears to be correct IMO.

kanidmd/src/lib/idm/server.rs
```diff
@@ -1356,6 +1359,46 @@ impl<'a> IdmServerProxyWriteTransaction<'a> {
...
+        &mut self,
+        au: &mut AuditScope,
+        gbe: &GenerateBackupCodeEvent,
+    ) -> Result<String, OperationError> {
+        let account = self.target_to_account(au, &gbe.target)?;
+
+        // Generate a new set of backup code.
+        let s = backup_code_from_random();
+        // TODO: into_iter() consumes s, find a way to get the joined string without cloning?
+        let s_str = s.clone().into_iter().collect::<Vec<String>>().join(", ");
```
Firstyear, line 1373: Why not just iter() and .join?

kanidmd/src/lib/idm/server.rs
```diff
@@ -1356,6 +1359,46 @@ impl<'a> IdmServerProxyWriteTransaction<'a> {
             })
     }
 
+    /// Generate a new set of backup code and remove the old ones.
+    pub fn generate_backup_code(
+        &mut self,
+        au: &mut AuditScope,
+        gbe: &GenerateBackupCodeEvent,
+    ) -> Result<String, OperationError> {
```
Firstyear, line 1367: Shouldn't we return a Vec<String> here to the client, and it's up to the front end utility to format this? Given this is structured data after all, and we don't want the client to have to str split on, when they can just process the json which is structured.

kanidmd/src/lib/utils.rs
```diff
@@ -37,6 +38,14 @@ pub fn password_from_random() -> String {
     rand_string
 }
 
+pub fn backup_code_from_random() -> HashSet<String> {
+    let mut set = std::collections::HashSet::with_capacity(8);
```
Firstyear, line 42: Minor, but can't you do:

```
(1..8).iter().map(|_| {
    readable_password_from_random()
})
.collect()
```

There is some type-hint magic in rust that pre-hints the capacity to the hashset during the collect so you don't need to do it manually.

kanidmd/src/lib/idm/authsession.rs
```diff
@@ -301,6 +309,23 @@ impl CredHandler {
...
+                        if backup_code.verify(&code_chal) {
+                            pw_mfa.mfa_state = CredVerifyState::Success;
+                            lsecurity!(
+                                au,
+                                "Handler::PasswordMfa -> Result::Continue - BackupCode OK, password -"
+                            );
+                            CredState::Continue(vec![AuthAllowed::Password])
```
Firstyear, line 319: Consider here adding a `DelayedAction` from kanidmd/src/lib/idm/delayed.rs which will remove code String from account with uuid. See "WebauthnCounterIncrement` for inspiration.

kanidmd/src/lib/idm/server.rs
Firstyear: Also consider adding a test into `kanidm_client/tests/proto_v1_test.rs` which actually tests these api interactions to create and retrieve the codes, then to use them from an external client perspective.

kanidmd/src/lib/idm/account.rs
```diff
@@ -357,6 +358,39 @@ impl Account {
...
+    // pub(crate) fn gen_backup_code_remove_mod(&self) -> Result<ModifyList<ModifyInvalid>, OperationError> {
+    //     match &self.primary {
+    //         // Change the cred
+    //         Some(primary) => {
+    //             let ncred = primary.remove_totp();
+    //             let vcred = Value::new_credential("primary", ncred);
+    //             Ok(ModifyList::new_purge_and_set("primary_credential", vcred))
+    //         }
+    //         None => {
+    //             // No credential exists, we can't remove what is not real.
```
Firstyear, line 388: You can remove anything if you try hard enough :)

Firstyear (commented):
Looking really good mate!

kanidm_proto/src/v1.rs
```diff
@@ -310,7 +310,7 @@ pub enum CredentialDetailType {
     GeneratedPassword,
     Webauthn(Vec<String>),
     /// totp, webauthn
-    PasswordMfa(bool, Vec<String>),
+    PasswordMfa(bool, Vec<String>, bool),
```
Firstyear, line 313: You may want this to be a number of how many codes are remaining so they know they may need to reset it soon if they are low.

kanidmd/src/lib/credential/mod.rs
```diff
@@ -214,30 +214,33 @@ impl Password {
 }
 
 #[derive(Clone, Debug)]
-pub struct BackupCode {
-    // TODO: any other attributes?
+pub struct BackupCodes {
+    // WIP_TODO: any other attributes?
```
Firstyear, line 218: For now this is okay :) but still good for it to be in a struct incase we add moe.

kanidmd/src/lib/credential/mod.rs
```diff
@@ -214,30 +214,33 @@ impl Password {
...
     code_set: HashSet<String>,
 }
 
-//TODO: No need a separate DbBackupCodeV1 struct probably?
-impl TryFrom<DbBackupCodeV1> for BackupCode {
+//WIP_TODO: No need a separate DbBackupCodeV1 struct probably?
```
Firstyear, line 222: No I think we should still have it as seperate incase we extend or change the type in the future.

kanidmd/src/lib/credential/mod.rs
```diff
@@ -768,58 +773,40 @@ impl Credential {
...
 
-    pub(crate) fn update_backup_code(&self, backup_code: HashSet<String>) -> Self {
-        let type_ = match &self.type_ {
-            // TODO: we should panic/reject CredentialType != PasswordMfa?
-            CredentialType::Password(pw) | CredentialType::GeneratedPassword(pw) => {
-                debug_assert!(false);
-                CredentialType::PasswordMfa(
+    pub(crate) fn update_backup_code(
+        &self,
+        backup_codes: BackupCodes, // Should I be passing reference??
```
Firstyear, line 778: No, because this function will take ownership of the backup_codes so it's correct to NOT use a reference here (if you used a reference you'd need to .clone() )

kanidmd/src/lib/idm/account.rs
```diff
@@ -360,14 +360,19 @@ impl Account {
 
     pub(crate) fn gen_backup_code_mod(
         &self,
-        backup_code: HashSet<String>,
+        backup_codes: BackupCodes, // Should I be passing reference??
```
Firstyear, line 363: See above :)

kanidmd/src/lib/idm/authsession.rs
```diff
@@ -309,8 +310,22 @@ impl CredHandler {
...
-                        if backup_code.verify(&code_chal) {
+                    (AuthCredential::BackupCode(code_chal), _, _, Some(backup_codes)) => {
+                        // Clone backup_codes because it's not mutable. If I make it a mutable ref,
+                        // then I have to pass it as &BackupCodes and specify lifetime on BackupCodeRemoval
+                        let mut cloned_codes = backup_codes.clone();
+                        if cloned_codes.verify(&code_chal) {
+                            if let Err(_e) =
+                                async_tx.send(DelayedAction::BackupCodeRemoval(BackupCodeRemoval {
+                                    target_uuid: who,
+                                    updated_codes: cloned_codes,
```
Firstyear, line 321: There is a problem here. Imagine we have two logins, and they each use a seperate backup code, A and B. We'll have two delayed actions queued. One set where A is NOT present, one where B is not present. So we'll "lose" one of the code removals. 

Also imagine the person reset's their backup codes, before this delayed action is processed. We would then process his and "reset back" all their codes! 

The correct action here is not to clone the codes to send in the delayed action, but to send the *code that was used*. and then the delayed action looks at the set within a write txn and removes that code, if present, or if not present, skips (because the codes were reset). 

This avoids your cloning issue too :) 

tl;dr - 
```
BackupCodeRemoval {
    target_uuid: Uuid,
    code_to_remove: String
}
```

kanidmd/src/lib/utils.rs
```diff
@@ -39,11 +39,10 @@ pub fn password_from_random() -> String {
 }
 
 pub fn backup_code_from_random() -> HashSet<String> {
-    let mut set = std::collections::HashSet::with_capacity(8);
-    for _ in 1..8 {
-        set.insert(readable_password_from_random());
-    }
-    set
+    (1..9)
```
Firstyear, line 42: Does this generate 8 codes? Just checking.

Firstyear (changes requested):

kanidmd/src/lib/idm/authsession.rs
victorcwai: Good catch. So when the delayed action is processed, I should retrieve the backup codes and remove `code_to_remove` from it, right?

kanidmd/src/lib/utils.rs
victorcwai: It inserts 8 codes. But on second thought, if the `readable_password_from_random()` returns the same string, then the HashSet won't have 8 codes.

victorcwai: I will change it to insert until set.size() <= 8

kanidmd/src/lib/core/https.rs
Firstyear: In the future there is a plan to have multiple credentials per account, but it's not fully fleshed out, but we want the routes to reflect this possible change for now.

kanidmd/src/lib/utils.rs
Firstyear: Just make it 0..8, that's clearer .... Also readable_password_from_random returns a new random PW each call.

kanidmd/src/lib/idm/authsession.rs
Firstyear: Correct! :D

kanidmd/src/lib/credential/mod.rs
Firstyear: Ahhh yes, your right, I misread :)

Firstyear (commented):
Hope these comments help :)

kanidmd/src/lib/idm/authsession.rs
```diff
@@ -1950,4 +1987,200 @@ mod tests {
...
+                &async_tx,
+                &webauthn,
+                Some(&pw_badlist_cache),
+                &hs512,
+            ) {
+                Ok(AuthState::Denied(msg)) => assert!(msg == BAD_PASSWORD_MSG),
+                _ => panic!(),
+            };
+        }
+        // Can't process BackupCodeRemoval without the server instance
```
Firstyear, line 2106: That's fine, you just have to deque here and check that the removal message *was* sent.

Firstyear (approved):
This looks really good. Is there anything else still missing here?

victorcwai:
Feel free to merge!

Firstyear:
Sure, do you also mind making a ticket for the kanidm_tools to be extended to create and reset these too? Shouldn't be a long jump from here to add that to the CLI.

PR #470, opened 2021-06-09
20210607 orca ldap

Firstyear:
This adds support for Orca to conduct load tests against 389-ds. This also resolves a few warnings and clippy lints. In the future this will allow us to perform performance comparisons between 389-ds and kanidm, and potentially even freeipa or similar. 

### DIFF
--- a/orca/src/ds.rs
+++ b/orca/src/ds.rs
@@ -1,0 +1,125 @@
+use crate::data::*;
+use crate::ldap::{LdapClient, LdapSchema};
+use crate::profile::DsConfig;
+use crate::{TargetServer, TargetServerBuilder};
+use ldap3_server::proto::*;
+use std::time::{Duration, Instant};
+use std::collections::{HashMap, HashSet};
+use uuid::Uuid;
+
+#[derive(Debug)]
+pub struct DirectoryServer {
+    ldap: LdapClient,
+    dm_pw: String,
+}
+
+impl DirectoryServer {
+    fn construct(uri: String, dm_pw: String,
+        basedn: String,
+    ) -> Result<Self, ()> {
+        let ldap = LdapClient::new(uri, basedn, LdapSchema::Rfc2307bis)?;
+
+        Ok(DirectoryServer { ldap, dm_pw })
+    }
+
+    pub fn build(uri: String, dm_pw: String,
+        basedn: String,
+    ) -> Result<TargetServer, ()> {
+        Self::construct(uri, dm_pw, basedn).map(TargetServer::DirSrv)
+    }
+
+    pub fn new(lconfig: &DsConfig) -> Result<TargetServer, ()> {
+        Self::construct(lconfig.uri.clone(), lconfig.dm_pw.clone(), lconfig.base_dn.clone()).map(TargetServer::DirSrv)
+    }
+
+    pub fn info(&self) -> String {
+        format!("Directory Server Connection: {}", self.ldap.uri)
+    }
+
+    pub fn builder(&self) -> TargetServerBuilder {
+        TargetServerBuilder::DirSrv(
+            self.ldap.uri.clone(),
+            self.dm_pw.clone(),
+            self.ldap.basedn.clone(),
+     )
+    }
+
+    pub async fn open_admin_connection(&self) -> Result<(), ()> {
+        self.ldap.open_dm_connection(&self.dm_pw).await
+    }
+
+    pub async fn setup_admin_delete_uuids(&self, targets: &[Uuid]) -> Result<(), ()> {
+        // We might hit admin limits depending on the dataset size, so we probably
+        // need to do this iteratively eventually. Or just change the limits ...
+
+        let filter = LdapFilter::Or(
+                targets.iter()
+                    .map(|u| LdapFilter::Equality("uid".to_string(), u.to_string()))
+                    .collect());
+
+        let res = self.ldap.search(filter).await?;
+
+        for ent in res.iter() {
+            self.ldap.delete(ent.dn.clone()).await?;
+        }
+        Ok(())
+    }
+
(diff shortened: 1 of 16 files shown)


Firstyear:
And just for you @yaleman it fixes all the orca clippy lints 📎

orca/src/ds.rs
```diff
@@ -0,0 +1,414 @@
...
+                    LdapAttribute {
+                        atype: "ou".to_string(),
+                        vals: vec!["people".to_string()],
+                    },
+                ],
+            };
+            self.ldap.add(ou_people).await?;
+        }
+
+        let res = self
```
victorcwai, line 109: Why do we need to run this twice?

orca/src/ds.rs
```diff
@@ -0,0 +1,414 @@
...
+        priv_account.sort_unstable();
+        priv_group.sort_unstable();
+        priv_account.dedup();
+        priv_group.dedup();
+        // Do the mod in one pass.
+        info!("Setting up cn=priv_group_manage");
+        let req = LdapModifyRequest {
+            dn: format!("cn=priv_group_manage,{}", self.ldap.basedn),
+            changes: vec![LdapModify {
+                operation: LdapModifyType::Delete,
```
victorcwai, line 370: This is deleting the priv_group?

orca/src/ds.rs
```diff
@@ -0,0 +1,414 @@
...
+                },
+            }],
+        };
+        let _ = self.ldap.modify(req).await;
+
+        info!("Setting up cn=priv_account_manage");
+        let req = LdapModifyRequest {
+            dn: format!("cn=priv_account_manage,{}", self.ldap.basedn),
+            changes: vec![LdapModify {
+                operation: LdapModifyType::Delete,
```
victorcwai, line 383: same as above

orca/src/kani.rs
```diff
@@ -330,11 +330,11 @@ impl KaniLdapServer {
         admin_pw: String,
         ldap_uri: String,
         basedn: String,
-    ) -> Result<Self, ()> {
+    ) -> Result<Box<Self>, ()> {
```
victorcwai, line 333: Why do we need `Box`?

orca/src/ldap.rs
```diff
@@ -197,26 +222,21 @@ impl LdapClient {
...
         })?;
 
+        let mut results = Vec::new();
         // It takes a lot more work to process a response from ldap :(
         while let Some(Ok(msg)) = inner.framed.next().await {
             match msg.op {
-                LdapOp::SearchResultEntry(_) => {
-                    count.fetch_add(1, Ordering::Relaxed);
-                }
+                LdapOp::SearchResultEntry(ent) => results.push(ent),
```
victorcwai, line 234: So we don't need Atomic operation for this anymore?

orca/src/runner/search.rs
```diff
@@ -30,7 +30,7 @@ struct CsvRow {
 
 fn basic_arbiter(
     mut broadcast_rx: tokio::sync::broadcast::Receiver<TestPhase>,
-    raw_results_rx: crossbeam::channel::Receiver<(Duration, Duration, usize)>,
+    raw_results_rx: &crossbeam::channel::Receiver<(Duration, Duration, usize)>,
```
victorcwai, line 33: What is the difference between `&crossbeam::channel::Receiver<(Duration, Duration, usize)>` and `crossbeam::channel::Receiver<(Duration, Duration, usize)>`?

victorcwai (commented):
Could you explain in general how orca works? Like how it performs load testing, what kind of API are we testing, how to measure the metrics, etc.?

orca/src/ds.rs
Firstyear: They search different things - one is searching for ou=people, the other, ou=groups.

orca/src/ds.rs
Firstyear: No, it's deleting the members of the priv_group (its an ldap modify so it's changing attributes, where if it was an LdapDelete it would be deleting the priv_group)

orca/src/kani.rs
Firstyear: Clippy! Clippy warned that this struct was getting "large" so it prefered to box the content.

orca/src/ldap.rs
Firstyear: Nope, because we return the results now so we can count the length that way instead.

orca/src/runner/search.rs
Firstyear: Again clippy - rather than pass the whole thing, pass by reference to save stack/call space.

Firstyear:
> Could you explain in general how orca works? Like how it performs load testing, what kind of API are we testing, how to measure the metrics, etc.?

The goal is to be able to compare and contrast the behaviour of Kanidm to other IDM services such as 389-ds, FreeIPA etc. This is different to say the criterion benches in kanidmd which are for improving kanidm's individual performance in isolation, allowing us to measure the impact of individual changes we make in the code. 

To do this we need to be able to show the behaviours that an administrator may consider valuable. The current test (search basic) measures search latency under moderate load. IE what is the average time of a search operation to complete, along with it's variance/SD. 

Some other goals for tests are:

* Maximum throughput (as many searches, authentications, writes  as possible)
* Test latency of searches while a write operation occurs in parallel IE latency impact due to mixed operation behaviour
* Latency of other operation types etc

The idea is that with the DS/Kanidm current components in place, we can write "generic" tests on top that know how to work with many types of backends so we can do these in a fair(ish) manner, that we can then compare. 

Does that help?

victorcwai:
Yep that helps a lot. Please go ahead and merge. :)

Firstyear:
Thank you!

Issue #456, opened 2021-05-24
When running restores, need to validate access to the file first and bail if not.


yaleman:
I tried to run a restore as a user who didn't have access and `r2d2` just kept throwing I/O errors until I killed the process.

Probably need to validate access to things before trying to use them.

Log:
```
kanidm:/var/lib/kanidm # sudo -u kanidm kanidmd restore kanidm.backup.json --config /etc/kanidm/kanidmd.toml
<snipping errors about other permissions>
Running in restore mode ...
[r2d2] disk I/O error
[r2d2] disk I/O error
[r2d2] disk I/O error
[r2d2] disk I/O error
[r2d2] disk I/O error
[r2d2] disk I/O error
```

yaleman:
Happens in `kanidm_unixd` too, when the user doesn't have permissions to create the database file:

```
May 27 22:32:37 monitoring kanidm_unixd[5233]: [2021-05-27T12:32:37Z ERROR r2d2] unable to open database file: /var/cache/kanidm-unixd/kanidm.cache.db
```

Firstyear:
Does your PR resolve this?

yaleman:
I think it only solves the latter unixd bits, I need to test restores more 😊

yaleman:
Yep, still... doesn't fail on trying to restore.
```
yaleman@kanidm:~> /usr/local/sbin/kanidmd restore /tmp/kanidm.backup.json  --config ~/kanidmd.toml
WARNING: permissions on /home/yaleman/kanidmd.toml may not be secure. Should be readonly to running uid. This could be a security risk ...
WARNING: /home/yaleman/kanidmd.toml owned by the current uid, which may allow file permission changes. This could be a security risk ...
Running in restore mode ...
[r2d2] unable to open database file: /var/lib/kanidm/kanidm.db
[r2d2] unable to open database file: /var/lib/kanidm/kanidm.db
<repeat>
```

PR #476, opened 2021-06-14
unixd will now bail if startup tests fail

yaleman:
Relates to #456

Checks to see if `kanidm_unixd` can actually start and run and write to the DB path

- [x] cargo fmt has been run

### DIFF
--- a/kanidm_unix_int/src/daemon.rs
+++ b/kanidm_unix_int/src/daemon.rs
@@ -121,9 +121,13 @@ impl TaskCodec {
 }
 
 fn rm_if_exist(p: &str) {
-    let _ = std::fs::remove_file(p).map_err(|e| {
-        warn!("attempting to remove {:?} -> {:?}", p, e);
-    });
+    if Path::new(p).exists() {
+        let _ = std::fs::remove_file(p).map_err(|e| {
+            warn!("attempting to remove {:?} -> {:?}", p, e);
+        });
+    } else {
+        debug!("Path {:?} doesn't exist, not attempting to remove.", p);
+    }
 }
 
 async fn handle_task_client(
@@ -381,6 +385,15 @@ async fn main() {
             std::process::exit(1);
         }
     };
+    if !cfg_path.exists() {
+        // there's no point trying to start up if we can't read a usable config!
+        error!(
+            "Client config missing from {} - cannot start up. Quitting.",
+            cfg_path_str
+        );
+        std::process::exit(1);
+    }
+
     if cfg_path.exists() {
         let cfg_meta = match metadata(&cfg_path) {
             Ok(v) => v,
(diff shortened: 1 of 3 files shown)


kanidm_unix_int/src/daemon.rs
```diff
@@ -121,9 +121,13 @@ impl TaskCodec {
 }
 
 fn rm_if_exist(p: &str) {
-    let _ = std::fs::remove_file(p).map_err(|e| {
-        warn!("attempting to remove {:?} -> {:?}", p, e);
-    });
+    if Path::new(p).exists() {
+        let _ = std::fs::remove_file(p).map_err(|e| {
+            warn!("attempting to remove {:?} -> {:?}", p, e);
```
Firstyear, line 126: This should be an error then if it fails shouldn't it since we know it exists.

kanidm_unix_int/src/daemon.rs
```diff
@@ -410,7 +423,14 @@ async fn main() {
             std::process::exit(1);
         }
     };
-    if unixd_path.exists() {
+    if !unixd_path.exists() {
+        // there's no point trying to start up if we can't read a usable config!
+        error!(
+            "Client config missing from {} - cannot start up. Quitting.",
```
Firstyear, line 429: Should be Unix config missing I think

kanidm_unix_int/src/daemon.rs
```diff
@@ -513,6 +533,36 @@ async fn main() {
...
+                Err(e) => {
+                    error!(
+                        "Unable to read metadata for {} - {:?}",
+                        db_path.to_str().unwrap_or_else(|| "<db_path invalid>"),
+                        e
+                    );
+                    std::process::exit(1);
+                }
+            };
+            if db_metadata.permissions().readonly() {
```
Firstyear, line 558: The rust readonly flag is really bad, so you may not want to check that here. It checks if ALL write bits do not exist, not the write bits for this USER to be able to access it.  So I think it's not really possible to do this.

Firstyear (changes requested):

Firstyear:
Other than some minor comments, looks great mate 🎉

kanidm_unix_int/src/daemon.rs
yaleman: Ack, I re-read the docs and ew. It's possible to do the right dance, but out of scope for this one. :(

kanidm_unix_int/src/daemon.rs
yaleman: Ack, fixing.

Firstyear (approved):

Firstyear:
Always, appreciate your work :D

Issue #473, opened 2021-06-11
`DbValue` serde renames


QnnOkabayashi:
The field names in `kanidmd/src/lib/be/dbvalue.rs` types are very short to minimize information that is serialized, but this makes it much more confusing for new contributors. We can use `serde` macros to have the structs serialized with much shorter names, while allowing us to use more human readable code. I imagine it would use something like `#[serde(rename = "x")]` in the struct definition.

For example, the `DbValueV1` type could be rewritten to something like:
```rust
#[derive(Serialize, Deserialize, Debug)]
pub enum DbValueV1 {
    ...
    #[serde(rename = "BO")]
    Bool(bool),
    #[serde(rename = "SY")]
    SynType(usize),
    #[serde(rename = "IN")]
    IdxType(usize),
    ...
}
```

And in `DbValueCredV1` for example, the 1-letter variable names completely obfuscate the meaning, and should be changed. 

```rust
#[derive(Serialize, Deserialize, Debug)]
pub struct DbValueCredV1 {
    #[serde(rename = "t")]
    pub this_starts_with_t: String,
    #[serde(rename = "d")]
    pub dunno_what_this_is: DbCredV1,
}
```

(I have no idea what most of these are short for. Hence the issue here.)

QnnOkabayashi:
I think `serde` renaming could actually come in handy in a lot more places too. `kanidmd/src/lib/value.rs` defines the `SyntaxType` and `IndexType`, which have methods to convert their variants to and from unique `usize`s. This causes a lot of clutter, and is only used in one place (according to VSCode) for serialization. Someone might have to dig around more here, though. Also, although fieldless `enum`s can be casted to `usize`s, we don't want anything to break if the fields are reordered.

For example:
```rust
pub enum SyntaxType {
    #[serde(rename = "0")] UTF8STRING,
    #[serde(rename = "1")] Utf8StringInsensitive,
    ...
    #[serde(rename = "16")] DateTime,
    #[serde(rename = "17")] EmailAddress,
}
```

Firstyear:
There is also repr(usize) potentially that we could use too.

QnnOkabayashi:
> There is also repr(usize) potentially that we could use too.

True, but as you mentioned before this could potentially screw up things if the `enum` is reordered when things are already stored in the db. Can we just assume that this won't happen? If possible, could we rename them to shorter string variations like I proposed for `DbValueV1` above?

I do agree that renaming them in numeric ascending order does seem silly though...

PR #477, opened 2021-06-14
Renamed fields in `dbvalue`

QnnOkabayashi:
Fixes https://github.com/kanidm/kanidm/issues/473

- [x] cargo fmt has been run

How do I check that the shortened names are actually what's getting serialized, and not the variant names directly? Also, There are some fields that I still can't infer the meaning of, help appreciated :)

### DIFF
--- a/kanidmd/src/lib/be/dbvalue.rs
+++ b/kanidmd/src/lib/be/dbvalue.rs
@@ -4,13 +4,12 @@ use webauthn_rs::proto::COSEKey;
 
 #[derive(Serialize, Deserialize, Debug)]
 pub struct DbCidV1 {
-    //? what do `d`, `s`, and `t` stand for? I wasn't able to infer it.
     #[serde(rename = "d")]
-    pub d: Uuid,
+    pub domain_id: Uuid,
     #[serde(rename = "s")]
-    pub s: Uuid,
+    pub server_id: Uuid,
     #[serde(rename = "t")]
-    pub t: Duration,
+    pub timestamp: Duration,
 }
 
 #[derive(Serialize, Deserialize, Debug)]
@@ -19,10 +18,6 @@ pub enum DbPasswordV1 {
     SSHA512(Vec<u8>, Vec<u8>),
 }
 
-//? This type is basically an alias for `TotpAlgo`.
-//? Why don't we do
-//? `type DbTotpAlgoV1 = TotpAlgo` and then
-//? derive `Serialize`/`Deserialize` on `TotpAlgo`?
 #[derive(Serialize, Deserialize, Debug)]
 pub enum DbTotpAlgoV1 {
     S1,
@@ -33,9 +28,9 @@ pub enum DbTotpAlgoV1 {
 #[derive(Serialize, Deserialize, Debug)]
 pub struct DbTotpV1 {
     #[serde(rename = "l")]
-    pub l: String, //? is this short for "label"?
+    pub label: String,
     #[serde(rename = "k")]
-    pub k: Vec<u8>, //? is this short for "secret"?
+    pub key: Vec<u8>,
     #[serde(rename = "s")]
     pub step: u64,
     #[serde(rename = "a")]
@@ -45,15 +40,15 @@ pub struct DbTotpV1 {
 #[derive(Serialize, Deserialize, Debug)]
 pub struct DbWebauthnV1 {
     #[serde(rename = "l")]
-    pub l: String, //? is this short for "label"?
+    pub label: String,
     #[serde(rename = "i")]
-    pub cred_id: Vec<u8>,
+    pub id: Vec<u8>,
     #[serde(rename = "c")]
     pub cred: COSEKey,
     #[serde(rename = "t")]
     pub counter: u32,
     #[serde(rename = "v")]
-    pub is_verified: bool,
+    pub verified: bool,
 }
 
 #[derive(Serialize, Deserialize, Debug)]
@@ -103,8 +98,7 @@ pub struct DbValueTaggedStringV1 {
 
 #[derive(Serialize, Deserialize, Debug)]
 pub struct DbValueEmailAddressV1 {
-    #[serde(rename = "d")]
-    pub email_addr: String,
+    pub d: String,
 }
 
 #[derive(Serialize, Deserialize, Debug)]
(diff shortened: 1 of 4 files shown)


kanidmd/src/lib/be/dbvalue.rs
```diff
@@ -4,8 +4,12 @@ use webauthn_rs::proto::COSEKey;
 
 #[derive(Serialize, Deserialize, Debug)]
 pub struct DbCidV1 {
+    //? what do `d`, `s`, and `t` stand for? I wasn't able to infer it.
+    #[serde(rename = "d")]
```
Firstyear, line 8: Domain ID, Server ID, TimeStamp

kanidmd/src/lib/be/dbvalue.rs
```diff
@@ -15,6 +19,10 @@ pub enum DbPasswordV1 {
     SSHA512(Vec<u8>, Vec<u8>),
 }
 
+//? This type is basically an alias for `TotpAlgo`.
+//? Why don't we do
+//? `type DbTotpAlgoV1 = TotpAlgo` and then
+//? derive `Serialize`/`Deserialize` on `TotpAlgo`?
```
Firstyear, line 25: Because we may need to change the on disk format - it doesn't matter if they are equivalent, we need them seperate so we have the flexibility to change these in the future if they diverge and we need to perform transformations.

kanidmd/src/lib/be/dbvalue.rs
```diff
@@ -24,19 +32,28 @@ pub enum DbTotpAlgoV1 {
 
 #[derive(Serialize, Deserialize, Debug)]
 pub struct DbTotpV1 {
-    pub l: String,
-    pub k: Vec<u8>,
-    pub s: u64,
-    pub a: DbTotpAlgoV1,
+    #[serde(rename = "l")]
+    pub l: String, //? is this short for "label"?
```
Firstyear, line 36: label

kanidmd/src/lib/be/dbvalue.rs
```diff
@@ -24,19 +32,28 @@ pub enum DbTotpAlgoV1 {
...
+    #[serde(rename = "k")]
+    pub k: Vec<u8>, //? is this short for "secret"?
```
Firstyear, line 38: key or secret

kanidmd/src/lib/be/dbvalue.rs
```diff
@@ -24,19 +32,28 @@ pub enum DbTotpAlgoV1 {
...
 
 #[derive(Serialize, Deserialize, Debug)]
 pub struct DbWebauthnV1 {
-    pub l: String,
-    pub i: Vec<u8>,
-    pub c: COSEKey,
-    pub t: u32,
-    pub v: bool,
+    #[serde(rename = "l")]
+    pub l: String, //? is this short for "label"?
```
Firstyear, line 48: label

kanidmd/src/lib/be/dbvalue.rs
```diff
@@ -24,19 +32,28 @@ pub enum DbTotpAlgoV1 {
...
+    #[serde(rename = "i")]
+    pub cred_id: Vec<u8>,
```
Firstyear, line 50: id rather than cred id maybe

kanidmd/src/lib/be/dbvalue.rs
```diff
@@ -24,19 +32,28 @@ pub enum DbTotpAlgoV1 {
...
+    #[serde(rename = "c")]
+    pub cred: COSEKey,
+    #[serde(rename = "t")]
+    pub counter: u32,
+    #[serde(rename = "v")]
+    pub is_verified: bool,
```
Firstyear, line 56: verified rather than is_verified I think

kanidmd/src/lib/be/dbvalue.rs
```diff
@@ -70,41 +87,64 @@ pub struct DbCredV1 {
...
+    pub tag: String,
+    #[serde(rename = "d")]
+    pub data: String,
 }
 
 #[derive(Serialize, Deserialize, Debug)]
 pub struct DbValueEmailAddressV1 {
-    pub d: String,
+    #[serde(rename = "d")]
+    pub email_addr: String,
```
Firstyear, line 107: d for data here I think, email address is already clear from the struct name.

kanidmd/src/lib/be/dbvalue.rs
```diff
@@ -70,41 +87,64 @@ pub struct DbCredV1 {
...
+    #[serde(rename = "I8")]
+    Iutf8(String),
+    #[serde(rename = "N8")]
+    Iname(String),
+    #[serde(rename = "UU")]
+    Uuid(Uuid),
+    #[serde(rename = "BO")]
+    Bool(bool),
+    #[serde(rename = "SY")]
+    SynType(usize),
```
Firstyear, line 123: SyntaxType

kanidmd/src/lib/be/dbvalue.rs
```diff
@@ -70,41 +87,64 @@ pub struct DbCredV1 {
...
+    #[serde(rename = "IN")]
+    IdxType(usize),
```
Firstyear, line 125: IndexType

kanidmd/src/lib/be/dbvalue.rs
```diff
@@ -70,41 +87,64 @@ pub struct DbCredV1 {
...
+    #[serde(rename = "RF")]
+    Refer(Uuid),
```
Firstyear, line 127: Reference

kanidmd/src/lib/be/dbvalue.rs
```diff
@@ -70,41 +87,64 @@ pub struct DbCredV1 {
...
+    #[serde(rename = "JF")]
+    JsonFilter(String),
+    #[serde(rename = "CR")]
+    Cred(DbValueCredV1),
```
Firstyear, line 131: Credential

kanidmd/src/lib/be/dbvalue.rs
```diff
@@ -70,41 +87,64 @@ pub struct DbCredV1 {
...
+    #[serde(rename = "RU")]
+    RadiusCred(String),
```
Firstyear, line 133: I'm about to rename this to "SecretValue" in my oauth2 pr, so change RadiusCred to SecretValue, but leave it as RU on the db format.

Firstyear (changes requested):
Beside some nits/names, this looks really good!

kanidmd/src/lib/be/dbvalue.rs
QnnOkabayashi: I'm going to use `key` because `s` is already used for `step`.

kanidmd/src/lib/be/dbvalue.rs
QnnOkabayashi: Why not use a tuple struct then? (I'm not sure how this would effect serialization though.)

QnnOkabayashi:
I'm a bit confused about the tests. GitHub says they failed here, and it failed on my machine at first, but when I ran them again without changing anything they all passed. What's going on?

yaleman:
Some of the tests are flaky :( There’s issues raised about them, still trying to work out the root cause.

Firstyear:
@QnnOkabayashi There seems to be a case in random number generation in openssl that sometimes fails to generate a valid key for the webauthn softtoken. This is what causes the test failure :(

Firstyear (approved):

Issue #472, opened 2021-06-11
Cleanup `kanidm_client` bool/return values


QnnOkabayashi:
In `kanidm_client`, there are many functions that return `Result<bool, ClientError>`, yet the `bool` is never used on `Ok`. We should change these to `Result<(), ClientError>` to decrease confusion. Inspired by @Firstyear's comment:

> It represents past williams mistakes. I thought that there would be a case where you could set and it would be Ok() but change nothing (false), but I think we never indicate that so it should be cleaned up. That could be a new issue from this (cleanup kanidm_client bool/return values).
> _Originally posted by @Firstyear in https://github.com/kanidm/kanidm/pull/465#discussion_r647877749_

Firstyear:
@QnnOkabayashi did you want to do this one? Or do you want me to fix it?

QnnOkabayashi:
I was gonna fix it Monday

PR #479, opened 2021-06-14
`kanidm_client` bool/return values

QnnOkabayashi:
Fixes https://github.com/kanidm/kanidm/issues/472

- [x] cargo fmt has been run

My changes were able to satisfy the type checker but I'm having trouble with some JSON somewhere getting deserialized where it expects a `unit` and finds a `bool`. I need help with finding the cause of this and how to fix it. Thanks!

### DIFF
--- a/kanidm_client/src/asynchronous.rs
+++ b/kanidm_client/src/asynchronous.rs
@@ -273,7 +273,7 @@ impl KanidmAsyncClient {
             .map_err(|e| ClientError::JsonDecode(e, opid))
     }
 
-    async fn perform_delete_request(&self, dest: &str) -> Result<bool, ClientError> {
+    async fn perform_delete_request(&self, dest: &str) -> Result<(), ClientError> {
         let dest = format!("{}{}", self.addr, dest);
 
         let response = self
@@ -315,11 +315,12 @@ impl KanidmAsyncClient {
             .await
             .map_err(|e| ClientError::JsonDecode(e, opid))
     }
+
     async fn perform_delete_request_with_body<R: Serialize>(
         &self,
         dest: &str,
         request: R,
-    ) -> Result<bool, ClientError> {
+    ) -> Result<(), ClientError> {
         let dest = format!("{}{}", self.addr, dest);
 
         let req_string = serde_json::to_string(&request).map_err(ClientError::JsonEncode)?;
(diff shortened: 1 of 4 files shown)


kanidm_client/src/asynchronous.rs
```diff
@@ -96,13 +96,13 @@ impl KanidmAsyncClient {
             let mut sguard = self.auth_session_id.write().await;
             *sguard = headers
                 .get(KSESSIONID)
-                .map(|hv| hv.to_str().ok().map(|s| s.to_string()))
+                .map(|hv| hv.to_str().ok().map(str::to_string))
```
Firstyear, line 99: As much as I love the cleanups, maybe we should all get better at cleaning them in seperate PR's so it's clearer to review. I'm as much guilty, so I need to improve too :)

kanidm_client/src/asynchronous.rs
```diff
@@ -656,22 +657,22 @@ impl KanidmAsyncClient {
...
         let c = CreateRequest { entries };
         let r: Result<OperationResponse, _> = self.perform_post_request("/v1/raw/create", c).await;
-        r.map(|_| true)
+        r.map(|_| ())
     }
 
-    pub async fn modify(&self, filter: Filter, modlist: ModifyList) -> Result<bool, ClientError> {
+    pub async fn modify(&self, filter: Filter, modlist: ModifyList) -> Result<(), ClientError> {
         let mr = ModifyRequest { filter, modlist };
         let r: Result<OperationResponse, _> = self.perform_post_request("/v1/raw/modify", mr).await;
```
Firstyear, line 668: It may be worth another ticket to remove OperationResponse as I think it does nothing.

Firstyear (approved):

Firstyear:
Do you mind opening a ticket to remove OperationResponse too? Otherwise looks great, I'm happy to merge it if you want. :)

QnnOkabayashi:
> Do you mind opening a ticket to remove OperationResponse too? Otherwise looks great, I'm happy to merge it if you want. :)

This currently fails a bunch of tests, are we ignoring those?

kanidm_client/src/asynchronous.rs
QnnOkabayashi: Agreed, I can open a ticket once this gets merged to avoid conflicts

Firstyear:
> > Do you mind opening a ticket to remove OperationResponse too? Otherwise looks great, I'm happy to merge it if you want. :)
> 
> This currently fails a bunch of tests, are we ignoring those?

I think these failures are introduced because of this change :) 

```
thread 'test_default_entries_rbac_users' panicked at 'called `Result::unwrap()` on an `Err` value: JsonDecode(reqwest::Error { kind: Decode, source: Error("invalid type: boolean `true`, expected unit", line: 1, column: 4) }, "d9ab7178-5124-416a-8183-81d277f1d9a3")', kanidm_client/tests/default_entries.rs:67:56
```

These probably need to be resolved by you before we merge these. Most likely I think that you may have missed bool -> () in async client for idm_group_add_members as I can't see this in your diff?

QnnOkabayashi:
> > > Do you mind opening a ticket to remove OperationResponse too? Otherwise looks great, I'm happy to merge it if you want. :)
> > 
> > 
> > This currently fails a bunch of tests, are we ignoring those?
> 
> I think these failures are introduced because of this change :)
> 
> ```
> thread 'test_default_entries_rbac_users' panicked at 'called `Result::unwrap()` on an `Err` value: JsonDecode(reqwest::Error { kind: Decode, source: Error("invalid type: boolean `true`, expected unit", line: 1, column: 4) }, "d9ab7178-5124-416a-8183-81d277f1d9a3")', kanidm_client/tests/default_entries.rs:67:56
> ```
> 
> These probably need to be resolved by you before we merge these. Most likely I think that you may have missed bool -> () in async client for idm_group_add_members as I can't see this in your diff?

I did change `idm_group_add_members` in `asynchronous.rs` [here](https://github.com/kanidm/kanidm/pull/479/files/3c5cc00a3e55db1b04a50b8acc7e1779eeca797a#diff-65b0009dfc1559162e33b7c9e43e1d45c2857c9c181423210c86dee3770fe368R721-R724), the function name is just not previews so it doesn't show when doing cmd+F. 

The error msg makes me thing there's a JSON object that's being loaded, which makes sense because it makes it past the type checker as a runtime error. The issue is I don't even know where this JSON object is coming from.

PR #480, opened 2021-06-15
fixes #478 - adds note about web ui already being packaged

### LINKED ISSUE #478: Mention Precompiled WASM Artefact in Docs
Hi @yaleman,

as requested, here is an issue for the WASM artefacts.

For creating kanidm package as well as getting started developing, especially the server, it would be good to mention, that there are precompiled wasm artefacts in `kanidmd_web_ui/pkg`. 

So there would be no need to build it, as it can just be used from there.
For server packages, it would be important to include this in order to use the UI. (and to successfully startup the server, if not turned off https://github.com/kanidm/kanidm/pull/432/files )

We might want to update the [DEVELOPER_README.md ](https://github.com/kanidm/kanidm/blob/master/DEVELOPER_README.md)?

Let me know if I can help here.

yaleman:
Fixes #478

### DIFF
--- a/DEVELOPER_README.md
+++ b/DEVELOPER_README.md
@@ -125,7 +125,7 @@ In a new terminal, you can now build and run the client tools with:
     cargo run -- login -H https://localhost:8443 -D admin -C ../insecure/ca.pem
     cargo run -- self whoami -H https://localhost:8443 -D admin -C ../insecure/ca.pem
 
-## Building the Web UI
+### Building the Web UI
 
 __NOTE:__ there is a pre-packaged version of the Web UI at `/kanidmd_web_ui/pkg/`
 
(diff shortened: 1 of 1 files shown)


DEVELOPER_README.md
```diff
@@ -129,12 +127,14 @@ In a new terminal, you can now build and run the client tools with:
 
 ### Building the Web UI
 
-The web ui uses rust wasm rather than javascript. To build this you need to setup the environment.
+__NOTE:__ there is a pre-packaged version of the Web UI at `/kanidmd_web_ui/pkg/`
```
cuberoot74088, line 130: Cool. How about adding this:

```suggestion
__NOTE:__ There is a pre-packaged version of the Web UI at `/kanidmd_web_ui/pkg/`, which can be used directly.
```

cuberoot74088 (commented):
Thanks for creating the note. :+1:

Firstyear:
Thank you all!

PR #490, opened 2021-06-19
WIP check user shell (#392)

### LINKED ISSUE #392: Add fall back shell option
If the shell doesn't exist, we should have a fallback option available.

cuberoot74088:
* this is work in progress
  * asking for your comments / feedback
 
* check if the provided `shell` exists
*  if not set it to fall back shell
* test using
  `cargo test -- --nocapture test_cache_account` 
  will give some insights.

Fixes #392

- [ ] cargo fmt has been run

### DIFF
--- a/kanidm_unix_int/src/cache.rs
+++ b/kanidm_unix_int/src/cache.rs
@@ -1,4 +1,3 @@
-use crate::constants::DEFAULT_SHELL;
 use crate::db::Db;
 use crate::unix_config::{HomeAttr, UidAttr};
 use crate::unix_proto::{HomeDirectoryInfo, NssGroup, NssUser};
@@ -278,7 +277,7 @@ impl CacheLayer {
             // bad luck - we have to put our fall back DEFAULT_SHELL
             eprintln!("x392 WIP - BAD user_shell <{}> missing.", user_shell);
             // using constants.DEFAULT_SHELL
-            DEFAULT_SHELL.to_string()
+            self.default_shell.clone()
         };
         eprintln!(
             "x392 WIP - FIN we will set <{}> as shell_to_use.",
@@ -666,11 +665,7 @@ impl CacheLayer {
                     name: self.token_uidattr(&tok),
                     gid: tok.gidnumber,
                     gecos: tok.displayname,
-                    // WIP #392: default shell
-                    // this should now be valid, so simply unwrap it?
-                    // where and how is `self.default_shell` set?
-                    shell: tok.shell.unwrap_or_default(),
-                    //shell: tok.shell.unwrap_or_else(|| self.default_shell.clone()),
+                    shell: tok.shell.unwrap_or_else(|| self.default_shell.clone()),
                 })
                 .collect()
         })
--- a/kanidm_unix_int/src/cache.rs
+++ b/kanidm_unix_int/src/cache.rs
@@ -667,7 +667,7 @@ impl CacheLayer {
                     gid: tok.gidnumber,
                     gecos: tok.displayname,
                     // WIP #392: default shell
-                    // this should now be valid, so simply unwrap?
+                    // this should now be valid, so simply unwrap it?
                     // where and how is `self.default_shell` set?
                     shell: tok.shell.unwrap_or_default(),
                     //shell: tok.shell.unwrap_or_else(|| self.default_shell.clone()),
--- a/kanidm_unix_int/src/cache.rs
+++ b/kanidm_unix_int/src/cache.rs
@@ -1,3 +1,4 @@
+use crate::constants::DEFAULT_SHELL;
 use crate::db::Db;
 use crate::unix_config::{HomeAttr, UidAttr};
 use crate::unix_proto::{HomeDirectoryInfo, NssGroup, NssUser};
@@ -8,6 +9,7 @@ use lru::LruCache;
 use reqwest::StatusCode;
 use std::collections::BTreeSet;
 use std::ops::{Add, Sub};
+use std::path::Path;
 use std::string::ToString;
 use std::time::{Duration, SystemTime};
 use tokio::sync::{Mutex, RwLock};
@@ -264,17 +266,36 @@ impl CacheLayer {
             .map_err(|e| {
                 error!("time conversion error - ex_time less than epoch? {:?}", e);
             })?;
+        // WIP #392: check user/default shell
+        // check if the provided `shell` exists on the system.
+        let mut new_token = token.clone();
+        let user_shell = new_token.shell.unwrap_or_default();
+        let shell_to_use = if Path::new(&user_shell).exists() {
+            // good - the shell exists
+            eprintln!("x392 WIP - OK user_shell <{}> exists.", user_shell);
+            user_shell
+        } else {
+            // bad luck - we have to put our fall back DEFAULT_SHELL
+            eprintln!("x392 WIP - BAD user_shell <{}> missing.", user_shell);
+            // using constants.DEFAULT_SHELL
+            DEFAULT_SHELL.to_string()
+        };
+        eprintln!(
+            "x392 WIP - FIN we will set <{}> as shell_to_use.",
+            shell_to_use
+        );
+        new_token.shell = shell_to_use.into();
 
         let dbtxn = self.db.write().await;
         // We need to add the groups first
-        token
+        new_token
             .groups
             .iter()
             .try_for_each(|g| dbtxn.update_group(g, offset.as_secs()))
             .and_then(|_|
                 // So that when we add the account it can make the relationships.
                 dbtxn
-                    .update_account(token, offset.as_secs()))
+                    .update_account(&new_token, offset.as_secs()))
             .and_then(|_| dbtxn.commit())
     }
 
@@ -645,7 +666,11 @@ impl CacheLayer {
                     name: self.token_uidattr(&tok),
                     gid: tok.gidnumber,
                     gecos: tok.displayname,
-                    shell: tok.shell.unwrap_or_else(|| self.default_shell.clone()),
+                    // WIP #392: default shell
+                    // this should now be valid, so simply unwrap?
+                    // where and how is `self.default_shell` set?
+                    shell: tok.shell.unwrap_or_default(),
+                    //shell: tok.shell.unwrap_or_else(|| self.default_shell.clone()),
                 })
                 .collect()
         })
--- a/kanidm_unix_int/src/cache.rs
+++ b/kanidm_unix_int/src/cache.rs
@@ -257,7 +257,7 @@ impl CacheLayer {
         }
     }
 
-    async fn set_cache_usertoken(&self, token: &UnixUserToken) -> Result<(), ()> {
+    async fn set_cache_usertoken(&self, token: &mut UnixUserToken) -> Result<(), ()> {
         // Set an expiry
         let ex_time = SystemTime::now() + Duration::from_secs(self.timeout_seconds);
         let offset = ex_time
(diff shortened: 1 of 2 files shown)


cuberoot74088:
Of course I will remove the `eprintln` once everyone is happy with it.

yaleman:
Looks good so far, the current messages could be dropped to `debug!()` or `info!()` once tested :)

cuberoot74088:
@yaleman, thanks for having a look!
Sure, if we want to keep it, we can do it like that.

About tests:
I have just pushed a simple test (https://github.com/kanidm/kanidm/pull/490/commits/e41cc46cf21d0cec97127de7b6d43d36a76f1026) to see, if the `DEFAULT_SHELL` is set, for the given example with `shell=None`.
Do we want to test, if it would also keep existing shells, i.e., `/bin/bash`, or replace some non-existing `/bin/not-a-real-shell`?

If so, how to do it? Add another `testaccount`?

yaleman:
> @yaleman, thanks for having a look!
> Sure, if we want to keep it, we can do it like that.
> 
> About tests:
> I have just pushed a simple test ([e41cc46](https://github.com/kanidm/kanidm/commit/e41cc46cf21d0cec97127de7b6d43d36a76f1026)) to see, if the `DEFAULT_SHELL` is set, for the given example with `shell=None`.
> Do we want to test, if it would also keep existing shells, i.e., `/bin/bash`, or replace some non-existing `/bin/not-a-real-shell`?
> 
> If so, how to do it? Add another `testaccount`?

That one could be done with the DEFAULT_TEST_ACCOUNT, and more test options are better than less, I'd say 😄 

The main issue for testing would be things that reference the filesystem since kanidm is developed on multiple platforms and you don't want it failing if someone ran tests on a windows machine or a cut down docker container without all the "usual" shells.

Firstyear:
> The main issue for testing would be things that reference the filesystem since kanidm is developed on multiple platforms and you don't want it failing if someone ran tests on a windows machine or a cut down docker container without all the "usual" shells.

Given this is the unixd integration, and it relies on unix sockets, it already won't work on windows. So I think it's okay to keep unix specific behaviours in this area of the code.

kanidm_unix_int/src/cache.rs
```diff
@@ -264,17 +266,36 @@ impl CacheLayer {
             .map_err(|e| {
                 error!("time conversion error - ex_time less than epoch? {:?}", e);
             })?;
+        // WIP #392: check user/default shell
+        // check if the provided `shell` exists on the system.
+        let mut new_token = token.clone();
+        let user_shell = new_token.shell.unwrap_or_default();
```
Firstyear, line 272: I don't think this is very clear - UnixUserToken doesn't define a default for the `shell: Option<String>`, which means this will hit `https://doc.rust-lang.org/std/string/struct.String.html#impl-Default` so it will return an empty string. It's a weird behaviour to rely on, so it's likely better to change this logic to show the possibilities more clearly:

* The user did not request a shell, so we set our fall back
* The requested shell does not exist, so we set our fall back
* The requested shell does exist, so we respect it.

```
let requested_shell_exists: bool = new_token.shell
    .as_ref()
    .map(|shell| {
        let exists = Path::new(&shell).exists();
        if !exists {
            info!("User requested shell is not present on this system - {}", shell)
        }
        exists
    })
    unwrap_or_else(|| {
        info!("User has not specified a shell, using default");
        false
    });

if !requested_shell_exists {
    token.shell = Some(self.default_shell.clone())
}
```

kanidm_unix_int/src/cache.rs
```diff
@@ -264,17 +266,36 @@ impl CacheLayer {
             .map_err(|e| {
                 error!("time conversion error - ex_time less than epoch? {:?}", e);
             })?;
+        // WIP #392: check user/default shell
+        // check if the provided `shell` exists on the system.
+        let mut new_token = token.clone();
```
Firstyear, line 271: You don't need this clone. Change this line:

```
async fn set_cache_usertoken(&self, token: UnixUserToken) -> Result<(), ()> {
```

to

```
async fn set_cache_usertoken(&self, mut token: UnixUserToken) -> Result<(), ()> {
```

kanidm_unix_int/src/cache.rs
```diff
@@ -264,17 +266,36 @@ impl CacheLayer {
...
+        eprintln!(
+            "x392 WIP - FIN we will set <{}> as shell_to_use.",
+            shell_to_use
+        );
+        new_token.shell = shell_to_use.into();
 
         let dbtxn = self.db.write().await;
         // We need to add the groups first
-        token
+        new_token
```
Firstyear, line 291: If you remove the clone, you won't need this change :)

kanidm_unix_int/src/cache.rs
```diff
@@ -645,7 +666,11 @@ impl CacheLayer {
                     name: self.token_uidattr(&tok),
                     gid: tok.gidnumber,
                     gecos: tok.displayname,
-                    shell: tok.shell.unwrap_or_else(|| self.default_shell.clone()),
+                    // WIP #392: default shell
+                    // this should now be valid, so simply unwrap it?
+                    // where and how is `self.default_shell` set?
+                    shell: tok.shell.unwrap_or_default(),
```
Firstyear, line 672: This line needs to be removed. There may be old, cached information in the sqlite store for accounts where the account is None, and this would cause a panic when the update is deployed, so the unwrap_or_else logic is still requried here as a safety barrier.

kanidm_unix_int/src/cache.rs
```diff
@@ -1,3 +1,4 @@
+use crate::constants::DEFAULT_SHELL;
```
Firstyear, line 1: You shouldn't be importing this, rather you sohuld use the `self.default_shell` field on the instance. This way later we can have the default_shell be configurable, rather than hardcoded. This line should be removed, and all instances of DEFAULT_SHELL should be self.default_shell

See also, kanidm_unix_int/src/daemon.rs line 571 where we set the default shell based on the configuration file.

Firstyear (changes requested):
Looks good, but a few changes to logic to resolve some issues here. Hope the comments help!

kanidm_unix_int/src/cache.rs
cuberoot74088: alright, I will remove it. I was not sure about `self.default_shell`.

kanidm_unix_int/src/cache.rs
cuberoot74088: I am trying to get this running. It is a `&UnixUserToken` 

`async fn set_cache_usertoken(&self, mut token: &UnixUserToken) -> Result<(), ()> {`

bit if it is a `&` ref, I cannot change the value. Therefore, I cloned it ... 

If I try to do it via 
`async fn set_cache_usertoken(&self, mut token: UnixUserToken) -> Result<(), ()> {`

I get some other problems ... (sorry this is not very specific)

kanidm_unix_int/src/cache.rs
cuberoot74088: Took me some time to understand. Interesting. :-)
Because I saw `Path::new()` with None does not work, but seems like the map just returns None, that is captured by `.unwrap_or_else()`. As a non-expert, this is also not so obvious, but ok. As I really like your list of cases above, I was looking for the "Nothing requested" case, just to mention.

kanidm_unix_int/src/cache.rs
cuberoot74088: I guess I need some help here. I don't see how this can be done without `clone()`. See comment above.

kanidm_unix_int/src/cache.rs
cuberoot74088: Good point. I will revert.

cuberoot74088:
Hi @Firstyear,
thanks for the detailed and valuable feedback! :+1: 
I have included the changes you mentioned, but I still need to `clone` the token. Maybe I missed something.

Do you have some more thoughts on this comment  https://github.com/kanidm/kanidm/pull/490#discussion_r655585271 about the `requested_shell_exists` logic. As said, I am quite new to it, so it might be fine - don't worry too much.
Thanks.

kanidm_unix_int/src/cache.rs
```diff
@@ -256,14 +257,37 @@ impl CacheLayer {
         }
     }
 
-    async fn set_cache_usertoken(&self, token: &UnixUserToken) -> Result<(), ()> {
+    async fn set_cache_usertoken(&self, token: &mut UnixUserToken) -> Result<(), ()> {
```
Firstyear, line 260: Actually, I like the &mut better anyway, so just leave it like this :)

kanidm_unix_int/src/cache.rs
```diff
@@ -732,9 +756,9 @@ impl CacheLayer {
             .idm_account_unix_cred_verify(account_id, cred)
             .await
         {
-            Ok(Some(n_tok)) => {
+            Ok(Some(mut n_tok)) => {
                 debug!("online password check success.");
-                self.set_cache_usertoken(&n_tok).await?;
+                self.set_cache_usertoken(&mut n_tok).await?;
```
Firstyear, line 761: To answer the clone question - I think here you could have removed the &n_tok in the call to n_tok to move the ownership, but actually, passing it as &mut is better here IMO.

Firstyear (approved):

Firstyear:
I think this looks really good now, is there anything else you want to change before we merge it?

cuberoot74088:
> I think this looks really good now, is there anything else you want to change before we merge it?

How about the test case, should we keep it like this? Just remove the WIP comment for sure. I think a comment about why we check it there, would be good for later reference though.

Any other ideas/wishes from your side?

Firstyear:
> > I think this looks really good now, is there anything else you want to change before we merge it?
> 
> How about the test case, should we keep it like this? Just remove the WIP comment for sure. I think a comment about why we check it there, would be good for later reference though.
> 
> Another thing, as we discussed about this [3eb9363#diff-1a198423e95e79069198c1484e2850686de9a8329b0cdc64a58142309ef1d8aaL648](https://github.com/kanidm/kanidm/commit/3eb9363df02d70f51b9bfe25b0aa7b0fcc8243df#diff-1a198423e95e79069198c1484e2850686de9a8329b0cdc64a58142309ef1d8aaL648)
> 
> * if there would be a cached item, that has no proper (not existing) shell set, how big would be the problem for such a user?
> * or when would this be "fixed" in that case?

It can't "really" be fixed. Think of this as an "on upgrade" problem.

With an older version, it's valid to write out a user token with a shell of "None".. then we upgrade to this version, and we are offline and we attempt to load the user. Since it went to unwrap, it would then panic, and the resolver would crash.

So it would be a huge problem. :) 


> 
> Any other ideas/wishes from your side?

As the code looks now, I'm happy with it.

If you wanted to add another test, you could in cache_layer_test add something after this like:

https://github.com/kanidm/kanidm/pull/490/files#diff-b7fab3e5d9880ab68c7220a14d2cb18d330480df3b8f4c66d9c178e10d57e875R243

```

adminclient.auth_simple_password("admin", ADMIN_TEST_PASSWORD).await.expect();
// with the admin client, set the shell to `/bin/tnaohtnuoesoetsuns`

assert!(cachelayer.invalidate().await.is_ok());

// assert the shell is still DEFAULT since that file can't exist.

```

Just to check the file non-existant logic. I'm not sure about the 'file exists' test though. Yo ucould do /bin/sh but that could be confusing as that's also the result that is returned if /bin/sh was MISSING .... but I'm not sure what eles is so univerasl that we can guarantee the file "exists". I guess you could do /bin/false as the shell ?

cuberoot74088:
Hi @Firstyear,
thanks for the explanations.

Well, then let's not over engineer the tests then. It will kick in if there is a `None` or if it not exists. So, it is okay IMHO.

Thus, I would be okay with it and consider it done.

Firstyear:
Thank you very much @cuberoot74088! And thanks for the contribution!

PR #496, opened 2021-06-25
Fix readonly check

### LINKED ISSUE #381: Fix readonly check for startup
The current rust "readonly" api is different to what we need. We should rework this check to be platform specific and clearer.

Firstyear:
Fixes #381 - this makes the readonly check aware of unix permissions and can better/correctly advise on these.


Firstyear:
Tested on macos/linux to make sure it works across unix platforms. I think we haven't properly supported windows for quite some time so we'll need to do some work to make that work in the future.

yaleman (approved):
Logic looks sane and matches what I was thinking would be the solution, nice :)

PR #497, opened 2021-06-26
`tracing` for `kanidmd/src/lib/plugins/memberof.rs`

### LINKED ISSUE #453: 🪵 The 🪵 Great 🪵 Log 🪵 Issue 🪵
Logs should be:

1. Fast.
2. Timestamped.
3. JSON.
4. Correlatable.

[json_env_logger](https://lib.rs/crates/json_env_logger) seems sensible from an import-a-crate-make-go perspective, env_logger is already being used in a bunch of places and replacing it with json\_env\_logger is largely a find-and-replace matter - then the long tail of `%s/eprintln/error/g`.

It doesn't do everything, but... it might be able to once the formatting options are considered.

I have a much longer rant about logging [here](https://yaleman.org/post/2021/2021-05-22-the-great-log-post-/) if you want to read it. It got pretty long, so ... let's start with ^ that.

QnnOkabayashi:
Fixes #453

- [x] cargo fmt has been run

An implementation sample for the [`tracing`](https://docs.rs/tracing/0.1.26/tracing) crate in `memberof.rs`. This PR is intended to illustrate two key points:
* `tracing`s first-class logging capabilities. It has a ton of features that we can use universally in Kanidm, allowing us to remove boilerplate code that we currently use to handle logging like `AuditScope`.
* The `LogBuffer` type, which is compatible with `tracing` and allows us to capture logs in memory to avoid `stderr` locking, and write them all at once.

Notes:
My initial changes only have an effect on the first test in `memberof.rs`. It can be tested with:
```
cargo test --package kanidm --lib -- plugins::memberof::tests::test_create_mo_single --exact --nocapture > out.log 2> err.log
```
Additionally, the `tracing` logs have been added in parallel to the existing `AuditScope` logs. I hope to eventually have it completely replace them before this gets merged.

### DIFF
--- a/kanidmd/src/lib/logbuffer.rs
+++ b/kanidmd/src/lib/logbuffer.rs
@@ -1,0 +1,68 @@
+use std::{io, ops};
+use tracing_subscriber::fmt::MakeWriter;
+
+// This type may be moved out of `kanidmd` later
+#[derive(Default)]
+pub struct LogBuffer {
+    buf: Vec<u8>,
+}
+
+// An "owned" type pointing to a buffer in a
+// `LogBuffer`. This type exists to satisfy the
+// `MakeWriter` trait restriction on `make_writer`,
+// which must return an owned `io::Write` object.
+// The benefit of using a raw ptr is that creating this
+// object of very cheap and fast, and unless `Subscriber`s
+// are opening a new thread of each log (I don't think they are),
+// this will never cross a thread boundry.
+pub struct LogBufferWriter(*mut Vec<u8>);
+
+impl ops::Drop for LogBuffer {
+    fn drop(&mut self) {
+        // logs are written to stderr when dropped
+        use io::Write;
+        #[allow(clippy::expect_used)]
+        io::stderr()
+            .write_all(&self.buf)
+            .expect("Failed to write logs")
+    }
+}
+
+impl MakeWriter for LogBuffer {
+    type Writer = LogBufferWriter;
+
+    fn make_writer(&self) -> Self::Writer {
+        LogBufferWriter(&self.buf as *const _ as *mut _)
+    }
+}
+
+impl LogBufferWriter {
+    fn get_buffer(&mut self) -> &mut Vec<u8> {
+        // SAFETY:
+        //
+        // `LogBufferWriter`s are logical references to a
+        // `LogBuffer`, which is held for the entire duration
+        // of a `Subscriber`. A `LogBufferWriter` is only held
+        // for the duration of a `write` call, which is shorter
+        // than the lifetime of the `Subscriber` that created it.
+        // Therefore, the data behind the ptr will always be safe.
+        //
+        // I think this is concurrency safe, because I'm pretty sure
+        // `Subscriber`s aren't sending each log to a new thread.
+        unsafe { &mut *self.0 }
+    }
+}
+
+impl io::Write for LogBufferWriter {
+    fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
+        self.get_buffer().write(buf)
+    }
+
+    fn flush(&mut self) -> io::Result<()> {
+        self.get_buffer().flush()
+    }
+
+    fn write_all(&mut self, buf: &[u8]) -> io::Result<()> {
+        self.get_buffer().write_all(buf)
+    }
+}
--- a/kanidmd/src/lib/macros.rs
+++ b/kanidmd/src/lib/macros.rs
@@ -241,6 +241,7 @@ macro_rules! run_create_test {
         use crate::schema::Schema;
         use crate::utils::duration_from_epoch_now;
         use crate::logbuffer::LogBuffer;
+        use crate::subscriber::KaniSubscriber;
 
         use tracing::{self, Level};
         use tracing_subscriber::FmtSubscriber;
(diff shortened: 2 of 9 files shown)


kanidmd/src/lib/macros.rs
```diff
@@ -240,41 +240,77 @@ macro_rules! run_create_test {
...
+        use tracing_subscriber::fmt::format::{FmtSpan, debug_fn};
+        use tracing_subscriber::field::MakeExt;
+
+        // https://docs.rs/tracing-subscriber/0.2.18/tracing_subscriber/fmt/struct.SubscriberBuilder.html
+        let subscriber = FmtSubscriber::builder()
+            // SETTINGS
+            .with_max_level(Level::TRACE)
+            .with_ansi(false) // no colors (good for .log files)
+
+            // JSON
```
Firstyear, line 256: We'll need some way to pass this configuration in in the future.

kanidmd/src/lib/logbuffer.rs
```diff
@@ -0,0 +1,68 @@
...
+        //
+        // `LogBufferWriter`s are logical references to a
+        // `LogBuffer`, which is held for the entire duration
+        // of a `Subscriber`. A `LogBufferWriter` is only held
+        // for the duration of a `write` call, which is shorter
+        // than the lifetime of the `Subscriber` that created it.
+        // Therefore, the data behind the ptr will always be safe.
+        //
+        // I think this is concurrency safe, because I'm pretty sure
+        // `Subscriber`s aren't sending each log to a new thread.
```
Firstyear, line 51: Today, it's not - in the future it will. The intent is to create "audit scopes" or "tracing spans" in a tide middleware / ldap thread. When these .await the query server they will cross threads, meaning that this is unsound and thus, incorrect.

kanidmd/src/lib/logbuffer.rs
```diff
@@ -0,0 +1,68 @@
...
+// are opening a new thread of each log (I don't think they are),
+// this will never cross a thread boundry.
+pub struct LogBufferWriter(*mut Vec<u8>);
+
+impl ops::Drop for LogBuffer {
+    fn drop(&mut self) {
+        // logs are written to stderr when dropped
+        use io::Write;
+        #[allow(clippy::expect_used)]
+        io::stderr()
```
Firstyear, line 25: In the future this will probably need to be sent to a queue for flushing to stderrl/stdout instead, rather than writing during drop.

kanidmd/src/lib/macros.rs
```diff
@@ -240,41 +240,77 @@ macro_rules! run_create_test {
...
+                        _ => write!(writer, "{}={:?}", key, value),
                     }
-                    Err(e) => {
-                        ladmin_error!(&mut au, "Rolling back => {:?}", e);
+                }).delimited(" | ")
+            })
+
+            .finish();
+
+        tracing::subscriber::with_default(subscriber, || {
```
Firstyear, line 274: Instead of using FmtSubscriber, why not create https://docs.rs/tracing/0.1.26/tracing/trait.Subscriber.html? This would let us capture all the data we want using tracing / spans, but then format and display it how we like instead. 

IE I think the solution here is what we want is to use tracing's machineries/facades, but then capture those events and process them in a way that is kanidm specific, and managing how we output these to a file/stdout/stderr.

In a way it feels like trying to capture the "output" buffers and trying to sync/manage that is working at the wrong level for the problem, when what we really want is to capture tracings *events* and then buffering these in a native format for post-processing.

Firstyear (changes requested):

kanidmd/src/lib/macros.rs
QnnOkabayashi: Where are we accessing the config in functions? Is it eventually going to be an object that's passed around? Or a global var? I agree that this should be configurable though.

kanidmd/src/lib/logbuffer.rs
QnnOkabayashi: Agreed. I was aiming to have the least amount of code possible for this initial PR, and using `Drop` meant that I didn't have to add any explicit dump function to call before cleanup, but I'm happy to change this.

kanidmd/src/lib/macros.rs
Firstyear: It's not so simple. There are some values that are global, and some that are "per request" so each subscriber may be uniquely configured - or could be re-configured during a request depending on which stage we are in of the process. 

For now we'll just assume some global defaults though, let's not overthink it yet. But we'll need a way to say "show me the current subscriber" for an event.

The main barrier right now will be the batching requirement. I think it may be time to make a new "side project" and experiment there in a small scale rather than here. We sohuld probably have a design doc and a talk about it?

kanidmd/src/lib/subscriber.rs
```diff
@@ -0,0 +1,154 @@
...
+    pub fn dump(mut self) -> String {
+        // TODO: update this to use the ctx to get span information
+        self.log(format!("INFO: {}: CLOSED", self.path));
+        self.buf.join("\n")
+    }
+}
+
+impl KaniSubscriber {
+    pub fn new() -> Self {
+        println!("OOGA NEW SUBSCRIBER");
```
QnnOkabayashi, line 119: Everything here is super temporary, I was just doing an initial implementation to check that I could insert my own arbitrary code that can do stuff when spans start and end.

QnnOkabayashi:
I started developing my own `Subscriber` on a mini project here:
https://github.com/QnnOkabayashi/tracing-tests

PR #499, opened 2021-06-26
Fixes #494 - password change user-facing responses

### LINKED ISSUE #494: Humanise result message for PasswordTooShort in kanidm credential set_password
This is a bit icky :) I'll have a go at fixing it.

```
> kanidm account credential set_password --name idm_admin test_splunk_user
Enter new password for test_splunk_user:
Retype the new password to confirm:
Error -> Http(500, Some(PasswordTooShort(10)), "f504c155-d13f-4d8e-bd87-4f50ad5a625f")
```

yaleman:
Fixes #494

- [x] cargo fmt has been run
- [x] cargo clippy has been run and is still very unimpressed with us

### DIFF
--- a/kanidm_tools/src/cli/account.rs
+++ b/kanidm_tools/src/cli/account.rs
@@ -9,6 +9,10 @@ use time::OffsetDateTime;
 
 use webauthn_authenticator_rs::{u2fhid::U2FHid, WebauthnAuthenticator};
 
+use kanidm::constants::PW_MIN_LENGTH;
+use kanidm_client::ClientError::Http as ClientErrorHttp;
+use kanidm_proto::v1::OperationError::{PasswordBadListed, PasswordTooShort, PasswordTooWeak};
+
 impl AccountOpt {
     pub fn debug(&self) -> bool {
         match self {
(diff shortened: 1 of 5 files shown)


kanidm_tools/src/cli/account.rs
```diff
@@ -9,6 +9,10 @@ use time::OffsetDateTime;
 
 use webauthn_authenticator_rs::{u2fhid::U2FHid, WebauthnAuthenticator};
 
+use kanidm::constants::PW_MIN_LENGTH;
```
Firstyear, line 12: It may be metter to move PW_MIN_LENGTH to proto for now instead? That way we dont need to include kanidm here.

kanidm_tools/src/cli/account.rs
```diff
@@ -68,7 +72,13 @@ impl AccountOpt {
                         acsopt.aopts.account_id.as_str(),
                         password.as_str(),
                     ) {
-                        eprintln!("Error -> {:?}", e);
+                        match e {
+                            // TODO: once the password length is configurable at a system level (#498), pull from the configuration.
+                            ClientErrorHttp(_, Some(PasswordBadListed), _) => error!("Password is banned by the administrator of this system, please try a different one."),
+                            ClientErrorHttp(_, Some(PasswordTooShort(PW_MIN_LENGTH)), _) => error!("Password was too short (needs to be at least {} characters), please try again.", PW_MIN_LENGTH),
```
Firstyear, line 78: We shouldn't need to match on min_length here - this is a variable, that indicates to the client what the min length is. So I think you can remove the import of PW_MIN_LEGNTH, and assign to say "min_len" and then display that. Seems simpler don't you think?

kanidm_tools/src/cli/account.rs
```diff
@@ -68,7 +72,13 @@ impl AccountOpt {
...
+                            ClientErrorHttp(_, Some(PasswordTooWeak), _) => error!("Password too weak, please try a more complex password."),
```
Firstyear, line 79: I think it could be good to open a subsequent issue to have better errors returned around password qual in general.

Firstyear (changes requested):

kanidm_tools/src/cli/account.rs
yaleman: Isn't needed once the check was fixed in 35076d3

kanidm_tools/src/cli/account.rs
yaleman: Fixed in 35076d3, thanks :)

kanidm_tools/src/cli/account.rs
yaleman: I've updated the wording a little, and yep, new issue created.

PR #505, opened 2021-06-28
Add statistical analysis to indexes

### LINKED ISSUE #238: Add statisical diversity hint to idxmeta
In idxmeta, we should use a "diversity" hint, rather than just "true/false" to determine indexing. This would be a range representing a percentage of how many values exist in the index. For example, these two terms would currently sort as:

```
And[ Eq('class', 'foo'), Eq('uuid', 'xxxx...')]
```

This is because we currently sort by attribute name in eq terms that both are indexed. But we can see quickly that class 'foo' as an index, does not yield an effective index, because class indexes tend to be deep, not broad (there are many class accounts). Contrast a uuid would would be broad and shallow, as each eq index of uuid is unique, giving us the result we want, quicker. 

This then plays to the filter test threshold in search, where we could evaluate the uuid term, load the entry, then apply the class=foo, rather than accessing the class foo index.

To achieve this, rather than "hardcoding" what attributes might be better, we can get information from the database about the breadth of their indexes - broader indexes have more diversity, and will help us to get a smaller candidate set quicker.

This would be stored in the idxmeta. It would not be refreshed constantly on writes (that's expensive), but could be generated on each schema reload aka server restart or schema modification. We could also consider reloading it on a significant change (greater than 10% of the number of entries of the db). 

We would take the class pres index as the "length" of the database, and then from there we can use that to count how many values exist in each index type. Given this, we can then calculate a percentage such as:

```
information = 1 - (idx_len / class_pres_len)
0.99996 = 1.0 - 1.0/25000.0
0.996 = 1.0 - 100.0/25000.0
0.98 = 1.0 - 500.0/25000.0
0.8 1.0 - 5000.0/25000.0
0.199 = 1.0 - 20000.0/25000.0
0.0 = 1.0 - 25000.0/25000.0
```

From this we know the index with 1 element gives us the most information, compared to the others, and allows sorting of indexes with much better knowledge of what they yield to our application.

Firstyear:
Fixes #238 - this adds statistical analysis capabilities to indexes allow the optimiser to make better decisions about queries and how to run them. It has some benefits in reducing the time to perform a query optimisation as well. 

### DIFF
--- a/kanidmd/src/lib/be/idl_arc_sqlite.rs
+++ b/kanidmd/src/lib/be/idl_arc_sqlite.rs
@@ -1022,7 +1022,7 @@ impl<'a> IdlArcSqliteWriteTransaction<'a> {
             let mean = data
                 .iter()
                 .take(u32::MAX as usize)
-                .fold(0.0, |acc, x| x + acc)
+                .sum::<f64>()
                 / c;
             let varience: f64 = data.iter().take(u32::MAX as usize).fold(0.0, |acc, i| {
                 let diff = mean - i;
--- a/kanidmd/src/lib/be/idl_arc_sqlite.rs
+++ b/kanidmd/src/lib/be/idl_arc_sqlite.rs
@@ -2,16 +2,22 @@ use crate::audit::AuditScope;
 use crate::be::idl_sqlite::{
     IdlSqlite, IdlSqliteReadTransaction, IdlSqliteTransaction, IdlSqliteWriteTransaction,
 };
-use crate::be::idxkey::{IdlCacheKey, IdlCacheKeyRef, IdlCacheKeyToRef};
+use crate::be::idxkey::{
+    IdlCacheKey, IdlCacheKeyRef, IdlCacheKeyToRef, IdxKey, IdxKeyRef, IdxKeyToRef, IdxSlope,
+};
 use crate::be::{BackendConfig, IdList, IdRawEntry};
 use crate::entry::{Entry, EntryCommitted, EntrySealed};
 use crate::value::IndexType;
 use crate::value::Value;
+
 use concread::arcache::{ARCache, ARCacheReadTxn, ARCacheWriteTxn};
 use concread::cowcell::*;
 use idlset::{v2::IDLBitRange, AndNot};
 use kanidm_proto::v1::{ConsistencyError, OperationError};
+
+use hashbrown::HashMap;
 use std::collections::BTreeSet;
+use std::convert::TryInto;
 use std::ops::DerefMut;
 use std::time::Duration;
 use uuid::Uuid;
@@ -798,6 +804,256 @@ impl<'a> IdlArcSqliteWriteTransaction<'a> {
                 }
             }
         })
+    }
+
+    pub fn is_idx_slopeyness_generated(
+        &self,
+        audit: &mut AuditScope,
+    ) -> Result<bool, OperationError> {
+        self.db.is_idx_slopeyness_generated(audit)
+    }
+
+    pub fn get_idx_slope(
+        &self,
+        audit: &mut AuditScope,
+        ikey: &IdxKey,
+    ) -> Result<Option<IdxSlope>, OperationError> {
+        self.db.get_idx_slope(audit, ikey)
+    }
+
+    /// Index Slope Analysis. For the purpose of external modules you can consider this as a
+    /// module that generates "weights" for each index that we have. Smaller values are faster
+    /// indexes - larger values are more costly ones. This is not intended to yield perfect
+    /// weights. The intent is to seperate over obviously more effective indexes rather than
+    /// to min-max the fine tuning of these. Consider name=foo vs class=*. name=foo will always
+    /// be better than class=*, but comparing name=foo to spn=foo is "much over muchness" since
+    /// both are really fast.
+    pub fn analyse_idx_slopes(&mut self, audit: &mut AuditScope) -> Result<(), OperationError> {
+        /*
+         * Inside of this analysis there are two major factors we need to understand
+         *
+         * * What is the variation of idl lengths within an index?
+         * * How man keys are stored in this index?
+         *
+         * Since we have the filter2idl threshold, we want to find "what is the smallest
+         * and most unique index asap so we can exit faster". This allows us to avoid
+         * loading larger most costly indexs that either have large idls, high variation
+         * or few keys and are likely to miss and have to go out to disk.
+         *
+         * A few methods were proposed, but thanks to advice from a Psychology researcher
+         * and statistics enthustiast, we were able to device a reasonable approach.
+         *
+         * These are commented in line to help understand the process.
+         */
+
+        /*
+         * Step 1 - we have an index like "idx_eq_member". It has data that looks somewhat
+         * like:
+         *
+         *  key    | idl
+         *  -------+------------
+         *  uuid_a | [1, 2, 3, ...]
+         *  -------+------------
+         *  uuid_b | [4, 5, 6, ...]
+         *
+         * We need to collect this into a single vec of "how long is each idl". Since we have
+         * each idl in the vec, the length of the vec is also the number of keys in the set.
+         * This yields for us:
+         *
+         *   idx_eq_member: [4.0, 5.0, ...]
+         * where each f64 value is the float representation of the length of idl.
+         *
+         * We then assemble these to a map so we have each idxkey and it's associated list
+         * of idl lens.
+         */
+
+        let mut data: HashMap<IdxKey, Vec<f64>> = HashMap::new();
+        self.idl_cache.iter_dirty().for_each(|(k, maybe_idl)| {
+            if let Some(idl) = maybe_idl {
+                let idl_len: u32 = idl.len().try_into().unwrap_or(u32::MAX);
+                // Convert to something we can use.
+                let idl_len = f64::from(idl_len);
+
+                let kref = IdxKeyRef::new(&k.a, &k.i);
+                if idl_len > 0.0 {
+                    // It's worth looking at. Anything len 0 will be removed.
+                    if let Some(lens) = data.get_mut(&kref as &dyn IdxKeyToRef) {
+                        lens.push(idl_len)
+                    } else {
+                        data.insert(kref.to_key(), vec![idl_len]);
+                    }
+                }
+            }
+        });
+
+        /*
+        * So now for each of our sets:
+        *
+        *   idx_eq_member: [4.0, 5.0, ...]
+        *   idx_eq_name  : [1.0, 1.0, 1.0, ...]
+        *
+        * To get the variability, we calculate the normal distribution of the set of values
+        * and then using this variance we use the 1st deviation (~85%) value to assert that
+        * 85% or more of the values in this set will be "equal or less" than this length.*
+        *
+        * So given say:
+        *  [1.0, 1.0, 1.0, 1.0]
+        * We know that the sd_1 will be 1.0. Given:
+        *  [1.0, 1.0, 2.0, 3.0]
+        * We know that it will be ~2.57 (mean 1.75 + sd of 0.82).
+        *
+        * The other factor is number of keys. This is thankfully easy! We have that from
+        * vec.len().
+        *
+        * We can now calculate the index slope. Why is it a slope you ask? Because we
+        * plot the data out on a graph, with "variability" on the y axis, and number of
+        * keys on the x.
+        *
+        * Lets plot our data we just added.
+        *
+        *    |
+        *  4 +
+        *    |
+        *  3 +
+        *    |
+        *  2 +           *  eq_member
+        *    |
+        *  1 +           *  eq_name
+        *    |
+        *    +--+--+--+--+--
+        *       1  2  3  4
+        *
+        * Now, if we were to connect a line from (0,0) to each point we get a line with an angle.
+        *
+        *    |
+        *  4 +
+        *    |
+        *  3 +
+        *    |
+        *  2 +           *  eq_member
+        *    |
+        *  1 +           *  eq_name
+        *    |/---------/
+        *    +--+--+--+--+--
+        *       1  2  3  4
+
+        *    |
+        *  4 +
+        *    |
+        *  3 +
+        *    |
+        *  2 +           *  eq_member
+        *    |        /--/
+        *  1 +    /--/   *  eq_name
+        *    |/--/
+        *    +--+--+--+--+--
+        *       1  2  3  4
+        *
+        * (Look it's ascii art, don't judge.).
+        *
+        * Point is that eq_member is "steeper" and eq_name is "shallower". This is what we call
+        * the "slopeyness" aka the jank of the line, or more precisely, the angle.
+        *
+        * Now we need a way to numerically compare these lines. Since the points could be
+        * anywere on our graph:
+        *
+        *    |
+        *  4 +  *
+        *    |
+        *  3 +         *
+        *    |
+        *  2 +     *
+        *    |
+        *  1 +           *
+        *    |
+        *    +--+--+--+--+--
+        *       1  2  3  4
+        *
+        * While we can see what's obvious or best here, a computer has to know it. So we now
+        * assume that these points construct a triangle, going through (0,0), (x, 0) and (x, y).
+        *
+        *
+        *                Λ│
+        *               ╱ │
+        *              ╱  │
+        *             ╱   │
+        *            ╱    │
+        *           ╱     │
+        *          ╱      │
+        *         ╱       │ sd_1
+        *        ╱        │
+        *       ╱         │
+        *      ───────────┼
+        *         nkeys
+        *
+        * Since this is right angled we can use arctan to work out the degress of the line. This
+        * gives us a value from 1.0 to 90.0 (We clamp to a minimum of 1.0, but really 0.0 would
+        * be okay).
+        *
+        * The problem is that we have to go from float to u8 - this means we lose decimal precision
+        * in the conversion. To lessen this, we multiply by 2 to give some extra weight to each angle
+        * to minimise this loss and then we convert.
+        *
+        * And there we have it! A slope factor of the index! A way to compare these sets quickly
+        * at query optimisation time to minimse index access.
+        */
+        let slopes: HashMap<_, _> = data
+            .into_iter()
+            .filter_map(|(k, lens)| {
+                let slope_factor = Self::calculate_sd_slope(lens);
+                if slope_factor == 0 {
+                    None
+                } else {
+                    Some((k, slope_factor))
+                }
+            })
+            .collect();
+        ltrace!(audit, "Generated slopes -> {:?}", slopes);
+        // Write the data down
+        self.db.store_idx_slope_analysis(audit, &slopes)
+    }
+
+    fn calculate_sd_slope(data: Vec<f64>) -> IdxSlope {
+        let (n_keys, sd_1) = if data.len() >= 2 {
+            // We can only do SD on sets greater than 2
+            let l: u32 = data.len().try_into().unwrap_or(u32::MAX);
+            let c = f64::from(l);
+            let mean = data
+                .iter()
+                .take(u32::MAX as usize)
+                .fold(0.0, |acc, x| x + acc)
+                / c;
+            let varience: f64 = data.iter().take(u32::MAX as usize).fold(0.0, |acc, i| {
+                let diff = mean - i;
+                acc + (diff * diff)
+            }) / (c - 1.0);
+            let sd = varience.sqrt();
+
+            // This is saying ~85% of values will be at least this len or less.
+            let sd_1 = mean + sd;
+            (c, sd_1)
+        } else if data.len() == 1 {
+            (1.0, data[0])
+        } else {
+            // Cant resolve.
+            return IdxSlope::MAX;
+        };
+
+        // Now we know sd_1 and number of keys. We can use this as a triangle to work out
+        // the angle along the hypotenuse. We use this angle - or slope - to show which
+        // elements have the smallest sd_1 and most keys available. Then because this
+        // is bound between 0.0 -> 90.0, we "unfurl" this around a half circle by multipling
+        // by 2. This gives us a little more precision when we drop the decimal point.
+        let sf = (sd_1 / n_keys).atan().to_degrees() * 2.0;
+
+        // Now these are fractions, and we can't use those in u8, so we resolve where this
+        // point would be at 100.
+        let sf = sf.clamp(1.0, 180.0);
+        if !sf.is_finite() {
+            IdxSlope::MAX
+        } else {
+            unsafe { sf.to_int_unchecked::<IdxSlope>() }
+        }
     }
 
     pub fn create_name2uuid(&self, audit: &mut AuditScope) -> Result<(), OperationError> {
--- a/kanidmd/src/lib/be/idl_arc_sqlite.rs
+++ b/kanidmd/src/lib/be/idl_arc_sqlite.rs
@@ -988,8 +988,8 @@ impl<'a> IdlArcSqliteWriteTransaction<'a> {
         *         nkeys
         *
         * Since this is right angled we can use arctan to work out the degress of the line. This
-        * gives us a value from 1.0 to 90.0 (We clamp to a minimum of 1.0, but really 0.0 would
-        * be okay).
+        * gives us a value from 1.0 to 90.0 (We clamp to a minimum of 1.0, because we use 0 as "None"
+        * in the NonZeroU8 type in filter.rs, which allows ZST optimisation)
         *
         * The problem is that we have to go from float to u8 - this means we lose decimal precision
         * in the conversion. To lessen this, we multiply by 2 to give some extra weight to each angle
@@ -1002,7 +1002,7 @@ impl<'a> IdlArcSqliteWriteTransaction<'a> {
             .into_iter()
             .filter_map(|(k, lens)| {
                 let slope_factor = Self::calculate_sd_slope(lens);
-                if slope_factor == 0 {
+                if slope_factor == 0 || slope_factor == IdxSlope::MAX {
                     None
                 } else {
                     Some((k, slope_factor))
@@ -1047,8 +1047,8 @@ impl<'a> IdlArcSqliteWriteTransaction<'a> {
         // by 2. This gives us a little more precision when we drop the decimal point.
         let sf = (sd_1 / n_keys).atan().to_degrees() * 2.0;
 
-        // Now these are fractions, and we can't use those in u8, so we resolve where this
-        // point would be at 100.
+        // Now these are fractions, and we can't use those in u8, so we clamp the min/max values
+        // that we expect to be yielded.
         let sf = sf.clamp(1.0, 180.0);
         if !sf.is_finite() {
             IdxSlope::MAX
--- a/kanidmd/src/lib/be/idl_arc_sqlite.rs
+++ b/kanidmd/src/lib/be/idl_arc_sqlite.rs
@@ -840,8 +840,9 @@ impl<'a> IdlArcSqliteWriteTransaction<'a> {
          * loading larger most costly indexs that either have large idls, high variation
          * or few keys and are likely to miss and have to go out to disk.
          *
-         * A few methods were proposed, but thanks to advice from a Psychology researcher
-         * and statistics enthustiast, we were able to device a reasonable approach.
+         * A few methods were proposed, but thanks to advice from Perri Boulton (psychology
+         * researcher with a background in statistics), we were able to device a reasonable
+         * approach.
          *
          * These are commented in line to help understand the process.
          */
--- a/kanidmd/src/lib/be/idl_arc_sqlite.rs
+++ b/kanidmd/src/lib/be/idl_arc_sqlite.rs
@@ -1054,6 +1054,9 @@ impl<'a> IdlArcSqliteWriteTransaction<'a> {
         if !sf.is_finite() {
             IdxSlope::MAX
         } else {
+            // SAFETY
+            // `sf` is clamped between 1.0 and 180.0 above, ensuring it is
+            // always in range.
             unsafe { sf.to_int_unchecked::<IdxSlope>() }
         }
     }
--- a/kanidmd/src/lib/be/idl_arc_sqlite.rs
+++ b/kanidmd/src/lib/be/idl_arc_sqlite.rs
@@ -1024,10 +1024,11 @@ impl<'a> IdlArcSqliteWriteTransaction<'a> {
                 .take(u32::MAX as usize)
                 .sum::<f64>()
                 / c;
-            let varience: f64 = data.iter().take(u32::MAX as usize).fold(0.0, |acc, i| {
-                let diff = mean - i;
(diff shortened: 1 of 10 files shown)


kanidmd/Cargo.toml
```diff
@@ -63,6 +63,7 @@ structopt = { version = "0.3", default-features = false }
 time = { version = "0.2", features = ["serde", "std"] }
 
 hashbrown = "0.11"
+ahash = "0.7"
```
Firstyear, line 66: Actually I think this can be removed I don't actually use this.

kanidmd/src/lib/be/idl_arc_sqlite.rs
```diff
@@ -710,6 +716,256 @@ impl<'a> IdlArcSqliteWriteTransaction<'a> {
...
+         *
+         * * What is the variation of idl lengths within an index?
+         * * How man keys are stored in this index?
+         *
+         * Since we have the filter2idl threshold, we want to find "what is the smallest
+         * and most unique index asap so we can exit faster". This allows us to avoid
+         * loading larger most costly indexs that either have large idls, high variation
+         * or few keys and are likely to miss and have to go out to disk.
+         *
+         * A few methods were proposed, but thanks to advice from a Psychology researcher
```
Firstyear, line 753: Need to add their name for attribution

kanidmd/src/lib/filter.rs
```diff
@@ -137,22 +140,28 @@ enum FilterComp {
...
 pub enum FilterResolved {
-    // This is attr - value - indexed
-    Eq(AttrString, PartialValue, bool),
-    Sub(AttrString, PartialValue, bool),
-    Pres(AttrString, bool),
-    LessThan(AttrString, PartialValue, bool),
-    Or(Vec<FilterResolved>),
-    And(Vec<FilterResolved>),
+    // This is attr - value - indexed slope factor
+    Eq(AttrString, PartialValue, Option<IdxSlope>),
```
Firstyear, line 156: This could be NonZeroU8 instead to allow Option ZST optimisation

kanidmd/src/lib/filter.rs
```diff
@@ -1086,43 +1126,53 @@ impl FilterResolved {
...
         }
     }
 
     fn resolve_no_idx(fc: FilterComp, ev: &Identity) -> Option<Self> {
         match fc {
-            FilterComp::Eq(a, v) => Some(FilterResolved::Eq(a, v, false)),
-            FilterComp::Sub(a, v) => Some(FilterResolved::Sub(a, v, false)),
-            FilterComp::Pres(a) => Some(FilterResolved::Pres(a, false)),
-            FilterComp::LessThan(a, v) => Some(FilterResolved::LessThan(a, v, false)),
+            FilterComp::Eq(a, v) => {
```
Firstyear, line 1136: Could re-use the same fn for default slopes.

Firstyear: Does this also mean fast_optimise cuold be removed too?

kanidmd/src/lib/filter.rs
```diff
@@ -8,25 +8,28 @@
...
-use smartstring::alias::String as AttrString;
+// use smartstring::alias::String as AttrString;
 use std::cmp::{Ordering, PartialOrd};
 use std::collections::BTreeSet;
 use std::hash::Hash;
 use std::iter;
+use std::num::NonZeroU8;
 use uuid::Uuid;
 
+use hashbrown::HashMap;
```
QnnOkabayashi, line 29: How come we're using `hashbrown` when the `std` currently has the same implementation?

kanidmd/src/lib/filter.rs
```diff
@@ -137,22 +140,28 @@ enum FilterComp {
...
-// properly.
-#[derive(Debug, Clone, PartialEq, Eq)]
+/// This is the fully resolved internal representation. Note the lack of Not and selfUUID
+/// because these are resolved into And(Pres(class), AndNot(term)) and Eq(uuid, ...).
+/// Importantly, we make this accessible to Entry so that it can then match on filters
+/// internally.
+///
+/// Each filter that has been resolved also has been enriched with metadata about its
+/// index, and index slope. For the purpose of this module, consider slope as a "weight"
+/// where small value - faster index, larger value - slower index. This metadat is extremely
```
QnnOkabayashi, line 150:
```suggestion
/// where small value - faster index, larger value - slower index. This metadata is extremely
```

kanidmd/src/lib/filter.rs
```diff
@@ -137,22 +140,28 @@ enum FilterComp {
...
+/// important for the query optimiser to make decisions about how to re-arrange queries
+/// correctly.
+#[derive(Debug, Clone, Eq)]
 pub enum FilterResolved {
```
QnnOkabayashi, line 154: Can you briefly describe what this type is used for? I assume its for matching on something in a db, but I'm confused on the structure.

kanidmd/src/lib/filter.rs
```diff
@@ -1086,43 +1131,54 @@ impl FilterResolved {
...
         match fc {
-            FilterComp::Eq(a, v) => Some(FilterResolved::Eq(a, v, false)),
-            FilterComp::Sub(a, v) => Some(FilterResolved::Sub(a, v, false)),
-            FilterComp::Pres(a) => Some(FilterResolved::Pres(a, false)),
-            FilterComp::LessThan(a, v) => Some(FilterResolved::LessThan(a, v, false)),
+            FilterComp::Eq(a, v) => {
+                // Since we have no index data, we manually configure a reasonable
+                // slope and indicate the presence of some expected basic
+                // indexes.
+                let idx = a == "name" || a == "uuid";
```
QnnOkabayashi, line 1148: I can't help myself
```suggestion
                let idx = matches!(a, "name" | "uuid");
```

kanidmd/src/lib/filter.rs
```diff
@@ -1131,31 +1187,25 @@ impl FilterResolved {
...
 
     // This is an optimise that only attempts to optimise the outer terms.
     fn fast_optimise(self) -> Self {
         match self {
-            FilterResolved::Inclusion(mut f_list) => {
+            FilterResolved::Inclusion(mut f_list, _) => {
                 f_list.sort_unstable();
                 f_list.dedup();
-                FilterResolved::Inclusion(f_list)
+                let sf = f_list.last().and_then(|f| f.get_slopeyness_factor());
```
QnnOkabayashi, line 1201: Why do we need the slopeyness factor of the "greatest" element? How are these even being compared when `sort_unstable` is called?

kanidmd/src/lib/filter.rs
```diff
@@ -1164,29 +1214,31 @@ impl FilterResolved {
...
+                    if let FilterResolved::Inclusion(mut l, _) = fc {
                         f_list_new.append(&mut l)
                     }
                 });
                 // finally, optimise this list by sorting.
                 f_list_new.sort_unstable();
                 f_list_new.dedup();
-                FilterResolved::Inclusion(f_list_new)
+                // Inclusions are similar to or, so what's our worst case?
+                let sf = f_list_new.last().and_then(|f| f.get_slopeyness_factor());
```
QnnOkabayashi, line 1233: Can we replace the closure with:
```rust
<idk type of f>::get_slopeyness_factor
```
I also don't think this would hurt considering that it's impossible to tell the type contained in `f_list_new`s `Vec` just by looking at this function (unless I'm missing something).

kanidmd/src/lib/filter.rs
```diff
@@ -1452,8 +1523,8 @@ mod tests {
         // antisymmetry: if a < b then !(a > b), as well as a > b implying !(a < b); and
         // These are unindexed so we have to check them this way.
         let f_t3b = unsafe { filter_resolved!(f_eq("userid", PartialValue::new_iutf8(""))) };
-        assert_eq!(f_t1a.partial_cmp(&f_t3b), Some(Ordering::Less));
-        assert_eq!(f_t3b.partial_cmp(&f_t1a), Some(Ordering::Greater));
+        assert_eq!(f_t1a.partial_cmp(&f_t3b), Some(Ordering::Greater));
+        assert_eq!(f_t3b.partial_cmp(&f_t1a), Some(Ordering::Less));
```
QnnOkabayashi, line 1527: Why are these flipped? Previous typo?

QnnOkabayashi (commented):
I'm still a bit confused on the whole filtering thing. We're sending filter objects to search for something in a db and the slopeyness factor tells us how to order things to move faster searched to the front?

kanidmd/src/lib/be/idl_arc_sqlite.rs
```diff
@@ -710,6 +716,257 @@ impl<'a> IdlArcSqliteWriteTransaction<'a> {
...
+         *
+         *   idx_eq_member: [4.0, 5.0, ...]
+         * where each f64 value is the float representation of the length of idl.
+         *
+         * We then assemble these to a map so we have each idxkey and it's associated list
+         * of idl lens.
+         */
+
+        let mut data: HashMap<IdxKey, Vec<f64>> = HashMap::new();
+        self.idl_cache.iter_dirty().for_each(|(k, maybe_idl)| {
```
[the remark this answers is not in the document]
Firstyear: it's part of https://docs.rs/concread/0.2.16/concread/arcache/struct.ARCacheWriteTxn.html#method.iter_dirty which shows you each element in the current working set that is dirty or has been written. Since this is called inside reindex, then "every index" is inside the dirty set at this time, which is why we can do this.

Also because for loops imply .into_iter, not .iter(), and I think .for_each looks prettier.

kanidmd/src/lib/be/idl_arc_sqlite.rs
```diff
@@ -710,6 +716,257 @@ impl<'a> IdlArcSqliteWriteTransaction<'a> {
...
+         *  uuid_a | [1, 2, 3, ...]
+         *  -------+------------
+         *  uuid_b | [4, 5, 6, ...]
+         *
+         * We need to collect this into a single vec of "how long is each idl". Since we have
+         * each idl in the vec, the length of the vec is also the number of keys in the set.
+         * This yields for us:
+         *
+         *   idx_eq_member: [4.0, 5.0, ...]
+         * where each f64 value is the float representation of the length of idl.
```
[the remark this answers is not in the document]
Firstyear: As float because to do the standard deviation calc it has to be a float, not a usize. 

id list == idl.

kanidmd/src/lib/be/idl_arc_sqlite.rs
```diff
@@ -710,6 +716,257 @@ impl<'a> IdlArcSqliteWriteTransaction<'a> {
...
+         */
+
+        let mut data: HashMap<IdxKey, Vec<f64>> = HashMap::new();
+        self.idl_cache.iter_dirty().for_each(|(k, maybe_idl)| {
+            if let Some(idl) = maybe_idl {
+                let idl_len: u32 = idl.len().try_into().unwrap_or(u32::MAX);
+                // Convert to something we can use.
+                let idl_len = f64::from(idl_len);
+
+                let kref = IdxKeyRef::new(&k.a, &k.i);
```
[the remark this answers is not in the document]
Firstyear: it's an index key reference, and that's the attribute and index type. Yet again, shorthand bitees :)

kanidmd/src/lib/be/idl_arc_sqlite.rs
```diff
@@ -710,6 +716,257 @@ impl<'a> IdlArcSqliteWriteTransaction<'a> {
...
+                if idl_len > 0.0 {
+                    // It's worth looking at. Anything len 0 will be removed.
+                    if let Some(lens) = data.get_mut(&kref as &dyn IdxKeyToRef) {
+                        lens.push(idl_len)
+                    } else {
+                        data.insert(kref.to_key(), vec![idl_len]);
+                    }
```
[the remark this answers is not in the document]
Firstyear: Whoa boy no. The set holds an IdxKey not an IdxKeyRef. And the ref has a lifetime, and that lifetime has to survive multiple transactions. So we have to store IdxKeys. Then if we want to do a lookup we need K: Borrow<Q> where the input is Q. https://docs.rs/hashbrown/0.11.2/hashbrown/struct.HashMap.html#method.get

So either:
* We have to allocate and clone a new IdkKey every time in the loop to do the lookup
* We create a way that a reference can borrow into IdxKey. 

https://github.com/kanidm/kanidm/blob/master/kanidmd/src/lib/be/idxkey.rs#L7

This blog here explains the magci. Both IdxKey and IdxKeyRef impl dyn IdxKeyToRef, and that's the common Borrow implementation that we can use. 

This actually saves *tons* of time in lookups compared to the cloning / allocation strategy. Especially in some of the hotter paths of the code.

kanidmd/src/lib/be/idl_arc_sqlite.rs
```diff
@@ -710,6 +716,257 @@ impl<'a> IdlArcSqliteWriteTransaction<'a> {
...
+        *           ╱     │
+        *          ╱      │
+        *         ╱       │ sd_1
+        *        ╱        │
+        *       ╱         │
+        *      ───────────┼
+        *         nkeys
+        *
+        * Since this is right angled we can use arctan to work out the degress of the line. This
+        * gives us a value from 1.0 to 90.0 (We clamp to a minimum of 1.0, because we use 0 as "None"
```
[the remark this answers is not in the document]
Firstyear: Not really.

kanidmd/src/lib/be/idl_arc_sqlite.rs
```diff
@@ -710,6 +716,257 @@ impl<'a> IdlArcSqliteWriteTransaction<'a> {
...
+            // Cant resolve.
+            return IdxSlope::MAX;
+        };
+
+        // Now we know sd_1 and number of keys. We can use this as a triangle to work out
+        // the angle along the hypotenuse. We use this angle - or slope - to show which
+        // elements have the smallest sd_1 and most keys available. Then because this
+        // is bound between 0.0 -> 90.0, we "unfurl" this around a half circle by multipling
+        // by 2. This gives us a little more precision when we drop the decimal point.
+        let sf = (sd_1 / n_keys).atan().to_degrees() * 2.0;
```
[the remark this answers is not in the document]
Firstyear: Well 255.6 is greater than u8::max :) 

And we could have that extra precision, but it's not suuuppeeeerrrr important. I'd consider something like 2.81 which gets it to 253.

kanidmd/src/lib/be/idl_sqlite.rs
```diff
@@ -997,6 +1002,94 @@ impl IdlSqliteWriteTransaction {
...
+                .execute(
+                    "INSERT OR REPLACE INTO idxslope_analysis (id, slope) VALUES(:id, :slope)",
+                    named_params! {
+                        ":id": &key,
+                        ":slope": &v,
+                    },
+                )
+                .map(|_| ())
+                .map_err(|e| {
+                    eprintln!("CRITICAL: rusqlite error {:?}", e);
```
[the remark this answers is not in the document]
Firstyear: How much should the admin be shitting their pants at exactly this moment that something has gone wrong :)

kanidmd/src/lib/be/idxkey.rs
```diff
@@ -4,6 +4,8 @@ use std::borrow::Borrow;
 use std::cmp::Ordering;
 use std::hash::{Hash, Hasher};
 
+pub type IdxSlope = u8;
```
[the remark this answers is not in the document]
Firstyear: the reason for u8 is that initially it was u32. But then after looking at it, the logs were really messy. and really to an admin reading that, or us trying to debug it, it was hard to really see a difference between say Some(9082401) and Some(12389742). Making it a u8 makes it both really human readable (0 to 255) and still doesn't lose information.

kanidmd/src/lib/be/mod.rs
```diff
@@ -1469,6 +1478,21 @@ impl<'a> BackendWriteTransaction<'a> {
...
 }
 
+// We have a number of hardcoded, "obvious" slopes that should
+// exist. We return these when the analysis has not been run, as
+// these are values that are generally "good enough" for most applications
+fn get_idx_slope_default(ikey: &IdxKey) -> IdxSlope {
+    match (ikey.attr.as_str(), &ikey.itype) {
+        ("name", IndexType::Equality)
+        | ("spn", IndexType::Equality)
+        | ("uuid", IndexType::Equality) => 1,
```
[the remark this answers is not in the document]
Firstyear: Correct, these are the cheap indexes.

kanidmd/src/lib/server.rs
```diff
@@ -987,6 +987,12 @@ impl QueryServer {
             .upgrade_reindex(audit, SYSTEM_INDEX_VERSION + 1)
             .and_then(|_| reindex_write_2.commit(audit))?;
 
+        // Force the schema to reload - this is so that any changes to index slope
+        // analysis are now reflected correctly.
+        let slope_reload = task::block_on(self.write_async(ts));
+        slope_reload.force_schema_reload();
+        slope_reload.commit(audit)?;
+
```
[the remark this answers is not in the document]
Firstyear: Not very often. It's only called during a schema reload (so server startup, or if the schema changes). We do cache them into the IdxMeta structure which is reloaded in the backend during a schema reload.

kanidmd/src/lib/filter.rs
Firstyear: Because it's not actually quite the same :) hashbrown uses ahash as the default hasher which is significantly faster than rust's std default hasher which is siphash. There are also some extra optimisations in hashbrown that aren't in the std version iirc.

kanidmd/src/lib/filter.rs
Firstyear: An inclusion is similar to an OR - so we want to know "what's the most expensive part of this query going to be" since we have to access EVERY index in the OR/Inclusion term. This way if we have a query like:

`x or inclusion(a, b, c)`

we want to check x first since it may *fail* faster allowing us to skip the more expensive inclusion. If we get to the expensive inclusion term, then we want to be at a point where our partial candidate set is smaller, allowing quicker merging of the data, and knowing that we will need to actually process it.

kanidmd/src/lib/filter.rs
Firstyear: we need to wait for the sort/dedup first then we can get the max elements, so we could only call the get slopeness after we construction the new or, but to construct the new or we need the slopefactor .... see the problem :)

kanidmd/src/lib/filter.rs
Firstyear: Change to the optimiser behaviour.

kanidmd/src/lib/filter.rs
Firstyear: There is a whole rant about how queries work. Each query has a lifecycle basicly. (There are quite a few types that have enforced compile time state machines to guarantee certain behaviour).

If it's from external, it starts as a "ProtoFilter" which is basicly an unsanisited user request.

we then turn that into a "Filter<FilterInvalid>". This is our "internal" structure of an unsanitised query. 

We can transform, alter, modify, or whatever else to the unsanisited query at this point. 

Next we then sanitise this. We process the filter through to "Filter<FilterValid>" which looks through the schema to ensure that every requested attribute, does exist in the schema (so it will be in the db or not). We also check that every request query key is the correct syntax type for the related element. This prevents a denial of service attack that affects most ldap servers (query an attribute that doesn't exist can force a full table scan). 

Once we have a FilterValid we can NOT change the content because any change introduce MAY be invalid and cause issues. 

A FilterValid is a query that is sanitised but it's not yet ready to work in the database. For example, a valid query request is "selfUUID" aka the UUID of the currently authenticated account. 

The next step is FilterValid to FilterValidResolved. This is where we do some internal transforms such as SelfUUID being extracted to an Eq term from the currently authenticated user (this is super important for access controls!). At this point we also *enrich* the query with metadata about the current indexing state of the database (which are the idx slopse, previously a bool of indexed/vs not). 

This metadata is critical for the underlying backend to know if a query component CAN have an index accessed or not. The point of this change changes that metadata from simply true/false to a U8 which allows indication of no-index (0, or in this case, None because of NonZero U8), or that it does have an index (1-255) and what the value of that index is. That's what's used to optimise the query to put "faster terms first".


The final piece of this that you may be missing is the filter threshold - in filter2idl in be/mod.rs, as a filter is being processed, if the partial candidate set falls below this threshold value, we shortcut and return the partial candidate set, and then perform the remaining filter tests on the entry directly.


This is why if you have:

`class=account and name=quinn`
vs
`name=quinn and class=account`

If you did the first query in that order, youd load a huge index, potentially 60,000 values, and then throw away most of it when you hit "name=quinn" which should be length 1. 

But in the other case, you hit name=quinn which is length one, we are below the threshold, so we shortcut return and we just check "does this entry also have class=account?". This avoids us hitting the really large index.

The goal - create queries that are going to get us below that threshold, faster, and if not, create queries that will narrow down more specific elements quicker that may fail so we avoid larger blocks ogf work unless really needed.

kanidmd/src/lib/be/idl_arc_sqlite.rs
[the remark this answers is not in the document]
QnnOkabayashi: So in theory, we could've made it take an `IdxKey` and then cloned for each lookup, and then the `get_mut` argument and the `insert` argument would've taken the same type, but instead we're doing this to avoid cloning since it's more expensive, especially since it contains `String`.

Also, I think we can delete this `impl` block because it doesn't seem to be used (?)
https://github.com/kanidm/kanidm/blob/93d5c5434d67ee6231526b9de9f0c6d79f012bb2/kanidmd/src/lib/be/idxkey.rs#L21-L25

kanidmd/src/lib/be/idl_arc_sqlite.rs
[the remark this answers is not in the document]
QnnOkabayashi: So we only care about the ordering, and not by how much? Because if there are two entries where `nkeys` is small but `sd_1` isn't and they're not even close to each other, the angle is still going to look pretty much the same, even though one might be much better than the other.

kanidmd/src/lib/filter.rs
QnnOkabayashi: > There is a whole rant about how queries work. Each query has a lifecycle basicly. (There are quite a few types that have enforced compile time state machines to guarantee certain behaviour).
> 
> If it's from external, it starts as a "ProtoFilter" which is basicly an unsanisited user request.
> 
> we then turn that into a "Filter". This is our "internal" structure of an unsanitised query.
> 
> We can transform, alter, modify, or whatever else to the unsanisited query at this point.

I saw some "optimizing" functions earlier, is this what those were handling?

> 
> Next we then sanitise this. We process the filter through to "Filter" which looks through the schema to ensure that every requested attribute, does exist in the schema (so it will be in the db or not). We also check that every request query key is the correct syntax type for the related element. This prevents a denial of service attack that affects most ldap servers (query an attribute that doesn't exist can force a full table scan).

What would you estimate the complexity of checking for the correct syntax type of a request query key is? O(1)?

> 
> Once we have a FilterValid we can NOT change the content because any change introduce MAY be invalid and cause issues.

How come changes here may be invalid but changes before were okay? Also what kinds of changes were we doing before?

> 
> A FilterValid is a query that is sanitised but it's not yet ready to work in the database. For example, a valid query request is "selfUUID" aka the UUID of the currently authenticated account.
> 
> The next step is FilterValid to FilterValidResolved. This is where we do some internal transforms such as SelfUUID being extracted to an Eq term from the currently authenticated user (this is super important for access controls!). At this point we also _enrich_ the query with metadata about the current indexing state of the database (which are the idx slopse, previously a bool of indexed/vs not).
> 
> This metadata is critical for the underlying backend to know if a query component CAN have an index accessed or not. The point of this change changes that metadata from simply true/false to a U8 which allows indication of no-index (0, or in this case, None because of NonZero U8), or that it does have an index (1-255) and what the value of that index is. That's what's used to optimise the query to put "faster terms first".
> 
> The final piece of this that you may be missing is the filter threshold - in filter2idl in be/mod.rs, as a filter is being processed, if the partial candidate set falls below this threshold value, we shortcut and return the partial candidate set, and then perform the remaining filter tests on the entry directly.
> 
> This is why if you have:
> 
> `class=account and name=quinn`
> vs
> `name=quinn and class=account`
> 
> If you did the first query in that order, youd load a huge index, potentially 60,000 values, and then throw away most of it when you hit "name=quinn" which should be length 1.
> 
> But in the other case, you hit name=quinn which is length one, we are below the threshold, so we shortcut return and we just check "does this entry also have class=account?". This avoids us hitting the really large index.
> 
> The goal - create queries that are going to get us below that threshold, faster, and if not, create queries that will narrow down more specific elements quicker that may fail so we avoid larger blocks ogf work unless really needed.

So essentially, this PR allows us to take a query and sort the attributes from rarest to most common, so we can search in a way that will most rapidly decrease the candidate entries we have to search through?

kanidmd/src/lib/filter.rs
QnnOkabayashi: Okay, the "what's the most expensive part of this query going to be" was the part that I wasn't sure about, but definitely aligns with the code. Also, what's the difference between `Inclusion` and `Or`?

QnnOkabayashi (approved):

kanidmd/src/lib/be/idl_arc_sqlite.rs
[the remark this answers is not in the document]
Firstyear: I think was experimenting with using it, but honestly, it's not harmful to leave it.

kanidmd/src/lib/be/idl_arc_sqlite.rs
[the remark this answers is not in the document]
Firstyear: We care abotu the ordering when they are extremely different, but not when the difference is low.

For example - name=foo or uuid=x are both going to yield an idl of length 1, so we don't care which is chosen. We just want to make sure either of those terms are looked at "first" before class=account.

kanidmd/src/lib/be/mod.rs
```diff
@@ -1333,35 +1381,6 @@ impl<'a> BackendWriteTransaction<'a> {
             OperationError::SerdeJsonError
         })?;
 
-        // Filter all elements that have a UUID in the system range.
```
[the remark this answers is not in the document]
Firstyear: Because it's all commented out :P

kanidmd/src/lib/filter.rs
Firstyear: Inclusion are an internal piece of magic for plugin behaviour - they can't be generated from external calls, as they don't "mean" anything in a search query. Specifically it's used in referential integrity https://github.com/kanidm/kanidm/blob/master/kanidmd/src/lib/plugins/refint.rs#L60 . The server has a second "lighter" search path for "existance" or rather "does this query return true/false". Inclusion is different to AND/OR because:

* AND expresses that all terms must hold true.
* OR expresses that at least one term holds true.
* Inclusion expresses that every term holds true at least ONCE.

Consider a query like:

```
uuid=A  OP uuid=B
```

If OP was AND, then this would always be false - we don't have an entry that could have both uuid's, since they are on seperate entries.

If OP was OR, then this would be true, provided A OR B exist.

But in referential integrity we need assert uuid=A and uuid=B both exist, but may be on seperate entries. That's what inclusion does. It allows expressing that each term must hold true on at least one entry, and any term that is false, will shortcut us out. There are some extra rules in inclusion/exists queries (they don't have a filter threshold because we need to force every indexed element of the inclusion to be evaluated else we'll have an inconsistent candidate set, exists queries disable shortcutting because we want to test indexes as much as possible as we wont load entries etc.)

The point is - Inclusion allows refint to assert the existance of many things in one query, rather than issuing many individual queries in sequence, and the fn exists allows testing a query for truth without loading entries from id2entry.

kanidmd/src/lib/filter.rs
Firstyear: > I saw some "optimizing" functions earlier, is this what those were handling?

Possibly? I'd need to see what you mean. But essential the call to optimise/fast_optimise can only be done on a ValidResolved query because that's the only one with enough metadata. 

> What would you estimate the complexity of checking for the correct syntax type of a request query key is? O(1)?

Big O and algorithmic complexity generally are actually pointless in any real world discussion. They only give you some hand waving theoretical properities about maybe performance - consider binary search tree vs b-tree/b+tree. bst has better algorithmic complexity, but b-tree absolutely wipes the floor with bst in real world. Turns out a whole generation of computer scientists deluded themself into thinking they were mathematicians and decided to literally ignore how computers actually work. Similar hashmaps are O(1) but let me tell you about "hash map performance" sometime. It's never so simple. Once you add threads, it's even more fun.

So I think the complexity is O(who knows). :) 

The more important metric here is about which datastructures, how we access them in the transaction, and how we do these lookups, have we got cache hot components, etc. 

Syntax validation is done here:

https://github.com/kanidm/kanidm/blob/master/kanidmd/src/lib/filter.rs#L599

We recurse through each filter element, and apply the schema attributes syntax validator to it. SchemaTransaction is a readonly in memory hashmap, which given it's shared between threads, copy-on-write, it will be in the S state of your cpu l1/l2/l3 cache (see MESI). The schema elements are tested to ensure they match their expected partial value type. You'll also notice that the attribute type is "AttrString" - this is actually "SmartString" which is an inline string, which prevents an allocation/dereference, which improves cache locality and performance too.

The bigger gain here, could be to investigate doing this with a stack/iteratively rather than recursion, but there is also a question of where are other gains to be found. 

> How come changes here may be invalid but changes before were okay? Also what kinds of changes were we doing before?

Any change could be invalid because even as programmers, we can make mistakes. Trust me, forcing this validation even to internal server code has already caught and prevented multiple bugs. 

https://external-content.duckduckgo.com/iu/?u=https%3A%2F%2Fwww.nesglobal.net%2Fwp-content%2Fuploads%2FHierarchy-of-Controls-1-1.png&f=1&nofb=1

Examples of an internal change we make is that most queries are wrapped in an exclusion of tombstone/recycled class types so that we can't see anything in the delete paths.

> So essentially, this PR allows us to take a query and sort the attributes from rarest to most common, so we can search in a way that will most rapidly decrease the candidate entries we have to search through?

Correct! And it means that the decision about how to do that sort is based on YOUR data, not just whatever "we" think will be best.

kanidmd/src/lib/be/mod.rs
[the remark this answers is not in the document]
QnnOkabayashi: Please forget that I wrote that comment :)

kanidmd/src/lib/be/idl_sqlite.rs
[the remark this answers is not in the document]
QnnOkabayashi: Would it be a good idea to add a feature in https://github.com/kanidm/kanidm/pull/497 so that regular logs go to a file, but we could have a special `alarm!` macro that goes through the subscriber and is written to a log but also immediately outputted to `stderr` or wherever is most appropriate/visible?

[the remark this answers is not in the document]
Firstyear: Yes, that's a very good idea, and you should definitely add that to the list of requirements/nice to have :)

victorcwai (approved):
Just finished reading you guys' discussion. Looks good to me :)

PR #507, opened 2021-06-29
Fixing kanidm windows client build

yaleman:
- [x] cargo fmt has been run

This removes some of the umask/uid specific checks to yeet this into prod as windows rust libs can't handle such checks.

### DIFF
--- a/kanidm_client/src/lib.rs
+++ b/kanidm_client/src/lib.rs
@@ -14,7 +14,6 @@ extern crate log;
 use serde_derive::Deserialize;
 use serde_json::error::Error as SerdeJsonError;
 use std::collections::BTreeSet as Set;
-use std::env::consts::FAMILY;
 use std::fs::{metadata, File, Metadata};
 use std::io::ErrorKind;
 use std::io::Read;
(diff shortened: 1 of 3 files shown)


kanidm_client/src/lib.rs
```diff
@@ -100,18 +104,23 @@ impl KanidmClientBuilder {
     fn parse_certificate(ca_path: &str) -> Result<reqwest::Certificate, ()> {
         let mut buf = Vec::new();
         // Is the CA secure?
-        let path = Path::new(ca_path);
-        let ca_meta = read_file_metadata(&path)?;
-
-        if !ca_meta.permissions().readonly() {
-            warn!("permissions on {} may not be secure. Should be readonly to running uid. This could be a security risk ...", ca_path);
-        }
+        if FAMILY == "windows" {
```
Firstyear, line 107: Why not use https://doc.rust-lang.org/std/macro.cfg.html ?

yaleman (commented):
Fixed

Firstyear:
Yep, seems fine to me. The other is there is a cfg!() macro you can use for if conditions instead if you need something "at run time" rather than compile time :)

PR #515, opened 2021-06-30
fixes #503 - TOTP prompt no longer drops a newline

### LINKED ISSUE #503: TOTP prompt input echoing on a second line in CLI, rather than same one
:3

Screen Shot 2021-06-28 at 11 21 35

yaleman:
Fixes #503 

- [x] cargo fmt has been run

### DIFF
--- a/kanidm_tools/src/cli/session.rs
+++ b/kanidm_tools/src/cli/session.rs
@@ -145,6 +145,7 @@ impl LoginOpt {
     fn do_totp(&self, client: &mut KanidmClient) -> Result<AuthResponse, ClientError> {
         let totp = loop {
             print!("Enter TOTP: ");
+            // We flush stdout so it'll write the buffer to screen, continuing operation. Without it, the application halts.
             io::stdout().flush().unwrap();
             let mut buffer = String::new();
             if let Err(e) = io::stdin().read_line(&mut buffer) {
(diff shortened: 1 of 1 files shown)


kanidm_tools/src/cli/session.rs
```diff
@@ -144,7 +144,8 @@ impl LoginOpt {
 
     fn do_totp(&self, client: &mut KanidmClient) -> Result<AuthResponse, ClientError> {
         let totp = loop {
-            println!("Enter TOTP: ");
+            print!("Enter TOTP: ");
+            io::stdout().flush().unwrap();
```
Firstyear, line 148: Curious what the stdout flush is for, but I trust it's needed :)

yaleman: Without the flush the app just hangs because stdout's like "yeah, I buffered it, now what?" 🙄

Firstyear: See I knew there would be a good reason :)

Firstyear: Actually, can you add that as a comment above that line so that future archelogists know this too?

yaleman: Done :)

Issue #506, opened 2021-06-28
TOTP registration fails


cuberoot74088:
Hi,
I am trying to register_totp on a test-account. Webauthn is working already. So I have this starting point:

> $  kanidm account credential status xxx --name xxx     
> uuid: 26xxx76
> claims:
> password: set
> totp: disabled
> backup_code: disabled
> webauthn:
>  * key

and using kanidm version from Jun 27, 1b146bd

If I go for the TOTP, I get this error after sending back my code ...
> Error Completing -> TotpVerifyFailed

What can I do to learn why it failed? It seems like it should work, right?

yaleman:
Was this with both apps mentioned in #501? There was a change to kanidm's hashing algorithm in #504 which reduced the hash algorithm from SHA512 to SHA256 which is like to solve the problem.

Firstyear:
It could simply be that the time window passed - there is some debounce code in there to handle this but perhaps if you take longer than ~60 seconds to enter the code, it can expire. If you tried again did it work? 

@yaleman There would be an issue registering the token then, rather than in the verification phase.

yaleman:
I'm a derp, I meant to include this in the other ticket - the error you get back is a verification fail - so it matches previous reported issues.

```Algorithm: SHA512
Period/Step: 30
Secret: X2DUK2WVY23I2
--------------------------------------------------------------
Enter a TOTP from your authenticator to complete registration:
TOTP: 850994
Error Completing -> TotpVerifyFailed(00000000-60da-7102-0b2d-f27961097c0d, 
TotpSecret { accountname: "yaleman", issuer: "yaleman@<myserver>", 
secret: [190, 135, 69, 106, 213, 198, 182, 141], algo: Sha512, step: 30 })
```

Firstyear:
verification failure would indicate wrong secret (unlikely) or timer expiry of the code you entered I think. I added some code to allow "TOTP minus one" so that means any code should be valid for 60 seconds. Did you possibly take longer than that to enter the code?

Firstyear:
Alternately, is there a clock sync issue between your server and the client? That would also affect this?

yaleman:
My above fail was fixed by moving from Authy to Okta client, but this is @cuberoot74088's thread so I'll wait for him 😆

cuberoot74088:
I gave it another try, took me around 25 seconds to scan and input the code ... it still fails verification, but I am really wondering about the time zones here:

> Algorithm: SHA512
[...]

PR #516, opened 2021-07-01
Fix totp registration workflow with broken authenticators

### LINKED ISSUE #501: TOTP app compatibility
Testing the CLI TOTP registration process this morning I couldn't get it to work with Authy's iOS app, which I'm guessing is a SHA512 support issue? Still researching.

Here's a few checks

❌ Authy iOS (v24.5.1)
✅ Okta iOS (v6.5.0)
✅ Google Authenticator (v3.2.0)

Firstyear:
Fixes #501 #506 - This changes the workflow in totp registration, so that we can detect broken authenticators. When we detect these, we request consent from the user to proceed in allowing this registration to proceed for the account. 

### DIFF
--- a/kanidm_tools/src/cli/account.rs
+++ b/kanidm_tools/src/cli/account.rs
@@ -192,7 +192,8 @@ impl AccountOpt {
                     eprintln!("--------------------------------------------------------------");
                     eprintln!("Enter a TOTP from your authenticator to complete registration:");
 
-                    loop {
+                    let mut attempts = 3;
+                    while attempts > 0 {
                         eprint!("TOTP: ");
                         let mut totp_input = String::new();
                         let input_result = io::stdin().read_line(&mut totp_input);
@@ -237,7 +238,7 @@ impl AccountOpt {
                                     break;
                                 };
 
-                                if confirm_input.trim() == "I am sure" {
+                                if confirm_input.to_lowercase().trim() == "i am sure" {
                                     match client.idm_account_primary_credential_accept_sha1_totp(
                                         acsopt.aopts.account_id.as_str(),
                                         session,
@@ -257,6 +258,7 @@ impl AccountOpt {
                             }
                             Err(ClientError::TotpVerifyFailed(_, _)) => {
                                 eprintln!("Incorrect TOTP code - try again");
+                                attempts -= 1;
                                 continue;
                             }
                             Err(e) => {
--- a/kanidm_tools/src/cli/account.rs
+++ b/kanidm_tools/src/cli/account.rs
@@ -9,6 +9,7 @@ use time::OffsetDateTime;
 
 use webauthn_authenticator_rs::{u2fhid::U2FHid, WebauthnAuthenticator};
 
+use kanidm_client::ClientError;
 use kanidm_client::ClientError::Http as ClientErrorHttp;
 use kanidm_proto::v1::OperationError::{PasswordBadListed, PasswordTooShort, PasswordTooWeak};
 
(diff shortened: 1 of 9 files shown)


Firstyear:
cc @cuberoot74088

yaleman (approved):
This works for me in prod, thanks :)

Firstyear:
I'll let @cuberoot74088 have a look in the next ~24 hours, then I'll merge it. Thanks for always reviewing @yaleman

cuberoot74088:
@Firstyear I will give it a try later today. Any thoughts on auto-magic downgrade vs. new user request with selected algo?

Firstyear:
> @Firstyear I will give it a try later today. Any thoughts on auto-magic downgrade vs. new user request with selected algo?

Many :) there is a reason I implemented this the way I did (request to downgrade after invalid authenticator).

The main reason is that we want the best defaults whenever possible. When someone writes documentation, or a blog, they'll say "kanidm account credentials register_totp". Now all it takes is one blogger, who hits this issue to say "need to add --algo=sha1" for this to work. People will copy paste it, and they'll have a different experience. People may question the need for the algo selection etc. They'll use sha1 without a consideration of the consequences (and why should they know those consequences? Not everyone is an authentication specialist, or knows the ins/outs of authenticator apps)

The intent here is to reduce options to guide people to a best default case, without the risk of weaking others. By having this as "auto-magic", we get:

* Strong defaults, for the majority of users, and less confusion because less options
* Clear, direct warnings for users who *are* affected by this issue, and that they have to consent to the downgrade
* An ability in the future for security policy to control if we answer this query allowing sha1 or not at all. 

So I've implemented the automagic method, where if you scan, get a sha1 authenticator, we warn you, and you have to consent to proceed. I don't want to be in a position where users have to "become experts" on a topic that isn't their area of expertise just to enable 2fa.

cuberoot74088:
Hi @Firstyear,
thanks for taking the time and sharing your thoughts! :+1: 

I think your points are very valid, especially reducing the complexity (more options etc.) and demanding the expertise. I also see another benefit in the downgrade workflow, as the user does not need to scan another code.

I can also confirm, it is working:

> Issuer: demo_user2@example.com
> Algorithm: SHA256
> Period/Step: 30
> Secret: K6VLL6GXBJQBQ
>
> Enter a TOTP from your authenticator to complete registration:
> TOTP: 635291
> 
> ⚠️  WARNING - It appears your authenticator app may be broken ⚠️  
>  The TOTP authenticator you are using is forcing the use of SHA1
> 
>  -- If you accept this risk, and wish to proceed, type 'I am sure' 
>  -- Otherwise press ENTER to cancel this operation
> 
> Are you sure: I am sure
> TOTP registration success.


Another comment is the input loop. Maybe 3 or 10 times would be enough? Seems like I can try this forever

> Incorrect TOTP code - try again
> TOTP: 3
> 
> 4Incorrect TOTP code - try again
> TOTP: 

Again, huge thanks for the quick fix and the explanations :1st_place_medal:

kanidm_tools/src/cli/account.rs
```diff
@@ -189,37 +190,81 @@ impl AccountOpt {
...
+                                eprintln!("");
+                                eprint!("Are you sure: ");
+
+                                let mut confirm_input = String::new();
+                                if let Err(e) = io::stdin().read_line(&mut confirm_input) {
+                                    eprintln!("Failed to read from stdin -> {:?}", e);
+                                    break;
+                                };
+
+                                if confirm_input.trim() == "I am sure" {
```
cuberoot74088, line 240: Should that be case sensitive?

kanidm_tools/src/cli/account.rs
```diff
@@ -189,37 +190,81 @@ impl AccountOpt {
...
+                                    };
+                                    break;
+                                } else {
+                                    eprintln!("Cancelling TOTP registration");
+                                    break;
+                                }
+                            }
+                            Err(ClientError::TotpVerifyFailed(_, _)) => {
+                                eprintln!("Incorrect TOTP code - try again");
+                                continue;
```
cuberoot74088, line 260: As far as I have seen and tried, this will loop over and over again. Do we want this? Maybe limit the maximum tries?

kanidm_tools/src/cli/account.rs
yaleman: I'd argue yes - you really want them to type it properly, or copy-paste it like every one of these things I come across 😆

kanidm_tools/src/cli/account.rs
Firstyear: I think a limit here is a good idea. :)

kanidm_tools/src/cli/account.rs
Firstyear: I think even just at a point they type 'i am sure' that's still strong enough "intent" that insensitive here is probably fine. We just need it to be a deliberate and intended action, rather than an accidental accept.

Firstyear:
Applied @cuberoot74088's feedback :)

PR #517, opened 2021-07-03
495 backup codes cli extension

### LINKED ISSUE #495: Extend kanidm_tools to create and reset backup codes in CLI
See https://github.com/kanidm/kanidm/pull/469#issuecomment-868157709.

cuberoot74088:
Fixes #495

Started to work on that.
I am able to generate and remove backup codes now using the latest server version.

What else should be done for this ticket?

- [X] cargo fmt has been run

### DIFF
--- a/kanidm_client/src/lib.rs
+++ b/kanidm_client/src/lib.rs
@@ -425,6 +425,10 @@ impl KanidmClient {
         tokio_block_on(self.asclient.auth_step_password(password))
     }
 
+    pub fn auth_step_backup_code(&self, backup_code: &str) -> Result<AuthResponse, ClientError> {
+        tokio_block_on(self.asclient.auth_step_backup_code(backup_code))
+    }
+
     pub fn auth_step_totp(&self, totp: u32) -> Result<AuthResponse, ClientError> {
         tokio_block_on(self.asclient.auth_step_totp(totp))
     }
--- a/kanidm_client/src/lib.rs
+++ b/kanidm_client/src/lib.rs
@@ -701,6 +701,27 @@ impl KanidmClient {
         )
     }
 
+    // backup codes
+    pub fn idm_account_primary_credential_generate_backup_code(
+        &self,
+        id: &str,
+    ) -> Result<Vec<String>, ClientError> {
+        tokio_block_on(
+            self.asclient
+                .idm_account_primary_credential_generate_backup_code(id),
+        )
+    }
+
+    pub fn idm_account_primary_credential_remove_backup_code(
+        &self,
+        id: &str,
+    ) -> Result<(), ClientError> {
+        tokio_block_on(
+            self.asclient
+                .idm_account_primary_credential_remove_backup_code(id),
+        )
+    }
+
     pub fn idm_account_get_credential_status(
         &self,
         id: &str,
--- a/kanidm_client/src/lib.rs
+++ b/kanidm_client/src/lib.rs
@@ -701,7 +701,6 @@ impl KanidmClient {
         )
     }
 
-    // backup codes
     pub fn idm_account_primary_credential_generate_backup_code(
         &self,
         id: &str,
--- a/kanidm_tools/src/cli/account.rs
+++ b/kanidm_tools/src/cli/account.rs
@@ -290,7 +290,9 @@ impl AccountOpt {
                     ) {
                         Ok(s) => {
                             println!("GenerateBackupCode success.");
-                            println!("{:#?}", s);
+                            println!("---");
+                            println!("{}", s.join("\n"));
+                            println!("---");
                         }
                         Err(e) => {
                             eprintln!("Error GenerateBackupCode for account -> {:?}", e);
--- a/kanidm_tools/src/cli/account.rs
+++ b/kanidm_tools/src/cli/account.rs
@@ -290,7 +290,7 @@ impl AccountOpt {
                     ) {
                         Ok(s) => {
                             println!("GenerateBackupCode success.");
-                            println!("{:?}", s);
+                            println!("{:#?}", s);
                         }
                         Err(e) => {
                             eprintln!("Error GenerateBackupCode for account -> {:?}", e);
--- a/kanidm_tools/src/cli/account.rs
+++ b/kanidm_tools/src/cli/account.rs
@@ -289,13 +289,13 @@ impl AccountOpt {
                         acsopt.aopts.account_id.as_str(),
                     ) {
                         Ok(s) => {
-                            println!("GenerateBackupCode success.");
+                            println!("Please store these Backup codes in a safe place");
                             println!("---");
                             println!("{}", s.join("\n"));
                             println!("---");
                         }
                         Err(e) => {
-                            eprintln!("Error GenerateBackupCode for account -> {:?}", e);
+                            eprintln!("Error generating Backup Codes for account -> {:?}", e);
                         }
                     }
                 }
--- a/kanidm_tools/src/cli/account.rs
+++ b/kanidm_tools/src/cli/account.rs
@@ -23,6 +23,8 @@ impl AccountOpt {
                 AccountCredential::RemoveWebauthn(acs) => acs.copt.debug,
                 AccountCredential::RegisterTotp(acs) => acs.copt.debug,
                 AccountCredential::RemoveTotp(acs) => acs.copt.debug,
+                AccountCredential::GenerateBackupCode(acs) => acs.copt.debug,
+                AccountCredential::BackupCodeRemove(acs) => acs.copt.debug,
                 AccountCredential::Status(acs) => acs.copt.debug,
             },
             AccountOpt::Radius(acopt) => match acopt {
@@ -278,6 +280,33 @@ impl AccountOpt {
                         }
                         Err(e) => {
                             eprintln!("Error Removing TOTP from account -> {:?}", e);
+                        }
+                    }
+                }
+                AccountCredential::GenerateBackupCode(acsopt) => {
+                    let client = acsopt.copt.to_client();
+                    match client.idm_account_primary_credential_generate_backup_code(
+                        acsopt.aopts.account_id.as_str(),
+                    ) {
+                        Ok(s) => {
+                            println!("GenerateBackupCode success.");
+                            println!("{:?}", s);
+                        }
+                        Err(e) => {
+                            eprintln!("Error GenerateBackupCode for account -> {:?}", e);
+                        }
+                    }
+                }
+                AccountCredential::BackupCodeRemove(acsopt) => {
+                    let client = acsopt.copt.to_client();
+                    match client.idm_account_primary_credential_remove_backup_code(
+                        acsopt.aopts.account_id.as_str(),
+                    ) {
+                        Ok(_) => {
+                            println!("BackupCodeRemove success.");
+                        }
+                        Err(e) => {
+                            eprintln!("Error BackupCodeRemove for account -> {:?}", e);
                         }
                     }
                 }
--- a/kanidm_tools/src/cli/session.rs
+++ b/kanidm_tools/src/cli/session.rs
@@ -142,6 +142,15 @@ impl LoginOpt {
         client.auth_step_password(password.as_str())
     }
 
+    fn do_backup_code(&self, client: &mut KanidmClient) -> Result<AuthResponse, ClientError> {
+        let mut backup_code = String::new();
+        if let Err(e) = io::stdin().read_line(&mut backup_code) {
+            eprintln!("Failed to read from stdin -> {:?}", e);
+            return Err(ClientError::SystemError);
+        };
+        client.auth_step_backup_code(backup_code.as_str())
+    }
+
     fn do_totp(&self, client: &mut KanidmClient) -> Result<AuthResponse, ClientError> {
         let totp = loop {
             print!("Enter TOTP: ");
@@ -264,6 +273,7 @@ impl LoginOpt {
             let res = match choice {
                 AuthAllowed::Anonymous => client.auth_step_anonymous(),
                 AuthAllowed::Password => self.do_password(&mut client),
+                AuthAllowed::BackupCode => self.do_backup_code(&mut client),
                 AuthAllowed::Totp => self.do_totp(&mut client),
                 AuthAllowed::Webauthn(chal) => self.do_webauthn(&mut client, chal.clone()),
             };
--- a/kanidm_tools/src/cli/session.rs
+++ b/kanidm_tools/src/cli/session.rs
@@ -143,12 +143,15 @@ impl LoginOpt {
     }
 
     fn do_backup_code(&self, client: &mut KanidmClient) -> Result<AuthResponse, ClientError> {
+        print!("Enter Backup Code: ");
+        // We flush stdout so it'll write the buffer to screen, continuing operation. Without it, the application halts.
+        io::stdout().flush().unwrap();
         let mut backup_code = String::new();
         if let Err(e) = io::stdin().read_line(&mut backup_code) {
             eprintln!("Failed to read from stdin -> {:?}", e);
             return Err(ClientError::SystemError);
         };
-        client.auth_step_backup_code(backup_code.as_str())
+        client.auth_step_backup_code(backup_code.trim())
     }
 
     fn do_totp(&self, client: &mut KanidmClient) -> Result<AuthResponse, ClientError> {
--- a/kanidm_tools/src/cli/session.rs
+++ b/kanidm_tools/src/cli/session.rs
@@ -147,10 +147,15 @@ impl LoginOpt {
         // We flush stdout so it'll write the buffer to screen, continuing operation. Without it, the application halts.
         io::stdout().flush().unwrap();
         let mut backup_code = String::new();
-        if let Err(e) = io::stdin().read_line(&mut backup_code) {
-            eprintln!("Failed to read from stdin -> {:?}", e);
-            return Err(ClientError::SystemError);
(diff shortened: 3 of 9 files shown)


cuberoot74088:
Hi @victorcwai and @Firstyear,
I was playing around with this a bit now and have a couple of open questions:

1. Formatting the backup code output?
* Is the current way of printing the generated codes using `{:#?}` okay? [here](https://github.com/kanidm/kanidm/pull/517/files#diff-4ea4d56b07466d12e7ad39e08667124dd3041e68b042d4ef38b5d68f8140198cR293)   
* Or somehow make use of `get_backup_code_view()` -> `BackupCodesView`? (not entirely sure if this would be the right one)


2. How to use it for `kanidm login`?
* If I have totp configured and backup codes enabled ... I don't know how to make use of the backup codes.
* a `login` directly requests the totp code and it allows only numbers there.
* looks like we would need to adjust someting like `auth_step`, e.g.,  https://github.com/kanidm/kanidm/blob/master/kanidm_client/src/asynchronous.rs#L484 or so. 
* Or maybe have it as even have it AuthAllowed? (I am a bit confused now)
* but that sounds more like a separate issue.

3. (Re)Naming
* to fit to the other entries in `SetCredentialRequest` enum
* https://github.com/kanidm/kanidm/blob/master/kanidm_proto/src/v1.rs#L684
* how about renaming `GenerateBackupCode` --> `BackupCodeGenerate`?
* could also be a separate ticket, but might also be done here.

What are your thoughts on these?

Firstyear:
Since @victorcwai implemented this I'd like them to give the first set of comments in this review :)

victorcwai:
> 1. Formatting the backup code output?
> 
> * Is the current way of printing the generated codes using `{:#?}` okay? [here](https://github.com/kanidm/kanidm/pull/517/files#diff-4ea4d56b07466d12e7ad39e08667124dd3041e68b042d4ef38b5d68f8140198cR293)

I think this is fine 👍 

> 1. How to use it for `kanidm login`?
> 
> * If I have totp configured and backup codes enabled ... I don't know how to make use of the backup codes.
> * a `login` directly requests the totp code and it allows only numbers there.
> * looks like we would need to adjust someting like `auth_step`, e.g.,  https://github.com/kanidm/kanidm/blob/master/kanidm_client/src/asynchronous.rs#L484 or so.

Backup codes were implemented only with HTTP API, using this endpoint:
https://github.com/kanidm/kanidm/blob/master/kanidmd/src/lib/core/https/mod.rs#L1242

So yeah you should make add an `auth_step` for backup code here, which should be quite similar to `auth_step_totp` (with AuthCredential::BackupCode).

> * Or maybe have it as even have it AuthAllowed? (I am a bit confused now)
> * but that sounds more like a separate issue.

Not sure about this one... @Firstyear 

> 1. (Re)Naming
> 
> * to fit to the other entries in `SetCredentialRequest` enum
> * https://github.com/kanidm/kanidm/blob/master/kanidm_proto/src/v1.rs#L684
> * how about renaming `GenerateBackupCode` --> `BackupCodeGenerate`?
> * could also be a separate ticket, but might also be done here.
> 
> What are your thoughts on these?

Agreed. I wrote that following `GeneratePassword` 😆  (https://github.com/kanidm/kanidm/blob/master/kanidm_proto/src/v1.rs#L673).
I think we should change both to `BackupCodeGenerate` and `GeneratePassword`.

kanidm_tools/src/cli/account.rs
```diff
@@ -281,6 +283,33 @@ impl AccountOpt {
...
                     }
                 }
+                AccountCredential::GenerateBackupCode(acsopt) => {
+                    let client = acsopt.copt.to_client();
+                    match client.idm_account_primary_credential_generate_backup_code(
+                        acsopt.aopts.account_id.as_str(),
+                    ) {
+                        Ok(s) => {
+                            println!("GenerateBackupCode success.");
+                            println!("{:#?}", s);
```
Firstyear, line 293: I know @victorcwai said this was oky, but I think we should have this as `{}` rather than `{:#?}`. The former is "Display" the latter is "Debug". When we are debbuging, or an error, we want that. But for "user facing" information, the default should be to provide it with "display". There are some subtle examples of this distinction in the proto files, where we seperate "debug" and "display" based on how and when we are presenting the information.

Firstyear:
> > * a `login` directly requests the totp code and it allows only numbers there.
> > * looks like we would need to adjust someting like `auth_step`, e.g.,  https://github.com/kanidm/kanidm/blob/master/kanidm_client/src/asynchronous.rs#L484 or so.
> 
> Backup codes were implemented only with HTTP API, using this endpoint:
> https://github.com/kanidm/kanidm/blob/master/kanidmd/src/lib/core/https/mod.rs#L1242
> 
> So yeah you should make add an `auth_step` for backup code here, which should be quite similar to `auth_step_totp` (with AuthCredential::BackupCode).

Correct, you'll need to have a different handler that knows how to prompt based on the next challenge step :) 

> 
> > * Or maybe have it as even have it AuthAllowed? (I am a bit confused now)
> > * but that sounds more like a separate issue.
> 
> Not sure about this one... @Firstyear

The auth process is challenge response - you start an authentication session, and the server says "here are the possible auth mechanisms". You select a mech, and then the server will repeatedly prompt with a set of possible "challenges" that can proceed at this point. The user selects the type of challenge, and then provides the details to fufil it. the server then either accepts that challenge step, or rejects the auth. When accepted the server will prompt for further challenge steps, until the authentication suceeds.

So here, you'll need to look at:

https://github.com/kanidm/kanidm/blob/master/kanidm_tools/src/cli/session.rs#L231

In this loop we prompt the user: ( https://github.com/kanidm/kanidm/blob/master/kanidm_tools/src/cli/session.rs#L249 ) to choose what they want to use next, then we use that to get the next factor ( https://github.com/kanidm/kanidm/blob/master/kanidm_tools/src/cli/session.rs#L264 )  

So in this case, you'll need to expand AuthAllowed to have a "BackupCode" option, and then create a handler that can collect that code (similar to do_password but likely without hiding the input). 

> 
> > 1. (Re)Naming
> > 
> > 
> > * to fit to the other entries in `SetCredentialRequest` enum
> > * https://github.com/kanidm/kanidm/blob/master/kanidm_proto/src/v1.rs#L684
> > * how about renaming `GenerateBackupCode` --> `BackupCodeGenerate`?
> > * could also be a separate ticket, but might also be done here.
> > 
> > What are your thoughts on these?
> 
> Agreed. I wrote that following `GeneratePassword` 😆 (https://github.com/kanidm/kanidm/blob/master/kanidm_proto/src/v1.rs#L673).
> I think we should change both to `BackupCodeGenerate` and `PasswordGenerate`.

I agreed, but this is a seperate issue IMO. Good spot @cuberoot74088

cuberoot74088:
Thanks for the feedback.
I have tried to get the potentially simplier task first - the output rendering. Well, this was not as easy as expected ... :disappointed: 

I started to `impl fmt::Display for BackupCodes` in kanidmd/src/lib/credential/mod.rs 
[here](https://github.com/kanidm/kanidm/pull/517/commits/7e845dc5fe2fb1df836c6186290ab7c3274ce05f#diff-1cf0fc6ce2e891dc21490977c4672519bcc3f38292de0229de604f6da8e8e5a0R222) Would that be the way to go?

I was not able to figure out how to use it in kanidm_tools/src/cli/account.rs for rendering, if this would be the way.
Therefore, there is now a [hacky manual formatting](https://github.com/kanidm/kanidm/pull/517/commits/7e845dc5fe2fb1df836c6186290ab7c3274ce05f#diff-4ea4d56b07466d12e7ad39e08667124dd3041e68b042d4ef38b5d68f8140198cR294) ... 

Would be great, if you could give me a hint how to do it properly.

There is also a `BackupCodesView` https://github.com/kanidm/kanidm/blob/master/kanidm_proto/src/v1.rs#L388 which also confused me. Moreover, there we have a type of `backup_codes: Vec<String>` and for `BackupCodes` in https://github.com/kanidm/kanidm/blob/master/kanidmd/src/lib/credential/mod.rs#L217 we have `HashSet<String>`. Why not `Vec<String>` for both?

cuberoot74088:
In 68d5efa625d49a707686e9eb542c836019665ff0 comes a lot of stuff for using the `BackupCode` for an auth.

I still cannot send them over via cli or web, because I am not asked for it :wink:. I cannot opt-in let's say. 
Seems like I missed somewhere, that the server sends the BackupCode as a `AuthMech`?

### auth flow
I think the auth flow needs to be changed a bit. For my understanding, the backup code "replaces" on of the second-factors (totp, webauthn). Currently we do TOTP first (maybe even webauthn, have not checked) and if you fail here, the auth process is stopped.

Maybe the auth-mech order should be (given you have MFA enabled):
`Password --[valid]--> mfa --[fail or skip]--> backup code --[valid]--> DONE.`

Or, if we keep it like that, have the chance after failing/skipping the totp/webauthn, to go for backup codes?

Firstyear:
@cuberoot74088 - The missing part looks like AuthAllowed in kanidm_proto/src/v1.rs - it needs a BackupCode option, which needs to be sent when the codes are present, and that's NOT currently generated by the server: src/lib/idm/authsession.rs line 497 (next_auth_allowed).

Once that's there, the rest will "fall into place". Normally when I do these features I do them "end to end" including the apis with tests, but in this case I asked @victorcwai to split it to two parts, and I missed it in review that the AuthAllowed enum wasn't extended.

So I think that's what is missing here :) 

Remember, the client doesn't drive the challenge response logic, the server does, so that's where the fix is here.

victorcwai:
> The missing part looks like AuthAllowed in kanidm_proto/src/v1.rs - it needs a BackupCode option, which needs to be sent when the codes are present, and that's NOT currently generated by the server: src/lib/idm/authsession.rs line 497 (next_auth_allowed).

Thanks for pointing that out, totally missed that part.

> There is also a `BackupCodesView` https://github.com/kanidm/kanidm/blob/master/kanidm_proto/src/v1.rs#L388 which also confused me. Moreover, there we have a type of `backup_codes: Vec<String>` and for `BackupCodes` in https://github.com/kanidm/kanidm/blob/master/kanidmd/src/lib/credential/mod.rs#L217 we have `HashSet<String>`. Why not `Vec<String>` for both?

The `BackupCodesView` is used for GET request (just for seeing the codes). 
I separate `BackupCodesView` and `BackupCodes` because I am not sure whether we will add more fields (that shouldn't be shown to users) to `BackupCodes`.

I changed to Vec as per the discussion here: https://github.com/kanidm/kanidm/pull/469#discussion_r646298349. Looking back at it, I think `generate_backup_code` could also return the backup codes as HashSet (to avoid confusion and unnecessarily transforming it into Vec).

Firstyear:
> > The missing part looks like AuthAllowed in kanidm_proto/src/v1.rs - it needs a BackupCode option, which needs to be sent when the codes are present, and that's NOT currently generated by the server: src/lib/idm/authsession.rs line 497 (next_auth_allowed).
> 
> Thanks for pointing that out, totally missed that part.
> 
> > There is also a `BackupCodesView` https://github.com/kanidm/kanidm/blob/master/kanidm_proto/src/v1.rs#L388 which also confused me. Moreover, there we have a type of `backup_codes: Vec<String>` and for `BackupCodes` in https://github.com/kanidm/kanidm/blob/master/kanidmd/src/lib/credential/mod.rs#L217 we have `HashSet<String>`. Why not `Vec<String>` for both?
> 
> The `BackupCodesView` is used for GET request (just for seeing the codes).
> I separate `BackupCodesView` and `BackupCodes` because I am not sure whether we will add more fields (that shouldn't be shown to users) to `BackupCodes`.

They need to be a hashset so we can look up in the BackupCodes during an auth if the code is present.

And @victorcwai is correct to seperate these as there is a difference between our protocol binds and our internal structures, so it does give us options to alter/change these. IE we can change the internal representation without breaking clients in the future. 

> 
> I changed to Vec as per the discussion here: [#469 (comment)](https://github.com/kanidm/kanidm/pull/469#discussion_r646298349). Looking back at it, I think `generate_backup_code` could also return the backup codes as HashSet (to avoid confusion and unnecessarily transforming it into Vec).

Sets have a different behaviour in json iirc, but I'm okay with it being a hashset in the proto for json. Im just thinking what is the "benefit" I guess? But I don't see drawbacks etiher .... 🤔

cuberoot74088:
Hi @Firstyear,

> @cuberoot74088 - The missing part looks like AuthAllowed in kanidm_proto/src/v1.rs - it needs a BackupCode option, which needs to be sent when the codes are present, and that's NOT currently generated by the server: src/lib/idm/authsession.rs line 497 (next_auth_allowed).
> 
> So I think that's what is missing here :)

Yes, this was the missing piece. Thanks for the hint! With 487e1c4 you can use backup codes from cli! :partying_face: 

### missing AuthMeth case selection on web-ui
I have to say, that the web-ui is now broken, since there are 2 options provided and this is not implemented. 
https://github.com/kanidm/kanidm/blob/master/kanidmd_web_ui/src/login.rs#L422

`next -> AuthResponse { sessionid: 00000000-60e3-fd34-0f5f-6b91ab227020, state: Continue([BackupCode, Totp]) }`

Firstyear:
That's fine - leave the webui broken for now. It's just intended as a proof-of-concept that something can exist. I'm currently working on a major refactor of this, so I'll resolve it there.

kanidm_client/src/lib.rs
```diff
@@ -701,6 +705,26 @@ impl KanidmClient {
         )
     }
 
+    pub fn idm_account_primary_credential_generate_backup_code(
+        &self,
+        id: &str,
+    ) -> Result<Vec<String>, ClientError> {
+        tokio_block_on(
+            self.asclient
+                .idm_account_primary_credential_generate_backup_code(id),
```
Firstyear, line 714: no need for the trailing ',' here

kanidm_tools/src/cli/account.rs
```diff
@@ -281,6 +283,35 @@ impl AccountOpt {
                         }
                     }
                 }
+                AccountCredential::GenerateBackupCode(acsopt) => {
+                    let client = acsopt.copt.to_client();
+                    match client.idm_account_primary_credential_generate_backup_code(
+                        acsopt.aopts.account_id.as_str(),
+                    ) {
+                        Ok(s) => {
+                            println!("GenerateBackupCode success.");
```
Firstyear, line 292: Probably need some instructions that these codes should be saved for future use.

kanidm_tools/src/cli/session.rs
```diff
@@ -142,6 +142,18 @@ impl LoginOpt {
...
 
+    fn do_backup_code(&self, client: &mut KanidmClient) -> Result<AuthResponse, ClientError> {
+        print!("Enter Backup Code: ");
+        // We flush stdout so it'll write the buffer to screen, continuing operation. Without it, the application halts.
+        io::stdout().flush().unwrap();
+        let mut backup_code = String::new();
+        if let Err(e) = io::stdin().read_line(&mut backup_code) {
+            eprintln!("Failed to read from stdin -> {:?}", e);
+            return Err(ClientError::SystemError);
+        };
```
Firstyear, line 153: Is it worth a check that backup_code.trim().len() > 0, so that we don't post empty strings where someone just accidentally hit return?

kanidmd/src/lib/credential/mod.rs
```diff
@@ -218,6 +219,17 @@ pub struct BackupCodes {
     code_set: HashSet<String>,
 }
 
+impl fmt::Display for BackupCodes {
```
Firstyear, line 222: Where is this display used? Do we need the server to be able to display/log these? I think that could be a security issue if we log someones backup codes ....

kanidmd_web_ui/src/login.rs
```diff
@@ -39,6 +39,7 @@ enum LoginState {
     // MechChoice
     // CredChoice
     Password(bool),
+    BackupCode(bool),
```
Firstyear, line 42: As much as I appreciate you adding this, I think you said there were some issues with the webui, so do you want to revert the changes in this file, and I'll resolve them in my current work? The webui atm is only a proof of concept of login, so it's not going to be breaking for most people.

Firstyear (changes requested):

kanidmd_web_ui/src/login.rs
cuberoot74088: Yes, the webui will fail if you have more than one AuthMech to choose from.

I tried to revert my changes, but cannot leave it like it is in master, as I get this:

> login.rs:406:35
>     |
> 406 | ...                   match allowed.pop().unwrap() {
>     |                             ^^^^^^^^^^^^^^^^^^^^^^ pattern `BackupCode` not covered
> 
That's why I started to modify the webui initially. 

If you are okay with it, I would say we leave it like this now. The project will compile and you can use the webui with  accounts that present only one AuthMech.
Then, you can work on the improved webui and just throw it away - no problem from my side.

kanidm_client/src/lib.rs
cuberoot74088: I cannot remove it - my editor puts it back :laughing:

kanidmd/src/lib/credential/mod.rs
cuberoot74088: Not used, this was a wrong idea. I will remove it.

cuberoot74088 (commented):
I hope I covered everthing.

Firstyear (approved):

Firstyear:
@victorcwai Leaving it to you to double check you are happy with this :)

victorcwai (approved):
Approved, thanks for the PR :)

PR #525, opened 2021-07-11
Prompting user for their username when multiple tokens exist

yaleman:
This prompts the user with an ncurses-based menu when they've got multiple tokens and they didn't specify `--name <username>`. It compiles on Ubuntu/macOS/openSUSE tumbleweed with no additional packages needed.

Note: when`--name <username>` *is* used, the current functionality doesn't change. 😄 

Screen Shot 2021-07-11 at 10 10 35

- [x] cargo fmt has been run

### DIFF
--- a/kanidm_client/src/lib.rs
+++ b/kanidm_client/src/lib.rs
@@ -65,8 +65,6 @@ struct KanidmClientConfig {
     verify_ca: Option<bool>,
     verify_hostnames: Option<bool>,
     ca_path: Option<String>,
-    // prompt users for tokens instead of just bailing
-    prompt_user_token: Option<bool>,
     // Should we add username/pw later? They could be part of the builder
     // process ...
 }
@@ -79,7 +77,6 @@ pub struct KanidmClientBuilder {
     ca: Option<reqwest::Certificate>,
     connect_timeout: Option<u64>,
     use_system_proxies: bool,
-    prompt_user_token: bool,
 }
 
 fn read_file_metadata<P: AsRef<Path>>(path: &P) -> Result<Metadata, ()> {
@@ -101,8 +98,6 @@ impl KanidmClientBuilder {
             ca: None,
             connect_timeout: None,
             use_system_proxies: true,
-            // default to prompting users with a menu if they have multiple tokens
-            prompt_user_token: true,
         }
     }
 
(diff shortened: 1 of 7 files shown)


kanidm_tools/src/cli/common.rs
```diff
@@ -76,9 +78,29 @@ impl CommonOpt {
...
+                            .items(&options)
+                            .interact()
+                            .unwrap();
+                        debug!("Index of the chosen menu item: {:?}", selection);
+
+                        let (f_uname, f_token) =
+                            tokens.iter().nth(selection).expect("Memory Corruption");
+                        info!("Using cached token for name {}", f_uname);
+                        f_token.clone()
+                        //String::from("lol")
```
Firstyear, line 102: lol.

kanidm_client/src/lib.rs
```diff
@@ -65,6 +65,8 @@ struct KanidmClientConfig {
     verify_ca: Option<bool>,
     verify_hostnames: Option<bool>,
     ca_path: Option<String>,
+    // prompt users for tokens instead of just bailing
+    prompt_user_token: Option<bool>,
```
Firstyear, line 69: Is there really a reason to make this configurable? To login you have to do an "interactive" process anyway from the cli tool. It also seems like a "layer" violation (config in the client, but it's used by the cli tool).

I think I'd rather remove the prompt_user_token bool, and just have the menu selector instead.

Firstyear (changes requested):
Looks good, really want to see this lib used in session.rs as a nicer alternate to the current index choice menu.

kanidm_client/src/lib.rs
yaleman: I figured some beardy types wouldn't want to be prompted like that, so made it an optional thing. Is there a cli tool config file? I thought the client config was the place for it 😄 

Happy to take that config part out if you'd prefer 😄

kanidm_tools/src/cli/common.rs
yaleman: lolwhups.

kanidm_client/src/lib.rs
Firstyear: I would rather remove this because it should either be a cli switch during login to remove the interactivity, but also, I don't want this to "become a scripting" tool. That's what the API should be for, and having people add tools like this to scripts just causes people to complain if we make future usability changes/improvements. I don't really want to lock us into that, so for now, I think remove the option because it sohuld be "interactive only". :)

yaleman: Sold, will remove it this evening :)

PR #526, opened 2021-07-14
Fix docu for generate password

cuberoot74088:
This updates the docs to reflect the latest kanidm-tools cli commands.

* old
`kanidm account credential generate_password [OPTIONS] <account-id>`

* new 
`kanidm account credential reset_credential [OPTIONS] <account-id>
`


Firstyear:
Thanks! Good find :)

PR #532, opened 2021-07-22
Update webauthn-rs to alpha.9

Firstyear:
Update to webauthn-rs alpha.9 which improves usage of references in the webauthn calls.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -84,9 +84,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.41"
+version = "1.0.42"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "15af2628f6890fe2609a3b91bef4c83450512802e59489f9c1cb1fa5df064a61"
+checksum = "595d3cfa7a60d4555cb5067b99f07142a08ea778de5cf993f7b75c7d8fabc486"
 
 [[package]]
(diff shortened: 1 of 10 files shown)


yaleman:
Looks like the usual level of goodness... two questions

- does it mean existing clients need to re-register?
- does this enforce the user verification policy?
- why can't I count?

Firstyear:
> Looks like the usual level of goodness... two questions
> 
> * does it mean existing clients need to re-register?
> * does this enforce the user verification policy?
> * why can't I count?

Four answers:

* Existing clients do NOT need to re-register. We only did discouraged before this, so all credentials will stay as discouraged.
* For more about the enforcing you need to check https://github.com/kanidm/webauthn-rs/blob/master/designs/authentication-use-cases.md but generally the answer is "yes" and stricter than the actual webauthn standard.
* Why *can't* you count indeed?
* Why can't I count?

yaleman (approved):

yaleman:
> * For more about the enforcing you need to check https://github.com/kanidm/webauthn-rs/blob/master/designs/authentication-use-cases.md but generally the answer is "yes" and stricter than the actual webauthn standard.

Yep, I figured it was - based on your previous talks 😄  Thanks.

> * Why _can't_ you count indeed?

I'm just drawn that way.

> * Why can't I count?

You can, as proved above. 😆

PR #536, opened 2021-07-23
Implement Online Backups (#25)

### LINKED ISSUE #25: Automated/Scheduled backups
It would be beneficial to deployments to have an automated/scheduled backup system in the server, to avoid the need to trigger a db backup before you perform a filesystem backup.

cuberoot74088:
This adds automatic online backups to a running kanidmd server.

Fixes #25 

- [X] cargo fmt has been run

### DIFF
--- a/examples/server.toml
+++ b/examples/server.toml
@@ -40,9 +40,11 @@ origin = "https://idm.example.com:8443"
 # [online_backup]
 #   The path to the output folder for online backups
 # path = "/var/lib/kanidm/backups/"
-#   How often the online_backup should run
-#   every day at 22:00 - see https://crontab.guru/ 
-# schedule = "0 22 * * *" 
+#   How often the online_backup should run - see https://crontab.guru/
+#   every day at 22:00
+# schedule = "00 22 * * *"
+#    four times a day at 3 minutes past the hour, every 6th hours
+# schedule = "03 */6 * * *"
 #   Number of backups to keep
 # versions = 3
 #
--- a/examples/server.toml
+++ b/examples/server.toml
@@ -36,17 +36,16 @@ db_path = "/var/lib/kanidm/kanidm.db"
 # origin = "https://idm.example.com"
 origin = "https://idm.example.com:8443"
 #
-#   TODO
-#   The path to the output folder for online backups
-#   Uncommenting this will enable the online backup.
-# online_backup_path = "/var/lib/kanidm/"
-online_backup_path = "/tmp/kanidm/"
 #
+# [online_backup]
+#   The path to the output folder for online backups
+# path = "/var/lib/kanidm/backups/"
 #   How often the online_backup should run
-#   might be switched to something cron like
-online_backup_interval = 5
+#   every day at 22:00 - see https://crontab.guru/ 
+# schedule = "0 22 * * *" 
 #   Number of backups to keep
-# online_backup_versions = 10
+# versions = 3
+#
 #
 #   The role of this server. This affects features available and how replication may interact.
 #   Valid roles are:
--- a/examples/server.toml
+++ b/examples/server.toml
@@ -51,13 +51,13 @@ origin = "https://idm.example.com:8443"
 #
 #   The role of this server. This affects features available and how replication may interact.
 #   Valid roles are:
-#   - write_replica
+#   - WriteReplica
 #     This server provides all functionality of Kanidm. It allows authentication, writes, and
 #     the web user interface to be served.
-#   - write_replica_no_ui
+#   - WriteReplicaNoUI
 #     This server is the same as a write_replica, but does NOT offer the web user interface.
-#   - read_only_replica
+#   - ReadOnlyReplica
 #     This server will not writes initiated by clients. It supports authentication and reads,
 #     and must have a replication agreement as a source of it's data.
-#   Defaults to "write_replica".
-# role = "write_replica"
+#   Defaults to "WriteReplica".
+# role = "WriteReplica"
--- a/examples/server.toml
+++ b/examples/server.toml
@@ -36,6 +36,18 @@ db_path = "/var/lib/kanidm/kanidm.db"
 # origin = "https://idm.example.com"
 origin = "https://idm.example.com:8443"
 #
+#   TODO
+#   The path to the output folder for online backups
+#   Uncommenting this will enable the online backup.
+# online_backup_path = "/var/lib/kanidm/"
+online_backup_path = "/tmp/kanidm/"
+#
+#   How often the online_backup should run
+#   might be switched to something cron like
+online_backup_interval = 5
+#   Number of backups to keep
+# online_backup_versions = 10
+#
 #   The role of this server. This affects features available and how replication may interact.
 #   Valid roles are:
 #   - write_replica
--- a/examples/server.toml
+++ b/examples/server.toml
@@ -41,12 +41,12 @@ origin = "https://idm.example.com:8443"
 #   The path to the output folder for online backups
 # path = "/var/lib/kanidm/backups/"
 #   How often the online_backup should run - see https://crontab.guru/
-#   every day at 22:00
+#   every day at 22:00 UTC (default)
 # schedule = "00 22 * * *"
 #    four times a day at 3 minutes past the hour, every 6th hours
 # schedule = "03 */6 * * *"
 #   Number of backups to keep
-# versions = 3
+# versions = 7
 #
 #
 #   The role of this server. This affects features available and how replication may interact.
--- a/examples/server.toml
+++ b/examples/server.toml
@@ -40,12 +40,12 @@ origin = "https://idm.example.com:8443"
 # [online_backup]
 #   The path to the output folder for online backups
 # path = "/var/lib/kanidm/backups/"
-#   How often the online_backup should run - see https://crontab.guru/
+#   The schedule to run online backups - see https://crontab.guru/
 #   every day at 22:00 UTC (default)
 # schedule = "00 22 * * *"
 #    four times a day at 3 minutes past the hour, every 6th hours
 # schedule = "03 */6 * * *"
-#   Number of backups to keep
+#   Number of backups to keep (default 7)
 # versions = 7
 #
 #
--- a/kanidmd/src/lib/actors/v1_read.rs
+++ b/kanidmd/src/lib/actors/v1_read.rs
@@ -1,5 +1,6 @@
 use tokio::sync::mpsc::UnboundedSender as Sender;
 
+use chrono::{DateTime, SecondsFormat, Utc};
 use std::sync::Arc;
 
 use crate::prelude::*;
@@ -177,7 +178,9 @@ impl QueryServerReadV1 {
         ltrace!(audit, "Begin online backup event {:?}", msg.eventid);
 
         // TODO: might be something like "backup-timestamp.json"
-        let dest_file = format!("{}/{}", outpath, "backup.json");
+        let now: DateTime<Utc> = Utc::now();
+        let timestamp = now.to_rfc3339_opts(SecondsFormat::Secs, true);
+        let dest_file = format!("{}/backup-{}.json", outpath, timestamp);
 
         // TODO: handle the max versions to keep here?
         let idms_prox_read = self.idms.proxy_read_async().await;
--- a/kanidmd/src/lib/actors/v1_read.rs
+++ b/kanidmd/src/lib/actors/v1_read.rs
@@ -4,7 +4,11 @@ use std::sync::Arc;
 
 use crate::prelude::*;
 
-use crate::event::{AuthEvent, AuthResult, SearchEvent, SearchResult, WhoamiResult};
+use crate::be::BackendTransaction;
+
+use crate::event::{
+    AuthEvent, AuthResult, LiveBackupEvent, SearchEvent, SearchResult, WhoamiResult,
+};
 use crate::idm::event::{
     CredentialStatusEvent, RadiusAuthTokenEvent, ReadBackupCodeEvent, UnixGroupTokenEvent,
     UnixUserAuthEvent, UnixUserTokenEvent,
@@ -165,6 +169,32 @@ impl QueryServerReadV1 {
             OperationError::InvalidState
         })?;
         res
+    }
+
+    pub async fn handle_live_backup(&self, msg: LiveBackupEvent) {
+        let mut audit = AuditScope::new("live backup", msg.eventid, self.log_level);
+
+        ltrace!(audit, "Begin live-backup event {:?}", msg.eventid);
+
+        let idms_prox_read = self.idms.proxy_read_async().await;
+        lperf_op_segment!(
+            &mut audit,
+            "actors::v1_read::handle<LiveBackupEvent>",
+            || {
+                let res = idms_prox_read
+                    .qs_read
+                    .get_be_txn()
+                    .backup(&mut audit, "/tmp/backup.json");
+
+                ladmin_info!(audit, "live backup result: {:?}", res);
+                #[allow(clippy::expect_used)]
+                res.expect("Live backup failed");
+            }
+        );
+        // At the end of the event we send it for logging.
+        self.log.send(audit).unwrap_or_else(|_| {
+            error!("CRITICAL: UNABLE TO COMMIT LOGS");
+        });
     }
 
     pub async fn handle_whoami(
--- a/kanidmd/src/lib/actors/v1_read.rs
+++ b/kanidmd/src/lib/actors/v1_read.rs
@@ -8,7 +8,7 @@ use crate::prelude::*;
 use crate::be::BackendTransaction;
 
 use crate::event::{
-    AuthEvent, AuthResult, LiveBackupEvent, SearchEvent, SearchResult, WhoamiResult,
+    AuthEvent, AuthResult, OnlineBackupEvent, SearchEvent, SearchResult, WhoamiResult,
 };
 use crate::idm::event::{
     CredentialStatusEvent, RadiusAuthTokenEvent, ReadBackupCodeEvent, UnixGroupTokenEvent,
@@ -172,17 +172,18 @@ impl QueryServerReadV1 {
         res
     }
 
-    pub async fn handle_online_backup(&self, msg: LiveBackupEvent, outpath: &str) {
+    pub async fn handle_online_backup(&self, msg: OnlineBackupEvent, outpath: &str, versions: u64) {
         let mut audit = AuditScope::new("online backup", msg.eventid, self.log_level);
 
         ltrace!(audit, "Begin online backup event {:?}", msg.eventid);
 
-        // TODO: might be something like "backup-timestamp.json"
         let now: DateTime<Utc> = Utc::now();
         let timestamp = now.to_rfc3339_opts(SecondsFormat::Secs, true);
+        // TODO: check outpath exists
         let dest_file = format!("{}/backup-{}.json", outpath, timestamp);
 
         // TODO: handle the max versions to keep here?
+        info!("online backup versions to keep = {}", versions);
         let idms_prox_read = self.idms.proxy_read_async().await;
         lperf_op_segment!(
             &mut audit,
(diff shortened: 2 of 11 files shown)


yaleman:
The interval/activity definitely needs to be configurable -  in clusters you don't want all nodes backing up every five seconds and some environments wouldn't be able to back up in that time either. 😄 

I'd probably call it a `BackupLive`, going by the AuthResult/SearchEvent nomenclature - `<Activity><SubType>`

kanidmd/src/lib/interval.rs
```diff
@@ -24,4 +26,14 @@ impl IntervalActor {
             }
         });
     }
+
+    pub fn start_live_backup(server: &'static QueryServerReadV1) {
```
Firstyear, line 30: I think it should be online backup not live backup

kanidmd/src/lib/interval.rs
```diff
@@ -24,4 +26,14 @@ impl IntervalActor {
...
+        tokio::spawn(async move {
+            let mut inter = interval(Duration::from_secs(5));
```
Firstyear, line 32: Rather than a duration, this should be triggered at a configured time in the server.toml I think. IE 10pm daily or hourly or similar.

kanidmd/src/lib/actors/v1_read.rs
```diff
@@ -167,6 +171,32 @@ impl QueryServerReadV1 {
...
+
+        let idms_prox_read = self.idms.proxy_read_async().await;
+        lperf_op_segment!(
+            &mut audit,
+            "actors::v1_read::handle<LiveBackupEvent>",
+            || {
+                let res = idms_prox_read
+                    .qs_read
+                    .get_be_txn()
+                    .backup(&mut audit, "/tmp/backup.json");
```
Firstyear, line 187: This destination needs to be configurable as well as /tmp is not considered a secure location. As well we should consider outputing these backups with a time stamp and rotation policy?

Firstyear (commented):
It's a good start mate! I think some more configuration elements will be needed though :)

kanidmd/src/lib/interval.rs
yaleman: cron calendar support, perhaps? :) [saffron](https://docs.rs/saffron/0.1.0/saffron/parse/index.html) is what CloudFlare made to parse things.

cuberoot74088: I was also thinking of a cron like thing, but then thought it might be too much, too complicated?

Thanks for the hint, will look into it.

It probably also adds a lot of flexibility.

How should it be? We could also start simple and extend. 😉

kanidmd/src/lib/actors/v1_read.rs
cuberoot74088: Sure. Should I put the config into server.toml then?

Was also thinking of the Rotation... Any good ideas how to do it? Keep last 3 versions?

Somehow like backup-timestamp.json?

cuberoot74088:
Note: This is still pretty much work in progress, far from ready.

I have started to put some config options in place. I would like to have your thoughts on that and some guidance how to do it better ;-) 
Where to parse and put the config together and how to pass it on to the subsequent functions. 

How about the keep versions setting? I think it could be integrated into the creation of the backup?

What do you think about the cron approach? Should we give saffron a try?

examples/server.toml
```diff
@@ -36,6 +36,18 @@ db_path = "/var/lib/kanidm/kanidm.db"
 # origin = "https://idm.example.com"
 origin = "https://idm.example.com:8443"
 #
+#   TODO
+#   The path to the output folder for online backups
+#   Uncommenting this will enable the online backup.
+# online_backup_path = "/var/lib/kanidm/"
+online_backup_path = "/tmp/kanidm/"
```
Firstyear, line 43: Should probably default to /data/backups/ for containers

examples/server.toml
```diff
@@ -36,6 +36,18 @@ db_path = "/var/lib/kanidm/kanidm.db"
 # origin = "https://idm.example.com"
 origin = "https://idm.example.com:8443"
 #
+#   TODO
+#   The path to the output folder for online backups
+#   Uncommenting this will enable the online backup.
+# online_backup_path = "/var/lib/kanidm/"
```
Firstyear, line 42: And this should be /var/lib/kanidm/backups/

kanidmd/src/lib/actors/v1_read.rs
```diff
@@ -167,6 +172,38 @@ impl QueryServerReadV1 {
         res
     }
 
+    pub async fn handle_online_backup(&self, msg: LiveBackupEvent, outpath: &str) {
+        let mut audit = AuditScope::new("online backup", msg.eventid, self.log_level);
+
+        ltrace!(audit, "Begin online backup event {:?}", msg.eventid);
+
+        // TODO: might be something like "backup-timestamp.json"
```
Firstyear, line 180: This todo comment is done :)

kanidmd/src/lib/actors/v1_read.rs
```diff
@@ -167,6 +172,38 @@ impl QueryServerReadV1 {
...
+        let now: DateTime<Utc> = Utc::now();
+        let timestamp = now.to_rfc3339_opts(SecondsFormat::Secs, true);
+        let dest_file = format!("{}/backup-{}.json", outpath, timestamp);
```
Firstyear, line 183: Need to ensure that outpath exists and is a directory before this too. Should also check that dest_file does NOT exist already.

kanidmd/src/lib/config.rs
```diff
@@ -71,6 +71,11 @@ pub struct Configuration {
     pub cookie_key: [u8; 32],
     pub integration_test_config: Option<Box<IntegrationTestConfig>>,
     pub log_level: Option<u32>,
+    // TODO: online backup settings
```
Firstyear, line 74: You could make this a section in the config by adding a struct called OnlineBackup and that will allow the config tou have a section like 

```
[online_backup]
path = 
interval = 
versions =
```

I think something to keep in mind is that if you make it an Option<OnlineBackup> then this section is optional and it's abscence means "don't run automatic backups".

kanidmd/src/lib/core/mod.rs
```diff
@@ -702,6 +702,23 @@ pub async fn create_server_core(config: Configuration) -> Result<(), ()> {
 
     // Setup timed events associated to the write thread
     IntervalActor::start(server_write_ref);
+    // Setup timed events associated to the read thread
+    // TODO: only start this if enable in config.
+    // TODO: maybe we should have better option processing here
+    // and pass it on to `start_online_backup()`
+    match &config.online_backup_path {
+        Some(p) => {
+            // TODO check path exists?
```
Firstyear, line 711: Should check in the event, not here :)

kanidmd/src/lib/event.rs
```diff
@@ -910,6 +910,27 @@ impl PurgeRecycledEvent {
     }
 }
 
+#[derive(Debug)]
+pub struct LiveBackupEvent {
```
Firstyear, line 914: OnlineBackup I think is a better name

kanidmd/src/lib/actors/v1_read.rs
```diff
@@ -167,6 +172,38 @@ impl QueryServerReadV1 {
...
+                    .qs_read
+                    .get_be_txn()
+                    .backup(&mut audit, &dest_file);
+
+                ladmin_info!(audit, "online backup result: {:?}", res);
+                #[allow(clippy::expect_used)]
+                res.expect("Online backup failed");
+            }
+        );
+        // At the end of the event we send it for logging.
```
Firstyear, line 201: I'm thinking we could do the cleanup here but I'm not sure what's the safest way that "won't" accidentally remove something the admin wants to keep.

examples/server.toml
```diff
@@ -36,6 +36,18 @@ db_path = "/var/lib/kanidm/kanidm.db"
...
+online_backup_path = "/tmp/kanidm/"
+#
+#   How often the online_backup should run
+#   might be switched to something cron like
+online_backup_interval = 5
```
Firstyear, line 47: Yes I think saffron looks okay. https://docs.rs/saffron/0.1.0/saffron/struct.Cron.html#method.next_after 

You can use this to feed in the current time, and it will tell you the next time the event will run. Then from that you can use https://docs.rs/tokio/1.9.0/tokio/time/fn.sleep_until.html inside of the IntervalActor so that between each operation you use next_after from saffron to work out the delay until the next event.

Firstyear (changes requested):

kanidmd/src/lib/actors/v1_read.rs
cuberoot74088: Yeah, checking the path is good. The function that writes the backup will overwrite existing files. I think this is okay, especially with timestamp, or?

kanidmd/src/lib/config.rs
cuberoot74088: Yes, a struct might be good. There are quite a few items related to the backup.

What would you like to have? Should it be optional? The main reason for me was th be able to switch it on and off. But it could als be a separate option in the new config section.

With the latter, everything could be required, but  will require action to upgrade config files. 

Another thought. How much do we want to be actively set in the config? I think there are also some good defaults that might be used auyomatically.

That said, i am also wondering if the online backup should be running by default or not.
Once a day might not be too bad.

Firstyear: I think it should be optional and not enabled by default. But I think we should show it in the example config in the book. 

But having it as a seperate section does allow it on/off by the presence /abscence of the section in the config :) 

And also remember, that within the struct for the Option<OnlineBackup> or whatever, OnlineBackup can also have values that ahve defaults within them so that just specifying "the section header" makes it present, then we get defaults for all the other values :) https://serde.rs/attr-default.html may help you because you can set the default values when not present. in this case. Does that help?

examples/server.toml
cuberoot74088: We are using saffron now. I did not manage to use `sleep_unitl`, so going with `sleep`. Maybe there is an easy way to switch, just let me know.

kanidmd/src/lib/core/mod.rs
cuberoot74088: Still todo, should we create the path if it is not there? I think we should try and fail if we cannot.

kanidmd/src/lib/config.rs
cuberoot74088: You meant I should add something like this `#[serde(default = "default_resource")]` to the OnlineBackup struct? Got some errors about (De)Serialisation (I think is would lead to implement FromString ToString?!) ... and I am not fully sure - would this be the place where we put default options for OnlineBackup?

With the current state, you will have to uncomment the entire online_backup section, but I think this is fair, as you might want to adjust everything there. We can also change that, but then I need some more hints ;-)

kanidmd/src/lib/actors/v1_read.rs
cuberoot74088: (wanted to post that here) Still todo, should we create the path if it is not there? I think we should try and fail if we cannot.

kanidmd/src/lib/actors/v1_read.rs
cuberoot74088: I also think this is a good place. Well, we would always keep the minimum versions requested. I was thinking of some sort of `ls backup*json`, sort it and remove what is above the limit. Other ideas? First research brought up `glob` is this an option or just do a `fs::read_dir()`?

Firstyear: I think it's better here to be very specific. Instead of a glob, we should list the dir and match a regex on the file names I think. 

IE if the admin moved backup.timestamp.json to backup.superimportant.json, the glob would still remove it. But a regex would match that it's no longer a timestamp so we don't remove it. 

Perhaps for the moment backup-rotation may be a second PR because it could be a fail bit of work on it's own. 

Another option is we write out a .backup.control.json file that contains metadata abotu the backups we have created ad what exists, and we can use that instead to know what files might exist that we need to remove/prune.

kanidmd/src/lib/actors/v1_read.rs
Firstyear: I think overwriting should fail, because every time stamp should be new, so we should never stomp something that exists. 

i think we should create the outhpath folder if it doesn't exist (and set it's permissions properly) and fail if doesn't exist or can't be created.

kanidmd/src/lib/config.rs
Firstyear: Yeah, but those `#[serde(default = "default_resource")]` are on the attributes. Something like:

```
fn default_to_keep() -> usize { 7 }

fn default_schedule() -> Cron { ... }

struct OnlineBackup {
    outpath = String,
    #[serde(default = "default_schedule")]
    schedule = Cron,
    #[serde(default = "default_to_keep")]
    number_to_keep = usize,
}
```

Does that example help?

kanidmd/src/lib/interval.rs
```diff
@@ -24,4 +29,51 @@ impl IntervalActor {
...
+            cron_expr.describe(English::default())
+        );
+
+        tokio::spawn(async move {
+            let ct = Utc::now();
+            let cron = Cron::new(cron_expr.clone());
+
+            let cron_iter = cron.clone().iter_after(ct);
+            for next_time in cron_iter {
+                // +1 to have even times, but we might anyway cut away the seconds from the timestamp.
```
Firstyear, line 61: What do you mean by "even times" here?

kanidmd/src/lib/interval.rs
```diff
@@ -24,4 +29,51 @@ impl IntervalActor {
...
+                let wait_seconds = 1 + (next_time - Utc::now()).num_seconds() as u64;
+                info!(
+                    "Online backup next run on {}, wait_time = {}s",
+                    next_time, wait_seconds
+                );
+
+                sleep(Duration::from_secs(wait_seconds)).await;
```
Firstyear, line 68: This basicly ends up doing the same as sleep_until I think anyway, so this is okay.

kanidmd/src/lib/interval.rs
```diff
@@ -24,4 +29,51 @@ impl IntervalActor {
...
+        let outpath = cfg.path.to_owned();
+        let schedule = cfg.schedule.to_owned();
+        let versions = cfg.versions;
+
+        // TODO: add some checks arount the provided cron pattern .any() etc.
+        let cron_expr = match schedule.as_str().parse::<CronExpr>() {
+            Ok(ce) => ce,
+            Err(e) => {
+                info!(
+                    "Online backup schedule parse error {}. Pattern set to < 00 22 * * * >",
```
Firstyear, line 43: If we fail to parse, we should return an error instead of proceeding with a default schedule.

kanidmd/src/lib/interval.rs
cuberoot74088: Assume you want to have the backup at 17:06(:00).

Without this I get

17:05:59Z

Instead of

17:06:00Z


Which will be displayed in the file name.

kanidmd/src/lib/interval.rs
cuberoot74088: Agreed. Someting like `eprintln`? I think stopping the entire server would be too much. On the other hand, we should make sure this error is configuration error recognized and fixed. Because there might be large delayes before it is run. How to do this best?

kanidmd/src/lib/config.rs
cuberoot74088: :facepalm:  Yes, thank you very much!

cuberoot74088:
In 7191380 I have implemented some very simple draft of backup version clean-up to maintain a maximum number of backups. As discussed, using a regexp to find exactly the automatically generated files.

How about the `remove_file()` here https://github.com/kanidm/kanidm/pull/536/commits/7191380619e0ab124064b9671da43825d624a161#diff-ee0f4dcb941062533c7d4a82cc13b6f76c40d2dc7df5f94868f264e7222b8f6eR249
should we do something with the `Result`?

kanidmd/src/lib/interval.rs
Firstyear: Ahhhhhhhhh yeah I see. Can you expand the comment with this example please :)

kanidmd/src/lib/config.rs
Firstyear: No need for "🤦 ", I'm happy to help :)

examples/server.toml
Firstyear: sleep works just as well here, so don't worry abotu this.

kanidmd/src/lib/interval.rs
Firstyear: If you change the `pub fn start_online_backup(...) -> Result<(), ()>`, then you can do:

```
let cron_expr = schedule.as_str().parse::<CronExpr>()
    .map_err(|e| {
        error!("Online backup schedule parse error {}", e);
    })?;
```

That will cause the function to return, and then in kanidmd/src/lib/core/mod.rs you can do somethig like:

```
start_online_backup(...)?;
```

When you call it, which will cause create server core to shortcut and retuirn without completing the server start up process.

kanidmd/src/lib/actors/v1_read.rs
cuberoot74088: This should be done now.

kanidmd/src/lib/actors/v1_read.rs
cuberoot74088: Should be done.
We are using the regex now and do a very simple cleanup of versions by max count

cuberoot74088:
My pleasure!

I tried to take care of rough edges as far as I could see. Let me know if you spot something.

 Maybe you have some time and give it a try, it should work quite good, I hope.

How about tests or docs on this?

kanidmd/src/lib/config.rs
```diff
@@ -14,6 +14,28 @@ pub struct IntegrationTestConfig {
...
+pub struct OnlineBackup {
+    #[serde(default = "default_online_backup_path")]
+    pub path: String,
+    #[serde(default = "default_online_backup_schedule")]
+    pub schedule: String,
+    #[serde(default = "default_online_backup_versions")]
+    pub versions: usize,
+}
+
+fn default_online_backup_path() -> String {
```
Firstyear, line 27: IMO there should be no default path here, we want the admin to specify it.

kanidmd/src/lib/interval.rs
```diff
@@ -24,4 +33,88 @@ impl IntervalActor {
...
+                )
+            })?;
+        }
+
+        if !op.is_dir() {
+            error!("Online backup output '{}' is not a directory or we are missing permissions to access it.", outpath);
+            return Err(());
+        }
+
+        // checking permissions (not sure about this)
```
Firstyear, line 86: I've just checked, the umask is set in the process startup so you don't need to worry about it.

kanidmd/src/lib/interval.rs
```diff
@@ -24,4 +33,88 @@ impl IntervalActor {
...
+        // TODO: we still might have a folder that we can read but not write in it.
+        let meta = op.metadata().unwrap();
+        if !file_permissions_readonly(&meta) {
```
Firstyear, line 89: This check isn't needed.

Firstyear (changes requested):
Beside a couple of tiny nit picks this looks really good.

For the docs, you'll probably want to add the new config parameters to installing_the_server.md in kanidm_book/src/ :)

Firstyear:
I think this looks good, and I'm willing to approve and merge. You'll need to do a git rebase master before that, or if you have issues with it, I can do the git rebase and merge process for you.

yaleman (approved):
This is great work

cuberoot74088:
Thank you very much @Firstyear and @yaleman!
I did the rebase and hope this works now.

BTW: 
I was running clippy and it mentioned, I might want to use `if let` in some places. 
And there are still some `unwrap()` calls (1,2,) - should I work on that as well? But they should be okay, as they are in the `Ok()` branch, right? 

Should I do something on these?

1. https://github.com/kanidm/kanidm/pull/536/files#diff-ee0f4dcb941062533c7d4a82cc13b6f76c40d2dc7df5f94868f264e7222b8f6eR236
2. https://github.com/kanidm/kanidm/pull/536/files#diff-ee0f4dcb941062533c7d4a82cc13b6f76c40d2dc7df5f94868f264e7222b8f6eR244

Firstyear:
I'm planning to do a clippy-fix-up tomorrow so feel free to leave these - I'll resolve them :)

Issue #468, opened 2021-06-05
Value should be rewritten


Firstyear:
In the past when I wrote the Value enum, I did not know how to effectively use Borrow for PartialEq within certain contexts. However, since https://github.com/kanidm/kanidm/commit/29566b8f997fffd135926be8747131b970a434aa there is a technique that I have learnt that can allow dynamic borrows into the same type.

This is why Value contains a PartialValue and then a "Data" type that's not truly associated.

We should rewrite this such that Value and PartialValue really are seperate types, and they can borrow into each other for eq/partialeq

Firstyear:
In addition rather than storing a set of Values' we should have a dedicated ValueSet type that has the type stored at a higher level rather than in every single Value to prevent duplication of the enum/types.

PR #538, opened 2021-07-24
468 valueset

Firstyear:
Relates #468 - this is stage one of converting to a proper ValueSet type inside of Entry. This replaces the direct usage of BTreeSet with ValueSet in entry, and asserts that all the current tests and behaviours pass. The intent is that by separating this with this layer, we now know what interfaces are required in ValueSet when we begin the conversation to a properly typed internal. 

### DIFF
--- a/kanidmd/src/lib/entry.rs
+++ b/kanidmd/src/lib/entry.rs
@@ -1765,24 +1765,20 @@ impl<VALID, STATE> Entry<VALID, STATE> {
     /// Assert if an attribute of this name is present, and one of it's values contains
     /// the following substring, if possible to perform the substring comparison.
     pub fn attribute_substring(&self, attr: &str, subvalue: &PartialValue) -> bool {
-        match self.attrs.get(attr) {
-            Some(v_list) => v_list
-                .iter()
-                .fold(false, |acc, v| if acc { acc } else { v.contains(subvalue) }),
-            None => false,
-        }
+        self.attrs
+            .get(attr)
+            .map(|vset| vset.substring(subvalue))
+            .unwrap_or(false)
     }
 
     #[inline(always)]
     /// Assert if an attribute of this name is present, and one of it's values is less than
     /// the following partial value
     pub fn attribute_lessthan(&self, attr: &str, subvalue: &PartialValue) -> bool {
-        match self.attrs.get(attr) {
-            Some(v_list) => v_list
-                .iter()
-                .fold(false, |acc, v| if acc { acc } else { v.lessthan(subvalue) }),
-            None => false,
-        }
+        self.attrs
+            .get(attr)
+            .map(|vset| vset.lessthan(subvalue))
+            .unwrap_or(false)
     }
 
     // Since EntryValid/Invalid is just about class adherenece, not Value correctness, we
--- a/kanidmd/src/lib/entry.rs
+++ b/kanidmd/src/lib/entry.rs
@@ -33,6 +33,7 @@ use crate::repl::cid::Cid;
 use crate::schema::{SchemaAttribute, SchemaClass, SchemaTransaction};
 use crate::value::{IndexType, SyntaxType};
 use crate::value::{PartialValue, Value};
+use crate::valueset::ValueSet;
 use kanidm_proto::v1::Entry as ProtoEntry;
 use kanidm_proto::v1::Filter as ProtoFilter;
 use kanidm_proto::v1::{OperationError, SchemaError};
@@ -151,7 +152,7 @@ pub struct EntryReduced {
     uuid: Uuid,
 }
 
-fn compare_attrs(left: &Map<AttrString, Set<Value>>, right: &Map<AttrString, Set<Value>>) -> bool {
+fn compare_attrs(left: &Map<AttrString, ValueSet>, right: &Map<AttrString, ValueSet>) -> bool {
     // We can't shortcut based on len because cid mod may not be present.
     // Build the set of all keys between both.
     let allkeys: Set<&str> = left
@@ -198,7 +199,7 @@ pub struct Entry<VALID, STATE> {
     valid: VALID,
     state: STATE,
     // We may need to change this to Set to allow borrow of Value -> PartialValue for lookups.
-    attrs: Map<AttrString, Set<Value>>,
+    attrs: Map<AttrString, ValueSet>,
 }
 
 impl<VALID, STATE> std::fmt::Debug for Entry<VALID, STATE>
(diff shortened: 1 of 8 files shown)


kanidmd/src/lib/entry.rs
```diff
@@ -1999,21 +2009,21 @@ impl<VALID, STATE> PartialEq for Entry<VALID, STATE> {
...
+        let name_v = valueset![Value::new_iutf8(s.name.as_str())];
+        let desc_v = valueset![Value::new_utf8(s.description.clone())];
 
-        let multivalue_v = btreeset![Value::from(s.multivalue)];
-        let unique_v = btreeset![Value::from(s.unique)];
+        let multivalue_v = valueset![Value::from(s.multivalue)];
+        let unique_v = valueset![Value::from(s.unique)];
 
-        let index_v: Set<_> = s.index.iter().map(|i| Value::from(i.clone())).collect();
+        let index_v: ValueSet = s.index.iter().map(|i| Value::from(i.clone())).collect();
```
QnnOkabayashi, line 2020:
```suggestion
        let index_v: ValueSet = s.index.iter().cloned().map(Value::from).collect();
```

kanidmd/src/lib/macros.rs
```diff
@@ -576,6 +577,33 @@ macro_rules! btreeset {
...
+        use crate::valueset::ValueSet;
+        let mut x: ValueSet = ValueSet::new();
+        assert!(x.insert($e));
+        x
+    });
+    ($e:expr,) => ({
+        use crate::valueset::ValueSet;
+        let mut x: ValueSet = ValueSet::new();
+        assert!(x.insert($e));
+        x
```
QnnOkabayashi, line 596: No need to repeat
```suggestion
        valueset!($e)
```

kanidmd/src/lib/entry.rs
```diff
@@ -1755,24 +1765,20 @@ impl<VALID, STATE> Entry<VALID, STATE> {
...
-        match self.attrs.get(attr) {
-            Some(v_list) => v_list
-                .iter()
-                .fold(false, |acc, v| if acc { acc } else { v.contains(subvalue) }),
-            None => false,
-        }
+        self.attrs
+            .get(attr)
+            .map(|vset| vset.substring(subvalue))
+            .unwrap_or(false)
```
QnnOkabayashi, line 1771: How come `self.attrs.get(attr)` used to return an iterable while it now returns a `PartialValue` (or `Value`, I can't infer here)? What were we iterating through before?

kanidmd/src/lib/valueset.rs
```diff
@@ -0,0 +1,210 @@
...
+    inner: BTreeSet<Value>,
+}
+
+impl Default for ValueSet {
+    fn default() -> Self {
+        ValueSet {
+            inner: BTreeSet::new(),
+        }
+    }
+}
```
QnnOkabayashi, line 17:
```suggestion
#[derive(Default)]
pub struct ValueSet {
    inner: BTreeSet<Value>,
}
```

kanidmd/src/lib/valueset.rs
```diff
@@ -0,0 +1,210 @@
...
+
+impl Clone for ValueSet {
+    fn clone(&self) -> Self {
+        ValueSet {
+            inner: self.inner.clone(),
+        }
+    }
+}
```
QnnOkabayashi, line 119: Is there a reason we didn't just derive `Clone`?

QnnOkabayashi (changes requested):

kanidmd/src/lib/entry.rs
```diff
@@ -1575,21 +1580,26 @@ impl<VALID, STATE> Entry<VALID, STATE> {
...
+            .or_insert_with(ValueSet::new);
         // Here we need to actually do a check/binary search ...
         v.insert(value);
         // Doesn't matter if it already exists, equality will replace.
     }
 
     /// Overwrite the current set of values for an attribute, with this new set.
-    pub fn set_ava_int(&mut self, attr: &str, values: Set<Value>) {
+    // pub fn set_ava_int(&mut self, attr: &str, values: Set<Value>) {
+    pub fn set_ava_int<T>(&mut self, attr: &str, iter: T)
```
victorcwai, line 1591: Why we pass iter here but not BTreeSet?

kanidmd/src/lib/valueset.rs
```diff
@@ -0,0 +1,210 @@
...
+        self.len() == 0
+    }
+
+    // We'll need to be able to do partialeq/intersect etc later.
+
+    pub fn iter(&self) -> Iter {
+        (&self).into_iter()
+    }
+
+    pub fn difference<'a>(&'a self, other: &'a ValueSet) -> Difference<'a> {
```
victorcwai, line 83: Why we need the lifetime annotations here (and in other functions)?

Firstyear: Because https://doc.rust-lang.org/std/collections/btree_set/struct.Difference.html needs a lifetime, and we need to then correctly set them all through the other elements so they are constrained correctly.

kanidmd/src/lib/entry.rs
Firstyear: Because in the future the idea is valueset will change from btreeset internally to a strongly typed interior with different possible representations.

kanidmd/src/lib/valueset.rs
Firstyear: Because when we go to change the interior of this we may want to change it. For now I'd rather the stub and we can derive later if we decide to .

kanidmd/src/lib/valueset.rs
Firstyear: See before - this is about stubs so that when we strongly type we can "fill in the gaps".

kanidmd/src/lib/entry.rs
Firstyear: Because the former code was really really old before I knew how to use map nicely. So the some(vlist) now changes to .map(|vset| and then we just call valueset substring's impl which has it's own internal iterator instead.

kanidmd/src/lib/macros.rs
Firstyear: I think there are issues with recursive definitions like this.

Firstyear:
Just as a note, remember the intent of this is that this adds an abstraction so we can see where valueset is used, and stage two will rewrite the inner from btreeset to be strongly type similar to the internal of Value.

victorcwai (approved):
Ok then!

kanidmd/src/lib/macros.rs
QnnOkabayashi: In general or in this specific scenario? Because I know they're used frequently like here https://docs.rs/tracing/0.1.26/src/tracing/macros.rs.html#586-792

QnnOkabayashi (approved):

kanidmd/src/lib/macros.rs
Firstyear: I think I've had issues with them before so maybe I should just try it :)

Firstyear: Well damn. It works. Doing this :)

PR #539, opened 2021-07-25
Improve cbor errors - fix accidental decode issue that prevents startup

Firstyear:
A change in webauthn broke startup but only for ecdsa keys which had a serialised field renamed. .10 adds an alias to allow successful deserialisation. 

- [ ] cargo fmt has been run


yaleman:
This works on my prod DB and fixes the problems I'm having with startup 😄

PR #545, opened 2021-07-31
Implement driftwood HTTP request logging

yaleman:
This is separate to the existing tracing work, adds an initial function to log HTTP requests to the server.

- [x] cargo fmt has been run

# Example output 

This is running locally with local queries (filtered for just the logs)... it'll log the remote IP address when queries happen.

Me running random cURL requests:
```
[::ffff:127.0.0.1]:52846 - - [31/Jul/2021:12:02:25 +1000] "GET /HELLO HTTP/1.1" 404 0 "" "curl/7.64.1"
```
Me `kanidm login`:
```
[::ffff:127.0.0.1]:52868 - - [31/Jul/2021:12:03:51 +1000] "POST /v1/auth HTTP/1.1" 200 87 "" ""
[::ffff:127.0.0.1]:52869 - - [31/Jul/2021:12:03:51 +1000] "POST /v1/auth HTTP/1.1" 200 410 "" ""
[::ffff:127.0.0.1]:52870 - - [31/Jul/2021:12:03:58 +1000] "POST /v1/auth HTTP/1.1" 200 86 "" ""
[::ffff:127.0.0.1]:52871 - - [31/Jul/2021:12:04:01 +1000] "POST /v1/auth HTTP/1.1" 200 715 "" ""
```

Future ideas I have include a proper UserAgent (with version) for `kanidmd` client requests and `userid`s being in these logs.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -645,6 +645,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "colored"
+version = "2.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b3616f750b84d8f0de8a58bda93e08e2a81ad3f523089b05f1dffecab48c6cbd"
+dependencies = [
+ "atty",
+ "lazy_static",
+ "winapi",
+]
+
+[[package]]
 name = "concread"
 version = "0.2.16"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


Firstyear:
i'm not sure about this. I like the intent, to help us improve our request logging. But we are also in the middle of @QnnOkabayashi adding their tracing work, and we could very likely emulate this in our own middleware along, which would give us the ability to emit this same data, but inside of structured/pretty logs rather than apache format. 

IMO I think we let @QnnOkabayashi get their tracing changes/middleware into the server, we do some work to refactor to that, and then we revisit this as a topic to improve logging of http requests and we can make a better choice as to if we use driftwood or something else. Is that reasonable?

yaleman:
Totes reasonable, this was a quick one to get it going with the view to expansion 😄

Firstyear:
@yaleman Given the work with tracing did we want to close and re-do this later?

yaleman:
Sure

PR #546, opened 2021-07-31
Add stricter headers

### LINKED ISSUE #374: Enforce stronger client usage headers
We should ensure that the following settings are enabled or relevant to prevent classes of attacks. 

* https://resourcepolicy.fyi/
* https://web.dev/fetch-metadata/
* https://web.dev/coop-coep/
* https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Frame-Options
* https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Content-Type-Options
* https://web.dev/samesite-cookies-explained/

Firstyear:
Fixes #374 - this adds strict usage headers and cache control header middlewares to our routers. Additionally this refactors out the middleware/router code from the actual handlers of these to make it a bit easier to follow.

### DIFF
--- a/kanidm_client/tests/oauth2_test.rs
+++ b/kanidm_client/tests/oauth2_test.rs
@@ -7,6 +7,30 @@ use oauth2_ext::PkceCodeChallenge;
 use std::collections::HashMap;
 use url::Url;
 
+macro_rules! assert_no_cache {
+    ($response:expr) => {{
+        // Check we have correct nocache headers.
+        let cache_header: &str = $response
+            .headers()
+            .get("cache-control")
+            .expect("missing cache-control header")
+            .to_str()
+            .expect("invalid cache-control header");
+
+        assert!(cache_header.contains("no-store"));
+        assert!(cache_header.contains("max-age=0"));
+
+        let pragma_header: &str = $response
+            .headers()
+            .get("pragma")
+            .expect("missing cache-control header")
+            .to_str()
+            .expect("invalid cache-control header");
+
+        assert!(pragma_header.contains("no-cache"));
+    }};
+}
+
 #[test]
 fn test_oauth2_basic_flow() {
     run_test(|rsclient: KanidmClient| {
@@ -73,6 +97,7 @@ fn test_oauth2_basic_flow() {
                 .expect("Failed to send request.");
 
             assert!(response.status() == reqwest::StatusCode::OK);
+            assert_no_cache!(response);
 
             let consent_req: ConsentRequest = response
                 .json()
@@ -92,6 +117,7 @@ fn test_oauth2_basic_flow() {
 
             // This should yield a 302 redirect with some query params.
             assert!(response.status() == reqwest::StatusCode::FOUND);
+            assert_no_cache!(response);
 
             // And we should have a URL in the location header.
             let redir_str = response
@@ -134,6 +160,7 @@ fn test_oauth2_basic_flow() {
                 .expect("Failed to send code exchange request.");
 
             assert!(response.status() == reqwest::StatusCode::OK);
+            assert_no_cache!(response);
 
             // The body is a json AccessTokenResponse
 
--- a/kanidmd/src/lib/core/https/mod.rs
+++ b/kanidmd/src/lib/core/https/mod.rs
@@ -368,23 +368,16 @@ impl<State: Clone + Send + Sync + 'static> tide::Middleware<State> for StrictReq
         request: tide::Request<State>,
         next: tide::Next<'_, State>,
     ) -> tide::Result {
-        let proceed = {
-            match request.header("sec-fetch-site") {
-                None => true,
-                Some(hv) => {
-                    let hvs = hv.as_str();
-                    if hvs == "same-origin" || hvs == "same-site" || hvs == "none" {
-                        true
-                    } else {
-                        request.header("sec-fetch-mode").map(|v| v.as_str()) == Some("navigate")
-                            && request.method() == tide::http::Method::Get
-                            && request.header("sec-fetch-dest").map(|v| v.as_str())
-                                != Some("object")
-                            && request.header("sec-fetch-dest").map(|v| v.as_str()) != Some("embed")
-                    }
-                }
-            }
-        };
+        let proceed = request
+            .header("sec-fetch-site")
+            .map(|hv| {
+                matches!(hv.as_str(), "same-origin" | "same-site" | "none")
+                    || (request.header("sec-fetch-mode").map(|v| v.as_str()) == Some("navigate")
+                        && request.method() == tide::http::Method::Get
+                        && request.header("sec-fetch-dest").map(|v| v.as_str()) != Some("object")
+                        && request.header("sec-fetch-dest").map(|v| v.as_str()) != Some("embed"))
+            })
+            .unwrap_or(true);
 
         if proceed {
             Ok(next.run(request).await)
--- a/kanidmd/src/lib/core/https/mod.rs
+++ b/kanidmd/src/lib/core/https/mod.rs
@@ -1,22 +1,14 @@
 mod oauth2;
+mod v1;
 
 use self::oauth2::*;
+use self::v1::*;
 
 use crate::actors::v1_read::QueryServerReadV1;
 use crate::actors::v1_write::QueryServerWriteV1;
 use crate::config::{ServerRole, TlsConfiguration};
-use crate::event::AuthResult;
-use crate::filter::{Filter, FilterInvalid};
-use crate::idm::AuthState;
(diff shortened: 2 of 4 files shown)


kanidm_client/tests/oauth2_test.rs
```diff
@@ -7,6 +7,30 @@ use oauth2_ext::PkceCodeChallenge;
...
+
+        assert!(cache_header.contains("no-store"));
+        assert!(cache_header.contains("max-age=0"));
+
+        let pragma_header: &str = $response
+            .headers()
+            .get("pragma")
+            .expect("missing cache-control header")
+            .to_str()
+            .expect("invalid cache-control header");
```
victorcwai, line 28:
```suggestion
            .expect("missing pragma header")
            .to_str()
            .expect("invalid pragma  header");
```

kanidmd/src/lib/core/https/mod.rs
```diff
@@ -1191,14 +425,17 @@ pub fn create_https_server(
...
     // Add middleware?
-    tserver.with(tide::log::LogMiddleware::new()).with(
+    tserver.with(tide::log::LogMiddleware::new());
+    tserver.with(
         // We do not force a session ttl, because we validate this elsewhere in usage.
         tide::sessions::SessionMiddleware::new(tide::sessions::MemoryStore::new(), cookie_key)
             .with_cookie_name("kanidm-session")
             .with_same_site_policy(tide::http::cookies::SameSite::Strict),
     );
+    tserver.with(StrictResponseMiddleware::default());
```
QnnOkabayashi, line 435: Why do we use the `Default` constructor? I feel like it would make more sense to do something like `StrictResponseMiddleware {}`, because this shows that it's a ZST and hides less information.

kanidmd/src/lib/core/https/mod.rs
```diff
@@ -1211,10 +448,13 @@ pub fn create_https_server(
             std::process::exit(1);
         }
 
-        tserver.at("/").get(index_view);
-        tserver.at("/ui/").get(index_view);
-        tserver.at("/ui/*").get(index_view);
-        tserver
+        let mut static_tserver = tserver.at("");
```
QnnOkabayashi, line 451: What's going on here? What about it is static?

kanidmd/src/lib/core/https/mod.rs
```diff
@@ -1161,6 +262,139 @@ impl<State: Clone + Send + Sync + 'static> tide::listener::Listener<State> for T
...
+
+#[async_trait::async_trait]
+impl<State: Clone + Send + Sync + 'static> tide::Middleware<State> for CacheableMiddleware {
+    async fn handle(
+        &self,
+        request: tide::Request<State>,
+        next: tide::Next<'_, State>,
+    ) -> tide::Result {
+        let mut response = next.run(request).await;
+        response.insert_header("Cache-Control", "max-age=60,must-revalidate,private");
```
QnnOkabayashi, line 303: What does this line do and what/how is it caching?

kanidmd/src/lib/core/https/mod.rs
```diff
@@ -1226,16 +466,35 @@ pub fn create_https_server(
...
+        .get(account_get_id_radius_token);
+    // We allow clients to cache the unix token.
+    account_route_cacheable
+        .at("/:id/_unix/_token")
+        .get(account_get_id_unix_token);
+
+    // ==== These routes can not be cached
+
+    let mut appserver = tserver.at("");
+    appserver.with(NoCacheMiddleware::default());
```
QnnOkabayashi, line 487: Isn't there still a `CacheableMiddleware` still in `tserver.at("")`? Or does this somehow only apply to the `tserver_cacheable` reference? I'm confused how these are are enforced to be disjoint.

kanidmd/src/lib/core/https/v1.rs
```diff
@@ -0,0 +1,896 @@
+use crate::event::AuthResult;
```
QnnOkabayashi, line 1: Assuming this is just moved from `https/mod.rs`.

QnnOkabayashi (changes requested):

kanidmd/src/lib/core/https/mod.rs
yaleman: It tells the client to cache content for a maximum of 60 seconds

kanidm_client/tests/oauth2_test.rs
```diff
@@ -7,6 +7,30 @@ use oauth2_ext::PkceCodeChallenge;
 use std::collections::HashMap;
 use url::Url;
 
+macro_rules! assert_no_cache {
```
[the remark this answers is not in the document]
Firstyear: It could be either tbh, I just like the macros here for small re-used snippets.

kanidmd/src/lib/core/https/oauth2.rs
```diff
@@ -290,8 +287,6 @@ async fn oauth2_authorise_permit(
             tide::Response::new(500)
         }
     };
-    res.insert_header("Cache-Control", "no-store");
-    res.insert_header("Pragma", "no-cache");
```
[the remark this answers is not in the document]
Firstyear: Correct!

kanidmd/src/lib/core/https/mod.rs
```diff
@@ -1191,14 +425,17 @@ pub fn create_https_server(
     // tide::log::with_level(tide::log::LevelFilter::Debug);
 
     // Add middleware?
-    tserver.with(tide::log::LogMiddleware::new()).with(
+    tserver.with(tide::log::LogMiddleware::new());
```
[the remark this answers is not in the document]
Firstyear: Indeed :)

kanidmd/src/lib/core/https/mod.rs
Firstyear: because else clippy will whinge if I added a new() -> Self but no default, so I used default instead. I prefered to leave it like this as we may want to add config/customisation later and it's a bit more readable this way.

kanidmd/src/lib/core/https/mod.rs
Firstyear: Static web content (IE static files that never change)

kanidmd/src/lib/core/https/mod.rs
Firstyear: So what happens is this is building a tree basicly. When we have the appserver from at("") this means "apply this middle ware at this route point" and then we build a set of subroutes that don't get cached through the appserver builder. The same way that later we use a different middleware that does apply the cache headers.

kanidmd/src/lib/core/https/v1.rs
Firstyear: Indeed!

kanidmd/src/lib/core/https/mod.rs
QnnOkabayashi: Ah, so when we insert the `NoCacheMiddleware`, we're making a new branch from `tserver.at("")` that we can add route functionality on. Whereas when we insert `CacheableMiddleware` above, that's it's own branch off of `tserver.at("")` that doesn't affect the route at `""` directly. Since the two middlewares are on their own branches of the tree, they're disjoint?

Firstyear: Yep, exactly. So all the routes are still correct but this just lets us apply different middlewares.

kanidmd/src/lib/core/https/mod.rs
```diff
@@ -1161,6 +262,139 @@ impl<State: Clone + Send + Sync + 'static> tide::listener::Listener<State> for T
...
+                Some(hv) => {
+                    let hvs = hv.as_str();
+                    if hvs == "same-origin" || hvs == "same-site" || hvs == "none" {
+                        true
+                    } else {
+                        request.header("sec-fetch-mode").map(|v| v.as_str()) == Some("navigate")
+                            && request.method() == tide::http::Method::Get
+                            && request.header("sec-fetch-dest").map(|v| v.as_str())
+                                != Some("object")
+                            && request.header("sec-fetch-dest").map(|v| v.as_str()) != Some("embed")
```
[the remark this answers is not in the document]
Firstyear: I think I'd rather there be `( ) ` around the components after the `||` just because I've been bitten by assuming operator ordering in the past. Otherwise I'd commit it (maybe I'm too old, I can't find how to edit your suggestion >.>)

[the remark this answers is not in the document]
Firstyear: Actually I'll just copy-paste it into the code myself, duh.

PR #549, opened 2021-08-01
LDAP UserAccountControl support

yaleman:
Per discussions offline, the LDAP UserAccountControl field is a bitfield of flags, which doesn't have a direct mapping back to kanidm things. This PR adds two things

 - a temporary hack to respond with the gidnumber to allow queries while I'm working on other things
 - an enum which allows for flags to be built
 
- [x] cargo fmt has been run


kanidmd/src/lib/constants/ldap.rs
```diff
@@ -0,0 +1,192 @@
...
+    ErrorPasswordExpired,
+    ///  The account is enabled for delegation. This is a security-sensitive setting; accounts with this option enabled SHOULD be strictly controlled. This setting enables a service running under the account to assume a client identity and authenticate as that user to other remote servers on the network.
+    TrustedToAuthnForDelegation,
+    ///  (Windows Server 2008/Windows Server 2008 R2) The account is a Read-Only Domain Controller (RODC). This is a security-sensitive setting. Removing this setting from an RODC compromises security on that server.
+    PartialSecretsAccount,
+    ///  Restrict this UserPrincipalName to use only Advanced Encryption Standard (AES) encryption types for keys. This bit is ignored by Windows Client and Windows Servers.
+    UserUseAesKeys,
+}
+
+impl From<LdapUacFlag> for u32 {
```
Firstyear, line 69: I think you could do LdapUacFlag as repr=u32 instead here?

kanidmd/src/lib/constants/ldap.rs
```diff
@@ -0,0 +1,192 @@
+//! Constants for the LDAP server
+
```
Firstyear, line 2: This could be in https://github.com/kanidm/ldap3_server potentially?

kanidmd/src/lib/ldap.rs
```diff
@@ -529,6 +530,8 @@ pub(crate) fn ldap_vattr_map(input: &str) -> &str {
         "sshpublickey" => "ssh_publickey",
         "cn" => "name",
         "uidnumber" => "gidnumber",
+        // TODO: temporary hack to make it work until the [kanidm::constants::ldap::LdapUacFlag] enum can be turned into a filter
```
Firstyear, line 533: Why can't it be a filter right now?

kanidmd/src/lib/schema.rs
```diff
@@ -1121,6 +1121,19 @@ impl<'a> SchemaWriteTransaction<'a> {
                     syntax: SyntaxType::UINT32,
                 },
             );
+            self.attributes.insert(
+                AttrString::from("useraccountcontrol"),
+                SchemaAttribute {
+                    name: AttrString::from("useraccountcontrol"),
+                    uuid: *UUID_SCHEMA_ATTR_USERACCOUNTCONTROL,
+                    description: String::from("An LDAP Compatible UserAccountControl"),
```
Firstyear, line 1129: Probably should say it's MS-AD compat

Firstyear (changes requested):

kanidmd/src/lib/ldap.rs
yaleman: Because I don't know how to write them? 😅

kanidmd/src/lib/constants/ldap.rs
yaleman: It sure could, I didn't realise it was its own crate :)

kanidmd/src/lib/constants/ldap.rs
yaleman: I really don't know enough rust to know what you mean by that, and google's not helping sorry :)

kanidmd/src/lib/schema.rs
yaleman: Done

kanidmd/src/lib/constants/ldap.rs
yaleman: .. fixed :D

kanidmd/src/lib/ldap.rs
Firstyear: Isn't it the other way here? We need to synthesise the UACflags from other values?

kanidmd/src/lib/constants/ldap.rs
Firstyear: For now we can do it here until it works, then we can add this to ldap3_server later :)

kanidmd/src/lib/ldap.rs
yaleman: Yes, I don't know the right words for the thing, I just put in a hack to make it work for testing 😄  You mentioned elsewhere it'd need to be... oh, a plugin. 

Words are hard. Derp.

yaleman:
No longer needed

PR #563, opened 2021-08-10
updating well-known URI to meed OIDC spec

yaleman:
Per https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfig the URI should be `/.well-known/openid-configuration`

- [x] cargo fmt has been run


yaleman:
Well that fail wasn't *me* at least 😆 
Screen Shot 2021-08-11 at 08 39 26

PR #565, opened 2021-08-13
508 token introspect

### LINKED ISSUE #508: rfc7662 oauth2 token introspection
https://tools.ietf.org/html/rfc7662

Now that we have basic oauth2, consumers may want to inspect identiy information. We can expose this with rfc7662 to some clients (as an alternative to oidc).

Firstyear:
Fixes #508 - adds rfc7662 oauth2 token introspection. This changes the internal formats of some of our auth tokens to better support this behaviour.

### DIFF
--- a/kanidm_proto/src/v1.rs
+++ b/kanidm_proto/src/v1.rs
@@ -182,12 +182,15 @@ pub enum AuthType {
 pub struct UserAuthToken {
     pub session_id: Uuid,
     pub auth_type: AuthType,
+    // What time we issued this authentication token.
+    pub issued_at: time::OffsetDateTime,
     // When this token should be considered expired. Interpretation
     // may depend on the client application.
     pub expiry: time::OffsetDateTime,
     pub uuid: Uuid,
     // pub name: String,
     pub spn: String,
+    pub displayname: String,
     // pub groups: Vec<Group>,
     // pub claims: Vec<Claim>,
     // Should we just retrieve these inside the server instead of in the uat?
--- a/kanidmd/src/lib/idm/account.rs
+++ b/kanidmd/src/lib/idm/account.rs
@@ -172,11 +172,10 @@ impl Account {
 
         // TODO: Apply policy to this expiry time.
         let expiry = OffsetDateTime::unix_epoch() + ct + Duration::from_secs(AUTH_SESSION_EXPIRY);
-        let issued_at = OffsetDateTime::unix_epoch() + ct;
 
         Some(UserAuthToken {
             session_id,
-            issued_at,
+            issued_at: OffsetDateTime::unix_epoch() + ct,
             expiry,
             // name: self.name.clone(),
             spn: self.spn.clone(),
--- a/kanidmd/src/lib/idm/account.rs
+++ b/kanidmd/src/lib/idm/account.rs
@@ -172,13 +172,15 @@ impl Account {
 
         // TODO: Apply policy to this expiry time.
         let expiry = OffsetDateTime::unix_epoch() + ct + Duration::from_secs(AUTH_SESSION_EXPIRY);
+        let issued_at = OffsetDateTime::unix_epoch() + ct;
 
         Some(UserAuthToken {
             session_id,
+            issued_at,
             expiry,
             // name: self.name.clone(),
             spn: self.spn.clone(),
-            // displayname: self.displayname.clone(),
+            displayname: self.displayname.clone(),
             uuid: self.uuid,
             // application: None,
             // groups: self.groups.iter().map(|g| g.to_proto()).collect(),
@@ -193,7 +195,7 @@ impl Account {
         })
     }
 
-    pub fn check_within_valid_time(
+    fn check_within_valid_time(
         ct: Duration,
         valid_from: Option<&OffsetDateTime>,
         expire: Option<&OffsetDateTime>,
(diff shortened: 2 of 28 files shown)


Firstyear:
Rebased on latest middleware changes

kanidmd/src/lib/idm/account.rs
```diff
@@ -172,13 +172,15 @@ impl Account {
 
         // TODO: Apply policy to this expiry time.
         let expiry = OffsetDateTime::unix_epoch() + ct + Duration::from_secs(AUTH_SESSION_EXPIRY);
+        let issued_at = OffsetDateTime::unix_epoch() + ct;
```
QnnOkabayashi, line 175: Inline this declaration into the constructor?

kanidmd/src/lib/idm/oauth2.rs
```diff
@@ -146,14 +149,16 @@ pub struct Oauth2ResourceServersWriteTransaction<'a> {
...
+impl Oauth2ResourceServers {
+    pub fn try_new(
+        value: Vec<EntrySealedCommitted>,
+        origin: String,
+    ) -> Result<Self, OperationError> {
         let fernet =
             Fernet::new(&Fernet::generate_key()).ok_or(OperationError::CryptographyError)?;
         let oauth2rs = Oauth2ResourceServers {
             inner: CowCell::new(Oauth2RSInner {
+                origin,
```
QnnOkabayashi, line 161: What's this?

kanidmd/src/lib/idm/oauth2.rs
```diff
@@ -499,24 +511,56 @@ impl Oauth2ResourceServersReadTransaction {
...
+        let expires_in = if exp > iat {
+            exp - iat
         } else {
             lsecurity!(
                 audit,
                 "User Auth Token has expired before we could publish the oauth2 response"
             );
             return Err(Oauth2Error::AccessDenied);
         };
+        debug_assert!(expires_in.is_positive());
```
QnnOkabayashi, line 531: I don't know what type `exp` and `iat` are but you could probably do something like this to be more idiomatic:

```suggestion
let expires_in = checked_sub(exp, iat).ok_or_else(|| {
    lsecurity!(
        audit,
        "User Auth Token has expired before we could publish the oauth2 response"
    )
    Oauth2Error::AccessDenied
)?;
```

kanidmd/src/lib/idm/oauth2.rs
```diff
@@ -1018,4 +1136,141 @@ mod tests {
...
+                idms_prox_read.check_oauth2_token_introspect(audit, &client_authz, &good_req, exp);
+
+            if let Ok(IntrospectionResponse::Invalid { active }) = r2 {
+                assert!(!active);
+            } else {
+                panic!();
+            };
+
+            // expire the account, check it's enforced.
+            std::mem::drop(idms_prox_read);
```
QnnOkabayashi, line 1244: `drop` is in the prelude

```suggestion
            drop(idms_prox_read);
```

QnnOkabayashi (changes requested):
So. Many. Minor changes :(

kanidm_proto/src/v1.rs
```diff
@@ -182,12 +182,15 @@ pub enum AuthType {
...
     pub auth_type: AuthType,
+    // What time we issued this authentication token.
+    pub issued_at: time::OffsetDateTime,
     // When this token should be considered expired. Interpretation
     // may depend on the client application.
     pub expiry: time::OffsetDateTime,
     pub uuid: Uuid,
     // pub name: String,
     pub spn: String,
+    pub displayname: String,
```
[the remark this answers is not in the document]
Firstyear: Yep!

kanidmd/src/lib/idm/oauth2.rs
Firstyear: Origin is the URL origin of this instance, which can be used for domain name checks etc.

kanidmd/src/lib/idm/oauth2.rs
Firstyear: expiry and issued at

kanidmd/src/lib/idm/oauth2.rs
```diff
@@ -515,29 +515,32 @@ impl Oauth2ResourceServersReadTransaction {
...
-let expires_in = exp.checked_sub(iat).ok_or_else(|| {
-    lsecurity!(
-        audit,
-        "User Auth Token has expired before we could publish the oauth2 response"
-    )
-    Oauth2Error::AccessDenied
-)?;
+        let expires_in = expiry
+            .checked_sub(issued_at)
+            .and_then(|r| if r.is_negative() { None } else { Some(r) })
```
QnnOkabayashi, line 524: I'm still confused about what type `expiry` and `issued_at` are. If they're signed types (which seemed implied to `is_negative`, then the checked sub is kind of pointless. I kinda suggested the change assuming they were unsigned.

Firstyear: They have to be i64 because lol js standards dont seem to have unsigned types :( 

The cheked sub is probably still a good idea incase something wild happens, and the resulting code is pretty clean anyway, so it's a lot of muchness I think

PR #567, opened 2021-08-13
Fixes 566 limit body size

### LINKED ISSUE #566: HTTP calls don't have a max size leading to OOM/DoS
@Firstyear prompted me on this one, but if you send a big enough request, it will fill up RAM. This is bad. 😄 

Where testfile.zero is an 8GB file, and the machine has 4GB of RAM, 2GB of swap: 

```
date && time curl -vvvH "Content-Type: application/json" -X POST --data-binary @testfile.zero https://kanidm.example.com/v1/auth && date
<snip>
< HTTP/1.1 100 Continue
* LibreSSL SSL_read: SSL_ERROR_SYSCALL, errno 54
* Closing connection 0
curl: (56) LibreSSL SSL_read: SSL_ERROR_SYSCALL, errno 54
curl -vvvH "Content-Type: application/json" --data-binary    20.56s user 6.29s system 2% cpu 16:07.48 total
```

In the meantime, CPU/Memory usage on the server spiked and the process was eventually OOM terminated by the kernel.

yaleman:
Fixes #566

- [x] cargo fmt has been run

### DIFF
--- a/kanidmd/src/lib/constants/mod.rs
+++ b/kanidmd/src/lib/constants/mod.rs
@@ -43,3 +43,6 @@ pub const PW_MIN_LENGTH: usize = 10;
 
 // Default
 pub const AUTH_SESSION_EXPIRY: u64 = 3600;
+
+/// Default max request post size, 100MB
+pub const MAX_BODY_SIZE: usize = 104857600;
--- a/kanidmd/src/lib/core/https/mod.rs
+++ b/kanidmd/src/lib/core/https/mod.rs
@@ -1164,9 +1164,92 @@ impl<State: Clone + Send + Sync + 'static> tide::listener::Listener<State> for T
 }
 */
 
+#[cfg(test)]
+mod tests {
+
+    pub const TEST_STRING: &'static str = "OK";
+
+    #[async_std::test]
+    /// This is a simple test which should always pass, just testing that when the limiter is loaded, it doesn't bin *everything!
+    async fn request_limiter_ok() -> tide::Result<()> {
+        use crate::constants::MAX_BODY_SIZE;
+        use crate::core::https::RequestLimiter;
+        use tide::http::{Method, Request, Response, Url};
+
+        let mut app = tide::new();
+        app.with(RequestLimiter::from_size(MAX_BODY_SIZE));
+
+        app.at("/").post(|_| async { Ok(TEST_STRING) });
+        let url = Url::parse("https://example.com").unwrap();
+        let mut req = Request::new(Method::Post, url);
+        req.set_body("Hello Chashu");
+
+        let mut res: Response = app.respond(req).await?;
+        assert_eq!(TEST_STRING, res.body_string().await?);
+        Ok(())
+    }
+    #[async_std::test]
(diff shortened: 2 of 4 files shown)


yaleman:
I still want to add testing to this... just have to work out the test harness bit 😆

kanidmd/src/lib/core/https/mod.rs
```diff
@@ -1163,6 +1164,140 @@ impl<State: Clone + Send + Sync + 'static> tide::listener::Listener<State> for T
...
+        req.set_body(body_string);
+
+        let mut res: Response = app.respond(req).await?;
+
+        assert_eq!(tide::StatusCode::BadRequest, res.status());
+        assert_eq!("", res.body_string().await?);
+        Ok(())
+    }
+
+    #[async_std::test]
```
Firstyear, line 1216: You may want to use tokio::test instead here

kanidmd/src/lib/core/https/mod.rs
```diff
@@ -1163,6 +1164,140 @@ impl<State: Clone + Send + Sync + 'static> tide::listener::Listener<State> for T
 }
 */
 
+#[cfg(test)]
+mod tests {
```
Firstyear, line 1168: Tests should be at the bottom of the file.

Firstyear (changes requested):
https://github.com/http-rs/tide/issues/448 It may be better to look at or contribute to this instead of trying our own thing :)

kanidmd/src/lib/core/https/mod.rs
yaleman: They totally will be once I'm done writing the code - just easier to keep it close and save epic-scrolling on my tiny laptop 😄

kanidmd/src/lib/constants/mod.rs
```diff
@@ -43,3 +43,6 @@ pub const PW_MIN_LENGTH: usize = 10;
 
 // Default
 pub const AUTH_SESSION_EXPIRY: u64 = 3600;
+
+/// Default max request post size, 100MB
+pub const MAX_BODY_SIZE: usize = 104857600;
```
vDorst, line 48: I probably unaware why we need a large request size.
100MB seems quit high for this application.
1MB or even 64KB (Max TCP packet size) seems too me, more then enough.

yaleman:
I'm... not even sure why I pushed this as a PR, it's well too early in my own dev cycle for commenting. I'll close it for now, apologies folks 😊

Firstyear:
No apologies needed mate, thanks for looking into it. You should also subscribe to the tide limiting issue :)

kanidmd/src/lib/constants/mod.rs
Firstyear: If someone is doing batch changes we'll need more than 1MB in some cases, so it needs to be configurable either per route or per user in some cases.

yaleman:
> No apologies needed mate, thanks for looking into it. You should also subscribe to the tide limiting issue :)

Already have, looks like it's going to take a whole load of coordination amongst projects 😄

PR #573, opened 2021-08-24
Fix io capture in tests

Firstyear:
This fixes tracing such that IO is properly captured during tests.

### DIFF
--- a/kanidmd/src/lib/tracing_tree/subscriber.rs
+++ b/kanidmd/src/lib/tracing_tree/subscriber.rs
@@ -230,7 +230,7 @@ impl<P: Processor> Layer<Registry> for TreeLayer<P> {
 
         let name = attrs.metadata().name();
         let mut uuid = None;
-        let mut out = TreeIo::Stderr;
+        let mut out = TreeIo::Stdout;
 
         attrs.record(
             &mut |field: &Field, value: &dyn fmt::Debug| match field.name() {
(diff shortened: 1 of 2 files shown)


kanidmd/src/lib/tracing_tree/subscriber.rs
```diff
@@ -478,8 +478,20 @@ impl TreePreProcessed {
         let buf = &formatted_logs[..];
 
         match processed_logs.tree_io() {
-            TreeIo::Stdout => io::stdout().write_all(buf),
-            TreeIo::Stderr => io::stderr().write_all(buf),
+            TreeIo::Stdout => {
+                // BUG - we can't write to stdout/err directly because this breaks
+                // cargo test capturing of io.
+                // io::stdout().write_all(buf)
+                let s = unsafe { std::str::from_utf8_unchecked(buf) };
```
QnnOkabayashi, line 485: We really should just make the `writer` in the `formatter.rs` be a `String`, so we can avoid this. This could go in a future PR though.

QnnOkabayashi (approved):

kanidmd/src/lib/tracing_tree/subscriber.rs
Firstyear: Yes I had a look, but the issue is then serde_json with the buffer. :( So it would be a bigger pr I think

QnnOkabayashi: Ah, didn't realize. There's a checked variant, [`str::from_utf8`](https://doc.rust-lang.org/stable/std/str/fn.from_utf8.html), that we should probably use instead of introducing `unsafe` more.

PR #575, opened 2021-08-28
Swap to tide-openssl

### LINKED ISSUE #341: Re-enable openssl for tide
We currently have to use rustls with tide because async_h1 https://docs.rs/async-h1/2.3.0/async_h1/server/fn.accept.htm requires futures_io https://docs.rs/futures-io/0.3.8/futures_io/trait.AsyncRead.html 

However, https://docs.rs/tokio-openssl/0.6.1/tokio_openssl/ implements https://docs.rs/tokio/1.0.1/tokio/io/trait.AsyncRead.html which is not compatible.

We need either to create tide_openssl that does do the futures_io version of asyncrw, or we need to wait for async_h1 to change to tokio async traits.

Firstyear:
Fixes #341 - this merges @victorcwai (excellent) tide-openssl project and allows us to standardise on openssl again. 🎉

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -240,17 +240,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "async-rustls"
-version = "0.2.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "9c86f33abd5a4f3e2d6d9251a9e0c6a7e52eb1113caf893dae8429bf4a53f378"
-dependencies = [
- "futures-lite",
- "rustls",
- "webpki",
-]
-
-[[package]]
 name = "async-session"
 version = "2.0.1"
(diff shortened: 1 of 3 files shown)


victorcwai:
woohoo! I wonder if we should run some benchmark to compare before and after though?

Firstyear:
Honestly, it's more important to use openssl over rustls for the project, so the result of any benchmark wouldn't really factory in to if we merge this. Benchkmarking would still be *useful* though to ensure that our async code is performing well, and that's going to involve extending orca to be better :)

Firstyear:
Saying that it's worth noting that performance still isn't entirely on the roadmap - it's a nice to have not a must have still at this stage, as feature enrichment is probably what we should be going for. I only do performance improvements as needed to make tests run faster XD

victorcwai (approved):

PR #579, opened 2021-09-12
Arc track entry

Firstyear:
Experimental, builds on top of valueset to track entries with Arc to reduce clones.

- [x] cargo fmt has been run

### DIFF
--- a/kanidmd/src/lib/access.rs
+++ b/kanidmd/src/lib/access.rs
@@ -23,6 +23,7 @@ use std::collections::BTreeSet;
 // use hashbrown::HashSet;
 use std::cell::Cell;
 use std::ops::DerefMut;
+use std::sync::Arc;
 use uuid::Uuid;
 
 use crate::entry::{Entry, EntryCommitted, EntryInit, EntryNew, EntryReduced, EntrySealed};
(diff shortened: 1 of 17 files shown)


QnnOkabayashi (approved):
Not sure if `Arc` is a great alternative but it looks good to me

Firstyear:
So this change was actually inspired by something you said a long time ago - about how can we reduce clones and use more references? 

The main challenge we have is being multi-threaded, and due to how the concurrent datastructures work in tandem with some other parts, we can't return references from inside those caches. Previously we would clone the whole entry, but that is expensive both on memory usage (especially on a really large search) but also in hashing/rebuilding the valueset and attrs internals. Once entries grow in size we also end up paying larger and larger costs here. 

So this change is to achieve two things: First, it removes the need to clone the full entry (or entries) during a search path, meaning we only end up cloning the exact set of requested and allowed attributes after access controls are applied. It also reduces the overall memory footprint due to reduction of these clones and how many entries we are keeping in memory. 

So I hope that explains it! 

PS: The reason I did it was I needed a distraction on the weekend and this seemed as good as any given I'd just changed the valueset code.

PR #591, opened 2021-10-10
Update outputs for "group" commands

### LINKED ISSUE #590: Group modification commands via CLI don't print success on successful events
They typically print errors, but not successes, even in debug mode 😄 :uwu:

yaleman:
Fixes #590 

- [x] cargo fmt has been run

### DIFF
--- a/kanidm_tools/src/cli/group.rs
+++ b/kanidm_tools/src/cli/group.rs
@@ -24,51 +24,58 @@ impl GroupOpt {
             GroupOpt::List(copt) => {
                 let client = copt.to_client();
                 match client.idm_group_list() {
-                    Ok(r) => r.iter().for_each(|ent| println!("{}", ent)),
-                    Err(e) => eprintln!("Error -> {:?}", e),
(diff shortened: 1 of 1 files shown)


cuberoot74088:
Hi @yaleman ,


I was running `kanidm group list` and other group related commands and was surprised to see no output at first.

Then I used `RUST_LOG=INFO` and got my expected and usual list of groups and members etc. ...

Is there some special reason this normal output was changes from `println!` to `info!` ?

I would expect to see "normal" output to my command. 
This is especially confusing as other parts of the cli tool still just do `println!`, this produce "normal" output.
I would expect a consistent way of returning or displaying output.

Or have i missed some new config setting or so?


Best regards

yaleman:
@cuberoot74088 this is one of those "kanidm is still in alpha" things at the moment - I went for `info!` to allow for configurability/consistency of output - it's curious that you aren't seeing `info` level messages by default... sounds like something's been tweaked.

yaleman:
I've raised this as an issue here @cuberoot74088 #652

PR #593, opened 2021-10-10
Integrated `compiled-uuid` into `kanidmd/src/lib/constants/uuids.rs`

### LINKED ISSUE #592: Use `compiled-uuid` for `Uuid` constants
In `kanidmd/src/lib/constants/uuid.rs`, we have a ton of `Uuid` constants are that initialized via `lazy_static` and `Uuid::parse_str` + `upwrap`. We can avoid this overhead and make the file much shorter by using `compiled_uuid::uuid`, which would let us initialize them all at compile time.

QnnOkabayashi:
Fixes #592

- [x] cargo fmt has been run

Left all the original constants commented out. Also added the `_` prefix to several `&str` constants that were never being used.

Also, a lot of other files changed also because we no longer have to deref the constants, and this made clippy very sad. Should I implement its suggestions?

### DIFF
--- a/kanidm_tools/build.rs
+++ b/kanidm_tools/build.rs
@@ -14,24 +14,39 @@ fn main() {
         Some(outdir) => outdir,
     };
 
+    let comp_dir = PathBuf::from(outdir)
+        .ancestors()
+        .skip(2)
+        .next()
+        .map(|p| p.join("completions"))
+        .expect("Unable to process completions path");
+
+    if !comp_dir.exists() {
+        std::fs::create_dir(&comp_dir).expect("Unable to create completions dir");
+    }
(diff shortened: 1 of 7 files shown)


QnnOkabayashi:
> Is there anything else before I merge? Still marked as draft ...

I'm not satisfied with the unused constants- can we delete them? Or make them all `Uuid`s? I wasn't sure what to do with them.

There are also some uuids that are used in `&str` representation in some areas. Some of them get parsed later, but some go so deep that I can't trace what happens to them. We don't have to address those here, but I was still confused by them. I just can't imagine a scenario where we care more about the `&str` repr more than the `Bytes` repr.

Issue #60, opened 2019-07-27
Auth session cleanup (gc) process


Firstyear:
We have a concurrently readable structure for in progress authentications. However, we should clean up expired sessions after a period of time. This requires an associated heap that can show us the "next" expiring structure for GC.

PR #605, opened 2021-10-20
Dynamic log level filtering

### LINKED ISSUE #604: Runtime-configurable log levels
We should be able to dynamically configure log levels at runtime for spans based on the configuration. To reduce configuration complexity, filters at the lowest level will apply to all higher levels.

QnnOkabayashi:
Fixes #604

- [x] cargo fmt has been run

I haven't tested this because I forgot how to run the server. But it should work?


Firstyear:
So how we would get the loglevel from config.log_level (https://github.com/kanidm/kanidm/blob/master/kanidmd/src/lib/core/mod.rs#L580 ) and get that into the right places?

Firstyear:
https://github.com/kanidm/kanidm/blob/master/kanidmd/src/lib/audit.rs#L9

This is the current way it's done, but we can change it of course #60 
I think the global log level seems like that best idea for now!

Firstyear:
Yes that's right, but we would normally not just have that level. We'd combine these to make something like this https://github.com/kanidm/kanidm/blob/master/kanidmd/src/lib/audit_loglevel.rs

QnnOkabayashi:
Currently working on cleaning up the log levels for integration, can you remind me what `PerfOp`, `PerfCourse`, and `PerfTrace` are used for and what the differences are?

Firstyear:
> Currently working on cleaning up the log levels for integration, can you remind me what `PerfOp`, `PerfCourse`, and `PerfTrace` are used for and what the differences are?

At the time these were to enable and disable the performance tracking levels for span timings. That's because it was "slower" in the way it was previously done, so you didn't want to enable too many of them. So PerfOp was the "top level" of how long did any given operation taken. PerfCoarse was a "rough" timing of elements in the Operation. And Trace was extremely detailed timings inside. Think of it as more and more detailed spans.

Firstyear:
@QnnOkabayashi hey mate, I'm looking at this and I'm a bit confused, don't we just need something like https://docs.rs/tracing-subscriber/latest/tracing_subscriber/struct.EnvFilter.html here? So that when we call TreeSubscriber::pretty(); we pass in a default log-level we want, and then we store that in our TreeSubscriber and can return the max_level_hint or similar?

yaleman:
Is this still being worked on, or can we close the PR for now?

Firstyear:
I think we can close it.

PR #608, opened 2021-11-02
278 603 OIDC implementation

### LINKED ISSUE #278: oauth / oidc
https://openid.net/specs/openid-connect-core-1_0.html

https://crates.io/crates/oxide-auth

### LINKED ISSUE #603: oauth2: `/oauth2/userinfo` is required for OpenID Connect support
One other problem I encountered when tackling #600 is that kanidm seems to be missing an endpoint for [OpenID Connect Userinfo](https://openid.net/specs/openid-connect-core-1_0.html#UserInfo).

At least for [vouch-proxy](https://github.com/vouch/vouch-proxy) I have to provide an `user_info_url`, which is required when using an OIDC identity provider. When naively setting it to `https://my.kanidm.instance/oauth2/userinfo` my login process fails after the `Consent to proceed...` button with a 400 error to vouch-proxy: 

x-vouch-error | /auth Error while retrieving user info after successful login at the OAuth provider: unexpected end of JSON input
-- | --

Also the OIDC spec (linked above) seems to mandate the existence of that endpoint.

I'd be happy to help in implementing the `userinfo` endpoint, but all this auth stuff is really new to me so I don't know of how much use my skills are going to be. I'll definetly be able to test the changes though.

Firstyear:
This adds OIDC to Kanidm. This replaces our current oauth2 tokens with OIDC compatible ones, and supports the correct generation of the private key material. 

Fixes #278 
Fixes #603


Firstyear:
The only thing that may be missing is the ability to reset/roll the private keys atm.

yaleman:
Tagging #278 #603 for cross referencing 😄

Firstyear:
Okay, this now works but I want to finish adding some extended test cases to complete this.

kanidm_book/src/oauth2.md
```diff
@@ -47,21 +47,31 @@ decisions to Kanidm.
...
-* auth url: https://idm.example.com/ui/oauth2
+* user auth url: https://idm.example.com/ui/oauth2
+* api auth url: https://idm.example.com/oauth2/authorise
 * token url: https://idm.example.com/oauth2/token
 * token inspect url: https://idm.example.com/oauth2/inspect
 
+OpenID Connect discovery:
+
+* openid connect issuer uri: https://idm.example.com/oauth2/openid/:client\_id/
+* openid connect discovery:  https://idm.example.com/oauth2/openid/:client\_id/.well-known/openid-configuration
```
yaleman, line 68: Please call out that the client ID should be substituted ^ on these lines

kanidm_book/src/oauth2.md
```diff
@@ -47,21 +47,31 @@ decisions to Kanidm.
...
+
+For manual OpenID configuration:
+
+* openid connect userinfo:   https://idm.example.com/oauth2/openid/:client\_id/userinfo
+* token signing public key:  https://idm.example.com/oauth2/openid/:client\_id/public\_key.jwk
```
yaleman, line 73: Please call out that the client ID should be substituted ^ on these lines

yaleman:
Flow works good so far, I'm testing with CloudFlare access and the following error comes back due to a lack of email fields for the user 😃 
```
User email was not returned. API permissions are likely incorrect.
{
  "id": "2e1399da-61f5-47f5-becb-146cf477d043",
  "oidc_fields": {},
  "amr": [
    "\"passwordmfa\""
  ]
}
```

It'd be good to have a check on the kanidm side when `http.host` != the configured `origin` for signing, to avoid admin footguns 😄

Firstyear:
Need to also fix the session timeout.

yaleman:
What's the current timeout? I seem to stay logged in for aaaaaaaaaaages.

Firstyear:
> What's the current timeout? I seem to stay logged in for aaaaaaaaaaages.

Apparently william fat fingered this and made it 8 minutes XD 

The fact that nothing is correctly handling the exp value is already pretty luls IMO ....

Firstyear:
> It'd be good to have a check on the kanidm side when `http.host` != the configured `origin` for signing, to avoid admin footguns 😄

What do you mean by http.host here? Looking at this, origin is the only thing we configure and it comes from server.toml.

Firstyear:
Email: https://github.com/kanidm/kanidm/issues/612 
Logout: https://github.com/kanidm/kanidm/issues/613

yaleman:
> > It'd be good to have a check on the kanidm side when `http.host` != the configured `origin` for signing, to avoid admin footguns 😄
> 
> What do you mean by http.host here? Looking at this, origin is the only thing we configure and it comes from server.toml.

When you point multiple hostnames at a server, `http.host` is the [Host header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/host) sent when the client connects, which we log here:

https://github.com/kanidm/kanidm/blob/b0542c7e54c8e234a379ccb5658f776efa823c8c/kanidmd/src/lib/tracing_tree/middleware.rs#L63

But my understanding is that the request signing will be based on the `origin` in server.toml, so it'll be invalid if they connect to another hostname. 

For example, if you have both `internal.example.com` and `example.com` as valid hostnames, but `example.com` is the `origin` setting, the user connects to internal and then the OIDC RS won't be able to parse the key.

Firstyear:
> For example, if you have both `internal.example.com` and `example.com` as valid hostnames, but `example.com` is the `origin` setting, the user connects to internal and then the OIDC RS won't be able to parse the response.

Right that's not a simple fix here. I think it's a new issue as we need to think about how to do this securely and correctly because we need to dynamically set the issuer value in this situation and even a quick few seconds of thought I can see how it could be abused. So I would rather it be a seperate issue so we can have the time to think about, discuss, and design it correctly. :)

yaleman:
Oh, I was just suggesting an `admin.warn` of "hey, this might be why you're having issues" 😄

Firstyear:
> Oh, I was just suggesting an `warn` level message of "hey, this might be why you're having issues" 😄

It's still quite a bit more work than may be apparent, we need to pass the http.host down through the stack. And even in that case, that warning won't really help because we wont have a fix to set the issuer correctly either. So I think it's still a new issue ....

Firstyear:
Okay, new tickets have been made, cleaned up a few things remaining and expiries fixed. It's working with nextcloud and mod_openid with apache too, so pretty happy here. The main limitation is lack of email addresses but thats up next.

So if there are no objections, I'll merge in the next few days.

PR #617, opened 2021-11-24
Improve book and errors related to domain name and origin mismatch

### LINKED ISSUE #616: Possibility to set `server_name` from config
Currently (current master 0f4189a57ef9fb0f048cd2cc3f5eee3d132adbd1) it doesn't seem possible to set `server_name` from config, resulting in startup errors like this:

```
kanidmd[919]: 00000000-0000-0000-0000-000000000000 2021-11-24T02:27:40.841254897+00:00 ERROR    🚨 [admin.error]: Effective domain is not a descendent of server domain name (rp_id). You must change origin or domain name to be consistent. ed: "https://idm.acme.test" - rp_id: "example.com"
kanidmd[919]: [kanidm::core] Unable to setup query server or idm server -> InvalidState
```

server.toml:

```toml
bindaddress = "[::1]:8443"
db_path = "/var/lib/kanidm/kanidm.db"
# this setting doesn't seem to have an effect but maybe would be a good place for specifying it?
domain_name = "acme.test"
log_level = "default"
origin = "https://idm.acme.test"
role = "WriteReplica"
```

My current workaround is patching the source:

```patch
diff --git a/kanidmd/src/lib/plugins/domain.rs b/kanidmd/src/lib/plugins/domain.rs
index 7e1797f2..15b8737d 100644
--- a/kanidmd/src/lib/plugins/domain.rs
+++ b/kanidmd/src/lib/plugins/domain.rs
@@ -39,7 +39,7 @@ impl Plugin for Domain {
                 trace!("plugin_domain: Applying uuid transform");
                 // We only apply this if one isn't provided.
                 if !e.attribute_pres("domain_name") {
-                    let n = Value::new_iname("example.com");
+                    let n = Value::new_iname("acme.test");
                     e.set_ava("domain_name", btreeset![n]);
                     trace!("plugin_domain: Applying domain_name transform");
                 }


```

I find it a bit unintuitive to have `example.com` be a default value, this is probably not intended?

Firstyear:
Fixes #616 - improves the errors and book around domain name configuration. 

### DIFF
--- a/kanidm_book/src/server_configuration.md
+++ b/kanidm_book/src/server_configuration.md
@@ -1,5 +1,7 @@
 ## Configuring the Server
 
+### Configuring server.toml
+
 You will also need a config file in the volume named `server.toml` (Within the container it should be `/data/server.toml`). Its contents should be as follows:
 
     #   The webserver bind address. Will use HTTPS if tls_* is provided.
@@ -36,7 +38,8 @@ You will also need a config file in the volume named `server.toml` (Within the c
     # log_level = "default"
     #
     #   The origin for webauthn. This is the url to the server, with the port included if
-    #   it is non-standard (any port except 443)
+    #   it is non-standard (any port except 443). This must match or be a descendent of the
+    #   domain name you configure with `kanidmd domain_name_change`
     # origin = "https://idm.example.com"
     origin = "https://idm.example.com:8443"
     #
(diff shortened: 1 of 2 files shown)


Firstyear:
cc @erictapen

kanidm_book/src/server_configuration.md
```diff
@@ -68,16 +71,26 @@ You will also need a config file in the volume named `server.toml` (Within the c
 
 An example is located in [examples/server.toml](../../examples/server.toml).
 
-Then you can setup the initial admin account and initialise the database into your volume.
-
-    docker run --rm -i -t -v kanidmd:/data kanidm/server:latest /sbin/kanidmd recover_account -c /data/server.toml -n admin
+### Domain Name
 
-You then want to set your domain name so that security principal names (spn's) are generated correctly.
+You then *MUST* to set your domain name so that security principal names (spn's) are generated correctly.
```
erictapen, line 76:
```suggestion
You then *MUST* set your domain name so that security principal names (spn's) are generated correctly.
```

erictapen (commented):
Makes the situation better definetly. In my opinion the error message could be even clearer. The server shouldn't be configured with `example.com` in the first place, as it is never usable for anything. Instead it should be set to some kind of `None` value and an error message along the lines of `domain_name isn't configured. You must setup a domain_name before first startup, see this doc on how to do it.` would communicate an even better expectation to the user.

Firstyear:
> Makes the situation better definetly. In my opinion the error message could be even clearer. The server shouldn't be configured with `example.com` in the first place, as it is never usable for anything. Instead it should be set to some kind of `None` value and an error message along the lines of `domain_name isn't configured. You must setup a domain_name before first startup, see this doc on how to do it.` would communicate an even better expectation to the user.

There is a chicken an egg problem. To automatically create the DB content, we need a domain name in place. But we need the DB to allow you to change the domain name.

If we were to make it "optional" we'd need a much bigger rewrite and change to handle that possibility through the code, as well as the tests - yes, the test cases rely on this default example.com behaviour. 

The easiest way to "force"  a change from default could be a bool flag for "first run complete" or similar that we stash into the domain config, and until you run that the main server doesn't start? Then in tests we can have a way to force override that internally.

yaleman (approved):
Makes sense to me 😄

Firstyear:
@erictapen If you want me to do the enforced first-run setup process, then can you open that as a new issue? I think it'll take a bit more effort and thought to get it nice.

Issue #622, opened 2021-11-29
Make `kanidmd domain_name_change` cheap if the name stays unchanged


erictapen:
The context is that at https://github.com/NixOS/nixpkgs/pull/143134 @Flakebi and I are currently working on making kanidm easily deployable on NixOS. At NixOS we are very commited to declarative configuration. That goes smoothly for kanidm's TOML config, but not for `domain_name`, which [is considered database state](https://github.com/kanidm/kanidm/pull/617#issuecomment-978350062).

For cases like this, we usually make stuff configurable non-theless, but make the stateful nature of the change transparent to the end user. E.g. it is possible to automatically create databases for a Postgresql cluster by setting [`services.postgresql.ensureDatabases.mydb = [ "mydb ];`](https://search.nixos.org/options?channel=21.05&from=0&size=50&sort=relevance&type=packages&query=services.postgresql.ensureDatabases). Contrary to usual NixOS configuration, the database is not going to vanish if the line is removed from the config, therefore the "ensure" in the name.

For kanidmd we'd like to introduce an option `services.kanidm.ensureDomainName`, which would be only settable if `services.kanidm.config.settings.role != "ReadOnlyReplica"`. The easiest implementation would be to execute `kanidmd domain_name_change` at every service startup. But for that to work out, the implementation should just check if the name didn't change and then do nothing. The way I read the code that is currently not the case.

Alternatively you could provide a fast way to query the current domain_name via a kanidmd command? This way we could write a shell one-liner that executes the change only if the current name is different from the set one.

yaleman:
I've added prechecks in #624 for the domain name change thing - it wasn't checking beforehand 😄

erictapen:
Thanks!

Firstyear:
@erictapen No problem, I think this is a really good idea for it to be a stateful command and to be cheap on that change. I still think my only concern here would be when you have two replicas (in the future) if you haven't deployed this change to both then they will "fight" about what state the domain name will be in. But that's a risk no matter what we do I think ... I think this solves your issue and if we hit that problem, we'll address it then.

PR #624, opened 2021-11-29
Check before rename for #622

yaleman:
This checks you're actually changing the domain name and quits if it's not. Still pretty noisy from a logging perspective because it has to boot up the QueryServer, but it's quick (0.7s in my testing)

In the case of a non-change, you'll get:

    00000000-0000-0000-0000-000000000000 2021-11-29T23:13:56.450842+00:00 INFO     💬 [admin.info]: Domain name not changing, stopping.

And a return code of `125` (ECANCELED/Operation Cancelled) to note the change wasn't made.

- [x] cargo fmt has been run

### DIFF
--- a/kanidmd/src/lib/core/mod.rs
+++ b/kanidmd/src/lib/core/mod.rs
@@ -387,7 +387,7 @@ pub fn domain_rename_core(config: &Configuration, new_domain_name: &str) {
         Ok(old_domain_name) => {
             if &old_domain_name == &new_domain_name {
                 admin_info!("Domain name not changing, stopping.");
-                std::process::exit(2);
+                std::process::exit(125); // ECANCELED
             }
         }
         Err(e) => {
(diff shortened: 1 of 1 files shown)


kanidmd/src/lib/core/mod.rs
```diff
@@ -382,6 +382,20 @@ pub fn domain_rename_core(config: &Configuration, new_domain_name: &str) {
         }
     };
 
+    // make sure we're actually changing the domain name...
+    match task::block_on(qs.read_async()).get_domain_name() {
+        Ok(old_domain_name) => {
+            if &old_domain_name == &new_domain_name {
+                admin_info!("Domain name not changing, stopping.");
+                std::process::exit(125); // ECANCELED
```
Firstyear, line 390: I'm not sure about this exit here, just because we are in a transaction still at this point, and the return would at least cause that to be ended. Do we have an especial need to return a different exit code? I think for the purpose of this, if we are "changi" the name to the same name, it's still a "success" anyway ....

Firstyear (changes requested):

kanidmd/src/lib/core/mod.rs
Firstyear: Yeah, it's a little confusing to look at, but the qs.read_async() starts a txn, and it lasts for the duration of this block while you get the domain name.

PR #639, opened 2022-02-14
Change how domain names are handled in our configuration.

### LINKED ISSUE #623: Don't set default domain_name `example.com`
Continuing https://github.com/kanidm/kanidm/pull/617#issuecomment-978373115 here.

As I already laid out, I think it's an unintuitive situation that `domain_name` is currently set to `example.com` at first startup and then must be changed by the user. A server that is configured with `example.com` is unusable for every situation (except in integration tests), but pretends to be in a state where everything is configured properly.

From a user side the ideal solution imo would be:

When no `domain_name` is set yet, the server quickly fails, communicating what needs to be done to the user.

```
$ kanidmd server
Creating db...
There is currently no domain_name set. Run `kanidmd domain_name_change` to set one. Also see the docs here blabla
```

~~Also it would be nice to have `kanidmd domain_name_change` create the db at startup if it doesn't exist yet, so it would be optional to run `kanidmd server` first.~~

Firstyear:
Fixes #623 - Improve domain name handling on server startup. Rather than starting with a default domain and expecting a rename, we make the domain name a config value in server.toml. When the server starts up, we use this initially to create the first domain name. When a domain rename is requested, we read the value from server.toml (and skip the origin check by not starting the idms layer).  Finally, this adds a config tester that can check if the server "would start" with this configuration. As always, book chapters have been updated too :) 

### DIFF
--- a/kanidm_book/src/administrivia.md
+++ b/kanidm_book/src/administrivia.md
@@ -48,30 +48,19 @@ by including the `[online_backup]` section in the `server.toml`.
 This allows you to run regular backups, defined by a cron schedule, and maintain
 the number of backup versions to keep. An example is located in [examples/server.toml](../../examples/server.toml).
 
-# Rename the domain
+# Configuration Test
 
-There are some cases where you may need to rename the domain. You should have configured
-this initially in the setup, however you may have a situation where a business is changing
-name, merging, or other needs which may prompt this needing to be changed.
+You can test your configuration will correctly start with the server.
 
-> **WARNING:** This WILL break ALL u2f/webauthn tokens that have been enrolled, which MAY cause
-> accounts to be locked out and unrecoverable until further action is taken. DO NOT CHANGE
-> the `domain_name` unless REQUIRED and have a plan on how to manage these issues.
-
-> **WARNING:** This operation can take an extensive amount of time as ALL accounts and groups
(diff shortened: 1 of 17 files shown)


Firstyear:
I think the only think missing right now, is a check that domain in db == server config domain at startup. I'm going to think where to put this :)

yaleman:
All looks sensible here, especially once the "compare db to domain name" bit is in 😄 

it'll happily start/pass configtest regardless of what's in the various options, and if I use kanidm.example.com.au while origin is kanidm.example.com - which ... would technically ***start***, but not ***work***

Firstyear:
@yaleman I changed it a bit, it won't fail to start the server if the config mismatches, but it will always prefer a db domain name over the config one, and will warn in the case you have it mis-configured. 

```
+00:00 WARN     🚧 [admin.warn]: Using database configured domain name idm.example.com - was example.com
+00:00 WARN     🚧 [admin.warn]: If you think this is an error, see https://kanidm.github.io/kanidm/administrivia.html#rename-the-domain
```

Otherwise I'm pretty happy with it now.

Firstyear:
Okay, deployed and running this on my instance now, seem happy. Any objections to merging?

yaleman (approved):
I see no issue with the changes as proposed, with the testing I've done.

PR #643, opened 2022-03-06
106 auth concurrency

### LINKED ISSUE #106: Improve auth concurrency
At the moment in the IDM server, the session set is a single thread behind a mutex to uphold certain properties with regard to account locking. To improve performance here, we could make auth-init only need the single thread, which inits a struct keyed on credential id. Then when the auth session is continued, the tree is mvcc and we take an inner mutex on the credential session set, find our session and continue from there.

It would make session cache purging more complex, but it would allow multiple authentications to proceed in parallel without losing the serialisation of per-user auth attempts to maintain coherent account lockouts.

Firstyear:
Fixes #106 - Implements concurrency in authentication. This exposed a few minor issues in how we were handling async in auth functions, which will be important later once we implement remote-auth for trusts and read-only systems. This will improve auth throughput in high load situations, while retaining all the previous guarantees of softlocking over credentials. 

This was done in preparation for multi-credential auth and credential reset workflows to help me get a refresh of how this all fits together ^^; 


kanidmd/src/lib/config.rs
```diff
@@ -135,7 +135,9 @@ impl Configuration {
         let mut c = Configuration {
             address: String::from("127.0.0.1:8080"),
             ldapaddress: None,
-            threads: num_cpus::get(),
+            threads: std::thread::available_parallelism()
+                .expect("Unable to read number of available CPUs")
```
yaleman, line 139: Maybe fallback to 1 when it can't read it? 😄

kanidmd/src/lib/lib.rs
```diff
@@ -2,7 +2,7 @@
 //! which is used to process authentication, store identities and enforce access controls.
 
 #![recursion_limit = "512"]
-#![deny(warnings)]
+// #![deny(warnings)]
```
yaleman, line 5: 📎 😢

yaleman (commented):
seems legit, I'll test it if this version's ready to go?

yaleman (approved):
Works in prod as far as I can tell 😄

Issue #383, opened 2021-03-20
Allow credentials to be associated to a device


Firstyear:
In order to make it easier to activate or deactivate specific devices and their related credentials, it would be good to be able to allow users to associate some classes of credentials to their devices. This means that we'll end up with:

* Floating credentials (ip yubikey 5ci, the primary account password, the single TOTP device, radius pw)
* Device bound credentials (touch id, yubikey nano, application passwords, )

This would help in a situation such a lost or compromised device, allowing temporary or permanent revocation of the associated credentials, and allowing the user to consider rotation of the accounts floating credentials.

Firstyear:
IMO given what we have now with passkeys, and our direction heading to passwordless, I think this is complete.

Issue #170, opened 2020-01-20
Account recovery for unavailable users.


Firstyear:
At LCA it was discussed that we should have recovery workflows for:

* (1 device + webauthn) Lost MFA device, password retained
* Person not available (loss of device, webauthn and password). IE need to recovery an account on someones behalf

Issue #164, opened 2020-01-03
Multi device enrollment workflow


Firstyear:
We should have a clear workflow where between two devices using webauthn, we can enroll other devices. This is important as firefox softtokens, touchid, android key etc can't be attached to an existing device to allow the login. we don't want to use backup codes, those are the wrong mechanism here.

Instead we should have a workflow that has a private-totp on the account, and when a user logs in they can use the private-totp to do a webauthn enrollment. The totp is shown on an existing device, and entered on the new device.

Firstyear:
Mostly done, we just need a way to display the qrcode to the user so they can do the pw-reset flow on a new device or enter the code.

PR #649, opened 2022-03-12
Add design doc, revive the domain wide enc token

Firstyear:
Prep work for #383 #170 #164. This revives our domain wide encryption token, adds a design doc, removes bundy in favour of our compact jwt lib since they are almost identical, and cleans up some other docs.

### DIFF
--- a/designs/credential-update.rst
+++ b/designs/credential-update.rst
@@ -64,6 +64,14 @@ The credential update session can be rejected by the user, canceling it's progre
 
 A credential update session IS tied to a single server, similar to authentication.
 
+A credential update session has a unique-uuid assigned. When committed, this is added to a history-log
+on the users account so they can see and audit the change.
+
+If an intent token is used, the uuid of the intent token becomes the uuid of the credential update session.
+
+If an intent token is exchanged, and it's uuid already exists in the history-log of the account, the
+intent token is rejected.
+
 Credential Update Process
 =========================
 
--- a/designs/credential-update.rst
+++ b/designs/credential-update.rst
@@ -1,0 +1,154 @@
+
+Credential Update and Onboarding Workflow
+-----------------------------------------
+
+Letting users update their own credentials, but also allowing new users to create their own credentials
+needs a new workflow. Since these are nearly identical, these processes can be combined.
+
+Major considerations are:
+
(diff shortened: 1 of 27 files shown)


designs/credential-update.rst
```diff
@@ -0,0 +1,154 @@
...
+Consistency
+^^^^^^^^^^^
+
+Only one credential update process may exist at a time for a user. Attempts to create a second
+credential update session while an existing session exists is rejected until the previous session
+is committed or timed out.
+
+The credential update session can be rejected by the user, canceling it's progress.
+
+A credential update session IS tied to a single server, similar to authentication.
```
yaleman, line 65: This feels like a footgun, unless it'll be solved in the future - it'll cause issues for two cases I can think of:

 - load balanced IDM servers, or 
 - a session being created by an "internal" admin, for a user who's on the "public" network

Firstyear: these two are interesting cases.

> load balanced IDM servers, or

This is already an issue with authentication, because we need to prevent classes of replay attacks and such. The auth-cred-change is a bit less prone to that attack flow, but for consistency it still makes more sense to keep this locked to a single state. We can't use / rely on backend sync due to eventual consistency. So we currently do already expect a load balancer to perform sticky sessions, which honestly is pretty widely required. Even LDAP requires it .... so for now, I'm happy to leave us requiring single-server contact and sticky sessions, and if it becomes an issue later, we should confront it at that time. 

> a session being created by an "internal" admin, for a user who's on the "public" network

This is already (mostly) solved! In the design the admin creates the intent-token. The intent token is currently "stateless" and then when that's given to the user, that intent token is exchanged for the credential update session. Because the user was the one that exchanged the intent -> credupdatesession that's done on the public network, so the credupdatesession will be on the public side and the user can proceed without issue. 

The main issue there is "what happens if the intent token is used multiple times?" Which is something I need to think about and factor into the design. I would like it to be that it can be exchanged multiple times for credupdatessessions, but only until *at least* one update session succeeds. After that one succeeds, it's now 'invalid'. So I have to think about how to achieve that nicely. It's not a super big deal, but consider we are on boarding someone, that's basically a 24 hour "change your credentials" link.

Firstyear: Updated docs to address the intent token case.

PR #653, opened 2022-03-27
setting default log level of INFO when --debug is not set

### LINKED ISSUE #652: kanidm group list does not show results by default
We're using `info!()` to print the outputs but ... info-level messages aren't being printed?

yaleman:
fixes #652 - previously when things like `info!("blah")` are run nothing's output by default, ref @cuberoot74088's [comment here](https://github.com/kanidm/kanidm/pull/591#issuecomment-1079779727)

- [x] cargo fmt has been run

### DIFF
--- a/kanidm_tools/src/cli/main.rs
+++ b/kanidm_tools/src/cli/main.rs
@@ -21,6 +21,8 @@ fn main() {
             "RUST_LOG",
             "kanidm=debug,kanidm_client=debug,webauthn=debug",
         );
+    } else {
+        ::std::env::set_var("RUST_LOG", "kanidm=INFO,kanidm_client=INFO,webauthn=INFO");
     }
     tracing_subscriber::fmt::init();
 


kanidm_tools/src/cli/main.rs
```diff
@@ -21,6 +21,8 @@ fn main() {
             "RUST_LOG",
             "kanidm=debug,kanidm_client=debug,webauthn=debug",
         );
+    } else {
+        ::std::env::set_var("RUST_LOG", "kanidm=INFO,kanidm_client=INFO,webauthn=INFO");
```
Firstyear, line 25: Since we have the tracing bits, I think I might rework this if that's okay?

yaleman: Of course 😄

cuberoot74088: Hi @firstyear,

I do not know what you mean exactly.
 Are you suggesting to not put standard command output into a log level?

This is what I am wondering right now. Imagine some other cli tool, maybe even ls, I don't want to see timestamps and other potentially noisy stuff, until I ask for it. 
As you mention tracing, this is probably the right place for this type of information. But for a list of group members or groups queried by a user, I personally would like to get just this back, like in alpha6 and for other cli commands in the current version.

Let me know if I can help out somewhere.

Firstyear: @cuberoot74088 No, I want to change how this if/else works, as there is a way in tracing to set default level filters that is a bit different.

Firstyear:
@yaleman Im thinking maybe we also make it so that any output we want the user to see is println, where anything logging/tracing is stderr + tracing? I dunno why past us did it like this ....

yaleman:
Probably because I assume that info-level things will be user-facing messages...?

Firstyear:
Superceded by https://github.com/kanidm/kanidm/pull/659

PR #664, opened 2022-03-31
Specifying MSRV in package now

yaleman:
Instead of having specific build scripting for it, now we're using the [post-rust-1.56 ability to specify MSRV in `Cargo.toml`](https://doc.rust-lang.org/cargo/reference/manifest.html#the-rust-version-field)

- [x] cargo fmt has been run

Updated docs too.


Firstyear (approved):
Can't we also delete profiles/RUST_MSRV too?

yaleman:
I thought I had, but evidently not. Sorted now :smile:

PR #665, opened 2022-04-01
Use pkg-config to link against pam

euank:
Some distros, such as nixos, require more than just '-lpam' to locate
the library. Adding a naive pkg-config invocation to the pam FFI
crate allows pam_kanidm to build on my system, where before this change it did not.

- [x] cargo fmt has been run
- [ ] ~book chapter included (if relevant)~ not relevant
- [ ] ~design document included (if relevant)~ not relevant

### DIFF
--- a/kanidm_unix_int/pam_kanidm/build.rs
+++ b/kanidm_unix_int/pam_kanidm/build.rs
@@ -1,0 +1,8 @@
+fn main() {
+    // ignore errors here since older versions of pam do not ship the pkg-config `pam.pc` file.
+    // Not setting anything here will fall back on just blindly linking with `-lpam`,
+    // which will work on environments with libpam.so, but no pkg-config file.
+    let _ = pkg_config::Config::new()
+        .atleast_version("1.3.0")
+        .probe("pam");
+}
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -14,9 +14,9 @@
 * Samuel Cabrero (scabrero)
 * Victor Wai (vcwai)
 * James Hodgkinson (yaleman)
+* Euan Kemp (euank)
 
 ## Acknowledgements
 
 * M. Gerstner
 * Perri Boulton
-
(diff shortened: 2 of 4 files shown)


kanidm_unix_int/pam_kanidm/build.rs
```diff
@@ -0,0 +1,6 @@
+fn main() {
+    pkg_config::Config::new()
+        .atleast_version("1.3.0")
```
euank, line 3: The minimum pam version was chosen somewhat arbitrarily.

I checked packaged versions from https://repology.org/project/pam/versions and just picked the number openSUSE has under the assumption that's the version that the code's been tested with.

yaleman: Checks out from deb-ish land:

- Debian 10: 1.3.1
- Ubuntu 21.10: 1.3.1

Firstyear:
@euank It looks like the CI failed, but maybe that's a fault of our CI env?

euank:
Digging a little more, it looks like libpam only recently started including pkg-config files (https://github.com/linux-pam/linux-pam/pull/369).
The ubuntu `libpam-dev` package we install doesn't include one.

Maybe the right answer is to try pkg-config, but default to just spitting out `-lpam` and hoping?
I'll update this PR to that, and I expect that'll make CI happy at least.

Firstyear:
Yeah I think that sounds like the best option :)

euank:
What I've got now works with my NixOS setup, and with CI, so I think this is a reasonable approach to it

Firstyear:
@euank Looks good! Can you also add yourself to https://github.com/kanidm/kanidm/blob/master/CONTRIBUTORS.md then I'll merge :)

Firstyear:
Thank you very much @euank ! Really appreciate your help and work :)

euank:
Thanks for accommodating my possibly slightly wonky distro setup I use, and of course for the project as a whole :)

Firstyear:
No no, the more places that we test and run on, the better we become! So please continue to raise issues, give feedback and help out.

Issue #648, opened 2022-03-11
Self-service 'person' attribute modification


euank:
Currently, the canned access control policies don't include one for self-modification of your own email attribute.

The closest we have is this one:

https://github.com/kanidm/kanidm/blob/58fb559262163d9fcd748a8dda0e53ae41b9f266/kanidmd/src/lib/constants/acp.rs#L117-L122

I think I want to have self-service email modification, which naturally leads me to the questions:

1. Should there be a policy, perhaps `idm_self_acp_extend_person_mail`, to allow this?
2. Is there currently a way to create that without compiling it into the server binary?

Grepping around, I don't see an existing api (yet, I assume it'll be implemented at some point?), so I assume I just have to bake in any policies I want for now; is that understanding correct?

Thanks for the project btw, it's been great to play around with so far!

---

(new title, more issue description)

Self-service editing any person attribute via `kanidm account person set <self>` appears to not work.

See the first several comments below for more info on that.

yaleman:
> 2. Is there currently a way to create that without compiling it into the server binary?

Not currently, I think @Firstyear has some ideas about implementation.

Firstyear:
ACP allow modification, but it has to be done manually right now. Currently you can do this with:

```
kanidm login -D admin
kanidm raw modify '{"eq": ["uuid", "00000000-0000-0000-0000-ffffff000021"]}' changes.json
```

Where changes.json contains:

```
[
    { "present": ["acp_modify_removedattr", "mail"] },
    { "present": ["acp_modify_presentattr", "mail"] }
]
```

This will change the current definition of the self write access control to extend it to allow self write of mail. This IS upgrade safe, where during a migration we will NOT remove custom added attributes like this, we only assert the attributes that we supply are present. 

In the future we want to make access controls easier to supply and edit, but I'm considering an improvement to how they are currently definined, but I also plan to do on-upgrade migration so hopefully your use case will continue to work :) 


Hope that helps!

Firstyear:
Oh you can also show the current set of access controls that exist with:

```
kanidm raw search '{"eq": ["class", "access_control_profile"]}'
```

You'll need to be logged in as admin to see these as well.

Firstyear:
Hey @euank did you get a chance to try this out?

euank:
Thanks a ton for the help, and thanks for following up!

I ran the above, but I ended up hitting an issue (and only just now got back to digging into it enough to give more info).

[...]

PR #667, opened 2022-04-02
Add 'account person set' command

euank:
This command allows a user to modify, say, their legal name in a
self-service fashion.

This wasn't possible before by default since the 'extend' operation
required additional ACPs in order to operate which not every user would
have, specifically the ability to add a "class" of "person" to the account.

The new "person set" api is compatible with the default self_write ACP,
and so allows self-service modification.

See #648 for more info.

I didn't tackle improving the `account` vs `self` splitup of commands yet. I think I can reasonably do that as a separate PR, and that'll make things easier for me, and for reviewers. I would like to come back to tackle that in a followup.

Manual testing done:

```
$ kanidm account create test6 "user" -D idm_admin
$ kanidm account get test6 -D idm_admin
class: account
class: object
displayname: user
name: test6
spn: test6@example.com
uuid: b02fd2f3-36ec-4ed0-aa27-e7755d451731

$ kanidm account person set test6 --legalname "test" -D test6
2022-04-01T23:52:27.859834Z ERROR kanidm_cli::account: Error -> Http(400, Some(SchemaViolation(InvalidAttribute("legalname"))), "5f40e0f8-4253-40eb-9dd3-b4fbf403f46c")

$ kanidm account person extend test6 --legalname "initial name" -D idm_admin
$ kanidm account get test6 -D test6
class: account
class: object
class: person
displayname: user
legalname: initial name
name: test6
primary_credential: primary
spn: test6@example.com
uuid: b02fd2f3-36ec-4ed0-aa27-e7755d451731

$ kanidm account person set test6 --legalname "updated name" -D test6
$ kanidm account get test6 -D test6
class: account
class: object
class: person
displayname: user
legalname: updated name
name: test6
primary_credential: primary
spn: test6@example.com
uuid: b02fd2f3-36ec-4ed0-aa27-e7755d451731
```

- [x] cargo fmt has been run
- [ ] ~book chapter included (if relevant)~ I added a section, not a full chapter
- [ ] ~design document included (if relevant)~ Probably not relevant

### DIFF
--- a/kanidmd/src/lib/actors/v1_write.rs
+++ b/kanidmd/src/lib/actors/v1_write.rs
@@ -28,7 +28,7 @@ use kanidm_proto::v1::Entry as ProtoEntry;
 use kanidm_proto::v1::Modify as ProtoModify;
 use kanidm_proto::v1::ModifyList as ProtoModifyList;
 use kanidm_proto::v1::{
-    AccountPersonExtend, AccountUnixExtend, CreateRequest, DeleteRequest, GroupUnixExtend,
+    AccountPersonSet, AccountUnixExtend, CreateRequest, DeleteRequest, GroupUnixExtend,
     ModifyRequest, SetCredentialRequest, SetCredentialResponse,
 };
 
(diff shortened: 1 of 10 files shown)


kanidmd/src/lib/actors/v1_write.rs
```diff
@@ -1011,6 +1009,61 @@ impl QueryServerWriteV1 {
         res
     }
 
+    #[instrument(
+        level = "trace",
+        name = "idmaccountpersonset",
+        skip(self, uat, uuid_or_name, eventid)
+        fields(uuid = ?eventid)
+    )]
+    pub async fn handle_idmaccountpersonset(
```
euank, line 1018: The duplication between this and `idmaccountpersonextend` doesn't feel great, but it also seems sorta reasonable.

The idea of an admin initializing an account to a "person" with some initial values makes a lot of sense, and it also makes sense there'd be some overlap with "initialize a person" and "update a person".

If you have an idea for how to better structure this code to make it feel less duplicated, I'm happy to hear ideas!

Firstyear: I don't today. But I think once we start to work on scim it might become clear how to make it better :)

Firstyear (approved):

PR #672, opened 2022-04-12
Add a new ACP and group allowing self-service mail updates

euank:
This adds a new "idm_people_self_write_mail_priv" group which follows
the existing canned group+acp format closely.

This also adds a test for the functionality

See the discussion in #648 for a bit more background

- [x] cargo fmt has been run
- [ ] ~book chapter included (if relevant)~ | minor doc section added where "people" accounts are mentioned
- [ ] ~design document included (if relevant)~

------

As some meta-commentary, the acp/group code was straightforward enough to make this change feel pretty easy, woo! It's very clearly meant as an intermediate state, but even so it wasn't hard to work with.

The `_UUID` consts, presumably to reserve IDs, did feel a little weird. They weren't actually referenced anywhere, like inline in ACP poicies (rather the UUID itself was copy-pasted), and they don't actually validate uniqueness on their own, other than by having one file to search through.

I'm curious, is the intent just a place to grep for IDs in? Is there a different plan for those consts eventually?


kanidmd/src/lib/constants/acp.rs
```diff
@@ -126,6 +126,29 @@ pub const JSON_IDM_SELF_ACP_WRITE_V1: &str = r#"{
...
+pub const JSON_IDM_PEOPLE_SELF_ACP_WRITE_MAIL_V1: &str = r#"{
+    "attrs": {
+        "class": ["object", "access_control_profile", "access_control_modify"],
+        "name": ["idm_people_self_acp_write_mail"],
+        "uuid": ["00000000-0000-0000-0000-ffffff000041"],
+        "description": ["Builtin IDM Control for self write of mail for people accounts."],
+        "acp_receiver": [
+            "{\"eq\":[\"memberof\",\"00000000-0000-0000-0000-000000000033\"]}"
+        ],
+        "acp_targetscope": [
```
yaleman, line 138: Should this duplicate `acp_targetscope` be here? 🤔

yaleman (changes requested):

kanidmd/src/lib/constants/acp.rs
Firstyear: No, it shouldn't but the second one will over-ride it.

kanidmd/src/lib/constants/acp.rs
```diff
@@ -126,6 +126,29 @@ pub const JSON_IDM_SELF_ACP_WRITE_V1: &str = r#"{
...
+        ],
+        "acp_targetscope": [
+            "{\"and\": [{\"eq\": [\"class\",\"person\"]}, \"self\"]}"
```
Firstyear, line 141: I think it should be :

```
"{\"and\": [{\"eq\": [\"class\",\"person\"]}, {\"eq\": [\"class\",\"account\"]} , \"self\"]}"
```

Since a person can exist as a phone-book style entry, but can't authenticate, so only accounts (can auth) and people (are human) can update mail.

kanidm_book/src/accounts_and_groups.md
```diff
@@ -141,6 +141,16 @@ Whether an account is currently a "person" or not can be identified from the "ac
 The presence of a "class: person" stanza indicates that this account may have
 "people" attributes.
 
+### Allowing people accounts to change their mail
+
+By default, Kanidm only allows a "person" account to change their legal name,
```
Firstyear, line 146: Kanidm allow a person to change their legal name, display name, username .... :)

Firstyear (changes requested):

Firstyear:
> As some meta-commentary, the acp/group code was straightforward enough to make this change feel pretty easy, woo! It's very clearly meant as an intermediate state, but even so it wasn't hard to work with.

Yep, it's modeled off of other access control systems :) 

In the future I want to change these around a bit so that these acp rules are "nicer" for the human, but then the actual access is generated out onto the actual entries that recieve because then it's faster to lookup and easier to audit. But right now other things have taken priority :) 

> 
> The `_UUID` consts, presumably to reserve IDs, did feel a little weird. They weren't actually referenced anywhere, like inline in ACP poicies (rather the UUID itself was copy-pasted), and they don't actually validate uniqueness on their own, other than by having one file to search through.

That's pretty much what it is today is a registry of what we allocated. 

> 
> I'm curious, is the intent just a place to grep for IDs in? Is there a different plan for those consts eventually?

Yes, there is a plan for me to fix this up so you can use the group names rather than the consts in there. 


This is pretty much why I haven't exposed a lot of this to the user yet, is I want to rework these internals quite a bit and don't want people to add too many customised access controls today. :)

kanidmd/src/lib/constants/acp.rs
euank: It's a copy-paste error, thanks for catching that @yaleman

Fixed, removed the empty one.

kanidm_book/src/accounts_and_groups.md
euank: I was just thinking of person-type attributes, which I thought were just mail and legalname, while display name, username, etc are account attributes (iiuc?)

You're right that the wording's bad. I updated it to not mention legal name specifically, and just call out that mail isn't in the default write set.

kanidmd/src/lib/constants/acp.rs
euank: Sure, updated to that. That makes some sense.

It does seem like it might be inconsistent with the ACP above though, `idm_self_acp_write`. That one has `acp_targetscope: ["self"]`, and allows modifying `legalname`, among other things, with no requirement of `class: "account"` (or not that I see anyway?)

With my current understanding, I also don't see how anything but an account could auth and thus be subject to this ACP.

Is this change to prevent some actual thing now, or does it fit the model better, but in practice not have an observable difference now?
If these phone-book style "people" entries existed (or do they?), would the ACP be able to apply to them in some way?

euank:
Thanks for the answers/clarification, appreciated!

> I want to rework these internals quite a bit and don't want people to add too many customised access controls today. :)

Don't mind me just sneaking this one in... :wink:

------

Anyway, I think I've addressed the review comments, thanks for the review!

Firstyear (approved):

Firstyear:
No, thank you for contributing!

kanidmd/src/lib/constants/acp.rs
Firstyear: Actually, we should fix that too then 😅

euank: Sure, done, pushed as an additional commit to this branch.

It would be nice to write a unit test too if it's possible to create a non-account-class "self" that can also call the API.. which I guess is what I'm asking above already. I'm happy to wait for an answer there, or do more digging through code/docs if you've got pointers.

Firstyear: I don't think you can, not without some of the internal server magic methods. The server internally has the ability to impersonate *any* entry of the system, which allows testing some of these properties, but they aren't exposed (yet) to the API as the security properties and limits on that need clearer definition and design. There are already self write tests, so lets just leave them as is :)

Firstyear:
@yaleman You need to also do an approve (unless I want to be naughty and bypass you :P )

Firstyear:
Don't worry about the test failure, that was my fault, so I'll merge it and fix later :)

yaleman (approved):

yaleman:
Push da buttonz @Firstyear

Firstyear:
Thank you so much @euank !

PR #674, opened 2022-04-14
Resolve test failure with radius

Firstyear:
Resolve the test case failure that I introduced yesterday with @euank. 


euank (approved):
LGTM!

I wasn't totally confident that the goal was that only "people" accounts could self-service change their display name in my change, but from your fixup and this change, it sounds like that's the idea.

That model makes sense to me, and this test change definitely makes sense in light of that!

Firstyear:
> LGTM!
> 
> I wasn't totally confident that the goal was that only "people" accounts could self-service change their display name in my change, but from your fixup and this change, it sounds like that's the idea.
> 
> That model makes sense to me, and this test change definitely makes sense in light of that!

Yep! When you think about it, imagine a service account - that shouldn't be able to self change things because it should be administered. I should probably define that a bit better in the future ....

PR #675, opened 2022-04-14
webUI "this is alpha" message

### LINKED ISSUE #673: WebUI / Leap 15.3 / kanidm_tools 1.1.0-alpha.7
I installed kanidm on Leap 15.3 and it seems to work by cmdline.

When i login into webui either with admin or with idm_admin, the page looks strange. I guess the Page should contain data, but i do only see "section title" and a table containing two rows having (with header) like:

"1,001 | random | data | placeholder | text | -- | -- | -- | -- | --"

on the left side, there is "Account" selectable, but when clicked, not really much happens.

i checked user demo_group and the data does not seem to be related.

Maybe some data missing, so the UI does not work like this?

yaleman:
Fixes #673 

- [x] cargo fmt has been run


yaleman:
screenshot for reference ^

kanidmd_web_ui/pkg/LICENSE.md
```diff
@@ -0,0 +1,374 @@
+Mozilla Public License Version 2.0
```
Firstyear, line 1: We have the license at the project root? Is there a reason to duplicate it here?

kanidmd_web_ui/pkg/README.md
```diff
@@ -0,0 +1,101 @@
+
```
Firstyear, line 1: Similar, is there a reason to duplicate this?

Firstyear (changes requested):
Otherwise looks good, thank you!

kanidmd_web_ui/pkg/LICENSE.md
yaleman: It makes wasm-bindgen happier

kanidmd_web_ui/pkg/README.md
yaleman: see above, fixing whinge from wasm

Firstyear: It can't use relative paths?

yaleman: Not as far as this (and some other searching) would indicate... https://github.com/rustwasm/wasm-pack/issues/718

Firstyear: Cool cool.

Firstyear (approved):

kanidmd_web_ui/pkg/LICENSE.md
yaleman: found a license-file setting, which I put into 7b14dbe2 ... which just copies the file locally instead of referring to it 🙄

Issue #728, opened 2022-04-30
Orca documentation/examples


yaleman:
Plz can haz? Even just a simple "this command makes it go" because... the structopt things are confusing and I want to see if I'm making kani slow 😅

Firstyear:
It's not so simple, orca "takes over" your database and assumes it can reset it and write arbitrary content. 

If you want a simple crash course, we could make a readme for the tool, but I'd rather leave it a bit quiet atm given it's still a work in progress.

Anyway, orca works in a few steps.

1. (Optional) preprocess an anonymised 389-ds access log (created from an external tool) into an orca data set.
2. Create an orca config which defines the targets you want to be able to setup and load test. See example_profiles/small/orca.toml
3. setup the kanidm/389-ds instance from the orca data set. You can see an example of this in example_profiles/small/data.json. This will reset the database, and add tons of entries etc. 
4. Run one of the metrics, based on that data set. For example `orca run -p  example_profiles/small/orca.toml [ds|kanidm|kanidm_ldap] search-basic

Currently the only metric we have is search-basic, which basicly just checks overall ability to do searches. I planned to add other tests based on write performance and different kinds of metrics.

yaleman:
Sweet, this is entirely for speed-testing dev, so being able to blow up the DB's fine. 😄 I'll give that a go, thanks 😄

Issue #644, opened 2022-03-06
Investigate OpenTelemetry for metrics


yaleman:
https://docs.rs/opentelemetry/latest/opentelemetry/

expands on #221

Firstyear:
It looks like https://github.com/tokio-rs/console/issues/69 will probably be a better fit for what we need :)

yaleman:
Ah, I think one feeds into the other, from my understanding - OT's about a single format for collating (not necessarily collecting) and exposing things

Firstyear:
The other one I found while looking is dhat-rs and slab. So I think for now I'll use those since they seem closer to what I want for understanding our memory usage in dev vs runtime.

Issue #216, opened 2020-05-03
Healthcheck endpoint


pando85:
We could have a simple endpoint that check kanindmd status and return 200. I like `/v1/ping` which return 200 with a body of `pong` but anything could be useful.

Firstyear:
https://openmetrics.io/ was suggested by @pando85 as a standard we could consider following, so we should look at this for status.

yaleman:
For people looking for this in the future, there's a simple health check at `/status` added in #221

Firstyear:
I think this issue now is more about extending the endpoint to internally verify more parts.

matthew-salerno:
Just gonna through my opinion here since this is still in dev. I just read over openmetrics. I like it quite a bit. Not a fan at how hard it is to read, but it makes sense under their design constraints for high throughput of metrics. That being said, idk how well it lines up with our use case here. I imagine for detailed info people would use the logs and maybe an aggregator. We also probably don't want to expose too much over http. So all I can think of is an alive status, the configured uri, maybe a version number? At that low of a data volume we might as well use json, yaml, or ron, and then at least it will be easy to curl.

yaleman:
If you're monitoring the metrics of an IdP, then ingesting prometheus metrics (which is the base implementation of OpenMetrics) should be trivial - it *can* be a curl request and it gives you a list of key-value pairs of `metric value`, ie `memoryUsed 123456778`.

In the current alpha stage there's less to monitor, but as soon as any replication's occurring then there's plenty of things to capture for monitoring purposes that would be handy to expose for specific monitoring outside logs.

Don't get me wrong, there's plenty of ways to achieve monitoring through logs - that's literally my day job 😇 - but from an ops perspective, systems collating and exposing their metrics is super handy.

Firstyear:
well, there is also stuff related to cache behaviour and efficiency too that we could be capturing, but I just haven't exposed all of that either ... 

But yes, thanks for your comments @matthew-salerno, I think early on I'd probably just make a json end point, but as @yaleman points out, once we have more systems in a large topology it becomes a bit more interesting. 

We may even wanth to start to track stats of how many unique operations are occuring as well.

yaleman:
[...]

PR #729, opened 2022-04-30
Adding metrics listener

yaleman:
Ref #644, Ref #216 

Adds an optional metrics listener which exposes tide request counts, but is (I think?) Pretty easily extensible.

I've added metrics to the QueryServer behind a feature flag because they might trash performance - once #728 has some details I'll do more testing (and add automated reporting for said testing).

- [x] cargo fmt has been run

### DIFF
--- a/kanidmd/score/Cargo.toml
+++ b/kanidmd/score/Cargo.toml
@@ -39,7 +39,7 @@ async-std = { version = "^1.11.0", features = ["tokio1"] }
 compact_jwt = "^0.2.1"
 
 # used for metrics
-tide-prometheus = { version = "0.1.0", features = ["process"] }
+prometheus-client = "0.15.1"
 
 [build-dependencies]
 profiles = { path = "../../profiles" }
--- a/kanidmd/score/Cargo.toml
+++ b/kanidmd/score/Cargo.toml
@@ -1,7 +1,10 @@
 [package]
 name = "score"
 version = "0.1.0"
-authors = ["William Brown <william@blackhats.net.au>"]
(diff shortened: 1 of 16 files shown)


kanidmd/score/Cargo.toml
```diff
@@ -1,7 +1,10 @@
 [package]
 name = "score"
 version = "0.1.0"
-authors = ["William Brown <william@blackhats.net.au>"]
+authors = [
+    "William Brown <william@blackhats.net.au>",
+    "James Hodgkinson <james@terminaloutcomes.com>",
```
Firstyear, line 6: ❤️

kanidmd/score/src/lib.rs
```diff
@@ -686,6 +687,13 @@ pub async fn create_server_core(config: Configuration, config_test: bool) -> Res
     if config_test {
         admin_info!("this config rocks! 🪨 ");
     } else {
+        let enable_metrics = match &config.metrics_listener {
+            Some(_) => {
```
Firstyear, line 691: I think you could do "is_some()" instead.

Firstyear (changes requested):
Minor comment, looks really good. Is there a way we can add authentication to the metric endpoint?

kanidmd/score/src/lib.rs
yaleman: Oh neat, have done this! 👍🏻

yaleman:
bailing on this, turns out the original plan was hilariously wrong and I'm going to start again 🙄

PR #737, opened 2022-05-02
Orca output dir fix

yaleman:
Fixes an issue where if the output dir doesn't exist, orca was bailing - it now prompts the user to create the dir and does so if it can.

Screen Shot 2022-05-02 at 11 59 13

- [x] cargo fmt has been run

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2529,6 +2529,7 @@ version = "1.1.0-alpha.8"
 dependencies = [
  "crossbeam",
  "csv",
+ "dialoguer",
  "futures-util",
  "jemallocator",
  "kanidm_client",
--- a/kanidmd/daemon/run_insecure_dev_server.sh
+++ b/kanidmd/daemon/run_insecure_dev_server.sh
@@ -19,4 +19,10 @@ if [ ! -f "/tmp/kanidm/key.pem" ]; then
     exit 1
 fi
 
-cargo run --bin kanidmd server -c "${CONFIG_FILE}"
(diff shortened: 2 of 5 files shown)


Firstyear:
It could also be good to take the "workflow" I dumped in the orca issue into the help for the cli. The doc strings are based on the structopt comments I think.

Firstyear:
Did you want to add a quick set of notes to this for the --help on orca, or should I?

yaleman:
```
orca 1.1.0-alpha.8
Orca Load Testing Utility

Orca works in a few steps.

1. Create an orca config which defines the targets you want to be able to setup and load test. See
example_profiles/small/orca.toml

2. (Optional) preprocess an anonymised 389-ds access log (created from an external tool) into an orca data set.

3. 'orca setup' the kanidm/389-ds instance from the orca data set. You can see an example of this in
example_profiles/small/data.json. This will reset the database, and add tons of entries etc. For example:

    orca setup kanidm -p ./example_profiles/small/orca.toml

4. 'orca run' one of the metrics, based on that data set. For example:

    orca run -p example_profiles/small/orca.toml kanidm search-basic


USAGE:
    orca <SUBCOMMAND>

FLAGS:
    -h, --help       Prints help information
    -V, --version    Prints version information

SUBCOMMANDS:
    help          Prints this message or the help of the given subcommand(s)
    preprocess    Preprocess a dataset that can be used for testing
    run           Run the load test as defined by the test profile
    setup         Setup a server as defined by a test profile
```

PR #739, opened 2022-05-04
383 170 164 authentication updates 4

Firstyear:
Almost there! This adds the CLI repl, with support for passwords, backup codes, and totp. Next I need to do webauthn, but that will require it's own block of work, so I wanted to get this in :) 

Note: we will need to rewrite the book chapter on credentials here because this will be deleting the old pw/totp management commands!


yaleman (approved):
woot!

PR #740, opened 2022-05-05
Adding Content-Security-Policy Headers and auto-generating integrity hashes

yaleman:
- adds CSP headers to /ui/ URIs, also created the favicon and WASM loader as their own files

This makes some security checks happier.

- [x] cargo fmt has been run


kanidmd_web_ui/pkg/wasmloader.js
```diff
@@ -0,0 +1,7 @@
+// loads the module which loads the WASM. It's loaders all the way down.
+import init, { run_app } from '/pkg/kanidmd_web_ui.js';
```
Firstyear, line 2: Makes me wonder if we can do a csp hash on this .... 🤔

Firstyear (changes requested):
One maybe-change,  Ithink you can make your test case a bit smaller by using the common module :)

Firstyear:
Also, fuck yes, favicon go snippy snippy

kanidmd_web_ui/pkg/wasmloader.js
yaleman: on `kanidmd_web_ui.js` ? Sure, but (Edge, at least) didn't seem to complain.

Firstyear: Yeah, as well as the other bits, but I dunno how easy that is. Maybe that's a second PR .... can you add integrity headers inside js like that?

kanidmd/score/tests/https_middleware.rs
```diff
@@ -0,0 +1,95 @@
+use std::net::TcpStream;
+use std::sync::atomic::{AtomicU16, Ordering};
+
+use kanidm::audit::LogLevel;
+use kanidm::config::{Configuration, IntegrationTestConfig, ServerRole};
+use kanidm::tracing_tree;
+use score::create_server_core;
+use tokio::task;
+
```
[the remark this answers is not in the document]
yaleman: Done!

Firstyear (approved):

kanidmd_web_ui/pkg/wasmloader.js
yaleman: Yeah, I made the entire thing pretty flexible so it's doable - I'm just doing a super-lazy string replace on the body at the moment, could do it with a similar thing for that one. Given that it's in the `pkg` dir though - it might be worth looking at some kind of compile-phase thing that updates the file so we don't need *another* middleware handler. 

There's also the issue that I'm likely to run into where IIRC I've got JS minification on in CloudFlare and it's likely to break the hashes as it is...

Firstyear: Yeah, no stress mate, thanks for making the project that little bit better :)

PR #741, opened 2022-05-06
Dockerized containerybuilds

yaleman:
Brings the automagical build of containers back with Github actions, pushing to the registry only when the commit's for `master`.

Builds a container each for `kanidm/kanidmd` and `kanidm/radius`... though I'll have to see what the packages are named when we actually merge this 🙄 

Moved to using the official `opensuse/tumbleweed` rust packages because some 🦀 got it into `zypper`. 😄


.github/workflows/docker_build.yml
```diff
@@ -0,0 +1,81 @@
+name: Container for Kanidm
+
+# this will build regardless, but only push to the container registry when you're committing on the master branch.
```
Firstyear, line 3: I don't think we should always build this, I think the CI tests are enough, we shoulrd only build on master

.github/workflows/docker_build.yml
```diff
@@ -0,0 +1,81 @@
...
+        with:
+          args: cargo test
+
+  kanidm_build:
+    needs: test
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        target:
+         - linux/arm64
```
Firstyear, line 23: Which image tags is this building?

.github/workflows/docker_build.yml
pando85: I just want to add here that this can help to catch up bugs in the container build step.
It's important if you want to keep stability in your releases.
In my opinion tests are free so any little help should be appreciated.

yaleman: @pando85's point matches my opinion - if it's free and catches issues in the build. 🤷🏻

.github/workflows/docker_build.yml
yaleman: These are mentioned later:

`tags: ghcr.io/${{ github.repository }}/kanidmd:latest`

and 

`tags: ghcr.io/${{ github.repository }}/radius:latest`

from previous testing it *should* end up as `kanidm/radius:latest` etc, we could rename it as `dev` or the less... accurate `nightly` which seems to match people's usual expectations.

This incarnation's more about having a testable container available on the github container registry than a "this is the current stable one" which I'd probably push to dockerhub or something?

yaleman: Having said that, we could also have another github actions config to build the tagged versions to match the releases....

Firstyear: I think latest is our release tag name, we should call this devel i think. Check Makefile i think it has the name.

.github/workflows/docker_build.yml
Firstyear: Yep fair enough.

.github/workflows/docker_build.yml
Firstyear: Oh hang on ghcr.io is a github registry, and this is not pushed to docker hub correct?

Firstyear: Anyway calling it devel keeps it inline with the docker hub images

yaleman: Yep, devel's the one, missed that in all the Makefile templateyness. Sorted.

Firstyear (approved):

PR #750, opened 2022-05-09
Update docs with some personal experience

matthew-salerno:
Just changes a few bits in the documentation:
- Fixed a typo and added a note about centos 9 compat in installing client tools  
- Added a clarification to posix accounts about podman having issues with kanidm client, not kanidm server
- Added a section on working with authselect instead of overriding symlinks

Hopefully this is up to par, the checklist provided seemed to be focused on adding functionality instead of docs so I apologize if I'm breaking any rules for your pull request process.

### DIFF
--- a/kanidm_book/src/installing_client_tools.md
+++ b/kanidm_book/src/installing_client_tools.md
@@ -1,6 +1,6 @@
 # Installing Client Tools
 
-> **NOTE** As this project is in a rapid development phase, running different release versions will likely present incompatibilities. Ensure you're running matching release versions of client and server binaries.
+> **NOTE** As this project is in a rapid development phase, running different release versions will likely present incompatibilities. Ensure you're running matching release versions of client and server binaries. If you have any issues check that you are running the latest software.
 
 ## From packages
 
@@ -9,7 +9,7 @@ Kanidm currently supports:
  * OpenSUSE Tumbleweed
  * OpenSUSE Leap 15.3/15.4
  * Fedora 34/35
- * Centos Stream 9 (alpha 7 only)
+ * CentOS Stream 9
 
 ### OpenSUSE Tumbleweed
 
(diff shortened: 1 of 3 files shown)


kanidm_book/src/installing_client_tools.md
```diff
@@ -9,7 +9,7 @@ Kanidm currently supports:
  * OpenSUSE Tumbleweed
  * OpenSUSE Leap 15.3/15.4
  * Fedora 34/35
- * Centos Stream 9
+ * Centos Stream 9 (alpha 7 only)
```
yaleman, line 12: This wording isn't clear to me - do you mean Kanidm alpha 7, or CentOS?

matthew-salerno:
My bad, that is kanidm alpha 7, not centos

May 9, 2022 17:19:57 James Hodgkinson ***@***.***>:

yaleman:
> My bad, that is kanidm alpha 7, not centos May 9, 2022 17:19:57 James Hodgkinson

I'd say we should update the note directly above the list to say something along the lines of "if you have any issues, please ensure you're running the latest code" instead. It's alpha software, so "stable" is a pretty shaky base at the moment - and alpha7's already n-1 on "stable".

kanidm_book/src/installing_client_tools.md
Firstyear: I think they mean alpha 7 only, because that was the first version that would work on centos. So I think in this case because we don't support "older" versions, we don't need this as we can safely assume everyone should be on alpha 8 and above.

Firstyear (changes requested):

Firstyear:
Yeah, I'd echo @yaleman 's comments here, beside that it looks great!

yaleman (approved):
Thanks!

matthew-salerno:
I just changed things around as suggested by @yaleman how's it look now? I tried selecting lines to quote in github but I can't figure it out :/

Firstyear (approved):

Firstyear:
> I just changed things around as suggested by @yaleman how's it look now? I tried selecting lines to quote in github but I can't figure it out :/

i trust Yaleman's suggestions :)

PR #773, opened 2022-05-24
Improve token readability, Fix issue with spn format

Firstyear:
Fixes a subtle bug in uuid2spn index, improves reset link tokens, docs fix. 

### DIFF
--- a/kanidmd/idm/src/idm/credupdatesession.rs
+++ b/kanidmd/idm/src/idm/credupdatesession.rs
@@ -11,7 +11,7 @@ use crate::credential::totp::{Totp, TOTP_DEFAULT_STEP};
 
 use kanidm_proto::v1::{CURegState, CUStatus, CredentialDetail, PasswordFeedback, TotpSecret};
 
-use crate::utils::{backup_code_from_random, uuid_from_duration};
+use crate::utils::{backup_code_from_random, readable_password_from_random, uuid_from_duration};
 
 use serde::{Deserialize, Serialize};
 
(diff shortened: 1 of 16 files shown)


kanidmd/idm/src/idm/credupdatesession.rs
```diff
@@ -391,70 +406,129 @@ impl<'a> IdmServerProxyWriteTransaction<'a> {
...
-            return Err(OperationError::SessionExpired);
-        }
+        let CredentialUpdateIntentToken { intent_id } = token;
+
+        /*
+            let entry = self.qs_write.internal_search_uuid(&token.target)?;
+        */
+        // ⚠️  due to a low, but possible risk of intent_id collision, if there are multiple
+        // entries, we will reject the intent.
+        // DO we need to force both to "Consumed" in this step?
```
yaleman, line 416: I'm not sure I'd use `Consumed` as the result, this feels like it needs its own state so it's really clear what happened - maybe a `ForcedExpired` so that if say ... a collision happened, or an admin killed the intent token at some point, it could be logged + terminated.

yaleman (approved):
One minor discussion point but otherwise approved 😄 Thanks for squishing bugs 💜

kanidmd/idm/src/idm/credupdatesession.rs
Firstyear: Yeah, that was my thinking was a seperate collision state, but for now, it just forces to a safe failure condition at least.

PR #774, opened 2022-05-24
Clean up Makefile documentation

### LINKED ISSUE #768: docs - explain the Makefile options for developers
**Documentation for Makefile usage**

As a developer, it would be handy to have documentation of options for the `Makefile` and a general theory of what's expected.

For example, I'm curious what [the `ARGS` reference is to a `redis` cache](https://github.com/kanidm/kanidm/blob/ec7c3cc6238459e7098c0ae948f4fb21fb27439e/Makefile#L7).

Identify whether it is preferable to update `DEVELOPER_README.md` or create a new document with a clear name such as `BUILDING_CONTAINERS.md` and document the various options available to build a `kanidmd` container image.

There's the Star Wars method - `use the source, Luke` - which is what I used my first time through the process.

kellinm:
```
Clean up Makefile documentation

- Add finalized changes that were not present at merge because I forgot
  to put the merge request into Draft mode while working.

Related https://github.com/kanidm/kanidm/issues/768

```

Fixes #768 

- [ ] cargo fmt has been run

### DIFF
--- a/DEVELOPER_README.md
+++ b/DEVELOPER_README.md
@@ -166,9 +166,9 @@ Setting different developer profiles while building is done by setting the envir
 
 For example: `KANIDM_BUILD_PROFILE=release_suse_generic cargo build --release --bin kanidmd`
 
-### Building the development container
+### Build a kanidm container
 
-Build a development container with the current branch using:
+Build a container with the current branch using:
 
     make <TARGET>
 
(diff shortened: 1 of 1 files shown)


kellinm:
@Firstyear - this is ready - I had just finished proofing the book output when I noticed the previous had merged.

DEVELOPER_README.md
```diff
@@ -166,30 +166,32 @@ Setting different developer profiles while building is done by setting the envir
...
+#### Container build examples
+
+Build a `kanidm` container using `podman`:
 
     CONTAINER_TOOL=podman make build/kanidmd
 
-Build a development container and specify a redis build cache:
+Build a `kanidm` container and use a redis build cache:
 
     CONTAINER_BUILD_ARGS="--build-arg "SCCACHE_REDIS=redis://redis.dev.blackhats.net.au:6379"" make build/kanidmd
```
yaleman, line 197: Now that I'm re-reading this, the quoting on that last line looks wrong - should it be:

```shell
CONTAINER_BUILD_ARGS='--build-arg "SCCACHE_REDIS=redis://redis.dev.blackhats.net.au:6379"'
```

Firstyear: @yaleman good catch, I'll leave this one to you to approve, I have some wasm to go wrestle :) 

Thanks again @kellinm, really appreciate it!

kellinm: It looks weird, but it is correct shell syntax and worked when tested.

Both variants pass muster with `ShellCheck` - though I prefer the double-double quote method because newcomers may try the single quote syntax, but also try to expand a local ENV variable in their invocation, and trip over shell expansion.

Your call as the maintainers, but it does work this way and is valid syntax.

Firstyear: I think we should make it the single wrapping on the outer, just to remove abiguity :)

kellinm: Changed. :)

PR #775, opened 2022-05-24
continuing review of Kanidm book

cjschroder:
small wording corrections, made commands consistent (no sudo, no command prompt)

There are several inline comments, and questions labeled TODO.
Overall this book is quite good, it follows a useful, logical order and gives a lot
of good examples

Fixes #

- [ ] cargo fmt has been run

### DIFF
--- a/kanidm_book/src/security_hardening.md
+++ b/kanidm_book/src/security_hardening.md
@@ -38,11 +38,9 @@ Each warning highlights an issue that may exist in your environment. It is not p
 prescribe an exact configuration that may secure your system. This is why we only present
 possible risks.
 
-### Should be readonly to running uid
+### Should be readonly to running UID
 
-TODO is the running UID the kanidm daemon?
-
-Files such as configurations should be read-only to this UID/GID. If an attacker is
+Files, such as configuration files, should be read-only to the UID of the Kanidm daemon. If an attacker is
 able to gain code execution, they are then unable to modify the configuration to write, or to over-write
 files in other locations, or to tamper with the systems configuration.
 
@@ -58,10 +56,10 @@ attacker can read Kanidm content and may be able to further attack the system as
 This can be prevented by removing "everyone: execute bits from parent directories containing the
 configuration, and removing "everyone" bits from the files in question.
 
-### owned by the current uid, which may allow file permission changes
+### owned by the current UID, which may allow file permission changes
 
-File permissions in unix systems are a discrestionary access control system, which means the
-named uid owner is able to further modify the access of a file regardless of the current
+File permissions in unix systems are a discretionary access control system, which means the
+named UID owner is able to further modify the access of a file regardless of the current
 settings. For example:
 
     [william@amethyst 12:25] /tmp > touch test
@@ -77,7 +75,6 @@ settings. For example:
 Notice that even though the file was set to "read only" to william, and no permission to any
 other users, user "william" can change the bits to add write permissions back or permissions
 for other users.
-<!-- gotta watch out for that william dude, he is trouble -->
 
 This can be prevent by making the file owner a different UID than the running process for kanidm.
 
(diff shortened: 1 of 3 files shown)


kanidm_book/src/security_hardening.md
```diff
@@ -39,8 +40,10 @@ possible risks.
 
 ### Should be readonly to running uid
 
-Files such as configurations should be readonly to this UID/GID. This is so that if an attacker is
-able to gain code execution, they are unable to modify the configuration to write or over-write
+TODO is the running UID the kanidm daemon?
```
yaleman, line 43: The user id running the kanidm daemon shouldn't be able to write to the configuration files.

kanidm_book/src/security_hardening.md
```diff
@@ -52,8 +55,8 @@ This means that given a permission mask, "everyone" or all users of the system c
 execute the content of this file. This may mean that if an account on the system is compromised the
 attacker can read Kanidm content and may be able to further attack the system as a result.
 
-This can be prevented by removed everyone execute bits from parent directories containing the
-configuration, and removing everyone bits from the files in question.
+This can be prevented by removing "everyone: execute bits from parent directories containing the
+configuration, and removing "everyone" bits from the files in question.
 
 ### owned by the current uid, which may allow file permission changes
```
yaleman, line 61: Missing capitalisation

kanidm_book/src/security_hardening.md
```diff
@@ -72,10 +75,11 @@ settings. For example:
     -rw-r--r--  1 william  wheel  0 29 Jul 12:25 test
 
 Notice that even though the file was set to "read only" to william, and no permission to any
-other users, that as "william" I can change the bits to add write permissions back or permissions
+other users, user "william" can change the bits to add write permissions back or permissions
 for other users.
+<!-- gotta watch out for that william dude, he is trouble -->
```
yaleman, line 80: Past William is trouble; future William's a real go-getter. He's going to fix everything!

kanidm_book/src/security_hardening.md
```diff
@@ -89,25 +93,29 @@ is to set configurations to:
 The kanidm server should be run as "kanidm:kanidm" with the appropriate user and user private
 group created on your system. This applies to unixd configuration as well.
 
+TODO is the kanidm user created the same way as other daemon users, without a /home dir and
```
yaleman, line 96: Correct (typically with systemd which dynamically creates the user and its associated uid, if I understand it properly) 👍🏻

kanidm_book/src/security_hardening.md
```diff
@@ -89,25 +93,29 @@ is to set configurations to:
...
 from the directory.
 
-Configurations for clients such as /etc/kanidm/config should be secured with the permissions
-as:
+Configurations for clients, such as /etc/kanidm/config, should be secured with read-only permissions:
 
     [william@amethyst 12:26] /etc/kanidm > ls -al config
     -r--r--r--    1 root  wheel    38 10 Jul 10:10 config
+    
+TODO must it be the wheel group? the wheel group controls sudo users on many linuxes
```
yaleman, line 116: `root:root` is probably better, thanks.

kanidm_book/src/security_hardening.md
```diff
@@ -121,26 +129,28 @@ This file should be everyone-readable which is why the bits are defined as such.
...
-    # chown root:root /data/server.toml
-    # chmod 644 /data/server.toml
+    mkdir /data/db/
+    chown 1000:1000 /data/db/
+    chmod 750 /data/db/
+    sed -i -e "s/db_path.*/db_path = \"\/data\/db\/kanidm.db\"/g" /data/server.toml
+    chown root:root /data/server.toml
+    chmod 644 /data/server.toml
+    
+    <!-- removed prompts for consistency with other command examples -->
```
yaleman, line 148: I feel like this needs a note to say the commands after `docker run...` need to run in the container. Is there an easy/coherent way to note this?

yaleman (commented):
Thank you very much!

kanidm_book/src/security_hardening.md
Firstyear: Yep, that's right, systemd dynamically creates it

Firstyear:
> Overall this book is quite good, it follows a useful, logical order and gives a lot of good examples

❤️ Yay! We did good!

kanidm_book/src/security_hardening.md
cjschroder: I need more clarification, please. How does this work, does systemd create the user & UID at installation, and then it is always the same? (I am not testing anything yet, just reviewing for clarity & typos)

kanidm_book/src/security_hardening.md
cjschroder: I am not docker-fluent, do the commands need to change? Or are the commands correct as-is, and is it enough to say something like "these commands must be in the container"

kanidm_book/src/security_hardening.md
cjschroder: > > Overall this book is quite good, it follows a useful, logical order and gives a lot of good examples
> 
> heart Yay! We did good!

Yes you did!

kanidm_book/src/security_hardening.md
yaleman: The latter - the commands are correct and the note would make sense as you have it there - apologies for being unclear 😄

kanidm_book/src/security_hardening.md
Firstyear: Ahhh it's much more magical, just like anything with systemd involved. 

http://0pointer.net/blog/dynamic-users-with-systemd.html

What happens is that when systemd starts up the service, it dynamically generates a uid/gid, and creates a home directory for the user. When the service stops the user "vanishes". This way, you can have the kanidm account appear/disappear basicly, because systemd will auto-magically chown your home dir and all that.


In the docker case, it's up to the admin to create and nominate a uid/gid to run the server as to provide to docker or their container engine.

cjschroder: Thanks! My #1 question is usually "Do I have to do something, or is there magick and I do not have to do something"

Firstyear:
> All done with this for now. I might take another look when I have a test system set up, as the server hardening chapter is full of potential perils :)

If you thought 389-ds was easy to setup, wait til you try this ^^;

PR #792, opened 2022-05-28
Set default value for SCCACHE_REDIS build arg

kellinm:
```
Set default value for SCCACHE_REDIS build arg

- Set a default value in the Dockerfile for SCCACHE_REDIS so that, at
  build time, it does not fail with the new behavior because nothing was
  actually being passed into the container.

```

From the, "_how the heck did I miss this when testing the behavior change_" department.

- [ ] cargo fmt has been run

### DIFF
--- a/kanidmd/Dockerfile
+++ b/kanidmd/Dockerfile
@@ -22,7 +22,7 @@ RUN zypper install -y \
 COPY . /usr/src/kanidm
 WORKDIR /usr/src/kanidm/kanidmd/daemon
 
-ARG SCCACHE_REDIS
+ARG SCCACHE_REDIS=""
 ARG KANIDM_FEATURES
 ARG KANIDM_BUILD_PROFILE
 


kellinm:
This will get replaced eventually, but for now this gets it working.

yaleman:
That's curious, I'm not setting a value and it's building using `make build/kanidmd` - how are you getting it to fail, and what's it doing?

Firstyear:
Yeah I tried it with this empty and it worked, but also I don't mind setting it like this.

@yaleman They are using podman, which is probably doing an incompatible behaviour, so I'm happy to accept this as is.

yaleman:
Sweet, I'll have to play with podman to see how I can make it pop at some point. Thanks @kellinm

kellinm:
You are correct, I am using `podman` @Firstyear 

Sorry for the delayed reply, things got busy here all of a sudden last week.

PR #797, opened 2022-05-29
WIP: Issue 668

### LINKED ISSUE #668: UX and api design for common functionality between "account" and "self"
### Background

This issue is to try and figure out the right way to model commands that modify accounts, and may also be self-service.

For a motivating example, let's look at the following existing pair of commands:

```
$ kanidm account credential set_password <account>
$ kanidm self set_password
```

Let's also consider a hypothetical pair of commands:
```
$ kanidm account person set --legalname "Name" <account>
$ kanidm self <set legal name subcommand; design TBD
```

There are several other attributes which may be self-service, and all of them have a similar potential split.

I think this shows that the `self` and `account` namespace should have some overlap, and the rest of this issue is to discuss options for that.

### Current technical solution / server api

The current solution used for `set_password` is effectively to duplicate the code:

`account credential set_password`

https://github.com/kanidm/kanidm/blob/4862b919de629afebf9133d2c82c1eb18490e6cb/kanidm_tools/src/cli/account.rs#L67-L78

`self set_password`

https://github.com/kanidm/kanidm/blob/4862b919de629afebf9133d2c82c1eb18490e6cb/kanidm_tools/src/cli/lib.rs#L61-L71

Notably, this includes two backend APIs on the server too:

(self): `/_credential/primary/set_password`

(account credential): `/v1/account/:id/_credential/primary`

### Alternative API

One alternative to the pair of server-side APIs would be to instead just have the account version of it.

The `self set_password` command, instead of calling `/_credential/primary/set_password`, could instead:

1. Call `/whoami` and get their own uuid
2. Call  `/v1/account/<uuid>/_credential/primary` to set the credential

Alternatively, we could have some reserved uuid or other identifier that we fill in to `:id` which indicates `self`, removing the whoami call. The server obviously already can figure out the authenticated account.

### Command UX

For the actual UX of the command, we have additional choices.

I think the biggest one is "do we mirror the 'account' subcommand, or do we make a bespoke self UX?"

I think there's a fair argument for making the account and self commands have a similar subcommand structure (with `self` being a subset of account). Having similar CLI UX will especially be helpful for an administrator who's used to using the 'account' commands, but wants to recommend a user run a corresponding `self` command.

If we followed the mirroring approach, I think it would currently look like so:

```
$ kanidm self
             ├ credential
             │           ├ generate_backup_codes
             │           ├ register_totp
             │           ├ register_webauthn
             │           ├ remove_backup_codes
             │           ├ remove_totp
             │           ├ reset_credential
             │           ├ remove_webauthn
             │           ├ set_password
             │           └ status
             │
             ├ person
             │       └ set --legalname --mail
             │
             ├ posix
             │      ├ set_password
             │      └ show
             │
             ├ radius
             │       ├ generate_secret
             │       ├ delete_secret
             │       └ show_secret
             │
             └ ssh
                  ├ add_publickey
                  ├ delete_publickey
                  └ list_publickeys
```

I think that this command structure is somewhat inconsistent (i.e. the credential grouping feels a little haphazard, the `person set` command has different ux compared to everything else, etc)....

But I don't think that's an argument not to mirror the structure, but is rather an argument to improve both the account and person structure to be laid out better, but still identically.

#### Hiding unused portions

One thing which would be nice for cleaning up the UX, but may not be viable, is excluding portions from 'self' that don't make sense for any logged in account.

For example, if a given account doesn't have the `posixaccount` class, none of the `posix` commands will do anything. This is similar for person.

Unfortunately, this seems like it's probably not technically feasible, especially when shipping static auto-completions etc.

### Initial decision points

The above raised two questions which I think are worth considering:

1. Should the server API include dedicated APIs for "self" commands, or should those commands typically lean on existing account APIs
     a. If "use account APIs", should they internally call a `whoami`/`get_account_id` api to fill in their account uuid (and possibly cache the uuid locally)? Or should they fill in some sigil/special uuid that indicates "currently authed user uuid here"?
2. Should the command structure for account vs person differ? Why/why not?

My current instincts are "1. No dedicated APIs, use the account ones, preferably with a 'placeholder' account id" and "2. Same structure".

---------------------

A final aside, this issue is too high level to actually talk about how we model 2 in structopt, but I did poke at that a bit.

It's possible to make an optional final arg (the account ID) in clap by making a generic structopt struct.

See what I mean [here](https://github.com/euank/kanidm/blob/c63a85ab919a3024dd7b9dac216d9b097c0eafdd/kanidm_tools/src/opt/kanidm.rs#L404-L408) (full diff [here](https://github.com/euank/kanidm/commit/c63a85ab919a3024dd7b9dac216d9b097c0eafdd), I'm quite aware the code's not well structured, I was just experimenting to see that it is possible to do that at all)

MateusAmin:
I started looking at the code today. Will push my work incrementally to this PR.

Fixes #668 

- [ ] cargo fmt has been run

### DIFF
--- a/kanidm_client/src/lib.rs
+++ b/kanidm_client/src/lib.rs
@@ -1130,9 +1130,9 @@ impl KanidmClient {
         self.perform_get_request("/v1/auth/valid").await
     }
 
+    // TODO(mea): self -> whoami?
     pub async fn whoami(&self) -> Result<Option<(Entry, UserAuthToken)>, ClientError> {
-        let whoami_dest = [self.addr.as_str(), "/v1/self"].concat();
-        // format!("{}/v1/self", self.addr);
+        let whoami_dest = [self.addr.as_str(), "/v1/whoami"].concat();
         debug!("{:?}", whoami_dest);
         let response = self.client.get(whoami_dest.as_str());
 
@@ -1521,7 +1521,7 @@ impl KanidmClient {
         }
     }
 
-    pub async fn idm_account_primary_credential_complete_webuthn_registration(
+    pub async fn idm_account_primary_credential_complete_webauthn_registration(
         &self,
         id: &str,
         rego: RegisterPublicKeyCredential,
--- a/kanidmd/score/src/https/mod.rs
+++ b/kanidmd/score/src/https/mod.rs
@@ -11,8 +11,8 @@ use kanidm::prelude::*;
 use kanidm::status::StatusActor;
 
 use serde::Serialize;
-use std::path::PathBuf;
 use std::fs::canonicalize;
+use std::path::PathBuf;
 use std::str::FromStr;
 use uuid::Uuid;
 
@@ -366,7 +366,6 @@ pub fn create_https_server(
     qe_w_ref: &'static QueryServerWriteV1,
     qe_r_ref: &'static QueryServerReadV1,
 ) -> Result<(), ()> {
-
     let jws_validator = jws_signer.get_validator().map_err(|e| {
         error!(?e, "Failed to get jws validator");
     })?;
@@ -401,7 +400,6 @@ pub fn create_https_server(
 
     // If we are no-ui, we remove this.
     if !matches!(role, ServerRole::WriteReplicaNoUI) {
-
         let pkg_path = PathBuf::from(env!("KANIDM_WEB_UI_PKG_PATH"));
         if !pkg_path.exists() {
             eprintln!(
@@ -555,7 +553,8 @@ pub fn create_https_server(
         .post(oauth2_id_scopemap_post)
         .delete(oauth2_id_scopemap_delete);
 
-    let mut self_route = appserver.at("/v1/self");
+    // TODO(mea): self -> whoami?
+    let mut self_route = appserver.at("/v1/whoami");
     self_route.at("/").get(whoami);
 
     self_route.at("/_attr/:attr").get(do_nothing);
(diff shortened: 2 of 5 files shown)


yaleman:
If this isn't ready to merge, could you please turn it a draft?

kanidm_client/src/lib.rs
```diff
@@ -1521,7 +1521,7 @@ impl KanidmClient {
         }
     }
 
-    pub async fn idm_account_primary_credential_complete_webuthn_registration(
+    pub async fn idm_account_primary_credential_complete_webauthn_registration(
```
Firstyear, line 1524: ROFL good spot!

kanidmd/score/src/https/mod.rs
```diff
@@ -555,7 +553,8 @@ pub fn create_https_server(
         .post(oauth2_id_scopemap_post)
         .delete(oauth2_id_scopemap_delete);
 
-    let mut self_route = appserver.at("/v1/self");
+    // TODO(mea): self -> whoami?
```
Firstyear, line 556: The "whoami" is just a reflector of details about the currently authenticated user. This is a bit of a "hold-over" from the past, where the auth-bearer tokens used to be opaque, so they couldn't be client-side introspected.

However, recently we have changed these so they CAN be client side introspected, meaning it's possible that we may NOT need the "whoami" route at all.

Firstyear: The lines around here:

https://github.com/kanidm/kanidm/blob/master/kanidm_tools/src/cli/common.rs#L115

You could use "uat.uuid" instead which gives you the ID of the token holder, since the tokens are JWT's internally.

Firstyear:
Closing due to age

PR #798, opened 2022-05-30
Windows automagical buildingtons

yaleman:
Because other OSen need 🦀🔪  too.

This makes sure that the `kanidm` client and `orca` build - I tried for `kanidmd` but there's a bunch of very unixy things in there that will likely need code changes, so I left it for now.

- [x] cargo fmt has been run and I'm offended you'd even ask, Your Honour.
- [x] cargo test has been run and passes, I think?
- [x] docs updated with a vague note about Windows support.

### DIFF
--- a/kanidmd/idm/src/lib.rs
+++ b/kanidmd/idm/src/lib.rs
@@ -14,6 +14,7 @@
 #![deny(clippy::needless_pass_by_value)]
 #![deny(clippy::trivially_copy_pass_by_ref)]
 
+#[cfg(not(target_family = "windows"))]
 #[cfg(all(jemallocator, test))]
 #[global_allocator]
 static ALLOC: jemallocator::Jemalloc = jemallocator::Jemalloc;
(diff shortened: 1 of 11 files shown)


kanidmd/idm/src/lib.rs
```diff
@@ -14,6 +14,7 @@
 #![deny(clippy::needless_pass_by_value)]
 #![deny(clippy::trivially_copy_pass_by_ref)]
 
+#[cfg(not(target_family = "windows"))]
```
Firstyear, line 17: Could this be merged down to the line below as  `#[cfg(all(jemallocator, test, not(target_family = "windows")))]`

Firstyear (approved):
One minor bit of feedback, but otherwise looks good!

kanidmd/idm/src/lib.rs
yaleman: Done! - I couldn't puzzle out the syntax to work that out and just copypasta'd around. 😇

Firstyear (approved):

PR #815, opened 2022-06-06
Add design session logout/expiry (affects oauth2)

Firstyear:
- [ ] cargo fmt has been run


Firstyear:
Still could be a draft, I'm still thinking it through still, but I want to implement this soon so figured I should write it up, especially with the other nearby oauth2 work soon.

Firstyear:
The other way to tackle this is instead of treating this like a revocation with short lived tokens + refresh, we have an initial "grace window" to allow replication to occur. during the grace window, the token is assumed valid and not revoked. Once the window passes, if the session id is "not found" in the database, it's assumed to have been revoked and invalid. That would resolve the issues around refresh and re-issuance of the tokens to clients, since we wouldn't need shortlived + refresh, since the session id would be present if valid, or removed if revoked.

designs/session_logout.rst
```diff
@@ -0,0 +1,105 @@
...
+Sessions can be revoked provided the user has the ability to "write" to the session attribute. This
+allows access controls to be used, and other users to administer another users session when delegation
+is required.
+
+Oauth2 rfc7009 revoke
+=====================
+
+Oauth2 will need refresh tokens issued along with access tokens. The same rules apply as above for
+user auth tokens. When a refresh is required, the oauth2 session id looked up to find the corresponding
+session and it's validity.
```
yaleman, line 88:
```suggestion
session and its validity.
```

designs/session_logout.rst
```diff
@@ -0,0 +1,105 @@
+
+Session Logout
+--------------
+
+Currently, Kanidm relies on "short session times" to manage and limit issues with session
+disclosure, but this is obviously not optimal long term! In addition, there are systems like
+oauth2 token revocation ( https://datatracker.ietf.org/doc/html/rfc7009 ) which we may want
```
yaleman, line 7:
```suggestion
OAuth 2.0 token revocation ( https://datatracker.ietf.org/doc/html/rfc7009 ) which we may want
```

designs/session_logout.rst
```diff
@@ -0,0 +1,105 @@
...
+to use to allow global logouts across linked applications.
+
+Goals
+=====
+
+* When a user selects "logout" the active session is canceled
+* A user can view their list of active sessions and audit or revoke them
+* A user can view what oauth2 applications an active session has interacted with
```
yaleman, line 15:
```suggestion
* A user can view which OAuth 2.0 applications an active session has been used with
```

designs/session_logout.rst
```diff
@@ -0,0 +1,105 @@
...
+* A oauth2 application can use rfc7009 logout to revoke a token that is associated to a session
```
yaleman, line 16:
```suggestion
* An OAuth 2.0 application can use the RFC7009  logout mechanism to revoke a token that is associated to a session
```

designs/session_logout.rst
```diff
@@ -0,0 +1,105 @@
...
+* Revokation of a session, implies revokation of all oauth2 sessions.
```
yaleman, line 17:
```suggestion
* Revocation of a session implies revocation of all OAuth 2.0 sessions.
```

designs/session_logout.rst
```diff
@@ -0,0 +1,105 @@
...
+To achieve this, on login we need a way to record the creation of the session. This session will be
+bound to the uuid of the session that is part of the user auth token. This will be added to the users
+entry along with metadata about the session such as the:
+
+* Source IP that created the session
+* Time the session was created
+* Time the session *may* expire, if an expiry exists
+* Any limits or details (for example, sudo mode) on the session.
+* A state flag defining that the session is *active*
+* Mappings from sub-sessions through other protocols, IE oauth2 and their related session id's
```
yaleman, line 32:
```suggestion
* Mappings from sub-sessions through other protocols, IE OAuth 2.0 and their related session IDs
```

designs/session_logout.rst
```diff
@@ -0,0 +1,105 @@
...
+
+The session will need to have a refresh-time, similar to an oauth2 refresh token, or kerberos
```
yaleman, line 34:
```suggestion
The session will need to have a refresh-time, similar to an OAuth 2.0 refresh token or kerberos
```

designs/session_logout.rst
```diff
@@ -0,0 +1,105 @@
...
+refresh window. This still allows each token to have a maximum usage time window, without
+needing a limit to the maximum duration of the session.
+
+On Logout
+=========
+
+Logout events have an impact with regard to replication.
+
+On logout, if the session id is found, it's state is transitioned to the expired state. No other
```
yaleman, line 43:
```suggestion
On logout, if the session id is found, its state is transitioned to the expired state. No other
```

designs/session_logout.rst
```diff
@@ -0,0 +1,105 @@
...
+details are changed.
+
+If the session id is NOT found, a session stub is created, with the expired state. Since we lack the
+metadata of the "creation" of the session, this is why we use the stub form that only lists the id and
+it's expiry.
```
yaleman, line 48:
```suggestion
its expiry.
```

designs/session_logout.rst
```diff
@@ -0,0 +1,105 @@
...
+On Logout
+=========
+
+Logout events have an impact with regard to replication.
+
+On logout, if the session id is found, it's state is transitioned to the expired state. No other
+details are changed.
+
+If the session id is NOT found, a session stub is created, with the expired state. Since we lack the
+metadata of the "creation" of the session, this is why we use the stub form that only lists the id and
```
yaleman, line 47:
```suggestion
metadata of the "creation" of the session, this is why we use the stub form that only lists the ID and
```

designs/session_logout.rst
```diff
@@ -0,0 +1,105 @@
...
+if the CID of expiry preceeds that of the active state. We merge the expiry into the metadata in
+this case.
+
+Token Usage
+===========
+
+This will require that during calls to /v1/auth/valid that the token can be re-issued with an
+updated refresh time for the client. Clients will need to be modified to accomodate this.
+
+When any api endpoint is called, if the token is valid and does not need a refresh:
```
yaleman, line 60:
```suggestion
When any API endpoint is called, if the token is valid and does not need a refresh:
```

designs/session_logout.rst
```diff
@@ -0,0 +1,105 @@
...
+
+* If no session id is stored, assume the token is VALID and we have a replication delay.
+* If a session is found and active, respond Ok
+* If a session is found and expired, respond unauthorized.
+
+When auth/valid is called, if the token requires a refresh:
+
+* If no session id is found, assume it is revoked and indicate unauthorised. (This can happen if the session is cleaned up post expiry, and the refresh token is used)
```
yaleman, line 68:
```suggestion
* If no session ID is found, assume it is revoked and indicate unauthorised. This can happen if the session is cleaned up after it has expired, and the refresh token is used.
```

designs/session_logout.rst
```diff
@@ -0,0 +1,105 @@
...
+Session Management
+==================
+
+Expired sessions are not displayed to the user. Only active sessions are shown.
+
+Sessions can be revoked provided the user has the ability to "write" to the session attribute. This
+allows access controls to be used, and other users to administer another users session when delegation
+is required.
+
+Oauth2 rfc7009 revoke
```
yaleman, line 83:
```suggestion
OAuth 2.0 (RFC7009) revoke
```

designs/session_logout.rst
```diff
@@ -0,0 +1,105 @@
...
+=====================
+
+Oauth2 will need refresh tokens issued along with access tokens. The same rules apply as above for
```
yaleman, line 86:
```suggestion
OAuth 2.0 will need refresh tokens issued along with access tokens. The same rules apply as above for
```

designs/session_logout.rst
```diff
@@ -0,0 +1,105 @@
...
+user auth tokens. When a refresh is required, the oauth2 session id looked up to find the corresponding
```
yaleman, line 87:
```suggestion
user auth tokens. When a refresh is required, the OAuth 2.0 session ID is found to identify the corresponding
```

designs/session_logout.rst
```diff
@@ -0,0 +1,105 @@
...
+Sessions can only be cleaned up once the refresh window has passed. We also need to consider the
+replication changelog window.
+
+As a result, the refresh window should be a small subset of the maximum changelog window, such
+that cleanups always occur correctly and allow the refresh windows to pass. 25% of the window is a reasonable amount, up to 16 hours maximum.
+
+Updates
+=======
+
+We may wish to allow updates to the session state to reflect things such as a change of IP of the session
```
yaleman, line 102:
```suggestion
We may wish to allow updates to the session state to reflect things such as a change of client IP of the session
```

yaleman (changes requested):
This edit thing is great... >.>


I'd avoid using IP address as an authentication factor or keeping it as an identifier for sessions - clients are increasingly using rolling IP address "privacy" features and/or NATs, making it useful only as something to log for metadata's sake.

yaleman:
Still pondering the auth/reauth flow bit ^_^

Firstyear:
> This edit thing is great... >.>

I'm glad you like the edit thing, but fuck I wish there was a way to select "okay, commit these X" in a batch. 

> 
> I'd avoid using IP address as an authentication factor or keeping it as an identifier for sessions - clients are increasingly using rolling IP address "privacy" features and/or NATs, making it useful only as something to log for metadata's sake.

Yeah, I'm thinking it's just for logging, and not a security thing. But of course, someone will ask for it ...... 

An example being the "limit anonymous to certain ips", then we need to see if the ip changes to deny the session.

Firstyear:
> 



> Still pondering the auth/reauth flow bit ^_^

I too r ponder.

MateusAmin:
>  "okay, commit these X" in a batch.

If we are talking about those PR edits Yaleman makes in mass that you can commit from the GUI. You can indeed do it in batch or my dreams are getting much more vivid.

yaleman:
> Yeah, I'm thinking it's just for logging, and not a security thing. But of course, someone will ask for it ......
> An example being the "limit anonymous to certain ips", then we need to see if the ip changes to deny the session.

I guess that'd be an ACL which triggers on `if(SPN and NOT (allowed_ips)) then block` rather than an "if IP changed" since the latter would be a scary state machine thing.

> For the purpose of this design, it's just for metadata though, not for IP limiting. That's a different basket of joy. I think we could alternately ditch the "last ip used" for a device / browser user-agent instead? Alternately, in the early versions, we could just omit this.

"Last IP used" would definitely be handy for user information, to get a list of sessions and be able to terminate them etc. 👍🏻

yaleman (approved):

Firstyear:
Yeah. I think the only thing I'm wondering is about the revocation still, and if we want to go refresh token or "initial grace window". I think we can do the grace window by checking the replication metadata internally so that if the entry is "past the point in time" of the session init, we can assume the revocation is up-to-date. We can adjust the grace-time in the future I think based on experience with replication delay etc. No matter what, we'll have some minor delays based on replication anyway, so I think this is the safer option. I'll update the PR to reflect.

Firstyear:
@yaleman I updated this a bit, I'm still not sure if I want to do the positive or negative control version. Thoughts?

yaleman:
Specifically regarding Pos/Neg validation, I'd go with _positive_. 

Having to hold a list of revoked tokens sounds like a bigger problem than a list of active sessions. With negative validation, this is possible.

1. system gets hacked
2. token is revoked
3. before the revocation is replicated/active, bad actor does bad things
4. kanidm db is restored to before point 2.
5. bad actor goes back to doing things

With positive, one can just blank out the session table before startup and you're ... less insecure than before 😄 

So, on to other bits.

---

> This allows "unlimited length" sessions since we rely on positive validation of the session existing.

I assume the "unlimited length" here means "while the user keeps renewing the token".

---

Grace window yak shaving

Grace windows feel icky if you're not running replicas - I guess this'd add a fair chunk of complexity but maybe make this an "only if running replicas" thing? A later feature flag, maybe... and if someone 😇 gets the metrics going, the replication delay can be monitored and used as a signal for grace window length.

---

Firstyear:
> Specifically regarding Pos/Neg validation, I'd go with _positive_.
> 
> Having to hold a list of revoked tokens sounds like a bigger problem than a list of active sessions. With negative validation, this is possible.
> 
> 1. system gets hacked
> 2. token is revoked
> 3. before the revocation is replicated/active, bad actor does bad things
> 4. kanidm db is restored to before point 2.
> 5. bad actor goes back to doing things

That's a really good point. I was already leaning a bit to the 'positive' form just because you can do exactly that, where you nuke all sessions and you are good to go. I think it also enables better access and inspection, and I think it feels easier to manage too. so I'll go that way. 

> 
> With positive, one can just blank out the session table before startup and you're ... less insecure than before 😄
> 
> So, on to other bits.
> 
> > This allows "unlimited length" sessions since we rely on positive validation of the session existing.
> 
> I assume the "unlimited length" here means "while the user keeps renewing the token".

Yeah. I need to sort out the refresh process. I've been reading up on a few things and I need to think more about it. Specifically, once we do a refresh of token A and we get a new token B, how do we prevent "forking" to also create token C? How do we track and prevent this kind of proliferation? How can we also ensure we aren't accidentally penalising users if they are doing the "right thing"? Or am I overthinking it 🤣

Like, an example in my head is we have the user with token A, and they refresh it to get token B. Now, the server they were talking to goes down, and has *not* yet replicated it's changes out. The loadbalancer fails over to another kani server. So what happens then if the refresh token A is now used on the server that hasn't received the replicated changes yet? Well, since it doesn't know any better, it would grant a divergent token. 

Anyway, I think I have a solution in my head, I just need to write it all out. 

> 
> Grace window yak shaving
> 
> Grace windows feel icky if you're not running replicas - I guess this'd add a fair chunk of complexity but maybe make this an "only if running replicas" thing? A later feature flag, maybe... and if someone 😇 gets the metrics going, the replication delay can be monitored and used as a signal for grace window length.

Yeah, I've certainly been thinking about this. I think in the "non replicated" version, because the session would appear "instantly", then the grace window would never really need to be used because then we have the positive validation available. At that point if a revoke occured, we'd get the revoked state which allows a short term "negative" control, and so again the gracewindow wouldn't be consulted because the revocation is present and active. That's also why the "cleanup delay" has to be longer than the gracewindow, to allow this scenario to be properly handled. 

I think I should detail that exact use case out in the design to show it, but in a single server deployment, the grace window should never be used, it's only needed in replicated scenarioes. 

So in theory until we have replication, we *could* set this window to be 0 initially, but for now I'd rather just implement it "right" and test the hell out of it :)

yaleman:
> Yeah. I need to sort out the refresh process. I've been reading up on a few things and I need to think more about it. Specifically, once we do a refresh of token A and we get a new token B, how do we prevent "forking" to also create token C? How do we track and prevent this kind of proliferation? How can we also ensure we aren't accidentally penalising users if they are doing the "right thing"? Or am I overthinking it 🤣

I probably need to read up on how other systems do a refresh, but I figured a refresh would just update the expiry timer on the existing token 🤔 

> Like, an example in my head is we have the user with token A, and they refresh it to get token B. Now, the server they were talking to goes down, and has not yet replicated it's changes out. The loadbalancer fails over to another kani server. So what happens then if the refresh token A is now used on the server that hasn't received the replicated changes yet? Well, since it doesn't know any better, it would grant a divergent token.

Sometimes things just break, I guess? 😄 

> gracewindow

Yep, implement now, put a lot of TODO flags everywhere and see how it goes.

Firstyear:
> > Yeah. I need to sort out the refresh process. I've been reading up on a few things and I need to think more about it. Specifically, once we do a refresh of token A and we get a new token B, how do we prevent "forking" to also create token C? How do we track and prevent this kind of proliferation? How can we also ensure we aren't accidentally penalising users if they are doing the "right thing"? Or am I overthinking it 🤣
> 
> I probably need to read up on how other systems do a refresh, but I figured a refresh would just update the expiry timer on the existing token 🤔

Not quite. It has to prevent the refresh token from being used again. There is a token-chain ... a chain of bl.... anyway it's discussed here:

https://www.oauth.com/oauth2-servers/making-authenticated-requests/refreshing-an-access-token/

The idea being that theft of the refresh token if it's already been exchanged shouldn't work, and that theft of a refresh token should then "invalidate" the users session so they kind of get alerted. 

> 
> > Like, an example in my head is we have the user with token A, and they refresh it to get token B. Now, the server they were talking to goes down, and has not yet replicated it's changes out. The loadbalancer fails over to another kani server. So what happens then if the refresh token A is now used on the server that hasn't received the replicated changes yet? Well, since it doesn't know any better, it would grant a divergent token.
> 
> Sometimes things just break, I guess? 😄

I think there is a solution that doesn't break, which is even nicer. 

> 
> > gracewindow
> 
> Yep, implement now, put a lot of TODO flags everywhere and see how it goes.

I think the design ATM won't need many TODO's at all tbh, I think I've got it sorted out in my head. I need to write it down today.

PR #817, opened 2022-06-07
Clap Migration

MateusAmin:
Status: No Static Errors
Next: Testing

Fixes #

- [x] cargo fmt has been run
- [x] cargo clippy has been run (I did not add any)
- [x] cargo test has been run and passes (minus pam_  becuase linking errors in master on my system)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,4 @@ vendor.tar.*
 orca/example_profiles/small/orca-edited.toml
 docs/
 kanidm_unix_int/pam_tester/Cargo.lock
+.vscode/
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -152,17 +152,16 @@ dependencies = [
 
 [[package]]
 name = "async-global-executor"
-version = "2.0.4"
+version = "2.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c290043c9a95b05d45e952fb6383c67bcb61471f60cfa21e890dba6654234f43"
+checksum = "fd8b508d585e01084059b60f06ade4cb7415cd2e4084b71dd1cb44e7d3fb9880"
 dependencies = [
  "async-channel",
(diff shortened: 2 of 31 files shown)


Firstyear:
Wow, this looks FANTASTIC! What a huge improvement!

What else do you think needs to happen before we merge?

MateusAmin:
@Firstyear The migration guide recommends  some specific testing: https://github.com/clap-rs/clap/blob/master/CHANGELOG.md#migrate-structopt

ATM, I am getting the oddest import/scope issues in kandim_tools. However, it is getting late here. Will probably be clear in the morning.

MateusAmin:
@Firstyear 

I took the kanidm_tools CLI around for a spin and everything was working as it should.

While testing was recommend this is very declarative code and testing would mainly be checking for "configuration" errors. I think a careful review would work as well. The testing guide recommends trycmd to test but it doesn't seem to be popular and I think most people don't bother. The most value you would get out of it would be specifically for migrations like this.

There should be some yak shaving here but for the sake of  clarity should probably happen in another PR. There are a few lingo changes that should probably be reflected consistently in the names. There looks like there could be some general simplifications in the code. But I think I will leave it to you guys since it is borderline stylistic and I think we differ.

Firstyear:
I'll give this a test today! Thank you SO MUCH! 

> There should be some yak shaving here but for the sake of clarity should probably happen in another PR. There are a few lingo changes that should probably be reflected consistently in the names. There looks like there could be some general simplifications in the code. But I think I will leave it to you guys since it is borderline stylistic and I think we differ.

Yeah, lets save that for another PR. Just because we have some differences, doesn't mean I don't want to hear the suggestions, I make plenty of mistakes :) So it would be good to hear the ideas you are thinking.

Firstyear:
So sorry for the review delay, things got busy with work. Tested it out and all good, great work!

PR #824, opened 2022-06-15
636 consent in oauth2

### LINKED ISSUE #636: oauth2 - "consent to proceed" prompt
I'm not sure if this is a debugging thing, but can this be disabled by default, or on an RS basis? It's configured, so it's allowed, and an unnecessary bump to the experience for users.

Firstyear:
Fixes #636 - This improves the oauth2 consent flow so that if the scopes requested have been previously consented to in the past, there is no need to re-request the proceed workflow from the user. This should make oauth2 just that little bit nicer to use :) 


kanidm_proto/src/oauth2.rs
```diff
@@ -57,18 +57,23 @@ pub struct AuthorisationRequestOidc {
     pub acr: Option<String>,
 }
 
-/// We ask our user to consent to this Authorisation Request with the
-/// following data.
+/// When we request to authorise, it can either prompt us for consent,
+/// or it can immediately be granted due te past grant.
```
yaleman, line 61:
```suggestion
/// or it can immediately be granted due the past grant.
```

kanidmd/score/src/https/oauth2.rs
```diff
@@ -219,12 +226,43 @@ async fn oauth2_authorise(
...
+            let mut res = tide::Response::new(302);
+
+            redirect_uri
+                .query_pairs_mut()
+                .clear()
+                .append_pair("state", &state)
+                .append_pair("code", &code);
+            res.insert_header("Location", redirect_uri.as_str());
+            // I think the client server needs this
+            // res.insert_header("Access-Control-Allow-Origin", redirect_uri.origin().ascii_serialization());
```
yaleman, line 264: not a bad idea... tl;dr "don't share the response with any other servers"

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin

kanidmd/score/src/https/oauth2.rs
```diff
@@ -266,7 +304,10 @@ pub async fn oauth2_authorise_permit_post(mut req: tide::Request<AppState>) -> t
         .await
         .map(|mut res| {
             // in post, we need the redirect not to be issued, so we mask 302 to 200
-            res.set_status(200);
+            if res.status() == 302 {
+                // in post, we need the redirect not to be issued, so we mask 302 to 200
```
yaleman, line 308:
```suggestion

```

yaleman (changes requested):
minor changes so far, just throwing these in - still have to review `kanidmd/idm/src/idm/oauth2.rs`

kanidmd/score/src/https/oauth2.rs
Firstyear: Yeah. I need to see how it interacts with various bits, I'm not sure really.

kanidmd/idm/src/idm/oauth2.rs
```diff
@@ -548,45 +570,106 @@ impl Oauth2ResourceServersReadTransaction {
...
+                req_scopes.eq(consent_scopes)
+            } else {
+                false
+            };
 
-        let consent_data = serde_json::to_vec(&consent_req).map_err(|e| {
-            admin_error!(err = ?e, "Unable to encode consent data");
-            Oauth2Error::ServerError(OperationError::SerdeJsonError)
-        })?;
+        admin_info!(
```
yaleman, line 580: Shouldn't this be after L585

kanidmd/idm/src/idm/oauth2.rs
```diff
@@ -548,45 +570,106 @@ impl Oauth2ResourceServersReadTransaction {
...
+            "User has previously consented, permitting. {:?}",
+            req_scopes
+        );
 
-        let consent_token = self
-            .inner
-            .fernet
-            .encrypt_at_time(&consent_data, ct.as_secs());
+        if consent_previously_granted {
```
yaleman, line 585: Shouldn't we check for previous consent, then log it?

kanidmd/idm/src/idm/oauth2.rs
```diff
@@ -706,6 +799,9 @@ impl Oauth2ResourceServersReadTransaction {
         token_req: &AccessTokenRequest,
         ct: Duration,
     ) -> Result<AccessTokenResponse, Oauth2Error> {
+        // todo: add refresh token grant type.
```
yaleman, line 802:
```suggestion
        // TODO: add refresh token grant type.
```

yaleman (changes requested):
Once I've played around with how it works, I might add a quick mention somewhere in the documentation just so it's explicit how/why this functionality works. It's cool, but when it plays up (eg, if it doesn't *always* prompt for consent) folks like us would notice and be like ...

> 🤔 what is going on¿

yaleman:
> Maybe it should actually be on the consent page itself. That page needs a bit of improvement to show what was actually requested, and maybe it can say 'you'll only be asked once, unless the requested permissions change'.

That's possible, though sometimes that can turn into a scare-factor for users.

> Worse, all the info is literally there, I've just been to lazy to move it a few more bytes to actually get it displayed lol.

💟 
 
> EDIT: Generally also worth noting anytime you have the urge to say "oh document this behaviour" it probably actually indicates some kind of user interaction or design/communication failure, so we should try to isolate that failure and resolve that first, and documentation is a last resort. :)

True, hence the "once played around" yale-flag. 😄

yaleman (approved):
❤️

Firstyear:
Smol update for the dialogs in the consent page.

Do you think there is a need for a user to be able to clear their consents they've granted? And to list them?

kanidmd_web_ui/src/oauth2.rs
```diff
@@ -440,11 +440,33 @@ impl Component for Oauth2App {
...
+                    html! {
+                      <div>
+                        <p>{ "This site will not have access to your personal information" }</p>
+                        <p>{ "If this site requests personal information in the future we will check with you" }</p>
+                      </div>
+                    }
+                } else {
+                    html! {
+                      <div>
+                        <p>{ "This site has requested to see the following personal information" }</p>
```
yaleman, line 459:
```suggestion
                        <p>{ "This site has requested to see the following personal information:" }</p>
```

yaleman (commented):
Nits picked, the verbiage looks good so far 💟

PR #859, opened 2022-06-24
Update to account recovery UX

### LINKED ISSUE #825: recover_account  doesn't print password to stdout, and making it output in JSON

I'm trying to reset the password as part of doing some automation, and the CLI's outputting both logs and the password output to stderr, which makes that harder.

Ideally, an option to return the password/message as a JSON object like this (but on a single line):

```shell
kanidmd recover_account -n admin -c kanidmd --output json
```

Results in:

```json
{
  "result" : "success",
  "action" : "recover_account",
  "user" : "admin",
  "new_credential" : "<password>",
}
```

Implementing k8s secrets push from the `recover_admin` command, which .. is an integration far too deep into a particular ecosystem.

**Additional context**

Crabz R0ol. 🦀 ❤️

yaleman:
Fixes #825 

- changed the CLI to just accept the user, don't need --name
  ie: `kanidm recover_account admin` 
- added `--output` to CLI, only affects recover_account so far but ... more in the future.

I think it'll be beneficial to move the `ConsoleOutputMode` struct to the `kanidm` crate to allow it to be used across `kanidmd` and the `kanidm` CLI.

- [x] cargo fmt has been run

### DIFF
--- a/kanidmd/score/src/lib.rs
+++ b/kanidmd/score/src/lib.rs
@@ -27,19 +27,17 @@ extern crate kanidm;
 
 mod https;
 mod ldaps;
-use libc::umask;
 
 // use crossbeam::channel::unbounded;
+use async_std::task;
+use compact_jwt::JwsSigner;
 use kanidm::prelude::*;
-use std::sync::Arc;
+use libc::umask;
 
-use kanidm::config::Configuration;
-
-// SearchResult
-// use self::ctx::ServerCtx;
 use kanidm::actors::v1_read::QueryServerReadV1;
 use kanidm::actors::v1_write::QueryServerWriteV1;
 use kanidm::be::{Backend, BackendConfig, BackendTransaction, FsType};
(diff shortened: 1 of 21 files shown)


kanidmd/score/src/lib.rs
```diff
@@ -484,6 +483,42 @@ pub fn verify_server_core(config: &Configuration) {
     // Now add IDM server verifications?
 }
 
+// TODO: should this go somewhere else
```
Firstyear, line 486: I think it can stay with account recover here, but in the future I might split the core.rs into each function as a little bundle of joy.

Firstyear (approved):

kanidmd/score/src/lib.rs
yaleman: Yeah, the idea might be to move the JSON-related structs for messages out into the `kanidm` crate since that should limit duplication.

PR #878, opened 2022-07-02
adding compression middleware to tide

yaleman:
Fixes `#;-_-;#` - adding deflate/gzip compression support for the web server on the `/ui/` and `/pkg`/ paths.

This reduces the bytes on the wire for the WASM file from 1.4MB to ~400Kb, and if you're a goof like me and running the dev WASM then it goes 7MB -> ~2MB.

I don't believe this requires any additional dev packages to be installed, I reinstalled my laptop from scratch last night so it's pretty clean.

- [x] cargo fmt has been run


Firstyear:
I'm a bit worried about this, I remember that there are issues with compression inside of TLS. Doesn't it open us up to https://en.wikipedia.org/wiki/BREACH ?

Firstyear (approved):
Offline discussion, this is all good, pls add comments in the code about why this isn't security sensitive.

Firstyear (approved):

Issue #68, opened 2019-07-27
Implement replication


Firstyear:
Implement replication and all associated metadata. Initially this would probably target active active writeable servers, but read-only is intended for support later.

yaleman:
This is so freaking cool.

PR #908, opened 2022-07-06
68 replication changelog basics

Firstyear:
Relates #68 - This includes the fundamentals of the replication changelog into in memory areas of entries. Currently we do not persist anything to disk, and re-build this all at a restart so that if there are issues, we can simply restart and move on. This also implements changelog trimming so that we don't grow memory (disk) without bound, and verification checks to ensure the changelog indexes (ruv) is all consistent. 


yaleman (changes requested):
Only one thing to sort out, otherwise awweeeesssoommmmeeeee 💟

yaleman (approved):
Thanks! ❤️

PR #910, opened 2022-07-07
WebUI Cleanup, icons, Profile Wireframing

yaleman:
- account person extend was showing failure when succeeding
- first run on a user profile page, did some other CSS tweaks to the UI
- UI neatening, profile wireframing, robots.txt, PWA manifest
- strips out some old junk code

- [x] cargo fmt has been run

Loading screen. This is part of the base HTML of the root page, and loads the SVG logo which is ~14KB while the 2 (release) or 7.9MB (dev) WASM loads and starts up.

image

Profile screen wireframe, it's not functional but gives some basic idea of something that might work:

Screen Shot 2022-07-07 at 18 06 56


kanidmd/score/src/https/mod.rs
```diff
@@ -448,6 +476,8 @@ pub fn create_https_server(
         static_tserver.with(compress_middleware.clone());
 
         static_tserver.at("/").get(index_view);
+        static_tserver.at("/robots.txt").get(robots_txt);
+        static_tserver.at("/manifest.webmanifest").get(manifest);
```
Firstyear, line 480: What does a webmanifest do?

Firstyear (approved):

Firstyear:
Looks really good, I'm happy with it even as a placeholder as is, because we can fill in the rest as we go :)

kanidmd/score/src/https/mod.rs
yaleman: It defines a bunch of stuff around what happens when you open the site, then click share -> "add to home screen" on your phone - sets the icon to the nice login square and the domain display name 😄

Firstyear: Ooooooooooo that's freaking awesome, I'm sold.

Firstyear (approved):

PR #911, opened 2022-07-10
More UI things

yaleman:
Some more updates things

- Instead of `wasm_bindgen` creating a JS snippet to externalize code, we're now loading pure-JS util functions from `wasmloader.js` (`#[wasm_bindgen(raw_module = "/pkg/wasmloader.js")]`)
- Sign out is now a confirmation box instead of "oh no I have to log back in because I'm clumsy and clicked a thing"
- Now using the `urlencoding` crate for encoding the TOTP URLs because string replacing encoded characters felt like writing our own crypto (and now you can call yourself whatever arbitrary string you want)
  - This fixed an issue in the web UI where the "Add a TOTP" interface would show URL-encoded things, but also made things easier for consistency.
- Moved the other web middleware objects into the middleware module because the main module was getting a bit unwieldy.
- Started auto-generating the integrity hashes in a different way on start up, which removes a middleware doing rando string replacements to inject them, and means we can update modules without having to manually update the string values in the HTML.

Checklist:

- [x] cargo fmt has been run


kanidmd/score/src/https/middleware.rs
```diff
@@ -39,16 +189,12 @@ impl<State: Clone + Send + Sync + 'static> tide::Middleware<State>
                 we should be loading, and should be really secure about that!
 
             */
-            // TODO: consider scraping the other js files that wasm-pack builds and including them too
             "content-security-policy",
             vec![
                 "default-src 'self'",
+                // TODO: have a dev/test mode where we can rebuild the hashes on page load, so when doing constant JS changes/rebuilds we don't have to restart the server every time. It'd be *terrible* to run in prod because of the constant disk thrashing, but nicer for devs.
```
Firstyear, line 195: Or in dev, just don't send hashes at all .....

kanidmd_web_ui/src/macros.rs
```diff
@@ -0,0 +1,25 @@
+///! Macros for the web UI
+
+/// Adds a default set of CSS classes to the body element
+#[macro_export]
+macro_rules! add_ui_form_classes {
```
Firstyear, line 5: should say it's "add_body_ui_form_classes" I think?

kanidmd_web_ui/src/utils.rs
```diff
@@ -62,15 +62,7 @@ pub fn get_value_from_element_id(id: &str) -> Option<String> {
         .map(|element| element.value())
 }
 
-#[wasm_bindgen(inline_js = "export function modal_hide(m) {
-    var elem = document.getElementById(m);
-    var modal = bootstrap.Modal.getInstance(elem);
-    modal.hide();
-}")]
+#[wasm_bindgen(raw_module = "/pkg/wasmloader.js")]
```
Firstyear, line 65: Very smart :)

Firstyear (changes requested):
Looks really good, :)

kanidmd/score/src/https/middleware.rs
yaleman: yeah, I was pondering which way to go on it, I'll raise it as an issue as it's a side quest at this point

yaleman: Raised as #912

kanidmd_web_ui/src/macros.rs
yaleman: Met you in the middle with `add_body_form_classes` 😄 - ui is redundant identifier in context

Firstyear (approved):

PR #913, opened 2022-07-10
Tide route map things

yaleman:
Adds an endpoint which provides a full list of available routes on the server.

- [x] cargo fmt has been run

This is a super early version, dumps a JSON blob of path/method. There's only minor changes to how the routes are set up, so it's relatively self-maintaining. 

Example output here: https://gist.github.com/yaleman/4fff1994bb1aee8e3f0d401bcc3cdf5f

### DIFF
--- a/kanidmd/score/src/https/routemaps.rs
+++ b/kanidmd/score/src/https/routemaps.rs
@@ -1,10 +1,10 @@
+use crate::https::AppState;
 ///! Route-mapping magic for tide
 ///
 /// Instead of adding routes with (for example) the .post method you add them with .mapped_post, pasing an instance of [RouteMap] and it'll do the rest...
 ///
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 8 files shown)


kanidmd/score/src/https/routemaps.rs
```diff
@@ -0,0 +1,95 @@
...
+
+impl RouteMaps for Route<'_, AppState> {
+    // add a mapped method to the list
+    fn mapped_method(
+        &mut self,
+        routemap: &mut RouteMap,
+        method: http_types::Method,
+        ep: impl Endpoint<AppState>,
+    ) -> &mut Self {
+        // TODO: truly weird things involving ASTs and sacrifices to eldritch gods to figure out how to represent the Endpoint
```
Firstyear, line 33: ALL HAIL

Firstyear (approved):

PR #920, opened 2022-07-14
.deb packaging config and scripting

yaleman:
Changes

* fixes a race condition on the Docker container builds where they could stomp each other's platforms.
* adds all the scripting and Makefile changes and hair-tearing-out evidence that I've been building .deb's again.

Checklist:

- [x] cargo fmt has been run


Makefile
```diff
@@ -147,3 +151,48 @@ docs/pykanidm/serve:
 	cd pykanidm && \
 	poetry install && \
 	poetry run mkdocs serve
+
+########################################################################
+
+release/kanidm: ## Build the Kanidm CLI
+	cargo build -p kanidm_tools --bin kanidm --release
+
+release/kanidmd: ## Build the Kanidm daemon
```
Firstyear, line 160: The problem here is this will be missing the release profiles so it'll try to use the relatively placed wasm resources rather than the usr/share bits.

Firstyear: Oh and it's missing CPU optimisations.

platform/debian/kanidm-unixd/postinst
```diff
@@ -0,0 +1,34 @@
...
+# postinst script for kanidm-unixd
+#
+# see: dh_installdeb(1)
+
+set -e
+
+
+case "$1" in
+    configure)
+        if [ ! -f /etc/kanidm/config ]; then
```
Firstyear, line 11: won't this run multiple times for -ssh and -unixd?

platform/debian/kanidm-unixd/rules
```diff
@@ -0,0 +1,57 @@
...
+
+%:
+	dh $@
+
+override_dh_auto_clean:
+
+override_dh_autoreconf:
+
+override_dh_auto_build:
+	dh_auto_build -- release/${PACKAGE}
```
Firstyear, line 22: Is this relying on the make release/ from above?

platform/debian/kanidmd/control
```diff
@@ -0,0 +1,14 @@
+Source: kanidmd
```
Firstyear, line 1: I've always been uncertain about packaging the kanidmd binary on distros, given the docker container exists. The unixd and nsswitch pam, is unavoidable, but having only a "single way to run the server" is pretty attractive? Especially when there are so many ways to build it ...

platform/debian/kanidmd/rules
```diff
@@ -0,0 +1,55 @@
...
+
+override_dh_auto_clean:
+# cargo clean
+
+override_dh_autoreconf:
+
+override_dh_auto_build:
+	./platform/debian/fix_wasm_opt.sh
+	cd kanidmd_web_ui && ./build_wasm.sh
+	dh_auto_build -- release/kanidmd
```
Firstyear, line 27: Again is this using the makefile above? If it's for debian, it'll need a platform profile and release opts, ?

Firstyear (changes requested):

platform/debian/kanidm-unixd/postinst
yaleman: sure will, there's one post-inst script per .deb - hence the check for the config file

platform/debian/kanidm-unixd/rules
yaleman: yep, the dh tooling assumes makefiles in the build process

platform/debian/kanidmd/control
yaleman: Some folks can't or won't run it on Docker, and I'm not sure I want to even go through the nightmares of *actually* submitting the packages to be blessed by the Debian/Ubuntu distributions, knowing the burden that that entails. (I might look at the PPA thing, but... not for a while)

This is more about enabling folks and providing dev builds since Github Actions are free, and adding `kanidmd` was a pretty simple addition.

platform/debian/kanidmd/rules
yaleman: yep, will continue that discussion above

platform/debian/kanidmd/control
Firstyear: Yeah, that's fair. It's not me saying "remove it", but just airing a general thought process :)

platform/debian/kanidm-unixd/postinst
Firstyear: Are these all part of "one deb"? Or is -unixd and -ssh separet debs?

Makefile
yaleman: Sorted 😄 As discussed elsewhere, these are intended to be the patterns for packages, with the builders applying distribution/build specific flags in the build pipelines.

Makefile
```diff
@@ -147,3 +151,27 @@ docs/pykanidm/serve:
 	cd pykanidm && \
 	poetry install && \
 	poetry run mkdocs serve
+
+########################################################################
+
+release/kanidm: ## Build the Kanidm CLI
+	cargo build -p kanidm_tools --bin kanidm --release
```
Firstyear, line 158: If we are using this, we still should indicate in the comments that a build profile ushould be set?

platform/debian/kanidmd/rules
```diff
@@ -0,0 +1,54 @@
...
+%:
+	dh $@
+
+override_dh_auto_clean:
+# cargo clean
+
+override_dh_autoreconf:
+
+override_dh_auto_build:
+	KANIDM_BUILD_PROFILE=release_suse_generic dh_auto_build -- release/kanidmd
```
Firstyear, line 25: Should be set on all the make release/* targets

Firstyear (changes requested):
Small feedbacks but looking good!

platform/debian/kanidmd/rules
yaleman: Do you mean in the debian `rules` files so the CPU flags are set?

Makefile
yaleman: So, to clarify, for each make target, should I replace: 

```shell
release/kanidm: ## Build the Kanidm CLI
cargo build -p kanidm_tools --bin kanidm --release
```

with

```shell
release/kanidm: ## Build the Kanidm CLI - ensure you include an environment variable KANIDM_BUILD_PROFILE
cargo build -p kanidm_tools --bin kanidm --release
```

?

platform/debian/kanidmd/rules
Firstyear: Every call to `dh_auto_build -- release/target` needs the profile set in all the various packages, because as you say, it affects cpu flags and potentially in the future, other values.

Makefile
Firstyear: yes! :)

platform/debian/kanidmd/rules
yaleman: Cool, just confirming that's what you mean, my inferring_meaning engine is a bit broken lately.

Firstyear (approved):

PR #942, opened 2022-07-20
RADIUS fixes

yaleman:
- fixes the fact that the RADIUS certs wouldn't load if you were actually using the config file
- actually writing cert files properly now
- adding an environment variable to the container to fix a [python module bug](https://bugzilla.redhat.com/show_bug.cgi?id=1917157)
- tweaked GitHub actions so they only login to the container registry if they need to

Checklist:


kanidm_rlm_python/Dockerfile
```diff
@@ -51,6 +52,7 @@ RUN python3 -m pip install --no-cache-dir --no-warn-script-location /pkg/kanidmr
 RUN rm -rf /pkg/*
 
 USER radiusd
+ENV LD_PRELOAD=/usr/lib64/libpython3.so
```
Firstyear, line 55: I'm sure there is a reason for this but .... da fuq?

kanidm_rlm_python/entrypoint.py
```diff
@@ -44,22 +50,27 @@ def setup_certs(
     kanidm_config_object: KanidmClientConfig,
     ) -> None:
     """ sets up certificates """
-    # copy ca to /etc/raddb/certs/ca.pem
-    if kanidm_config_object.ca_path:
-        cert_ca = Path(kanidm_config_object.ca_path).expanduser().resolve()
+    print(kanidm_config_object)
+    # sys.exit(1)
```
Firstyear, line 54: Left over?

Firstyear (changes requested):

kanidm_rlm_python/entrypoint.py
yaleman: sure is, have fixed.

kanidm_rlm_python/Dockerfile
yaleman: Yeah, python changed something about how they link modules

Firstyear: Discussed externally, I'm going to submit a fix for freeradius-server to opensuse :)

kanidm_rlm_python/Dockerfile
```diff
@@ -2,6 +2,7 @@ FROM opensuse/tumbleweed:latest
 
 EXPOSE 1812 1813
 
+
 RUN zypper --gpg-auto-import-keys refresh --force
```
Firstyear, line 6: Can you add "zypper ar -f obs://home:firstyear:branches:network home:firstyear:branches:network " here before you install the free radius server?

Firstyear:
```
7ef3a792a088:/etc/zypp/repos.d # ldd /usr/lib64/freeradius/rlm_python3.so
	linux-vdso.so.1 (0x00007ffd77fdf000)
	libpython3.10.so.1.0 => /lib64/libpython3.10.so.1.0 (0x00007fda4c262000)
	libc.so.6 => /lib64/libc.so.6 (0x00007fda4c033000)
	/lib64/ld-linux-x86-64.so.2 (0x00007fda4c5c2000)
	libm.so.6 => /lib64/libm.so.6 (0x00007fda4bf4b000)
```

looks resolved in a test, submitted to opensuse

Firstyear (approved):

yaleman:
This fixes the code but the github actions thing is currently broken, I'll fix this elsewhere

PR #957, opened 2022-07-29
Improve radius to support eap-tls with ca-dir

Firstyear:
Adds support for eap-tls with radius, along with a check that the CN of the certificate has a valid UUID matching in the database. Also improves the types to have optionality and better defaults. I need to update the book still but that will be a second PR because there are some other book changes to make .... 

- [ ] cargo fmt has been run

### DIFF
--- a/kanidm_rlm_python/Dockerfile
+++ b/kanidm_rlm_python/Dockerfile
@@ -2,11 +2,10 @@ FROM opensuse/tumbleweed:latest
 
 EXPOSE 1812 1813
 
-# TODO: remove this once the freeradius python fix has been rolled into tumbleweed main
-RUN zypper ar -f obs://home:firstyear:branches:network home:firstyear:branches:network
-
-RUN zypper --gpg-auto-import-keys refresh --force
-RUN zypper install -y \
+# These all need to be on one line else the rpm cache ends
+# up in the layers.
+RUN zypper refresh --force && \
+    zypper install -y \
     freeradius-client \
     freeradius-server \
     freeradius-server-python3 \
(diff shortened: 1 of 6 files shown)


yaleman (commented):
Looks good, I assume this doesn't *enforce* TLS auth, just supports it?

yaleman:
Added the env fix since the freeradius package is still broken upstream...

Firstyear:
@yaleman Fixed your comments, and tweaked the configs, I have it working with both mschap and tls eap now :)

mcoms:
I'm trying to work out how to use this — [the docs mention mapping "special configuration" to users](https://github.com/kanidm/kanidm/blob/master/book/src/integrations/radius.md?plain=1#L28-L29), but it seems to be any valid certificate signed by the CA would work, so long as the certificate's [CN is the user's UUID](https://github.com/kanidm/kanidm/blob/master/pykanidm/kanidm/radius/__init__.py#L97-L105) and the user is in an allowed group; is that right? Do the client certificates also have to exist in `/etc/raddb/certs/`? Thanks!

Firstyear:
> I'm trying to work out how to use this — [the docs mention mapping "special configuration" to users](https://github.com/kanidm/kanidm/blob/master/book/src/integrations/radius.md?plain=1#L28-L29), but it seems to be any valid certificate signed by the CA would work, so long as the certificate's [CN is the user's UUID](https://github.com/kanidm/kanidm/blob/master/pykanidm/kanidm/radius/__init__.py#L97-L105) and the user is in an allowed group; is that right? Do the client certificates also have to exist in `/etc/raddb/certs/`? Thanks!

Ahh yes, I think we never really documented this.

* The user must be a member of the radius access allowed group
* The client ca certificate (and optionally it's crls) must be in the folder configured with `radius_ca_dir` 
* The client certificate must have a CN of the users UUID
* The client certificate must have the proper client extensions configured such as:
   -  Basic Constraints: CA FALSE
   - Subject Key Identifier
   - Authority Key Identifier
   - Key Usage (critical):   Digital Signature, Non Repudiation, Key Encipherment
   - Extended Key Usage: Client Authentication

PR #958, opened 2022-07-31
A pile of UI tweaks

yaleman:
- updates to a bunch of the UI elements, layout, prettiness etc.
- updated the workspace Cargo.toml to set the build options
- renamed the build_wasm_release.sh to build_wasm.sh and linked to it (reversing how things were, that poor balding yak)
- changed the WASM build to stop it building typescript files we don't use
- hiding a lot of console debug logging when in the release build
- rando test OAuth2 example stuff so one can test things

Checklist

- [x] cargo fmt has been run
- [x] cargo clippy has been run and I refuse to read the output

# OAuth "proceed" first view

Screen Shot 2022-07-31 at 22 34 54

# Cred reset prompt

Screen Shot 2022-07-31 at 22 38 47

# Cred management interface

Screen Shot 2022-07-31 at 22 39 16

# OAuth2 Interstitial

Screen Shot 2022-07-31 at 22 42 23

^ There's a version of the "consent" screen that is similar.


kanidmd_web_ui/src/credential/reset.rs
```diff
@@ -444,20 +462,22 @@ impl CredentialResetApp {
 
                     <hr class="my-4" />
 
+                    <h4>{"Password / MFA"}</h4>
                     { pw_html }
 
                     <hr class="my-4" />
 
+                    // TODO: this could probably just be a link back home, currently it breaks and sends the user to an error..
```
Firstyear, line 470: On cancel we need to send an abort to the server for the update session which also unlocks the intent token if it was used. So perhaps your server version needs update to accept that? It was cancelling correctly for me ....

Firstyear (approved):

kanidmd_web_ui/src/credential/reset.rs
yaleman: this was all tested with the current codebase 🤔

yaleman: If you click the "password and security settings" button from the dashboard to open the modal, then click cancel, it'll throw an error.

PR #970, opened 2022-08-05
UI updates

yaleman:
- user profile is functional now
- credential reset sessions will resume if you refresh the page and you've started one from the dashboard (a real pain if a CERTAIN browser does weird things with passkeys).
- smoothed out some of the credential reset UI, hiding fields when they're not used etc
- did battle with CSS again, trying to make things more consistent

Checklist:
- [x] cargo fmt has been run
- [x] cargo clippy has been run and is now silent in `kanidmd_web_ui`


kanidmd_web_ui/src/credential/passkey.rs
```diff
@@ -269,7 +269,9 @@ impl Component for PasskeyModalApp {
                                     Msg::Generate
                                 })
                         }
-                    >{ "Start Creating a New Passkey" }</button>
+                    >
+                    // TODO: they probably already want to create one if they've clicked the button but I won't judge for making really sure.
```
Firstyear, line 273: The extra buttons are needed because of bugs in safari and how fetching works.

kanidmd_web_ui/src/credential/totpmodal.rs
```diff
@@ -278,29 +279,36 @@ impl Component for TotpModalApp {
...
+                            Msg::TotpSubmit
+                        })
                     }
+                    type="button"
                 >{ "Submit" }</button>
             },
         };
 
         let totp_secret_state = match &self.secret {
+            // TODO: the user's already clicked start, why are we asking them to press it again? :D
```
Firstyear, line 298: Because we have to delay the fetch. The component is built on page load and hidden, so we have to wait for interaction to start a fetch and I'm too stoopid to work out how to make it "happen on display of the modal" so this dumb solution works.

kanidmd_web_ui/src/credential/totpmodal.rs
```diff
@@ -370,31 +380,50 @@ impl Component for TotpModalApp {
...
+                        TotpValue::Secret(secret)  => {
+                    html! {
+                        <form class="row g-3 needs-validation" novalidate=true
                         onsubmit={ ctx.link().callback(|e: FocusEvent| {
                             e.prevent_default();
                             Msg::TotpSubmit
                         } ) }
                     >
-                      <label for="totp" class="form-label">{ "Enter a TOTP" }</label>
+                    // TODO: the wording on this needs some improving.
```
Firstyear, line 393: Smoosh the codes into the apps or it gets the force quit again.

Firstyear (approved):

kanidmd_web_ui/src/credential/totpmodal.rs
yaleman: cool cool, I have ideas around this so will poke at it, having played with other bits recently ❤️

PR #971, opened 2022-08-05
Initial thoughts about oauth2 app listing

Firstyear:
Fixes #

- [ ] cargo fmt has been run

### DIFF
--- a/kanidm_book/src/developers/designs/oauth2_app_listing.md
+++ b/kanidm_book/src/developers/designs/oauth2_app_listing.md
@@ -1,0 +1,69 @@
+
+Oauth2 Application Listing
+==========================
+
+A feature of some other IDM systems is to also double as a portal to linked applications. This allows
+a convinent access point for users to discover and access linked applications without having to
+navigate to them manually. This naturally works quite well since it means that the user is already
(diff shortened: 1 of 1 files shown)


yaleman:
Looks sensible so far, thanks for putting it into words 😄 

## Implicit scope removal

For the migration to the new config and removal of implicit scopes, would it be relatively trivial to:

- create an `all_people` (or similar) group
- automagically add all existing `Person` accounts to it 
- add it as the access scope for any existing OAuth2 RS

For a one-off change during Alpha-mode it might be a load of work (and thus ... meh?) But would save a little surprise.

## Images

That's going to be fun 😄 I'd definitely go with PNG/SVGs from a size perspective and limit the size to... rather small either way. User images can be done in a similar way 🥳

Firstyear:
I've already done the all people migrate so that's done :) 

Now just to add the access controls, but for now I think all users can read the outh2 public config parts?

yaleman:
It does look that way - were we going to make the origin public?

Firstyear:
I think it has to be so we can present the link to the users to actually render and follow from the front page.

Firstyear:
I'll submit some acp changes for this today.

Issue #989, opened 2022-08-11
"version" output for all clients


yaleman:
It'd be handy to be able to run `kanidmd version` and get the version (and git commit, at compile). 

This'd help with identifying which specific code you're running, to work out how @yaleman managed to break it this time.

**Describe the solution you'd like**

Where possible for all CLI programs, run `<program> version` (or `<program> --version` for the unixd-related ones?) and get something like:

```shell
$ kanidmd version
1.1.0-alpha.9 (7e617a63)
```

Bin list:

- [x] kanidmd
- [x] kanidm
- [x] kanidm_unixd
- [x] kanidm_cache_clear
- [x] kanidm_cache_invalidate
- [x] kanidm_ssh_authorizedkeys
- [x] kanidm_ssh_authorizedkeys_direct
- [x] kanidm_unixd
- [x] kanidm_unixd_status
- [ ] kanidm_unixd_tasks
- [ ] orca
- [x] kanidm_test_auth?

Note:

The [git-version](https://crates.io/crates/git-version) crate supports showing the commit, but it only works when you're running it from within a git repository. It'd take some modifications to `build.rs` or other ... build-time curiosities to make it work.

Firstyear:
git-version has "fallback: If all else fails, this string will be given instead of reporting an error.". So we can embed something in there like the cargo/crate version I think. I think that's accessible via let version = env!("CARGO_PKG_VERSION"); per https://doc.rust-lang.org/cargo/reference/environment-variables.html

yaleman:
Updated OP to mention how git-version works (it only works if you're actually in the repository) and updated the list.

PR #991, opened 2022-08-14
Version argument for kanidm and kanidmd 

yaleman:
Relates to #989, implements `version` for kanidm and kanidmd.

The tweaks in `sketching` were me not being able to leave a clippy lint alone. :)

- [x] cargo fmt has been run

Example output:

```
# kanidmd version
kanidmd 1.1.0-alpha.9 (b8d09ab)
```

### DIFF
--- a/kanidm_proto/src/utils.rs
+++ b/kanidm_proto/src/utils.rs
@@ -1,0 +1,14 @@
+/// Shows the version string and current git commit status at build
+pub fn show_version(name: &str) {
+    println!("{}", get_version(name));
+}
+
+pub fn get_version(name: &str) -> String {
+    let version = env!("CARGO_PKG_VERSION");
+    let lgc = match last_git_commit::LastGitCommit::new().build() {
+        Ok(value) => value.id().short(),
+        Err(_) => String::from("Unknown git commit"),
+    };
+
+    format!("{} {} ({})", name, version, lgc)
+}
--- a/kanidm_proto/src/utils.rs
+++ b/kanidm_proto/src/utils.rs
@@ -5,10 +5,8 @@ pub fn show_version(name: &str) {
 
 pub fn get_version(name: &str) -> String {
     let version = env!("CARGO_PKG_VERSION");
-    let lgc = match last_git_commit::LastGitCommit::new().build() {
-        Ok(value) => value.id().short(),
-        Err(_) => String::from("Unknown git commit"),
-    };
-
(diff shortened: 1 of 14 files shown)


yaleman:
I don't know how I managed to break every. single. rust. build. but... it works on my laptop so somethin's weird.

Firstyear:
@yaleman You didn't break it, I did. :)

Firstyear:
@yaleman Actually, try a cargo update before you recommit?

yaleman:
> @yaleman You didn't break it, I did. :)

AHA! I was trying not to blame past-William, but I see he strikes again!

kanidm_proto/src/utils.rs
```diff
@@ -0,0 +1,14 @@
+/// Shows the version string and current git commit status at build
+pub fn show_version(name: &str) {
+    println!("{}", get_version(name));
+}
+
+pub fn get_version(name: &str) -> String {
+    let version = env!("CARGO_PKG_VERSION");
+    let lgc = match last_git_commit::LastGitCommit::new().build() {
+        Ok(value) => value.id().short(),
+        Err(_) => String::from("Unknown git commit"),
```
Firstyear, line 10: This will always display "unknown git commit" when packaged, so maybe something more subtle like a `--` ?

kanidmd/daemon/src/main.rs
```diff
@@ -447,6 +455,9 @@ async fn main() {
                     eprintln!("Running in vacuum mode ...");
                     vacuum_server_core(&config);
                 }
+                _ => {
```
Firstyear, line 458: Should match specifically on ::Version here, and just call to return;

Firstyear (changes requested):

kanidm_proto/src/utils.rs
yaleman: I was going for something that was baked in, like the CARGO_PKG_VERSION, which... will need more testing for sure.

yaleman: I wonder if I can sneak it in through a call in `build.rs`

Firstyear (approved):

Firstyear:
Is this ready to eview?

yaleman:
Yeah, it'll print the cargo version if you're running it normally and if it's running from within a git repo it'll print the current commit. It'll do until I work out a better way 😄

PR #994, opened 2022-08-16
Allow disabling networked tests in python client library

mweinelt:
Distros often build packages in sandboxes, which will make tests that
access the network fail. Marking these tests with a network mark is a
good way to allow downstreams to disable just this subset of tests.

Fixes tests of python client library test in package build sandboxes.

- [ ] cargo fmt has been run


yaleman:
Thanks!

PR #1008, opened 2022-08-29
change for KanidmClientConfig to make it public

yaleman:
I wanna use this in [kanidm-profiles](https://github.com/yaleman/kanidm-profiles), and it should be the canonical definition for a Kanidm client config file 😄 

Also, minor 🖇️-calming.

- [x] cargo fmt has been run

### DIFF
--- a/kanidm_client/src/lib.rs
+++ b/kanidm_client/src/lib.rs
@@ -61,13 +61,15 @@ pub enum ClientError {
 }
 
 #[derive(Debug, Deserialize)]
-struct KanidmClientConfig {
+pub struct KanidmClientConfig {
     uri: Option<String>,
     verify_ca: Option<bool>,
     verify_hostnames: Option<bool>,
     ca_path: Option<String>,
-    // Should we add username/pw later? They could be part of the builder
-    // process ...
+    #[allow(dead_code)]
+    username: Option<String>,
+    #[allow(dead_code)]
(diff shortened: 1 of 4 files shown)


kanidm_client/src/lib.rs
```diff
@@ -61,13 +61,15 @@ pub enum ClientError {
...
 #[derive(Debug, Deserialize)]
-struct KanidmClientConfig {
+pub struct KanidmClientConfig {
     uri: Option<String>,
     verify_ca: Option<bool>,
     verify_hostnames: Option<bool>,
     ca_path: Option<String>,
-    // Should we add username/pw later? They could be part of the builder
-    // process ...
+    #[allow(dead_code)]
```
Firstyear, line 69: I think we just get rid of these two fields for now since we don't use them today.

Firstyear (approved):
Minor comment, but looks good.

kanidm_client/src/lib.rs
yaleman: Maybe comment them out? I was going for 'the canonical definition for a Kanidm client config file' to try to keep `pykanidm` and the core def together. Happy either way.

Firstyear: We don't use them, so remove them and we can readd if needed

Firstyear (approved):

PR #1014, opened 2022-09-03
feat: add unix passwod reset to security web ui

theSuess:
Adds the functionality to change your own unix password. Currently, the button is displayed for all users as I did not want to copy the user fetching code from `profile.rs`. If anyone has an idea on how to solve this in a clean manner, I'd be happy to implement it.

- [x] cargo fmt has been run
- [x] cargo clippy has been run (only existing issues showed up)


theSuess:
Forgot to clear the input on success/failure.

Here's a short screenrecord on how it looks like

[pw_change_flow.webm](https://user-images.githubusercontent.com/1364576/188279288-2f3e4202-2ef0-4cec-b260-3a4c83f0d4e7.webm)

kanidmd_web_ui/src/views/security.rs
```diff
@@ -46,10 +51,12 @@ enum State {
     Init,
     Waiting,
     Error { emsg: String, kopid: Option<String> },
+    Success { msg: String },
 }
 
 pub struct SecurityApp {
     state: State,
+    unix_input_value: String,
```
yaleman, line 59: This feels like it should be an `Option<String>` just so we're never passing *any* string around unless it's actually in use.

kanidmd_web_ui/src/views/security.rs
```diff
@@ -111,6 +121,36 @@ impl Component for SecurityApp {
...
+                ctx.link().send_future(async {
+                    match Self::update_unix_password(id, token, newpw).await {
+                        Ok(v) => v,
+                        Err(v) => v.into(),
+                    }
+                });
+                true
+            }
+            Msg::UnixPasswordInput(mut inputvalue) => {
+                std::mem::swap(&mut self.unix_input_value, &mut inputvalue);
```
yaleman, line 153: Why not just use an assignment here? That feels weird.

yaleman (changes requested):
Thanks for submitting this so quickly 😄 

This definitely needs the usual "password" + "confirm password" boxes; even though it's trivial to update when the user goes "oh no I can't log in, let me fix it" it's one of those UI patterns that saves some headaches for everyone.

Regarding the "shows for everyone" thing, that's... something that needs pondering. There's a bunch of helper functions for this kind of introspection that need writing at the moment... 🤔

kanidmd_web_ui/src/views/security.rs
theSuess: I've taken this from the input field on the login page. I'm pretty new to this framework so I assumed it has a reason. Will have to change this anyways when switching to options.

kanidmd_web_ui/src/views/security.rs
theSuess: This became obsolete due to refactoring

kanidmd_web_ui/src/views/security.rs
theSuess: Also obsolete now

theSuess:
I've refactored the `ViewsApp` to be the owner of the profile data (this also allows us to display the username in the header, etc.).
Now the change password dialog is displayed conditionally

Currently, this is all realized using `props` but ideally we should introduce a global `Agent` (yew term for `Service` or `Store`) to take care of global state/api requests. But this could be the topic of a separate PR

Firstyear:
Well for now I'm happy for it to "display for everyone", but @yaleman is right, we'll need to refactor once we can integrate some better permissions of "what to display" into the user auth token.

kanidm_proto/src/v1.rs
```diff
@@ -324,7 +324,7 @@ impl fmt::Display for AuthType {
 ///
 /// It's likely that this must have a relationship to the server's user structure
 /// and to the Entry so that filters or access controls can be applied.
-#[derive(Debug, Serialize, Deserialize, Clone)]
+#[derive(Debug, Serialize, Deserialize, Clone, PartialEq, Eq)]
```
Firstyear, line 327: Why does this need eq?

kanidmd_web_ui/src/components/change_unix_password.rs
```diff
@@ -0,0 +1,237 @@
...
+                        ctx.link().callback(|e: FocusEvent| {
+                          e.prevent_default();
+                          let form = e.target().and_then(|t| t.dyn_into::<HtmlFormElement>().ok()).unwrap();
+                          Msg::Submit(FormData::new_with_form(&form).unwrap())
+                        })
+                      }
+                  >
+                    <div class="modal-content">
+                    <div class="modal-header">
+                        <h5 class="modal-title">{"Update your password"}</h5>
```
Firstyear, line 136: I think it would be good here to have the title to match saying "Update your unix password" along with some short text about what this password is used for. Something like "This password is used when logging into a unix-like system"

kanidmd_web_ui/src/views/security.rs
```diff
@@ -175,7 +188,7 @@ impl SecurityApp {
         opts.method("GET");
         opts.mode(RequestMode::SameOrigin);
 
-        let uri = format!("/v1/account/{}/_credential/_update", id);
+        let uri = format!("/v1/person/{}/_credential/_update", id);
```
Firstyear, line 191: 😅 Thanks for fixing this, I was in the middle of a large account refactor.

kanidmd_web_ui/src/components/change_unix_password.rs
```diff
@@ -0,0 +1,237 @@
...
+    type Properties = ChangeUnixPasswordProps;
+
+    fn create(_ctx: &Context<Self>) -> Self {
+        Self { state: State::Init }
+    }
+    fn update(&mut self, ctx: &Context<Self>, msg: Self::Message) -> bool {
+        match msg {
+            Msg::Submit(data) => {
+                let fd: FormValues = data.into();
+                if fd.password_input != fd.password_repeat_input {
```
Firstyear, line 69: Totally up to you if you want to do this, but there is a way to prevent submit being possible until both fields match, I think it's in the credential update code. If you want you can update to use a similar pattern here which saves the need to error on the fields not matching.

kanidmd_web_ui/src/components/change_unix_password.rs
```diff
@@ -0,0 +1,237 @@
...
+                        ctx,
+                        Msg::Error {
+                            emsg: "Password fields did not match".to_string(),
+                            kopid: None,
+                        },
+                    );
+                }
+                let tk = ctx.props().token.clone();
+                ctx.link().send_future(async {
+                    match Self::update_unix_password(tk, fd.password_input).await {
```
Firstyear, line 80: Worth checking the password update code in the other credential update path since I think this can return a password quality errer from zxcvbn, and maybe there is a way to share some code between the two?

Firstyear (changes requested):
Looks really good some minor comments and questions, and some suggestions for possible improvements, but I'd also be happy to merge and add the other improvements in a second PR too :) 

Please also add yourself to the CONTRIBUTORS.md file since you are contributing!

kanidm_proto/src/v1.rs
theSuess: We need this since we pass the `WhoamiResponse` as a property now (to enable conditional display of the dialog)

I'm not too sure of the implications but if you'd like this to stay the same, we can always add a new struct for the property (which will become obsolete once better state management is implemented in the UI)

theSuess:
> Well for now I'm happy for it to "display for everyone", but @yaleman is right, we'll need to refactor once we can integrate some better permissions of "what to display" into the user auth token.

With the latest changes, the button is displayed conditionally (see https://github.com/kanidm/kanidm/pull/1014#issuecomment-1236301628 for how it's implemented now and how it could be implemented in a better way)

kanidmd_web_ui/src/components/change_unix_password.rs
theSuess: Resolved. Also added a little message to the password change modal in the credentials update why the field is marked as invalid

kanidmd_web_ui/src/components/change_unix_password.rs
theSuess: While implementing the blocking of the submit button, I figured out just how similar the two modals are. I'll probably work on a shared component after this is merged.

theSuess:
After my last changes, I figured out that I just reimplemented the existing modal, so I'll work on merging these two after this PR is merged

kanidm_proto/src/v1.rs
Firstyear: I think it's okay, I was just curious.

Firstyear:
I think we just need you to rebase to latest master and then we can merge :)

theSuess:
The branch is rebased now. Should I also commit the built frontend (in pkg) or is there some kind of CI in place for that?

yaleman:
Please ensure the frontend WASM’s been built and committed :)

theSuess:
Done! I just ran the `./build_wasm.sh` script. Hope this is correct

Firstyear (approved):

Firstyear:
@yaleman Can you double check, your request change is blocking merge :)

yaleman (approved):

yaleman:
yolooooo

PR #1024, opened 2022-09-07
#896 kanidm_unixd UX updoots

### LINKED ISSUE #896: Test/force-run mode for kanidm_unixd
While testing recently I couldn't force kanidm_unixd to run as root, it'd be nice if we could do this. :smile:

- [ ] `--configtest` flag, to validate/dump config then exit
- [ ] `--force` flag to skip the "won't run as root" mode

yaleman:
Fixes #896 - Test/force run mode for kanidm_unixd

* Adds `--skip-root-check` to allow running as root
* Adds `--configtest` to dump the parsed config
* Adds some other bits which you can read in the help output below
* Moves some file paths into constants
* Fixes an issue where if the program hits an error and bails, it doesn't always log (due to tracing not being able to flush its messages before bailing)
  * **NOTE:** the daemon no longer exits with a code 1, I can't work out how to handle this cleanly in `tokio`'s event loop magic😞 
* Adds a [clap](https://crates.io/crates/clap)-based CLI interface:

```
kanidm_unixd 1.1.0-alpha.9
Kanidm Unix daemon

USAGE:
    kanidm_unixd [OPTIONS]

OPTIONS:
    -c, --client-config <client-config>
            Set the client config file path [env: KANIDM_CLIENT_CONFIG=] [default:
            /etc/kanidm/config]

    -d, --debug
            Show extra debug information

    -h, --help
            Print help information

    -r, --skip-root-check
            Allow running as root

    -t, --configtest
            Display the configuration and exit

    -u, --unixd-config <unixd-config>
            Set the unixd config file path [env: KANIDM_UNIX_CONFIG=] [default: /etc/kanidm/unixd]

    -V, --version
            Print version information
```

Checklist:
- [x] cargo fmt has been run


kanidm_unix_int/src/daemon.rs
```diff
@@ -374,9 +370,63 @@ async fn main() {
...
 
-    if cuid == 0 || ceuid == 0 || cgid == 0 || cegid == 0 {
-        eprintln!("Refusing to run - this process must not operate as root.");
-        std::process::exit(1);
+    let clap_args = Command::new("kanidm_unixd")
+        .version(env!("CARGO_PKG_VERSION"))
+        .about("Kanidm Unix daemon")
+        .arg(
+            Arg::new("skip-root-check")
+                .help("Allow running as root")
```
Firstyear, line 378: Needs a warning to say it shouldn't be used in production as it's a risk.

kanidm_unix_int/src/daemon.rs
```diff
@@ -374,9 +370,63 @@ async fn main() {
...
+        warn!("Skipping root user check, if you're running this for testing, ensure you clean up temporary files.")
+        // TODO: this wording is not great m'kay.
+    } else {
+        if cuid == 0 || ceuid == 0 || cgid == 0 || cegid == 0 {
+            error!("Refusing to run - this process must not operate as root.");
+            return;
+        }
+    };
+    if clap_args.get_flag("debug") {
+        std::env::set_var("RUST_LOG", "DEBUG");
```
Firstyear, line 429: minor nit, debug should be lower case.

kanidm_unix_int/src/daemon.rs
```diff
@@ -393,29 +443,23 @@ async fn main() {
...
 
-            let cfg_path = Path::new("/etc/kanidm/config");
-            let cfg_path_str = match cfg_path.to_str() {
-                Some(cps) => cps,
-                None => {
-                    error!("Unable to turn cfg_path to str");
-                    std::process::exit(1);
-                }
-            };
+            // we can do a blind unwrap because there's a default in clap
```
Firstyear, line 446: clippy will be sad,  you may need the #[allow(clippy::expect_used)]

Better to use expect in this case than unwrap.

kanidm_unix_int/src/daemon.rs
```diff
@@ -393,29 +443,23 @@ async fn main() {
...
-
-            if cfg_path.exists() {
+                return
+            } else {
                 let cfg_meta = match metadata(&cfg_path) {
                     Ok(v) => v,
                     Err(e) => {
                         error!("Unable to read metadata for {} - {:?}", cfg_path_str, e);
-                        std::process::exit(1);
+                        return
```
Firstyear, line 462: https://doc.rust-lang.org/std/process/struct.ExitCode.html

You can return exitcodes now!

kanidmd/idm/src/macros.rs
```diff
@@ -168,7 +168,7 @@ macro_rules! run_idm_test {
     }};
 }
 
-#[cfg(test)]
+// #[cfg(test)]
```
Firstyear, line 171: Oh, are these used somewhere else?

Firstyear (changes requested):
Looks good, some comments as usual :)

kanidm_unix_int/src/daemon.rs
yaleman: Sorted.

kanidm_unix_int/src/daemon.rs
yaleman: Do you mean in the set environment variable? It's lowercase in the highlighted line...?

Firstyear: Yeah, I meant in ('RUST_LOG', 'debug')

I still have NFI how to "add suggestions" that you can commit else I would have.

kanidmd/idm/src/macros.rs
yaleman: Yep, in `cargo bench`

Firstyear: Ahhhh okay. Maybe they should be #[cfg(test, bench)] ?

kanidm_unix_int/src/daemon.rs
yaleman: Click the line to highlight, then this widget in the comment box 😄 : Screen Shot 2022-09-08 at 10 29 16

Sorted!

kanidm_unix_int/src/daemon.rs
yaleman: Yas, but since this is hidden inside the tracing worker magic... then inside tokio's wrapped event loop thing, I can't work out what to do... :sad:

kanidm_unix_int/src/daemon.rs
```diff
@@ -393,29 +444,24 @@ async fn main() {
...
-
-            if cfg_path.exists() {
+                return
+            } else {
                 let cfg_meta = match metadata(&cfg_path) {
                     Ok(v) => v,
                     Err(e) => {
                         error!("Unable to read metadata for {} - {:?}", cfg_path_str, e);
-                        std::process::exit(1);
+                        return ExitCode::from(1)
```
Firstyear, line 464: Did you rebuilt/test? I think the return has to have every return path as exitcode and the main function signature needs to change too? my guess is this won't build as is.

yaleman: ah poop I didn't mean to include this in the most recent push >.>

kanidm_unix_int/src/daemon.rs
Firstyear: :(

Firstyear (approved):

kanidmd/idm/src/macros.rs
yaleman: that doesn't seem to be an option as far as I can tell? :(

PR #1043, opened 2022-09-14
feat: allow switch between spn and name for claims

### LINKED ISSUE #1023: Configurable value of `preferred_username` claim

When connecting to external systems using OIDC, the preferred_username attribute equals the full spn, including `@idm.example.com`. In my opinion, this results in "ugly" usernames for third party applications, so I would like to have this set to just the name.

Either introducing a configuration option or changing the default for the claim.

Introducing a claim-mapping system (seen for example in KeyCloak) using either a key<->value mapping or embedded scripting language. This seems too much for the issue at hand though

theSuess:
Fixes #1023 

I'll have little to no connectivity for the next two weeks, so I'll be slow to respond to feedback

### DIFF
--- a/kanidm_client/src/lib.rs
+++ b/kanidm_client/src/lib.rs
@@ -1755,6 +1755,29 @@ impl KanidmClient {
             .await
     }
 
+    pub async fn idm_oauth2_rs_prefer_short_username(&self, id: &str) -> Result<(), ClientError> {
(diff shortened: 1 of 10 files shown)


Firstyear (approved):

Firstyear:
I actually think it looks really good as is, you've done it perfectly! Great stuff!

Issue #996, opened 2022-08-21
WebUI - admin interface beginnings


yaleman:
As a user with admin privileges, I'd like to be able to look at the details of other users in the Web UI.

**Describe the solution you'd like**

 - [ ] add an "admin" link in the nav
 - [ ] present a list of options (accounts, groups, feral crab management, oauth2 integrations)
 - [ ] when clicking on "accounts", be prompted with a list of users
   - [ ] which when an account is clicked, will present a page using the same UI as the "profile" page for users
 - [ ] when clicking on "groups", be prompted with a list of groups
   - [ ] which when a group is clicked, will present a page showing details of the group
 - [ ] when clicking on feral crab management, be scolded for judging crabs this way, there's no such thing as a feral crab
 - [ ] when clicking on oauth2 applications, be presented a list of RPs
   - [ ] which when clicked will show details of the RP (the secret should maybe be hidden by default?)

**Describe alternatives you've considered**

Using the CLI but I like clicking things.

**Additional context**

Yeah, we were talking about changing the API with respect to people vs service accounts but I wanna click the things.

Firstyear:
So this is not super easy because:
* We need the server to indicate if a user can or can not "see" the admin ui link (which is kinda permission dependent but kinda not)
* The ability to show what you *can* and *cannot* edit (partially started via the permission test code as part of credential update)

It's been something in the back of my mind for a while, but we do need to really consider how to determine who would be shown what.

yaleman:
Yep, for sure - similar to the OAuth2 RP list thing, it's going to force some investment into these kinds of interface things. I was going to go with a "assume the user has access, and handle permissions errors when making the calls" approach.

Editing's a whole other problem/issue/bag of nightmares 🤣

TheRealGramdalf:
Just thought I'd add my two cents here:

- Spacedrive, another rust-based project, is using a [PPRTT stack](https://github.com/spacedriveapp/spacedrive#architecture) - I've heard good things about tauri, and it may be worth checking out
- Depending on various factors, I thought it might be worth going a route semi-similar again to portainer - but slightly different.
    - Keep Kanidm Server as it is right now - a standalone, primarily CLI server with a WebUI for SSO login and password reset
    - Use a separate client tool (WebUI, local desktop/mobile application, etc.) that interacts with the server directly with the API
This would make managing security much simpler because of the client/server nature of things - as long as the API is secure, the client tool will be inherently secure (given you manage cookies and whatnot properly, of course).
[...]

PR #1044, opened 2022-09-17
Initial admin UI things

yaleman:
Relates to #996. There's still a lot of work to be done, but opening the draft so people can look.

- adds the admin menu and some icons
- and the admin -> accounts list
- and the admin -> groups list

Checklist

- [x] cargo fmt has been run

Admin window:
Screen Shot 2022-09-17 at 11 34 14

Accounts list:
Screen Shot 2022-09-18 at 13 07 35

Groups list:
Screen Shot 2022-09-17 at 11 34 56


kanidmd_web_ui/src/components/admin_accounts.rs
```diff
@@ -0,0 +1,258 @@
...
+}
+
+/// Pulls all accounts (service or person-class) from the backend and returns a HashMap
+/// with the "name" field being the keys, for easy human-facing sortability.
+pub async fn get_accounts(token: &str) -> Result<AdminListAccountsMsg, GetError> {
+    // TODO: the actual pulling and turning into a BTreeMap in this and get_groups could *probably* be rolled up into one function? The result object differs but all the widgets are the same.
+    let mut all_accounts = BTreeMap::new();
+
+    // we iterate over these endpoints
+    let endpoints = [
```
Firstyear, line 69: I think there is still an /v1/account endpoint that does this for you.

kanidmd_web_ui/src/components/change_unix_password.rs
```diff
@@ -170,8 +177,10 @@ impl Component for ChangeUnixPassword {
                       onsubmit={
                         ctx.link().callback(|e: FocusEvent| {
                           e.prevent_default();
-                          let form = e.target().and_then(|t| t.dyn_into::<HtmlFormElement>().ok()).unwrap();
-                          Msg::Submit(FormData::new_with_form(&form).unwrap())
+                          #[allow(clippy::expect_used)]
+                          let form = e.target().and_then(|t| t.dyn_into::<HtmlFormElement>().ok()).expect("Failed to pull the form data from the browser");
```
Firstyear, line 181: We can probably swap to expect_throw here instead?

kanidmd_web_ui/src/views/profile.rs
```diff
@@ -1,3 +1,4 @@
+use crate::constants::CSS_PAGE_HEADER;
```
Firstyear, line 1: We could also consider importing and allowing a customised css file too somewhere? But I'm not smart enough at web things for that.

kanidmd_web_ui/src/components/admin_accounts.rs
```diff
@@ -0,0 +1,258 @@
...
+    // for filtering and pagination
+    // #[allow(dead_code)]
+    // search: Option<String>,
+    // #[allow(dead_code)]
+    // page: Option<u32>,
+}
+
+/// Pulls all accounts (service or person-class) from the backend and returns a HashMap
+/// with the "name" field being the keys, for easy human-facing sortability.
+pub async fn get_accounts(token: &str) -> Result<AdminListAccountsMsg, GetError> {
```
Firstyear, line 64: I think the more we put into  the webadmin like this, the more important it'll be we start to test it, because else there is a lot of risk of breakage especially around api changes.

kanidmd_web_ui/src/components/admin_accounts.rs
yaleman: Nope, 'tis gone. 😄 This allows me to really easily go "this is an x account type" too.

kanidmd_web_ui/src/components/change_unix_password.rs
yaleman: Fixed in bfc92d1b

kanidmd_web_ui/src/components/admin_accounts.rs
yaleman: Yep, that's going to be... fun :)

I've got some testing automation to build up (create 10K random accounts, to test how the API/UI blows up, etc) which is going to come after this PR's done, then I'll probably throw myself on the browser-testing grenade.

kanidmd_web_ui/src/views/profile.rs
yaleman: That's possible, and I can imagine some branding/UI customisation will come further down the track, but these constants are the classes that get applied to individual UI elements, so it'd still need to be defined somewhere.

kanidmd_web_ui/src/components/admin_accounts.rs
Firstyear: Well I actually don't think that'll be too bad, I think you can probably re-use some of the ideas from https://github.com/kanidm/kanidm/blob/master/kanidm_unix_int/tests/cache_layer_test.rs and use that to spin up a server and then test the "apis" here without actually needing a browser tester like selenium at all. That will cover a big chunk of the "wht can go wrong" at least. And I think yew can be tested from cargo test like this, which can cover the rest of the interactions. So I think it's probably not too bad. but we might need to consider breaking ` async fn setup_test` into it's own crate because it's a duplicate of the setup for proto_test in the server itself.

kanidmd_web_ui/src/views/profile.rs
Firstyear: yeah, zero pressure now :)

kanidmd_web_ui/src/components/admin_accounts.rs
Firstyear: Yep, you're right :) 

Well really they should be two seperate parts of admin ui anyway, a person admin and a robot/service account admin rather than "just accounts".

kanidmd_web_ui/src/components/admin_accounts.rs
yaleman: Yeah, that's fair - I guess that's one of the advantages of using rust/WASM, it *should* be largely testable outside a browser of some kind 😃

kanidmd_web_ui/src/components/admin_accounts.rs
yaleman: 🤔 I might argue that at the "list accounts" level:

- an account's an account - they've got usernames and auth methods and go into groups etc
- since they can (technically) move between `Service Account` <> `Person`, their root classification makes them similar enough to list them in one place
- having to iterate through two places for "something that authenticates through Kanidm" would be a little annoying for me as an admin

Obviously this is real early days for everything, but the difference between the two is also indicated visually:

Screen Shot 2022-09-18 at 13 05 41

Firstyear:
Okay, so I spent some time this morning thinking about this, and wanted to request some changes / propose a future proofing idea?

 Can we add an extra type to UserAuthToken called "ui_hints" which is a BTreeSet of a new type "permission marker". These would be things like Oauth2, Groups, Persons, ServiceAccounts etc.  The idea is that when that ui_hint is set in the btreeset, we show that link / set of admin uis. Currently we can hardcode this to set "all the ui hints" in all accounts, but later we can actually make it derive from group membership to access controls so we can have the ui elements show up / hide as needed.

Would that be too much to ask?

yaleman:
> Can we add an extra type to UserAuthToken called "ui_hints" which is a BTreeSet of a new type "permission marker". These would be things like Oauth2, Groups, Persons, ServiceAccounts etc. The idea is that when that ui_hint is set in the btreeset, we show that link / set of admin uis. Currently we can hardcode this to set "all the ui hints" in all accounts, but later we can actually make it derive from group membership to access controls so we can have the ui elements show up / hide as needed.

I was going to suggest a REST endpoint that responds with similar information and can be queried periodically to update the "permissions state" - either solution would work 😄 

I'll spin this out into its own discussion

Firstyear (approved):

PR #1071, opened 2022-09-23
20220911 api tokens

### LINKED ISSUE #990: LDAP service accounts with additional permissions / Access mail attribute via LDAP for Postfix
I'm trying to deploy Kanidm in a self-hosting/hobbyist scenario with a handful of users, as a replacement for our exising FreeIPA installation. We run our own mail server with user authentication and mailbox mappings provided through LDAP. As the `mail` attribute in Kanidm is not readable by the anonymous user, LDAP lookups cannot read it.

Since giving anonymous access to the `mail` attribute isn't exactly great from a security perspective, one option I've considered is the use of service accounts. Bind DNs are already usable for auth purposes in Kanidm, so maybe this could be extended to also include extra attributes the given user can read as long as the user has a "read_from_ldap" permission or something like that. This would also be useful for user federation when using IdPs like Keycloak.

Writing a Postfix plugin to use the Kanidm API directly? :skull:

Firstyear:
Fixes #990 - adds support for api tokens to service accounts. Generally lays a lot of the ground work for sessions, with tons of small improvements all over the codebase too. 


kanidm_proto/src/v1.rs
```diff
@@ -313,18 +314,21 @@ impl fmt::Display for AuthType {
...
 ///
 /// This structure and how it works will *very much* change over time from this
-/// point onward!
-///
-/// It's likely that this must have a relationship to the server's user structure
-/// and to the Entry so that filters or access controls can be applied.
-#[derive(Debug, Serialize, Deserialize, Clone, PartialEq, Eq)]
+/// point onward! This means on updates, that sessions will invalidate in many
+/// cases.
+#[derive(Debug, Serialize, Deserialize, Clone)]
```
yaleman, line 331: As discussed, we need to change expiry in UserAuthToken to be serialized as seconds since epoch.

[the time serde docs](https://docs.rs/time/latest/time/serde/index.html) indicate adding this should sort it:

```
 #[serde(with = "time::serde::timestamp")
 ```

yaleman (changes requested):
some thing needed on this one

yaleman (approved):

PR #1081, opened 2022-09-28
Python updoots

yaleman:
* Adds support for token auth
* Merged RADIUS module into a submodule of `pykanidm` so there's only one hive of villa... Python module in this repo.
* More tests, a prototype JWS parser to pull fields etc.
* Functionality to load/save/strip tokens from the user's store 
* Stripped out use of cookies


pykanidm/kanidm/__init__.py
```diff
@@ -218,13 +256,24 @@ async def auth_begin(
...
         if response.status_code != 200:
-            # TODO: write mocked test for this
+            # TODO: mock test for auth_begin raises AuthBeginFailed
             raise AuthBeginFailed(response.content)
 
         retobject = AuthBeginResponse.parse_obj(response.data)
         retobject.response = response
         return response
 
+    async def authenticate_with_token(self, token: str) -> ClientResponse:
```
Firstyear, line 266: I don't think we should rely on the cookie here. I think it was there as a "legacy" thing, but the tokens are getting bigger and we'll hit the 4096 byte limit soon. So authenticate_with_token should just set self.token and then you should use bearer from then on? That makes more sense here I think that doing an exchange step.

Firstyear (changes requested):

pykanidm/kanidm/__init__.py
yaleman: Yep, removing it as per our other discussions. The authentication will go through ... the other methods. :D

yaleman:
@Firstyear as far as I can tell this all works now, including testing in a RADIUS container. One day I'll get a full end-to-end test jig running for that 🙄

Firstyear (approved):

PR #1084, opened 2022-09-29
Fix issues with radius

Firstyear:
Also cleans some noise with logs

### DIFF
--- a/pykanidm/kanidm/__init__.py
+++ b/pykanidm/kanidm/__init__.py
@@ -165,6 +165,7 @@ class KanidmClient:
                 elif "authorization" not in headers:
                     logging.debug("Setting auth headers as Authorization not in keys")
                     headers.update(self._token_headers)
+            logging.debug("headers --> %s" % headers)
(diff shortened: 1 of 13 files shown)


pykanidm/kanidm/__init__.py
```diff
@@ -165,6 +165,7 @@ async def _call(
                 elif "authorization" not in headers:
                     logging.debug("Setting auth headers as Authorization not in keys")
                     headers.update(self._token_headers)
+            logging.debug("headers --> %s" % headers)
```
yaleman, line 168: I specifically wasn't logging this because it leaks auth tokens, and that's bad...

yaleman (changes requested):
Security note, but we also need to update the docs to reflect the change.

pykanidm/kanidm/__init__.py
Firstyear: Fixed!

Firstyear:
Doc fixes :)

yaleman (approved):
I touched it lasts, so I might as well approve it! 🦈

PR #1090, opened 2022-10-03
20221001 refactor

Firstyear:
A second huge cleanup, should tidy some modules, renames some libs, and removes some tech debt. 


kanidmd/daemon/Cargo.toml
```diff
@@ -18,9 +18,11 @@ name = "kanidmd"
 path = "src/main.rs"
 
 [dependencies]
-kanidm.workspace = true
+# Remove these!
```
yaleman, line 21: orly? like now, or later?

kanidmd/lib/src/macros.rs
```diff
@@ -736,3 +736,9 @@ macro_rules! vs_cid {
         x
     });
 }
+
+macro_rules! limmediate_warning {
```
yaleman, line 740: l'immediate? Je ne comprends pas le Français?

yaleman (approved):

kanidmd/daemon/Cargo.toml
Firstyear: It's in progress!

kanidmd/lib/src/macros.rs
Firstyear: Hah - it was an old leftover from log_immediate. It can be removed soon because tracing-forest has a handler for this. I guess even a 4 year old crab can have tech debt :)

kanidmd/daemon/Cargo.toml
yaleman: Didn't know if it was a leftover or not 😄

PR #1094, opened 2022-10-05
Stricter name rules

Firstyear:
Sadly there are too many systems that break with utf8 usernames, so we have to make this stricter. It doesn't affect legalname/displayname though. Makes the disallowed name check more efficient too. 


yaleman (approved):
🥳

PR #1098, opened 2022-10-07
Add test and comments about pkce

### LINKED ISSUE #1097: PKCE Downgrade Attack possible when PKCE enforcement mode not enabled
### Short description of issue

I think Kanidm's implementation of the authorization code flow is vulnerable to [PKCE Downgrade Attack](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-security-topics#section-4.8)

In order to not be vulnerable, the token endpoint must reject requests that contain a `code_verifier` but the `code` does not have an associated `code_challenge`. Currently kanidm doesn't do that but silently downgrades to not requiring PKCE at all.

I think the following patch should address the issue (not tested):

```patch
From 7d3661698cdb092e627f620c4198ff67a7d13de6 Mon Sep 17 00:00:00 2001
From: Arian van Putten <arian.vanputten@gmail.com>
Date: Thu, 6 Oct 2022 15:16:02 +0200
Subject: [PATCH] oauth2: check for presence of code_verifier before
 code_challenge

If a code_verifier is present but a code_challenge was not we should not
silently fall back to non-PKCE. If we do, we do not get any of the protections
that PKCE is supposed to provide.

This attack and mitigation is described in "4.8.  PKCE Downgrade Attack" in
https://datatracker.ietf.org/doc/html/draft-ietf-oauth-security-topics

> to prevent PKCE downgrade attacks, the AS MUST ensure that if there was no
code_challenge in the authorization request, a request to the token endpoint
containing a code_verifier is rejected.
---
 kanidmd/lib/src/idm/oauth2.rs | 8 ++++----
 1 file changed, 4 insertions(+), 4 deletions(-)

diff --git a/kanidmd/lib/src/idm/oauth2.rs b/kanidmd/lib/src/idm/oauth2.rs
index 0f690378..f45129e1 100644
--- a/kanidmd/lib/src/idm/oauth2.rs
+++ b/kanidmd/lib/src/idm/oauth2.rs
@@ -850,12 +850,12 @@ impl Oauth2ResourceServersReadTransaction {
                 })
             })?;
 
-        if let Some(code_challenge) = code_xchg.code_challenge {
+        if let Some(code_verifier) = token_req.code_verifier {
             // Validate the code_verifier
-            let code_verifier = token_req.code_verifier
+            let code_challenge  = code_xchg.code_challenge
                     .as_deref()
                     .ok_or_else(|| {
-                        security_info!("PKCE code verification failed - code challenge is present, but not verifier was provided");
+                        security_info!("PKCE code verification failed - code verifier is present, but not challenge was provided");
                         Oauth2Error::InvalidRequest
                     })?;
             let mut hasher = sha::Sha256::new();
@@ -870,7 +870,7 @@ impl Oauth2ResourceServersReadTransaction {
             }
         } else if o2rs.enable_pkce {
             security_info!(
-                "PKCE code verification failed - no code challenge present in PKCE enforced mode"
+                "PKCE code verification failed - no code verifier present in PKCE enforced mode"
             );
             return Err(Oauth2Error::InvalidRequest);
         }
-- 
2.37.0 (Apple Git-136)
```

main branch

Firstyear:
Fixes #1097. It was reported we were vulnerable to a pkce downgrade attack however, that may not be the case. The server provides a code_xchg (exchange) structure which is *encrypted* with a per-resource-server key. This code exchange is what contains the original code challenge. If the client alters or sends no code_verifier in it's token request, then the session is terminated with invalid request. In the case that the code_exchange for some reason is absent the code_challenge in it's encrypted blob AND the rs is configured to enforce pkce, then the session is also terminated. 

### DIFF
--- a/kanidmd/lib/src/idm/oauth2.rs
+++ b/kanidmd/lib/src/idm/oauth2.rs
@@ -471,6 +471,9 @@ impl Oauth2ResourceServersReadTransaction {
         }
 
         let code_challenge = if let Some(pkce_request) = &auth_req.pkce_request {
+            if !o2rs.enable_pkce {
+                security_info!(?o2rs.name, "Insecure rs configuration - pkce is not enforced, but rs is requesting it!");
+            }
             // CodeChallengeMethod must be S256
             if pkce_request.code_challenge_method != CodeChallengeMethod::S256 {
                 admin_warn!("Invalid oauth2 code_challenge_method (must be 'S256')");
@@ -874,6 +877,11 @@ impl Oauth2ResourceServersReadTransaction {
         } else if o2rs.enable_pkce {
             security_info!(
                 "PKCE code verification failed - no code challenge present in PKCE enforced mode"
+            );
+            return Err(Oauth2Error::InvalidRequest);
+        } else if token_req.code_verifier.is_some() {
+            security_info!(
+                "PKCE code verification failed - a code verifier is present, but no code challenge in exchange"
             );
             return Err(Oauth2Error::InvalidRequest);
         }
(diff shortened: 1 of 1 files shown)


Firstyear:
@arianvp This PR adds a test case to show the handling of the pkce. See the explanation above.

yaleman (approved):

Firstyear:
Thanks @yaleman I want @arianvp to have  alook first.

arianvp:
From your description i'm a bit skeptical if this demonstrates that kandim is not vulnerable.

the `code_verifier` is _not_ modified at any point; or ommited. The idea of the attack is as ofllows:


Mallory starts a flow without PKCE and gets a callback url back `https://kanidm?code=abc`

Alice starts a flow with PKCE (code_challenge=sha256(foo)) And is supposed to get a callback url back `https://kanidm?code=xyz` where `xyz` is indeed bound to `sha256(foo)`.

Mallory now either modifies the response from kanidm server (hard) or just delays the response from kanidm indefinitely (easy) and convinces Alice to click on the link `https://kanidm?code=abc`

Alice now makes a request to `/token` with `code=abc` and `code_verifier=foo`

Because `abc` is not bound to a code_challenge; kanidm accepts it, and now Alice has a session that is under Mallory's control.

If kanidm would bail out if a `code_verifier` is present whilst a `code_challenge` was not; it can detect this kind of attack.

Firstyear:
> From your description i'm a bit skeptical if this demonstrates that kandim is not vulnerable.
> 
> the `code_verifier` is _not_ modified at any point; or ommited. The idea of the attack is as follows:
> 
> Mallory starts a flow without PKCE and gets a callback url back `https://kanidm?code=abc`

If the rs configuration has enable_pkce=true, then it's not possible to start a flow without pkce, and will be rejected without a callback url being issued. 

> 
> Alice starts a flow with PKCE (`code_challenge=sha256(foo)`) And is supposed to get a callback url back `https://kanidm?code=xyz` where `xyz` is indeed bound to `sha256(foo)`.
> 
> Mallory now either modifies the response from kanidm server (hard) or just delays the response from kanidm indefinitely (easy) and convinces Alice to click on the link `https://kanidm?code=abc`

Code exchange tokens have a 60 second validity window.

> 
> Alice now makes a request to `/token` with `code=abc` and `code_verifier=foo`
> 
> Because `abc` is not bound to a `code_challenge`; kanidm accepts it, and now Alice has an `access_token` and `id_token` that is under Mallory's control even though Alice thought they logged in to their own account. Mallory is smart and will set up this profile to look like Alice's and Alice might be convinced to disclose information that they do intend to.

In this case kani would see code and code verifier don't match and reject the operation? 

> 
> If kanidm would bail out if a `code_verifier` is present whilst a `code_challenge` was not; it can detect this kind of attack.

The primary defence here is that if pkce_enable is true it is not possible to issue the code exchange in the first place, and that if the code verifier is not present this also fails.

I think to really prove this modifying or constructing a test case is needed here.

arianvp:
The attack is only possible if PKCE mode is not enforced. Correct.  It's only possible if `pkce_enable = false`.  So Kani definitely is not vulnerable in its default configuration.

I assumed that when `pkce_enable = false` PKCE is still possible; but only if the client decides to use it.  I do think that in this case it's still important that PKCE is correctly implemented.

> In this case kani would see code and code verifier don't match and reject the operation?

How? It never gets to this branch because the `code_xchg` does not contain a `code_challenge` so it never tests for the presence of `code_verifier`.

arianvp:
> I think to really prove this modifying or constructing a test case is needed here.

I will give a shot at this!

Firstyear:
> > I think to really prove this modifying or constructing a test case is needed here.
> 
> I will give a shot at this!

You can probably clone / modify the test case in the PR. I generally try to copy-paste a lot in tests tbh to make it clearer whats going on.

I think that this might only be possible when enable pkce = false, but if the rs can do pkce then why not enforce it? 

Perhaps the better check is "if you do pkce and enable pkce is false, then error and make noise to enable it"

arianvp:
> Perhaps the better check is "if you do pkce and enable pkce is false, then error and make noise to enable it"

hmm I'm also not sure if this is the best solution.  One would disable PKCE enforcement to support some legacy clients that don't implement it. But say you have multiple clients connected to kanidm; some implement PKCE and others don't. Then suddenly the more secure clients can not connect to kanidm anymore. 

I really think the flag should be about _enforcement_ not about disabling the feature altogether.

Firstyear:
> > Perhaps the better check is "if you do pkce and enable pkce is false, then error and make noise to enable it"
> 
> hmm I'm also not sure if this is the best solution. One would disable PKCE enforcement to support some legacy clients that don't implement it. But say you have multiple clients connected to kanidm; some implement PKCE and others don't. Then suddenly the more secure clients can not connect to kanidm anymore.
> 
> Especially many confidential clients don't implement PKCE whilst public clients almost always do
> 
> I really think the flag should be about _enforcement_ not about disabling the feature altogether.

I think it's important to point out the "clients" doing the pkce here are the relying servers, not the users browser. So it does kind of end up as a boolean, where your relying server etiher does or does not do pkce.

That's part of the benefit to splitting relying servers to have a seperate config for each one is rules like pkce can be enforced like then per application integration. Were this X number of varied applications per rs config, I think it could be different. 

We can probably fix it to "do the right thing" here, but it's also still a question of "if your rs can do pkce, why isn't it enforced?"

arianvp:
Aha that is exactly the last piece of context I missed. if this boolean is per Relying Party and not global then this a perfectly reasonable approach!

Firstyear:
> Aha that is exactly the last piece of context I missed. if this boolean is per Relying Party and not global then this a perfectly reasonable approach!

Okay, I still think we can improve this area of the code, so I'll do that later :)

Firstyear:
@arianvp Okay, I have improve this, where if a code verifier is submitted, but no code challenge is in the encrypted token, then we fail the exchange because "something is weird". The attack as per the rfc example isn't possible because that relies on stripping the code *challenge* not the code *verifier* from the request though, so our use of encryption and short lived tokens here has paid off!

Issue #967, opened 2022-08-04
OAuth2 `oauth2_rs_origin` - can we make this accessible to an authenticated user?


yaleman:
**Is your feature request related to a problem? Please describe.**

If I'm a pubbie I run `kanidm system oauth2 list` and get a list of OAuth2 apps. As `admin` I see the origin, but as `idm_admin` or `testuser` I only see displayname/Vec<class>/uuid. 

I'd like to be able to access the "Apps" page on the Web UI and get a list of apps, with a link to each one.

**Describe the solution you'd like**

Please to be tweaking:

- the ACPs so authenticated users can see the `oauth2_rs_origin` field of oauth2 objects.
- also a way of identifying which ones are *actually* accessible the current logged-in user, that'd be handy too - I'm guessing there's some kind of iterative check that can be done?

**Describe alternatives you've considered**

Rendering the list as an "internal" server-side thing but we really still need to be checking the user has access - Steve from accounting doesn't need to see the link for the super secret crab-cake making machine admin panel.

**Additional context**

Nah, not today.

Firstyear:
This has been on my radar for a bit. I have "ideas" about how to do this right :)

PR #1102, opened 2022-10-09
1063 967 oauth2 improvements

### LINKED ISSUE #1063: OAuth2 API endpoints return secrets by default
### Short description of issue

The following endpoints are query-able and return the `oauth2_basic_secret` field to users with permission.

* `/v1/oauth2`
* `/v1/oauth2/:id` (Where `:id` is currently the string-type `oauth_rs_name`)

I suggest this should require an explicit call to an API endpoint (such as `/v1/oauth2/:id/secret`, or `/v1/oauth2/:id/_attr/oauth2_basic_secret`) to get it. 

The driver for this is that I'm querying the current endpoints to return basic information for UI things, and... getting the secret should be an explicit call.

The only current dependency on this API element (I hope!) is the CLI, where it'd have to make an extra call in `kanidm system oauth2 get <name>` - and I suggest an invocation flag be added to *that* to return the key as well.

Current as of 2022-09-19

All

Firstyear:
Fixes #1063, works toward #967 - remove implicit scopes, add supplemental scopes, and make basic secrets secret by default. 


Firstyear:
Deployed and working in my environment too! Hooray for in place upgrades

yaleman (approved):
Looks good 🥳

PR #1111, opened 2022-10-10
Design doc for SCIM and migration planning

Firstyear:
Write out some initial thoughts about the migration processes and what needs to be developed. My thinking is the next step will be to start to sync from freeipa as a step of the process, especially to work out the testing flows that will be required. 

### DIFF
--- a/kanidm_book/src/developers/designs/scim_migration_planning.md
+++ b/kanidm_book/src/developers/designs/scim_migration_planning.md
@@ -1,0 +1,153 @@
+
+# Scim and Migration Tooling
+
+We need to be able to synchronise content from other directory or identity management systems.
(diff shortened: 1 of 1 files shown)


yaleman:
This is a great idea, from a bunch of perspectives. 

Extreme cat-amongst-the-pigeons comment: Adding SCIM support could also let us deprecate a chunk of the current REST API. SCIM is extensible and requires self-documenting schema, and supporting duplicate methods for the same thing seems redundant? 😃

Firstyear:
Not controversial at all, I actually already had similar thoughts.

yaleman:
Yep, for sure - but for the things SCIM can do, we might as well just do that 🥳

PR #1113, opened 2022-10-10
Improvements and notes adjacent to kanidm-unixd and packaging

kalebo:
This PR makes unixd-tasks populate the home directory using `/etc/skel`. This introduces one direct dependency that was already an indirect dependency (https://github.com/BurntSushi/walkdir).

This PR also includes a working pam configuration for debian derivatives leveraging the `pam-auth-update` tooling.

I've also included an alternative packaging script that can be used for improving debian packaging and a new page in the book for traefik integration. 

Let me know if there's any changes you would prefer or if I should break this PR in two three more specific PRs.

- [x] cargo fmt has been run
- [-] cargo clippy has been run
- [x] cargo test has been run and passes (test_password_from_samba_nt_hash and test_password_from_ipa_nt_hash fail but that is probably me missing some libraries and unrelated to the changes above.)

### DIFF
--- a/kanidm_book/src/integrations/pam_and_nsswitch.md
+++ b/kanidm_book/src/integrations/pam_and_nsswitch.md
@@ -45,6 +45,7 @@ You can also configure some unixd-specific options with the file /etc/kanidm/uni
     home_prefix = "/home/"
     home_attr = "uuid"
     home_alias = "spn"
+    use_etc_skel = false
     uid_attr_map = "spn"
     gid_attr_map = "spn"
 
@@ -72,6 +73,9 @@ Valid choices are `none`, `uuid`, `name`, `spn`. Defaults to `spn`.
 > in your system. We recommend that you have a stable ID (like the UUID), and symlinks
 > from the name to the UUID folder. Automatic support is provided for this via the unixd
 > tasks daemon, as documented here.
+
+`use_etc_skel` controls if home directories should be prepopulated with the contents of `/etc/skel` 
+when first created. Defaults to false.
 
 `uid_attr_map` chooses which attribute is used for domain local users in presentation. Defaults
 to `spn`. Users from a trust will always use spn.
--- a/kanidm_unix_int/Cargo.toml
+++ b/kanidm_unix_int/Cargo.toml
@@ -70,6 +70,7 @@ tokio-util = { workspace = true, features = ["codec"] }
 tracing.workspace = true
 reqwest.workspace = true
 users.workspace = true
+walkdir.workspace = true
 
 [features]
 # default = [ "libsqlite3-sys/bundled" ]
--- a/kanidm_unix_int/src/tasks_daemon.rs
+++ b/kanidm_unix_int/src/tasks_daemon.rs
@@ -12,6 +12,7 @@
 
 use std::ffi::CString;
 use std::os::unix::fs::symlink;
+use std::os::unix::ffi::OsStrExt;
 use std::path::Path;
 use std::time::Duration;
 use std::{fs, io};
@@ -29,6 +30,7 @@ use tokio::net::UnixStream;
 use tokio::time;
 use tokio_util::codec::{Decoder, Encoder, Framed};
 use users::{get_effective_gid, get_effective_uid};
+use walkdir::WalkDir;
 
 struct TaskCodec;
 
@@ -68,6 +70,17 @@ impl TaskCodec {
     }
 }
 
+fn chown(path: &Path, gid: u32) -> Result<(), String> {
+    let path_os = CString::new(path.as_os_str().as_bytes())
+        .map_err(|_| "Unable to create c-string".to_string())?;
+
+    // Change the owner to the gid - remember, kanidm ONLY has gid's, the uid is implied.
+    if unsafe { lchown(path_os.as_ptr(), gid, gid) } != 0 {
+        return Err("Unable to set ownership".to_string());
+    }
+    Ok(())
+}
+
 fn create_home_directory(info: &HomeDirectoryInfo, home_prefix: &str) -> Result<(), String> {
     // Final sanity check to prevent certain classes of attacks.
     let name = info
@@ -96,14 +109,10 @@ fn create_home_directory(info: &HomeDirectoryInfo, home_prefix: &str) -> Result<
         return Err("Invalid/Corrupt home directory path - no prefix found".to_string());
     }
 
-    let hd_path_os =
-        CString::new(hd_path_raw.clone()).map_err(|_| "Unable to create c-string".to_string())?;
-
     // Does the home directory exist?
     if !hd_path.exists() {
         // Set a umask
         let before = unsafe { umask(0o0027) };
-        // TODO: Should we copy content from /etc/skel?
         // Create the dir
         if let Err(e) = fs::create_dir_all(hd_path) {
             let _ = unsafe { umask(before) };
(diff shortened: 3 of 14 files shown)


yaleman (changes requested):
Thanks for the PR! 

I've added some fix suggestions which you can easily tweak/merge in the Github interface - or hand-edit if that's your thing 😃

Two other things:

To include the Traefik page in the book, please add the following line to `kanidm_book/src/SUMMARY.md`:

```markdown
- [Traefik](integrations/traefik.md)
```

Please feel free to add yourself to `CONTRIBUTORS.md` in the repository root!

kanidm_unix_int/Cargo.toml
```diff
@@ -70,6 +70,7 @@ tokio-util = { workspace = true, features = ["codec"] }
 tracing.workspace = true
 reqwest.workspace = true
 users.workspace = true
+walkdir = "2"
```
Firstyear, line 73: should be walkdir.workspace = true, and then in the top level Cargo.toml set in workspace.dependencies walkdir = "2"

Firstyear (changes requested):
Looks like a really good start, one suggestion could be a bool to control if /etc/skel is copied from at all?

Firstyear:
Thanks for the PR! Really appreciate you contributing. If you aren't already, please also ensure you add your details to the CONTRIBUTORS.md in tthe project root as part of this PR :)

platform/debian/simple_pkg.sh
```diff
@@ -0,0 +1,93 @@
...
+if [ 1 -eq 0 ]; then
+  mkdir -p deployment-config
+
+  cat << EOF > deployment-config/config 
+uri = "https://idm.example.com"
+verify_ca = true
+verify_hostnames = true
+EOF
+
+  cat << EOF > deployment-config/unixd
```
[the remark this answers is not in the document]
kalebo: These are in a disable section section of the script. They were meant for customizing the built package for a specific deployment. Should I just make the heredocs empty?

platform/debian/simple_pkg.sh
```diff
@@ -0,0 +1,93 @@
...
+mkdir -p "$pkgdir"
+
+# build the project
+make release/kanidm release/kanidm-unixd
+
+# enable the following block to include deployment specific configuration files
+if [ 1 -eq 0 ]; then
+  mkdir -p deployment-config
+
+  cat << EOF > deployment-config/config
```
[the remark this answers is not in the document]
kalebo: See other comment.

kanidm_unix_int/src/tasks_daemon.rs
```diff
@@ -12,6 +12,7 @@
 
 use std::ffi::CString;
 use std::os::unix::fs::symlink;
+use std::os::unix::prelude::OsStrExt;
```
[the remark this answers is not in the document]
kalebo: Thanks for catching this!

kalebo:
Thanks @Firstyear and @yaleman for reviewing!

> Looks like a really good start, one suggestion could be a bool to control if /etc/skel is copied from at all?

I started looking into adding it to the config file but I wasn't sure if the setting should be a bool, or a path that should be used or what the defaults should be. For myself, the least surprising behavior is to respect /etc/skel, but are there other considerations that need to weighed here in choosing the default?

kanidm_unix_int/src/tasks_daemon.rs
```diff
@@ -96,25 +109,37 @@ fn create_home_directory(info: &HomeDirectoryInfo, home_prefix: &str) -> Result<
...
         }
         let _ = unsafe { umask(before) };
 
-        // Change the owner to the gid - remember, kanidm ONLY has gid's, the uid is implied.
-
-        if unsafe { lchown(hd_path_os.as_ptr(), info.gid, info.gid) } != 0 {
-            return Err("Unable to set ownership".to_string());
+        chown(hd_path, info.gid)?;
+
+        // Copy in structure from /etc/skel/ if present
```
Firstyear, line 125: Won't this copy in every time the task is run? Shouldn't it only happen once?

kanidm_unix_int/src/tasks_daemon.rs
```diff
@@ -96,25 +109,37 @@ fn create_home_directory(info: &HomeDirectoryInfo, home_prefix: &str) -> Result<
...
+                let dest = &hd_path.join(
+                    entry
+                        .path()
+                        .strip_prefix(skel_dir)
+                        .map_err(|e| e.to_string())?,
+                );
+                if entry.path().is_dir() {
+                    fs::create_dir_all(dest).map_err(|e| e.to_string())?;
+                } else {
+                    fs::copy(entry.path(), dest).map_err(|e| e.to_string())?;
```
Firstyear, line 139: fs::copy will *overwrite* the content if it already exists, so this will nuke user changes on login.

Firstyear (changes requested):

Firstyear:
> I started looking into adding it to the config file but I wasn't sure if the setting should be a bool, or a path that should be used or what the defaults should be. For myself, the least surprising behavior is to respect /etc/skel, but are there other considerations that need to weighed here in choosing the default?

I think initially a bool is sufficient, such as "use_home_dir_template = true/false". Defaults to false if unset. Remember we can always change the option in the future if we feel it's needed!

kanidm_unix_int/src/tasks_daemon.rs
kalebo: It shouldn't since this is all within the if-block checking `!hd_path.exists()`. Is there something missing in that logic?

kanidm_unix_int/src/tasks_daemon.rs
kalebo: Overwriting shouldn't be an issue if this only fires when the canonical home directory for a user isn't present as noted above

platform/debian/simple_pkg.sh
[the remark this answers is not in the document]
kalebo: I changed the heredocs to make it more obvious they are just placeholders that would need to be customized if that block is enabled.

kanidm_unix_int/src/tasks_daemon.rs
Firstyear: Oh you're right, my mistake :)

kanidm_unix_int/src/tasks_daemon.rs
Firstyear: Yep, my mistake :)

yaleman (approved):

Firstyear:
@kalebo hey there, the boolean for enable/disable the copying from /etc/skel is still missing, did you need some help to implement/add that?

kalebo:
Sure thing, I just added the configuration toggle.

kanidm_book/src/integrations/pam_and_nsswitch.md
```diff
@@ -73,6 +74,9 @@ Valid choices are `none`, `uuid`, `name`, `spn`. Defaults to `spn`.
 > from the name to the UUID folder. Automatic support is provided for this via the unixd
 > tasks daemon, as documented here.
 
+`use_etc_skel` controls if home directories should be prepopulated with the contents of `/etc/skel`
```
Firstyear, line 77: Should default to false IMO

Firstyear (changes requested):
If you make it default to false, then I think this is done :)

kanidm_book/src/integrations/pam_and_nsswitch.md
kalebo: Fixed

Firstyear (approved):

Firstyear:
Thanks so much @kalebo, appreciate your patience with us :)

PR #1119, opened 2022-10-11
Systemd Unit Files in Deb packages

### LINKED ISSUE #1093: Create systemd unit files for ubuntu packages

While building all the packaging bits I didn't make systemd unit files. Weird.

Copypasta them from elsewhere

Be a big meanie and not do it

**Additional context**

I declare hugs for everyone and small pieces of cake for anyone that doesn't want hugs. If you don't want hugs or cake, then sorry?

yaleman:
Fixes #1093, adds systemd service files for `kanidmd`, `kanidm-unixd` and `kanidm-unixd-tasks` in their respective folders

### DIFF
--- a/examples/systemd/kanidmd.service
+++ b/examples/systemd/kanidmd.service
@@ -1,0 +1,20 @@
+# You should not need to edit this file. Instead, use a drop-in file by running:
+#   systemctl edit kanidm-unixd-tasks.service
+
+[Unit]
+Description=Kanidm, the IDM for rustaceans
+After=network-online.target
+Wants=network-online.target
+
+[Service]
+Type=simple
+
+ExecStart=/usr/local/sbin/kanidmd server --config=/etc/kanidm/server.toml
+Restart=on-failure
+RestartSec=15s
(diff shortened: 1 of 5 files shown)


examples/systemd/kanidmd.service
```diff
@@ -0,0 +1,20 @@
+# You should not need to edit this file. Instead, use a drop-in file by running:
```
Firstyear, line 1: I think there are a lot of hardenings missing from https://github.com/kanidm/kanidm/blob/master/platform/opensuse/kanidmd.service

Firstyear (changes requested):

examples/systemd/kanidmd.service
yaleman: The challenge was that a lot of them weren't available on Ubuntu/Debian because their systemd versions weren't as bleeding-edge as SUSE seem to be.

Firstyear: Ahhhhhh! All good then :)

Firstyear (approved):

Issue #1115, opened 2022-10-11
User "sudo" or elevated permission mode


Firstyear:
Users should need to "re-authenticate" to access privileges in their account. This is similar to githubs sudo mode. This will allow longer lived sessions but with short-lived permission access.

PR #1120, opened 2022-10-12
20221011 sudo mode

Firstyear:
partial #1115 adds the "token" components of the sudo/priv mode. This allows an internal field on tokens denoting the purpose or priveleges of the session. Currently we provision all users with read write, but in the future we'll assign them readwrite but expired, and then do a session re-auth to enable "priv" mode. Since these are in the tokens, we enable these for api tokens as well.

### DIFF
--- a/kanidmd/lib/src/access.rs
+++ b/kanidmd/lib/src/access.rs
@@ -515,8 +515,7 @@ pub trait AccessControlsTransaction<'a> {
 
         match se.ident.access_scope() {
             AccessScope::IdentityOnly | AccessScope::Synchronise => {
-                security_access!("identity access scope is not permitted to search");
-                security_access!("denied ❌");
+                security_access!("denied ❌ - identity access scope is not permitted to search");
                 return Ok(vec![]);
             }
             AccessScope::ReadOnly | AccessScope::ReadWrite => {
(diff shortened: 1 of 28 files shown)


kanidmd/lib/src/access.rs
```diff
@@ -514,7 +511,18 @@ pub trait AccessControlsTransaction<'a> {
             }
             IdentType::User(u) => &u.entry,
         };
-        trace!(event = %se.ident, "Access check for search (filter) event");
+        info!(event = %se.ident, "Access check for search (filter) event");
+
+        match se.ident.access_scope() {
+            AccessScope::IdentityOnly | AccessScope::Synchronise => {
+                security_access!("identity access scope is not permitted to search");
```
yaleman, line 518: Does this need to be a double log? Why not just 

`"denied ❌ - identity access scope is not permitted to search"`

yaleman (changes requested):
Logging locally into my kanidm server's a challenge because journald's choking on my (not that many) endpoints... double-logging denials seems bad 👎🏻

Firstyear:
Fixed :)

yaleman (approved):

PR #1153, opened 2022-10-30
Improve cookie/token handling

### LINKED ISSUE #1082: Cookies and Tokens are a bit ambiguous in usage
We currently issue both. The reason is a number of rust libraries don't properly support cookies, or rather persisting them in some fashion. As a result we use tokens in these cases. We extended that to tokens in browser, but that may expose us to xss attacks.

To resolve this we need clearer patterns about cookie and token issuance. This is tied into #406 because the ability to have sessions in the DB will allow us to create session cookies that are MUCH smaller which avoids the size limits we have previously worried about.

That would leave:

* Tokens as the primary method for auth for CLI/rust based client tools and utils
* Tokens as the primary auth for python
* Cookies for wasm based authentication in the UI (which also makes that code a lot simpler too). 

We will need a "reflector" that allows the wasm to get a non-signed UAT for replacing the lack of client side token introspection now.

Firstyear:
Fixes #1082 - this makes the ui use cookies exclusively while leaving tokens alone for the cli/apps. This changes some of the way the ui works to ensure the uat will fit into a single cookie. Further changes are likely needed in the future. 


yaleman (approved):
Woo, more cookies less bearer token BS. I like 🥳

yaleman:
Ignore those python fails, I think it was some dependency weirdness.

PR #1166, opened 2022-11-01
Further test improvements

Firstyear:
Further test improvements and work toward removing async_std. This adds the idm_test proc macro, which allows async and fixtures in idm tests. I still need to think about how to add the plugin tests after this. I didn't convert every test over yet, that's future williams problem. 


yaleman:
I like good tests and I cannot lie 🥳

PR #1175, opened 2022-11-07
Add version header and warnings

Firstyear:
Relates #989 add client / server version mismatch warning. 

### DIFF
--- a/kanidm_client/src/lib.rs
+++ b/kanidm_client/src/lib.rs
@@ -30,7 +30,7 @@ pub use reqwest::StatusCode;
 use serde::de::DeserializeOwned;
 use serde::{Deserialize, Serialize};
 use serde_json::error::Error as SerdeJsonError;
-use tokio::sync::RwLock;
+use tokio::sync::{Mutex, RwLock};
 use url::Url;
 use uuid::Uuid;
 use webauthn_rs_proto::{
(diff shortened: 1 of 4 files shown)


yaleman:
Checking for my head-canon - if the header's not returned by the server - the client will assume it's "" and complain about it?

It's going to be pretty common for folks to strip this header from server responses as OWASP and other security frameworks will have kittens over exposing server version information in headers.

Firstyear:
> Checking for my head-canon - if the header's not returned by the server - the client will assume it's "" and complain about it?

Yep, if not there, assume bad. 

> 
> It's going to be pretty common for folks to strip this header from server responses as OWASP and other security frameworks will have kittens over exposing server version information in headers.

The problem is that we expose the version in other ways, there will always be ways to fingerprint the server. If we add a special URL to detect this, that's still showing the version. We can't hash etc because both sides need to know what the expected values are. And really, an attacer could find the version by working out what protocol bits we support or looking at the current system/domain info versions. So I'm not sure if there really is a better way here?

yaleman:
Maybe "if not there, then then just don't mention it" because after a very short number of releases, the "old" versions without the header *should* have aged out.

9SMTM6:
I would at the very least still provide a warning of there was no header present. People can then, when somethings going wrong, consider looking up the versions in some other way, as I did, but I did that shortly after reading about versionmismatches in your book, with some time between IDK if that wouldve been among my first guesses.

9SMTM6:
Yeah, for the old versions before this gets merged this means no "proper" errors, but thats a temporary issue.

9SMTM6:
However you decide on that particular topic, checking for a version match programmatically is a great approach, thanks @Firstyear .

Firstyear:
> Maybe "if not there, then then just don't mention it" because after a very short number of releases, the "old" versions without the header _should_ have aged out.

I disagree, because if the server isn't showing a header then it's "older" so we are again in the mismatch scenario. The only concern I think here that's valid is header filtering, but we'll break if people filter out the opid and such anyway. 

I think I want to merge as is, and if we hit problems, then we change the approach?

yaleman:
Sure, sounds good to me

PR #1185, opened 2022-11-13
1116 UI hint

Firstyear:
Start of  #1116 - This reworks how acp's target receiving entries. We need this because during account uat issuance, we don't want to have to evaluate all the acp filters against the current entry as that will severly impact authentication performance. Instead, we update ACP such that the receivers are always group bound, meaning we can do a UUID lookup into the memberof set. There are likely still improvements here since currently in the way we approach this in the acces eval code we iterate over acp and test for memberof, but the inverse would likely be faster (or we could do a zip / filter). 

PS: I already tested that in place upgrades work flawlessly. As a bonus this PR improves some early boot phase logging to make it easier to pinpoint error locations. 


yaleman (approved):

yaleman:
I've started an [issue on the product discussion forum](https://github.com/orgs/community/discussions/38975) about all these failures due to network failures on actions 😡

PR #1225, opened 2022-11-23
Add ability to bind with token as a dn

### LINKED ISSUE #1208: LDAP Bind to service account with non-empty bind dn
Despite solving how to generally read mail attributes as described in https://github.com/kanidm/kanidm/issues/1207, it still doesnt work in Authelia, and perhaps also gitea with ldap.

First of all, were speaking about binding with access token from a service account to LDAP.

Unsure about gitea - what I'm noticing there is that searches etc get logged coming from anonymous - but empty bind dn is accepted in the form and appears to correctly bind. 

Authelia REQUIRES a non-empty bind dn, at least Ive not been able to get it to accept some variant of empty in the config. 

I've not had enough time to test all possible avenues, but I've experimented a decent bit, and havent got it to work.

Thus I would ask If it would be possible to also allow to provide the dn of the serviceaccount when using a token to log in with it.

I've not tried it with service account PWs as these are deprecated, but perhaps that would be a reason to keep it if the above is not possible. (EDIT: Right, that credential is also not allowed to access LDAP, at least thats what the help text on the command says).

I've now also tried all kinds of variations with ldapsearch and yet to find a way to make tokens work with a non empty bind dn.

EDIT2: Puh that was barely comprehensible, fixed a few things and highlighted that were speaking about access tokens, yaleman got it but that makes it clearer for others. 

-

-

**Additional context**

https://www.authelia.com/configuration/first-factor/ldap/#user

Firstyear:
Fixes #1208 - add an extra syntax to allow tokens in the binddn with the syntax "token=APITOKEN". We still allow the anonymous bind dn + token syntax as well. 

### DIFF
--- a/kanidmd/lib/src/ldap.rs
+++ b/kanidmd/lib/src/ldap.rs
@@ -56,7 +56,6 @@ pub struct LdapServer {
     basedn: String,
     dnre: Regex,
     binddnre: Regex,
-    tokenre: Regex,
 }
 
 impl LdapServer {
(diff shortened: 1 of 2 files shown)


Firstyear:
Fixed, there is now a "magic" bind dn of the value "dn=token" which treats the pw as a token.

yaleman:
Dear Oauth tests plz stop breaking

yaleman (approved):

Issue #735, opened 2022-05-02
Import users from 389-ds


Firstyear:
Support importing accounts from 389-ds

Issue #734, opened 2022-05-02
Import users from slapd


Firstyear:
Develop tooling to help import users from slapd - see openldap_to_ds for prior work in this area

mweinelt:
I tried getting the FreeIPA exporter to work with slapd, which led me to 

```
$ kanidm-ipa-sync -i ipasync.conf 
2023-06-06T13:15:49.593130Z ERROR kanidm_ipa_sync: Unsure how to handle refreshDeletes=True
2023-06-06T13:15:49.595217Z ERROR kanidm_ipa_sync: Sync completed with error e=Preprocess
```

I also had to remove filter parts, that lead to a filter including `(?expr)` pieces. Slapd does not seem to like those and returned no results.

slapd[72334]: conn=1003 op=0 BIND dn="cn=kanidm,dc=darmstadt,dc=ccc,dc=de" mech=SIMPLE bind_ssf=0 ssf=256
slapd[72334]: conn=1003 op=0 RESULT tag=97 err=0 qtime=0.000006 etime=0.065301 text=
slapd[72334]: conn=1003 op=1 SRCH base="ou=people,ou=accounts,dc=darmstadt,dc=ccc,dc=de" scope=2 deref=0 filter="(|(&(objectClass=person)(?objectClass=ipantuserattrs)(objectClass=posixAccount))(&(objectClass=groupOfNames)(?objectClass=ipausergroup)(!(?objectClass=mepmanagedentry))(!(cn=admins))(!(cn=ipausers)))(&(?objectClass=ipatoken)(?objectClass=ipatokentotp)))"
slapd[72334]: conn=1003 op=1 syncprov_op_search: got a search with a cookie=(null)
slapd[72334]: conn=1003 op=1 syncprov_search_response: cookie=rid=000,sid=002,csn=20220415194814.647229Z#000000#000#000000;20230605210722.820545Z#000000#001#000000;20230606131229.016435Z#000000#002#000000
slapd[72334]: conn=1003 op=1 SEARCH RESULT tag=101 err=0 qtime=0.000017 etime=0.000919 nentries=0 text=
slapd[72334]: conn=1003 fd=19 closed (connection lost)


After that I saw kanidm-ipa-sync do a search and get a reasonable number of users back.

```
conn=1030 op=1 SEARCH RESULT tag=101 err=0 qtime=0.000047 etime=0.002589 nentries=95 text=

```

The `refreshDeletes` issue above remains and needs more investigation.

[...]

PR #1258, opened 2022-12-13
20221123 iam migration work

### LINKED ISSUE #736: Import accounts and groups from freeipa
Support importing accounts and groups from freeipa

### LINKED ISSUE #1114: Marking of entries as "protected" during sync phase, remove protection in final sync
Entries that are still under the control of an external IDM authority should not be altered by Kanidm. We may allow "overrides" in the case that entries in a specific sync domain can have a subset of attrs altered?

Firstyear:
Fixes #736 Fixes #1114 Works toward #734 and #735. This adds support for freeipa to kanidm sync via a helper tool that can translate ldap sync repl operations into scim post requests. 

The scim request is "non-standard" but uses scim entries at it's core. It asserts a state over an entry, and allows a set of attributes to be delegated permission to kanidm which kanidm can control. For the current state, changes are blocked to these entries. 

Now that the tool and gateway exists, it will be easier to add other helper tools for other ldap deployments.


Firstyear:
For the record, I'll do a second follow up PR with book chapters and improvements as I want to use the tool "in anger" on my instance first.

yaleman (approved):

Firstyear:
Added the design doc reasoning.

PR #1259, opened 2022-12-13
fix(oauth2): use the short name in the userinfo

theSuess:
This implements #1023 for the userinfo endpoint, which was an oversight in #1043. I've also added a testcase to verify this behavior

- [x] cargo fmt has been run


yaleman:
LGTM, but @Firstyear has more insight into this flow

Firstyear (approved):

Firstyear:
Thanks @theSuess !

PR #1272, opened 2022-12-17
feat(oauth2): add support for a 'groups' claim

theSuess:
Adds support for a 'groups' claim. If supplied, it will add the UUID of the accounts groups to the id_token and userinfo endpoint.

Since we had a lot of duplicate logic between token exchange and userinfo, I also refactored out the common parts.

As discussed in the chat, I'll also add a toggle for the resource server to only provide this information in the userinfo endpoint and not the ID token, just wanted to submit this already to get some early feedback if this change is going in the right direction

- [x] cargo fmt has been run

### DIFF
--- a/kanidmd/lib/src/idm/account.rs
+++ b/kanidmd/lib/src/idm/account.rs
@@ -156,7 +156,7 @@ impl Account {
     #[instrument(level = "trace", skip_all)]
     pub(crate) fn try_from_entry_ro(
         value: &Entry<EntrySealed, EntryCommitted>,
-        qs: &mut QueryServerReadTransaction,
+        qs: &QueryServerReadTransaction,
     ) -> Result<Self, OperationError> {
         let groups = Group::try_from_account_entry_ro(value, qs)?;
         try_from_entry!(value, groups)
--- a/kanidmd/lib/src/idm/oauth2.rs
+++ b/kanidmd/lib/src/idm/oauth2.rs
@@ -34,6 +34,7 @@ use tracing::trace;
 use url::{Origin, Url};
 
 use crate::identity::IdentityId;
+use crate::idm::account::Account;
 use crate::idm::delayed::{DelayedAction, Oauth2ConsentGrant, Oauth2SessionRecord};
 use crate::idm::server::{
     IdmServerProxyReadTransaction, IdmServerProxyWriteTransaction, IdmServerTransaction,
@@ -955,6 +956,7 @@ impl Oauth2ResourceServersReadTransaction {
 
     pub fn check_oauth2_token_exchange(
         &self,
+        idms: &IdmServerProxyReadTransaction<'_>,
         client_authz: Option<&str>,
         token_req: &AccessTokenRequest,
         ct: Duration,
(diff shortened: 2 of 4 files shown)


kanidmd/lib/src/idm/oauth2.rs
```diff
@@ -1669,7 +1688,7 @@ mod tests {
             // System admins
             (
                 "oauth2_rs_scope_map",
-                Value::new_oauthscopemap(UUID_SYSTEM_ADMINS, btreeset!["read".to_string()])
+                Value::new_oauthscopemap(UUID_SYSTEM_ADMINS, btreeset!["groups".to_string()])
```
yaleman, line 1691: `setup_oauth2_resource_server` is used 18 times in tests, this change needs to be carefully considered

kanidmd/lib/src/idm/server.rs
```diff
@@ -620,7 +620,7 @@ pub trait IdmServerTransaction<'a> {
             security_info!("The token grace window is in effect. Assuming valid.");
         };
 
-        Account::try_from_entry_no_groups(entry.as_ref()).map(Some)
```
yaleman, line 623: Why not just use `Account::try_from_entry_ro` here?

yaleman (commented):
Looks good so far, some comments added along the way. 

As discussed in community chat, this only affects OAuth2 claims so won't blow up the cookie size; in which case, the RS flag for including them shouldn't be needed.

kanidmd/lib/src/idm/oauth2.rs
theSuess: As far as I was able to tell, this "read" scope was only used in one test and there was no real advantage of adding a new scope as opposed to replacing the existing one.

kanidmd/lib/src/idm/server.rs
theSuess: Because we do not need the group information in every case (e.g the token introspection endpoint) so this saves a few DB queries

kanidmd/lib/src/idm/oauth2.rs
Firstyear: Yeah, read was just a "placeholder" for if you had a scope or not in one test.

kanidmd/lib/src/idm/account.rs
```diff
@@ -156,7 +156,7 @@ impl Account {
     #[instrument(level = "trace", skip_all)]
     pub(crate) fn try_from_entry_ro(
         value: &Entry<EntrySealed, EntryCommitted>,
-        qs: &mut QueryServerReadTransaction,
+        qs: &QueryServerReadTransaction,
```
Firstyear, line 159: please leave the &mut here for now, as I'm working on some cleanups that will require it to be present (unless it generates a warning, then leave it removed )

kanidmd/lib/src/idm/oauth2.rs
```diff
@@ -1580,6 +1557,47 @@ fn parse_basic_authz(client_authz: &str) -> Result<(String, String), Oauth2Error
     Ok((client_id.to_string(), secret.to_string()))
 }
 
+fn s_claims_for_account(o2rs: &Oauth2RS, account: &Account, scopes: &Vec<String>) -> OidcClaims {
```
Firstyear, line 1560: Should be &[String] not Vec, clippy will be sad otherwise.

Firstyear (approved):
looks fantastic, thanks so much. I think the only thing is this needs a rebase due to a cleanup change I did in master.

kanidmd/lib/src/idm/account.rs
theSuess: Hm, if I keep the reference mutable, it requires a pretty large refactoring (adding a lot of `mut`s all the way to the top) and then causes some borrowing issues. I'm sure there's a better way but I don't know rust well enough to come up with a good solution

Firstyear: All good, leave it as is for now then :)

PR #1286, opened 2022-12-21
docs: reformat book and introduce workflow to ensure it stays formatted

jceb:
- I formatted the book with `deno fmt`
- added syntax name to all preformatted code sections
- extended the kanidm_book github workflow with a test to ensure the documentation stays formatted


yaleman:
Looks like the changes to the Makefile broke it - can you please check/resolve that?

Firstyear:
I don't mind this change, I'd prefer column width of 100, but an autoformatter is fine.

I'll trust @yaleman to make the decision on this as they are the github actions wizard.

Makefile
```diff
@@ -121,12 +121,20 @@ test/pykanidm/mypy:
 test/pykanidm: ## run the test suite (mypy/pylint/pytest) for the kanidm python module
 test/pykanidm: test/pykanidm/pytest test/pykanidm/mypy test/pykanidm/pylint
 
+.PHONY: test/doc/format
+test/doc/format: ## Format docs and the Kanidm book
+	find . -type f -name \*.md -print0 | xargs -r -0 deno fmt --check
```
yaleman, line 126: This doesn't need xargs and the problems that it can bring - `find . -type f -name \*.md -exec deno fmt --check "{}" \;` would do.

yaleman (changes requested):
This needs another rebase and removal of xargs please

yaleman:
> I don't mind this change, I'd prefer column width of 100, but an autoformatter is fine.
> 
> I'll trust @yaleman to make the decision on this as they are the github actions wizard.

I'm not a *huge* fan of adding another programming language/interpreter to the implicit requirements for the project, but if there's no equivalent in rust-land then it'll do. A standard formatter is nice either way 😃

Firstyear:
Pretty sure deno is written in rust at least? I thought it was a ts/js engine, didn't know it did formatting.

jceb:
Yes, it's written in rust.
We could also use https://rome.tools. I haven't used it before but it's also written in rust and focuses on formatting and packaging. A nice addition that we'd get with deno is a JS runtime that could be leveraged for scripting tasks in our CI workflows.

jceb:
> This needs another rebase and removal of xargs please

Okay, will do.

jceb:
> I don't mind this change, I'd prefer column width of 100, but an autoformatter is fine.

Okay, will adjust the line length.

yaleman (approved):
Looks great, thanks 😃

PR #1287, opened 2022-12-21
feat(kanidmd): add ldap support for mail primary and alternative address

### LINKED ISSUE #1270: Mail primary address handling for LDAP unclear / not implemented?
### I did this

I created a user and tried set one and then multiple email addresses as implemented in #461. I tried to determine the primary mail address but I failed to clearly identify it. The order of email addresses in the ldapsearch output seems to stay the same regardless of the order I use in the update command. There's no unique attribute that holds the primary email address, e.g. `mailPrimaryAddress`.

```bash
ldapsearch -D dn=token -y sogo-ldap '(name=myuser)' mail

dn: spn=myuser@idm.example.com,dc=idm,dc=example,dc=com

kanidm person update -m myuser@example.com myuser --name idm_admin
ldapsearch -D dn=token -y sogo-ldap '(name=myuser)' mail

dn: spn=myuser@idm.example.com,dc=idm,dc=example,dc=com
mail: myuser@example.com

kanidm person update -m myuser.surname@example.com -m myuser@example.com myuser --name idm_admin
ldapsearch -D dn=token -y sogo-ldap '(name=myuser)' mail

dn: spn=myuser@idm.example.com,dc=idm,dc=example,dc=com
mail: myuser.surname@example.com
mail: myuser@example.com

kanidm person update -m myuser@example.com -m myuser.surname@example.com myuser --name idm_admin
ldapsearch -D dn=token -y sogo-ldap '(name=myuser)' mail

dn: spn=myuser@idm.example.com,dc=idm,dc=example,dc=com
mail: myuser.surname@example.com
mail: myuser@example.com
```

### I expected the following

My goal is to query the email addresses from LDAP and retrieve the primary address as well as the alternative addresses. Ideally, I'd like to keep the primary email address in a separate field, e.g. `mailPrimaryAddress` so it's clearly separated from the other addresses. Furthermore, I'd like to see the primary mail address removed from the alternative email addresses in the `mail` attribute so that there are no duplicates in the list when joining `mail` and `mailPrimaryAddress`. For backward compatibility it might be helpful to keep the existing `mail` attributes and a two new ones: `mailPrimaryAddress` and `mailAlternativesAddresses`

kanidm 1.1.0-alpha.10
2022-12-16T08:46:28.476620Z DEBUG kanidm: Using 16 worker threads

kanidmd 1.1.0-alpha.11-dev (in docker)

Linux torch 6.0.12-arch1-1 #1 SMP PREEMPT_DYNAMIC Thu, 08 Dec 2022 11:03:38 +0000 x86_64 GNU/Linux

### Any other comments

jceb:
Fixes #1270

This is a first attempt. The retrieval of the primary/alternative mail works well. However, there's an issue with detection of the correct primary mail address when the update command (`kanidm person update -m myuser.surname@example.com -m myuser@example.com myuser --name idm_admin`) is run twice with a different order of addresses.

- [x] cargo fmt has been run
       - I ran `cargo clippy` but it failed with 13 issues - none seems to be related to my change
- [/] book chapter included (if relevant)
- [/] design document included (if relevant)


Firstyear:
What do you mean by there is an issue? Is it swapping the order of the addresses sometimes in the request even though the CLI params are not changing?

The person update command is a bit rough, it was a quick fix and it's something we need to improve.

kanidmd/lib/src/entry.rs
```diff
@@ -1795,6 +1795,24 @@ impl Entry<EntryReduced, EntryCommitted> {
                         atype: "entrydn".to_string(),
                         vals: vec![dn.as_bytes().to_vec()],
                     }),
+                    "mail;primary" | "mailprimary" | "emailprimary" => {
+                        attr_map.get(kani_a).map(|pvs| LdapPartialAttribute {
+                            atype: ldap_a.to_string(),
+                            vals: pvs
```
Firstyear, line 1801: I think this here isn't right. The pvs is a `Vec<Vec<u8>>` which contains a set of emails per https://github.com/kanidm/kanidm/blob/master/kanidmd/lib/src/valueset/address.rs#L321 but this isn't sorted so that the primary is first, it's just the set of addresses with the primary tagging removed. 

I think to make this "consistent" with your behaviour, you need to adjust to_proto_string_clone_iter in address.rs to put the primary first so that your .first() and .split_first() trick works correctly. 

This probably explains the issue you were seeing with the incorrect order of primary based on what you set on the cli. 

It feels a bit like a hack to do it this way and rely on the ordering, so I want to think about this some more for a bit too, but for now that will at least make this work.

Additionally, please add a test for this in kanidmd/lib/src/ldap.rs to ensure that the virtual attributes are being generated as we expect so that we don't regress or break this in future!

Firstyear (changes requested):

kanidmd/lib/src/entry.rs
```diff
@@ -1795,6 +1795,24 @@ impl Entry<EntryReduced, EntryCommitted> {
...
+                                .first()
+                                .map(|first| vec![first.clone()])
+                                .unwrap_or_default(),
+                        })
+                    }
+                    "mail;alternative" | "mailalternative" | "emailalternative" => {
```
Firstyear, line 1807: Since you only add the phantoms for `emailalternative` then you should remove `mailalternative`. I think that mail;alternative can stay though since that wouldn't be allow by iname rules

kanidmd/lib/src/entry.rs
jceb: Thanks for the hint. I added that feature. and also implemented tests.

kanidmd/lib/src/entry.rs
jceb: Done.

Firstyear (approved):

Firstyear:
Thanks @jceb ! Wonderful to see your contributions :)

jceb:
Thanks for being so welcoming and merging so quickly👍👍

Firstyear:
@jceb As an FYI in the next week I'm going to do a big refactor of some source code locations which make break in flight PR's (excluding docs). So hopefully this doesn't affect you too much, since I think most people around the world are on holidays :)

jceb:
Thanks for the heads-up. At the moment I don't have any other PR in the making.

PR #1331, opened 2023-01-22
Fix debian build path

### LINKED ISSUE #1329: Serval questions about ubuntu packaging
I finally got kanidm works on ubuntu. Here are some problems I found during the process. If they are confirmed, I'll be glad to write a PR to fix these.

1. in https://github.com/kanidm/kanidm/blob/master/platform/debian/simple_pkg.sh#L16, `kanidm-ssh` isn't built
2. In https://github.com/kanidm/kanidm/blob/master/platform/debian/kanidm-unixd/rules#L11, `bindir` is `/usr/sbin`, however, in https://github.com/kanidm/kanidm/blob/master/examples/systemd/kanidm-unixd.service, it uses `/usr/local/sbin/kanidm_unixd`
   should I update `bindir` in deb or `kanidm-unixd.service`?
4. pam and nss lib not built or installed in https://github.com/kanidm/kanidm/blob/master/platform/debian/kanidm-unixd/rules#L47

leoleoasd:
Fixes #1329

- [x] cargo fmt has been run


Firstyear:
I'll leave it to @yaleman to review as they are the debian build person, but also add yourself to contributors.md :)

leoleoasd:
https://manpages.debian.org/testing/debhelper/dh_usrlocal.1.en.html
https://unix.stackexchange.com/questions/409800/dh-usrlocal-is-not-a-directory

It seems that it isn't standard to put files under /usr/local. Should we override it or put the binaries under /usr/sbin?

yaleman:
Can you please provide an authoritative link to the Debian policy on that? I am pretty burnt by trying to follow their packaging stuff, which is a nightmare of vague or conflicting information.

leoleoasd:
I found this: https://www.debian.org/doc/debian-policy/ch-opersys.html#site-specific-programs
image

And wikipedia page: https://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard

leoleoasd:
Does it seem that `/usr/local` is only for applications that **aren't** installed from the system package manager (like manually built and locally installed)?

yaleman:
Thanks for that, I guess it's better to replace `/usr/local` with `/usr` in the scripts?

leoleoasd:
then the systemd service should be updated as well. Are they used anywhere else?

leoleoasd:
Is the systemd service file under `examples` directory meant to be referenced by users who manually build this? I copied them to `platform/debian` and updated paths.

yaleman (approved):

leoleoasd:
forgot to push this one. @yaleman  can you approve actions again?

yaleman:
`platform/debian/pam-config-kanidm` is used in the `simple_pkg.sh` script, so that'll need updating, and I'm not sure if just putting it into the package dir includes it into the OS PAM config somehow? (Also, that'll need testing, because ... automagically enabling PAM is risky)

leoleoasd:
it won't be automatically emabled. you need to run `pam-auth-config` and select kanidm

leoleoasd:
image

leoleoasd:
Putting it into package dir will make it installed by https://man7.org/linux/man-pages/man1/dh_installpam.1.html

yaleman:
Very cool, thanks for the context! 😄

leoleoasd:
I made a few more modifications:
1. Merged `kanidm-ssh` and `kanidm-unixd`, as discussed [here](https://github.com/kanidm/kanidm/issues/1329#issuecomment-1401120573)
2. Updated CI scripts so now it will build debs for ubuntu and debian
3. add `bash` in makefiles when invoking `.sh` files to make it work with lower version of `make` (used in ubuntu 18.04)
4. deleted `jetli/wasm-pack-action`, since it doesn't support running inside containers and `wasm-pack` will be run in later `make debs/xxx` anyway.
5. move `actions/download-artifact` to a new job that runs after build on all platforms to upload all debs to releases.

Firstyear (approved):

Firstyear:
This looks fantastic, thank you very much @leoleoasd

yaleman:
Thanks @leoleoasd 🥳

Issue #1355, opened 2023-02-01
Improve docker builds


Firstyear:
Currently it can be a bit awkward/slow to do docker builds, or you need a weird setup with buildx for multiple arch releases. 

To improve this we should consider:

* Vendor deps first
* Having a "setup" phase that setups a single builder that can be used to help with build caching.
* Setting up cross-complilation ( https://www.docker.com/blog/faster-multi-platform-builds-dockerfile-cross-compilation-guide/ )

PR #1357, opened 2023-02-01
1355 docker builds

Firstyear:
Relates #1355 - adds vendor and mold to improve build performance. Cross compilation would be the next step to improve performance I think when we don't have a cluster available. 


yaleman (commented):
seems good to me :D

PR #1385, opened 2023-02-17
Reduce the number of cow cells in idm

Firstyear:
Very minor cleanup to remove some cow cellus that are all replaced at the same time. I'm going to experiment some more with how to make this work better as a whole as I think we can significantly reduce the amount of locking in the whole transaction. 


yaleman (approved):
🐮 <(moo)

Nice! Less locks, more betterererer.

PR #1386, opened 2023-02-17
1115 setup credential ids

Firstyear:
Relates #1115 - we need to track credential ids used for sessions for two reasons. First to cancel (expire) sessions that have their authorising credentials removed (ie panic reset causes immediate session revoke). Second we need to track the cred id to know what credential to prompt for during re-auth in sudo mode.

This adds the first part. We add cred id tracking to sessions, as well as expire sessions if the originating credential is removed. Also tested was the behaviour of the ui in this case, where post a credential reset you are redirected to login again if you *delete* the credential that was used for the active session. 


Firstyear:
PS: It's likely the re-auth stuff will prompt some other auth cleanups

yaleman:
Usual question - will this change affect the downstream clients on older versions of kanidm_unixd etc?

Firstyear:
> Usual question - will this change affect the downstream clients on older versions of kanidm_unixd etc?

Nope! This only changes the storage of the sessions server side, which associates the credential id. No changes to the client side tool versions :)

yaleman (approved):

PR #1388, opened 2023-02-18
Windows build fixes

yaleman:
This is a start on trying to filter out things while building on windows.

There's a hack to do in `build.rs` that while automagically enable the `unix` feature in `kanidm_unix_int` when building on unix targets - we can't filter out `lib`s using features though.. sigh.

- [x] cargo fmt has been run


yaleman:
Ref https://github.com/rust-lang/cargo/issues/1197 for target-specific-features being not a thing officially yet :(

Firstyear:
@yaleman We almost need target specific [bin] or [[libs]], or even target specific workspace memberships.

yaleman:
> @yaleman We almost need target specific [bin] or [[libs]], or even target specific workspace memberships.

All of those things would be great, yep. 😢

iam_migrations/freeipa/Cargo.toml
```diff
@@ -34,6 +33,9 @@ uuid = { workspace = true, features = ["serde"] }
 # For file metadata, should this me moved out?
 kanidmd_lib.workspace = true
 
+[target.'cfg(not(target_family = "windows"))'.dependencies]
```
Firstyear, line 36: Probably makes more sense for this to be "target_family = unix" since the options are unix, windows, wasm, and neither windows or wasm support this.

kanidm_unix_int/nss_kanidm/src/lib.rs
```diff
@@ -10,22 +10,34 @@
 #![deny(clippy::needless_pass_by_value)]
 #![deny(clippy::trivially_copy_pass_by_ref)]
 
+#[cfg(target_family = "unix")]
```
Firstyear, line 13: Feels a bit rough to have so many cfg statements here, do we actually need to do something silly like break this into modules that are cfg gated? 

Something like:

```
#[cfg(target_family = "unix")]
#[macro_use]
extern crate ...

#[cfg(target_family) ]
mod blahblah {
     // all the use statements, code, hooks etc.
}

#[cfg(not(target_family)) ]
mod blahblah
{
     do nothing? 
}
```

Firstyear (commented):
I'm wondering if all this is worth it, or if we just need to break some modules out of the workspace? I'm not sure .... There seems to be no nice answer here.

Firstyear:
https://github.com/rust-lang/rust/issues/108251

kanidm_unix_int/nss_kanidm/src/lib.rs
yaleman: yeah, it's a big clusterf*ck and was me just trying to yolo the whole thing on those modules - I'm poking around to find better ways of doing it in general.

iam_migrations/freeipa/Cargo.toml
yaleman: True, Done!

kanidm_unix_int/nss_kanidm/src/lib.rs
workingjubilee: > Feels a bit rough to have so many cfg statements here, do we actually need to do something silly like break this into modules that are cfg gated?

That is generally how it is done, yes. You build a module that defines a shared common abstraction and then you have each platform provide the relevant implementations to back up this abstraction, so you only need to write something like `#[cfg] use platform::imp as module;`.

Firstyear: @workingjubilee The problem here is there is no common abstraction for pam or nsswitch on windows, which is why we raised the conditional compilation target issue. We want to completely skip these workspace members on windows rather than try to cfg define them out and make an empty .dll. So in this case the cfg around the high level module break out is still a bit of a "hack" for us :(

Firstyear:
That does seem like a cleaner fix with the nss impl @yaleman :)

kanidm_unix_int/nss_kanidm/src/lib.rs
workingjubilee: Oof. Yeah, that's fair. Sometimes doing things with computers is Like This.

Firstyear: Absolutely. Especially when we are writing something that is very platform dependent and targeted, we get bitten by things like this. In the opposite, we are starting to see interest in development of a windows specific authentication module, which means we'll have the reverse issue of a target that's windows only.

Thus, why we opened the rust (now cargo) issue about platform specific build targets for workspace members, since it would be the cleanest option for us, and quite likely, many others :)

yaleman:
Is there some tooling I can use (other than building a VM and installing the modules) to test these modules are actually compiling and working? Other than some kind of wild hacky FFI thing... which I wouldn't *hate*...

iam_migrations/freeipa/src/main.rs
```diff
@@ -177,39 +179,50 @@ async fn driver_main(opt: Opt) {
...
-                    tokio::signal::unix::signal(sigterm).unwrap().recv().await
-                } => {
-                    // Ignore
-                }
-                Some(()) = async move {
-                    let sigterm = tokio::signal::unix::SignalKind::user_defined1();
-                    tokio::signal::unix::signal(sigterm).unwrap().recv().await
-                } => {
-                    // Ignore
+            #[cfg(target_family = "unix")]
```
Firstyear, line 182: I wonder if we need to move this signal pattern into the junk draw given it's used in a few places.

iam_migrations/freeipa/src/main.rs
```diff
@@ -952,10 +958,14 @@ fn main() {
         .init();
 
     // Startup sanity checks.
+    #[cfg(target_family = "unix")]
```
Firstyear, line 961: Similar this for the junk draw for "check current user permission is not root/admin"

kanidm_unix_int/Cargo.toml
```diff
@@ -11,37 +11,51 @@ license.workspace = true
...
-[lib]
-name = "kanidm_unix_common"
-path = "src/lib.rs"
+[features]
+unix = []
 
 [[bin]]
 name = "kanidm_unixd"
 path = "src/daemon.rs"
+required-features = ["unix"]
```
Firstyear, line 20: Wait, this gates the target based on a feature?!

kanidm_unix_int/nss_kanidm/src/lib.rs
```diff
@@ -10,162 +10,15 @@
...
-
-fn group_from_nssgroup(ng: NssGroup) -> Group {
-    Group {
-        name: ng.name,
-        passwd: "x".to_string(),
-        gid: ng.gid,
-        members: ng.members,
-    }
-}
+#[cfg(target_family = "unix")]
```
Firstyear, line 23: Do we actually need this pub use? Given the way that the nsswitch module works, it doesn't matter what file the implementation is in because of howe the symbols are exported.

kanidm_unix_int/pam_kanidm/src/lib.rs
```diff
@@ -11,326 +11,8 @@
...
-        if opts.debug {
-            println!("sm_setcred");
-            println!("args -> {:?}", args);
-            println!("opts -> {:?}", opts);
-        }
-        PamResultCode::PAM_SUCCESS
-    }
-}
+#[cfg(target_family = "unix")]
+pub use crate::pam::*;
```
Firstyear, line 18: As above, not sure the pub use is needed due to how the symbols export.

kanidm_unix_int/pam_kanidm/src/pam/mod.rs
```diff
@@ -30,3 +30,332 @@ pub mod items;
 #[doc(hidden)]
 pub mod macros;
 pub mod module;
+
+#[cfg(target_family = "unix")]
```
Firstyear, line 34: Since this is the module, and we have the cfg(unix) above in lib.rs, do we need all these cfgs?

kanidmd/lib/src/valueset/session.rs
```diff
@@ -38,7 +38,7 @@ impl ValueSetSession {
                     // MISTAKE - Skip due to lack of credential id
                     // Don't actually skip, generate a random cred id. Session cleanup will
                     // trim sessions on users, but if we skip blazenly we invalidate every api
-                    // token ever issued. OPPS!
+                    // token ever issued. OOPS!
```
Firstyear, line 41: OPPS.

iam_migrations/freeipa/src/main.rs
yaleman: or maybe some kind of builder/macro thing, since there's a few different versions of it?

kanidm_unix_int/Cargo.toml
yaleman: yep, but only for bins not libs 😡

kanidm_unix_int/pam_kanidm/src/pam/mod.rs
yaleman: Nope, still need to clean it up (done)

kanidm_unix_int/Cargo.toml
Firstyear: What is a bin but a lib out of hand

yaleman: Truly a wise man, this William guy!

kanidm_unix_int/nss_kanidm/src/lib.rs
yaleman: :shrug: I'm not sure how any of it works TBH, I assumed it was based on being exported from the root of the module somehow?

kanidm_unix_int/nss_kanidm/src/lib.rs
yaleman: Cleaned this up a bunch. :D

kanidm_unix_int/nss_kanidm/src/lib.rs
Firstyear: Yes, but not in the way you might think. pub use is a rust-specific construct that is only understood by the rust compiler and has no impact on the resulting symbols in the shared library.

The actual magic is in the "pam_hooks" macro, which defines a c extern / no_mangle which is a function name that is NOT dead code eliminated, and also has a static symbol name in the dll/so. The location of it within the rust source tree doesn't matter, because it always exports that symbol name into the resulting object.

So the pub use here isn't needed, because we don't have any consuming rust code that links into this, and the pam_hooks call is exporting the symbol names correctly in a way that is indepedent to the rust source layout.

Does that help,

yaleman: Sounds good to me, I'll take it out and we'll see how it goes 😄

Firstyear: As discussed offline - william was wrong!

yaleman: Amazing! It does happen sometimes. ❤️

Firstyear:
When do you reckon this would be good to go?

yaleman:
> When do you reckon this would be good to go?

I was just about to ask if we wanted to make the `unix` feature a thing - because it kinda makes sense in context. You don't really want to build the UNIX-specific tools on most targets.

Firstyear:
> > When do you reckon this would be good to go?
> 
> I was just about to ask if we wanted to make the `unix` feature a thing - because it kinda makes sense in context. You don't really want to build the UNIX-specific tools on most targets.

I think we make unix a feature, but make it a default feature because that's "least disruptive" atm. We can undo or change that later.

yaleman:
> > > When do you reckon this would be good to go?
> > 
> > 
> > I was just about to ask if we wanted to make the `unix` feature a thing - because it kinda makes sense in context. You don't really want to build the UNIX-specific tools on most targets.
> 
> I think we make unix a feature, but make it a default feature because that's "least disruptive" atm. We can undo or change that later.

Sounds good to me :D

PR #1408, opened 2023-02-26
Version  output for 4 unix_int tools

cuberoot74088:
This adds a `--version` flag to
* kanidm_cache_clear
* kanidm_cache_invalidate
* kanidm_unixd_status
* kanidm_ssh_authorizedkeys
 
as listed in #989.


Firstyear:
Looks good so far :)

cuberoot74088:
Where else would you see this doable and useful?

- orca 

could be one. I need to see where to put it best.

looking at the unix_int, we would have:

- [x] "kanidm_unixd" (done before, but only CARGO_PKG_VERSION afaik)
- [ ] "kanidm_unixd_tasks" (no Opts so far)
- [x] "kanidm_ssh_authorizedkeys" (this pr)
- [x] "kanidm_cache_invalidate" (this pr)
- [x]  "kanidm_cache_clear" (this pr)
- [x] "kanidm_unixd_status" (this pr)
- [ ] "kanidm_test_auth"

Firstyear:
I think unixd_tasks is pretty tightly bonded to unixd so we don't need it there. We could alternately add a version mismatch check between these two over the unix socket instead so that it can only work with it's correct partner.

For pam test, that probably can be removed as a tool since I make a seperate pam tester repo.

cuberoot74088:
Ok. So we you like to keep it as it is right now?

On a side note:
As I went trough the 4 unix tools, I thought, some of them could also be combined and it would be more like commands for that unified tool:

`unix-util <command>`

with `<command>` being one of:
- cache-clean
- cache-invalidate
- unixd-status

or is there some special reason, why there should be separate binaries?

Firstyear:
> Ok. So we you like to keep it as it is right now?
> 
> On a side note: As I went trough the 4 unix tools, I thought, some of them could also be combined and it would be more like commands for that unified tool:
> 
> `unix-util <command>`
> 
> with `<command>` being one of:
> 
> * cache-clean
> * cache-invalidate
> * unixd-status
> 
> or is there some special reason, why there should be separate binaries?

I think they can be combined yes :) 

I'm just about to do a re-org of some of the source code and don't want to break any PR's, so I might merge this as is and we continue it in a follow up once I do the reorg.

PR #1415, opened 2023-03-01
User auth token session display implementation

### LINKED ISSUE #1359: User get user_auth_token_session field contains Debug values.
### I did this

Ran `kanidm person show nirya`

The output of fields regarding the user `nirya` with:
```
user_auth_token_session: 804f61db-4742-4799-8b21-fc7ecba51b4f: Session { label: "Auth Session", expiry: Some(OffsetDateTime { utc_datetime: PrimitiveDateTime { date: Date { year: 2023, ordinal: 35 }, time: Time { hour: 4, minute: 55, second: 54, nanosecond: 309616627 } }, offset: UtcOffset { seconds: 0 } }), issued_at: OffsetDateTime { utc_datetime: PrimitiveDateTime { date: Date { year: 2023, ordinal: 35 }, time: Time { hour: 3, minute: 55, second: 54, nanosecond: 309616627 } }, offset: UtcOffset { seconds: 0 } }, issued_by: User(db2342d3-f2f6-4cf0-bf68-f7fa17b5ea8f), scope: ReadWrite }
```

### I expected the following

An output of normal looking values:

Perhaps:
```
user_auth_token_session: Auth Session: id 804f61db-4742-4799-8b21-fc7ecba51b4f, expiry <UTC date>, issued at <UTC date>
```

kanidm 1.1.0-alpha.11
(release)

Darwin misato.local 22.2.0 Darwin Kernel Version 22.2.0: Fri Nov 11 02:03:51 PST 2022; root:xnu-8792.61.2~4/RELEASE_ARM64_T6000 arm64

But also appears on Linux machines.

### Any other comments

None

Seba-T:
Fixes #1359

- [x] cargo fmt has been run


book/src/developers/designs/identity-verification.md
```diff
@@ -0,0 +1,42 @@
+## Draft proposal for and Identity verification system with an out of bound challenge
```
yaleman, line 1: This should be cherry-picked out as it looks like you've got the draft doc from #1416 in this PR...?

yaleman (changes requested):

book/src/developers/designs/identity-verification.md
Seba-T: wops sure looks like I pushed it to the wrong branch :D

Seba-T:
Now it should be fixed

server/lib/src/valueset/session.rs
```diff
@@ -347,7 +347,7 @@ impl ValueSetT for ValueSetSession {
         Box::new(
             self.map
                 .iter()
-                .map(|(u, m)| format!("{}: {:?}", uuid_to_proto_string(*u), m)),
+                .map(|(u, m)| format!("Auth session id: {}, {:?}", uuid_to_proto_string(*u), m)),
```
Firstyear, line 350: We can't "format" things in "to proto string". This is meant to be a stable transmitted representation of the data, not a human/user friendly display. So this needs to be undone sorry :)

server/lib/src/value.rs
```diff
@@ -774,6 +776,25 @@ pub struct Session {
     pub scope: SessionScope,
 }
 
+impl fmt::Debug for Session {
+    fn fmt(&self, f: &mut Formatter) -> fmt::Result {
+        let issuer = match self.issued_by {
+            IdentityId::User(u) => uuid_to_proto_string(u),
```
Firstyear, line 782: I think this needs to display "User - uuid-here" or "Sync - uuid-here". The current match is losing a lot of information about the issuer and the token that created the session.

server/lib/src/value.rs
```diff
@@ -774,6 +776,25 @@ pub struct Session {
...
+            IdentityId::Synch(u) => uuid_to_proto_string(u),
+            _ => "invalid".to_string(),
```
Firstyear, line 784: It's not invalid - match the remaining cases, because Internal is a valid issuer!

Firstyear (changes requested):
Nice attempt, but I think there are plenty of ways to improve this :)

server/lib/src/value.rs
Seba-T: Wopsies my bad there, I meant internal instead of invalid :p 
I should have fixed everything now! 
And also thanks for all the helpful feedback :)

yaleman: And please don't `_`-match an enum - explicitly match the `Internal` variant.

Firstyear: Yeah, it's always better to explicitly match the variants @Seba-T because that way if we add a new variant or change things, the complier will flag we now need to handle the cas differently in this code path. It's always better to be explicit, there are very few cases where I'll use a '_' match because we need to be thorough.

Firstyear (approved):
Please also add yourself to the CONTRIBUTORS.md file and i'll merge this :)

Seba-T:
Done!! :)

server/lib/src/value.rs
Seba-T: Alright, thanks for the tip!

Firstyear (approved):

Firstyear (approved):

Firstyear:
@yaleman need your approve here :)

yaleman (approved):

PR #1416, opened 2023-03-01
possible design to implement an identity verification system

### LINKED ISSUE #337: Feature request: verify identity functionality

Seba-T:
Fixes #337

- [ ] cargo fmt has been run


Firstyear:
Email is not really possible or viable here IMO.

There is an easier way as well that doesn't require any super complex dh stuff.

The easier way is that each user's account has a "half" totp key stored as a secret in the DB. 

Then when alice and bob want to verify each other their both login on their own respective computers, and type in the username they want to verify (alice would type bob, and bob would type alice)

Alice is then displayed the TOTP which is the result of  TOTP ( (Alice Key Half + Bob Key Half) ). 

Bob is then displayed the TOTP which is the result of  TOTP ( (Bob Key Half + Alice Key Half) ). 

Alice would need to type in what is displayed to bob (and the server can verify that), and in returne bob would type in the return.

This avoids:

* Social engineering of a user where they simply say "yep that's valid" on a phone call or similar, since both parties can validate the other.
* Each users validation is unque.
* Each user validation is also unique to only this *pair* of users together.
* It doesn't require email
* It works with replication since the two servers can be disconnected from active communication at the time and still succeed the check since the knowledge was pre-sent.

yaleman:
"email" is probably supposed to be "SPN" in this context, having read it. 

It's an interesting use case that assumes federation between instances is configured etc, which is a long way off...

Firstyear:
In the federation case we'll need a different approach anyway so I think we leave it until then and rethink when we get to it.

Firstyear:
I thought about this more and maybe the dh approach could work. 

This way an account like alice/bob each have a private key / public key. both of these would be stored in their account record, and if we do trusts, we just send the public part to the trusted domain. 

Then when they want to do the verification, both servers can do their dh operations. 

So the question then is how to make ecdh work since I think dh params are simply too cumbersome. ecdh has a nice public/private seperation.

The other question is given ecdh is about arriving at a shared secret, once we have that secret how can we "invert" it on one server or the other so that bobs challenge to alice is differenc to alices' to bobs. Perhaps even something as simple as "derived secret + uuid of target entry" would add sufficient entropy to the totp generation.

Firstyear:
I think this is the correct api: https://docs.rs/openssl/latest/openssl/pkey_ctx/struct.PkeyCtxRef.html#method.derive_set_peer 

This will allow the use of private/public keys with ecdh..

Seba-T:
First of all thanks for all the feedback! 
Secondly, I'm not sure I'm getting this right: 
> This way an account like alice/bob each have a private key / public key. both of these would be stored in their account record, and if we do trusts, we just send the public part to the trusted domain.
>
Would those public keys be sent beforehand, when the servers establish trust, or just when an authentication request is performed? If option 1 were to be implemented, wouldn't size be a problem, as there would be keys for all users of possibly many servers? Also I don't know about the feasibility of sharing keys on a request basis.

Seba-T:
> "email" is probably supposed to be "SPN" in this context, having read it.
> 
Yes now that I see what SPN is it indeed would be a better fit than email, I just didn't know about it :p

Firstyear:
> First of all thanks for all the feedback! Secondly, I'm not sure I'm getting this right:
> 
> > This way an account like alice/bob each have a private key / public key. both of these would be stored in their account record, and if we do trusts, we just send the public part to the trusted domain.
> 
> Would those public keys be sent beforehand, when the servers establish trust, or just when an authentication request is performed? If option 1 were to be implemented, wouldn't size be a problem, as there would be keys for all users of possibly many servers? Also I don't know about the feasibility of option 2, ie sharing keys on a request basis.

We'd send them in the user records. Size *is* a concern, that's why you use ecdh not traditional dh params. But also remember we already have to send a bunch of data anyway.

Perhaps it'll help to see it in a picture.

```
                ┌──────────────────────────────────────────┐                
                │Domain A                                  │                
                │┌────────────┐              ┌────────────┐│                
                ││            │              │            ││                
                ││            │              │            ││                
┌──────────┐    ││            │              │            ││    ┌──────────┐
│  Alice   │    ││  Server 1  │◀────────────▶│  Server 2  ││    │   Bob    │
└──────────┘    ││            │              │            ││    └──────────┘
                ││            │              │            ││                
                ││            │              │            ││                
                │└────────────┘              └────────────┘│                
                │                                          │                
                └──────────────────────────────────────────┘                
```

So not only are alice and bob connected to different Kanidm servers, but those Kanidm servers may not have active connection between them. Imagine an internet outage in a branch office or something.

But we already *replicate* all the database content between server 1 and server 2. Because these two nodes are now fully disconnected from each other, how can they resolve the problem? 

In this case, because we consider these servers within the domain as "trusted" the entry record of alice contains their private/public parts, and bob contains their private/public parts (in the future we could be fancier and modify this but lets start simple).

This way each server can do their needed half of the operation without an active communication between each other!

Now lets make it more complex. A future idea is *trust* between kanidm domains.

In this case.

```
                ┌──────────────────────────────────────────┐               
                │Domain A                                  │               
                │┌────────────┐              ┌────────────┐│               
                ││            │              │            ││               
                ││            │              │            ││               
┌──────────┐    ││            │              │            ││               
│  Alice   │    ││  Server 1  │◀────────────▶│  Server 2  ││               
└──────────┘    ││            │              │            ││               
                ││            │              │            ││               
                ││            │              │            ││               
                │└────────────┘              └────────────┘│               
                │                                          │               
                └──────────────────────────────────────────┘               
                                      ▲                                    
                                      │                                    
                                      │                                    
                                      │                                    
                                      │                                    
                                      ▼                                    
                ┌──────────────────────────────────────────┐               
                │Domain B                                  │               
                │┌────────────┐              ┌────────────┐│               
                ││            │              │            ││               
                ││            │              │            ││               
                ││            │              │            ││   ┌──────────┐
                ││  Server 3  │◀────────────▶│  Server 4  ││   │   Bob    │
                ││            │              │            ││   └──────────┘
                ││            │              │            ││               
                ││            │              │            ││               
                │└────────────┘              └────────────┘│               
                │                                          │               
                └──────────────────────────────────────────┘               
```

So not only now is bob on a seperate domain, but again they are on a different server with no direct connection path. Part of the trust design we have is that from domain A to domain B we would send "stub" or "partial" records of users from the trusted domain. In these stub records we could include the *public* part of the ec key (excluding the private). That way bob can use the public part in Alice's record for computing the ecdh, and vice versa alice to bob.

So that's why I'm suggesting the design I did - we have to not only consider that the servers involved are *disconnected* from each other so both must arrive at the same result without communication, but also that it must be possible to do so over a trust boundary. 

And of course, coming back to size - we're already replicating the whole database! So it's not going to be excessively more data to include a hidden-private ec key on the account. In fact, we could use that key for many things beyond just id verification in the future too.

Seba-T:
Do you think that with this approach the TOTP would still be a good idea? I think it should be implemented nonetheless, to prevent replay attacks. I believe I more or less got the idea, I'll now update the design document!

Seba-T:
Btw I added your diagrams to the file as I found them very helpful, is there any way to recognize contributions of this kind in a commit? Is that even something people usually do?

Firstyear:
> Do you think that with this approach the TOTP would still be a good idea? I think it should be implemented nonetheless, to prevent replay attacks. Or should we provide more flexibility by allowing users to encrypt whatever they want with their public keys, in order to provide a true out of band challenge experience? Anyhow I believe I more or less got the idea, I'll now update the design document!

I think we use the ecdh exchange to arrive at a shared secret unique to the two parties that want to verify, then we use that shared secret as the totp seeds for the users. I still am not sure how to do the "secret inversion" so that each side can verify the other. 

The user experience should still be "call someone and they give a challenge response" where each side needs to enter some value that only the other side has.

Firstyear:
> Btw I added your diagrams to the file as I found them very helpful, is there any way to recognize contributions of this kind in a commit? Is that even something people usually do?

I'm not sure to be honest, but given all my contributions to this project already I am not too worried about extra contributions being made to me :)

book/src/developers/designs/identity-verification.md
```diff
@@ -0,0 +1,160 @@
+## Draft proposal for and Identity verification system with an out of band challenge
+
+Implements #337
+
+### Goal:
+
+Providing a way to allow 2 subject to authenticate each other without prior knowledge of each other
```
Firstyear, line 7: Well, yes and no. The idea is to validate that a person you are communicating with is exactly who they say they are. 

For example, if someone wants to you transfer a bunch of money, then you can phone them and validate who they are, with kanidm acting as the assessor that both parties can firstly login as who they claim to be, and that they have active and current access to that system indicating a valid session.

It's effectively a way for a person to remotely validate that a person can authenticate to Kanidm, but without having to actually invasively check that they are able to. 

So with this in mind we need:

* Something that allows both sides to validate the other. We don't want person A to read our their code and B "lies" and just says yep that was valid. We need A to validate B and vice versa.
* This means that codes have to be easy to read and type
* The codes have to work for enough time for someone on a phone call to muck about and login and such.
* The codes have to also *change* after a period of time. 
* The codes must be unique to all pairs of people as they validate - A and B should have different codes to A and C or B and C.
* The codes must be consistent between all Kanidm servers due to replication segmentation and netsplits etc. 

This is why I suggested ecdh private key on each account that can then be used for every pair of accounts to arrive at a unique secret. 

We can then use that unique secret as a TOTP seed, since TOTP changes on a time schedule and is easy to read and type. The TOTP in Kanidm also has a "grace" allowing a single previous TOTP to be valid meaning that small de-sync in time steps is not noticed by the parties involved. 

The "trick" is that we can't have each side get the same totp. We want them to have a totp for A to B, and a seperate code for B to A. 

This means we need a deterministic way to take our derive secret and have each server do the same derivation for A and B to get their seperate it/out secrets.

AN example could be the account with the lower uuid takes uses the derived secret "as is" and the higher uuid reverses the byte array. This is a way both servers without communication can derive the same totp seeds without communication. 

Does that help?

book/src/developers/designs/identity-verification.md
```diff
@@ -0,0 +1,160 @@
...
+││            │              │            ││
+│└────────────┘              └────────────┘│
+│                                          │
+└──────────────────────────────────────────┘
+```
+
+- Two servers interested in creating such trust bond will perform a ECDH key exchange to create a shared secret, encrypting the communication with TLS
+- a KBKDF, such as the [Scrypt](https://docs.rs/openssl/latest/openssl/pkey_ctx/struct.PkeyCtxRef.html#method.derive_set_peer) KDF will be used to generate a secret key from the shared secret,
+- the secret key will stored in the database of both servers,
+- Finally, they both store the current unix time in the database (which will be referred as _T0_); it will be used for TOTP generation. Every time a new authentication request is received, the _token-time_ will be equal to the number of 30 seconds interval that have passed since _T0_, ie _token-time = (current_unix_time - T0) / 30_.
```
Firstyear, line 44: We don't have to store the current time at all. We also don't need TLS at all either. The "arrows" here represent that replication *has* occured, but may not be currently occuring. 

Try not to think about this as something where there is any active communication at all.

Server 1 has accounts A and B. A and B both have a private key / public key.

Server 1 then replicates the full DB content to Server 2. Server 2 is now an identical copy.

At this point we *disconnect server 1 and 2* from each other. They can't talk to each other at all!

Person A logs into server 1

Person B logs into server 2

How can they validate each other?

Since server 1 can internally read the private secrets of account A and B it can get the private/public keys and derive the ecdh secrets. 

Server 2 can do the same process.

These secrets are now in memory on server 1 and server 2 and they can be used as the totp seeds.

See how there is no communication outside the original replication event? Because the DB is replicated, each server has everything needed to all arrive at the same results without communication which is absolutely critical in a distributed system like this :)

book/src/developers/designs/identity-verification.md
```diff
@@ -0,0 +1,160 @@
...
+  └──────────┘    ││            │              │            ││    └──────────┘
+                  ││            │              │            ││
+                  ││            │              │            ││
+                  │└────────────┘              └────────────┘│
+                  │                                          │
+                  └──────────────────────────────────────────┘
+  ```
+
+- ### Servers belonging to different domains:
+  This case is very similar to the previous one, except that the servers don't belong to the same domain, and therefore the database is not replicated between them. In this case, the servers will stil have to share all the users' public keys, but the way that will be achieved is (for now) out of the scope of this document.
```
Firstyear, line 121: Well, trusts over a domain is actually just a special case of replication where we replicate a partial entry to the partner domain. We can just send the public part of the ec key in the partial entry over the trust boundary so here server 1 has Alice's private key and public key, and would have bob's public key only. Inverse, Server 4 would have Alices public key, and bobs private and public key. 

So again, they can each use what they have (alice's private + bobs public || bobs private + alices public) for the ecdh operation and arrive at the same secret, again without direction communication or contact.

book/src/developers/designs/identity-verification.md
```diff
@@ -0,0 +1,160 @@
...
+                  ││            │              │            ││
+                  ││            │              │            ││
+                  │└────────────┘              └────────────┘│
+                  │                                          │
+                  └──────────────────────────────────────────┘
+  ```
+
+## Possible weaknesses and general notes:
+
+First of all, this designed is based on the assumption that the secret key is not leaked, and that both the servers involved are not compromised. If either of these assumptions is broken, then impersonation attacks would indeed be possible.
```
Firstyear, line 159: Trusts are only established to ... well, trusted entities. Systems like this don't just configure trusts on whims, it's for careful and planned reasons.

And yes, the server is considered a security boundary already. Since Kanidm is the authority in issuing ouath2 sessions, logging people in, validating credentials, exploit of the server would already mean "game over" on many different dimensions of badness.

Firstyear (changes requested):

book/src/developers/designs/identity-verification.md
Seba-T: > 
> Does that help?

Yess now it  is indeed more clear! I didn't think about users possibly lying about the correctness of their TOTP, now I see how we need two TOTPs to allow users to mutually authenticate each other. For the time being, I'm thinking about something of the kind: 
First of all let's supposed Alice has the lowest UUID
- Alice's server computes the hash of the current time (rounded to 30 seconds) + her UUID, and encrypts it with her private key and Bob's pubic key
- Bob's server computes the hash of the current time (same 30s rounding) + his UUID reversed, and encrypts it with his private key and Alice's public key
- They exchange the encrypted messages by ie phone call (they will be in a human friendly format obv), they type them in their own servers and eventually the servers will verify the correctness of the TOTPs by decrypting them with the corresponding private and public key and checking whether the plaintext matches the secret computed locally. Note that the servers will have to take into account that the other server has inverted the UUID.  

I used the UUID strategy you suggested just to check if I understood, but I intend to think about it a little more to see if I can come up with something on my own! 
For example I was thinking we could compute the hash using both of their UUID while also maintaining the inversion for the highest UUID, ie: 
Alice computes hash(time, AliceUUID, BobUUID)
Bob computes hash(time, ReversedAliceUUID, ReversedBobUUID)
Both of them compute the hash using both of the UUIDs, but the account with the lowest UUID reverses the byte array of the UUIDs.  
I was also thinking we could derive encryption keys from one UUID and use them to hash the other UUID. 
Alice would derive an encryption key from her UUID and would use it to compute an HMAC of Bob's UUID, like HMAC(KDF(AliceUUID), BobUUID), Bob would be able to verify the correctness of the HMAC by computing it again locally.

Seba-T:
I made some changes trying to convey all the feedback you gave me into it, I hope it's going in the right direction! :)

Seba-T:
Just to check in, is this last version somewhat feasible to implement, or should I add some more tweaks to it? Also I'd like to discuss some paperwork stuff, who can I email @Firstyear @yaleman?

yaleman:
As long as the lifetime of the "trusted keys" is pretty short, and we can limit the scope to `person` accounts, not *all* accounts (and also a group, even if it's `all_people`), that seems sensible so far.

Firstyear:
> Just to check in, is this last version somewhat feasible to implement, or should I add some more tweaks to it? Also I'd like to discuss some paperwork stuff, who can I email @Firstyear @yaleman?

I think it needs tweaks still, I'll review it shortly.

Firstyear:
> As long as the lifetime of the "trusted keys" is pretty short, and we can limit the scope to `person` accounts, not _all_ accounts (and also a group, even if it's `all_people`), that seems sensible so far.

Yes, it would only be for persons @yaleman.

book/src/developers/designs/identity-verification.md
```diff
@@ -0,0 +1,152 @@
...
+
+Providing a way to allow 2 subject to authenticate each other using an out of band challenge.
+
+### Assumptions:
+
+Both subjects will have to be connected to a Kanidm server (not necessarily the same or under the same domain), and the servers will have to mutually trust each other; the trust bond must be established beforehand.
+
+### High level overview:
+
+The idea is to associate a [ECDH](https://docs.rs/openssl/latest/openssl/pkey_ctx/struct.PkeyCtxRef.html#method.derive_set_peer) public/secret key pair with each user. If two users want to authenticate each other, their private key and the other subject's public keys will be used to encrypt an HMAC message that uses both of the users UUID. Kanidm currently doesn't support ECDH keys for users, but once they will be implemented we will be able to use them for other purposes as well.
```
Firstyear, line 15: I don't know why you keep saying "hmac" or "encryption" here. Neither of these are involved in this process. 

I think you should try implementing this but in isolation to "see" the process.

The steps will be.

* generate a private and public ec key for both parties.
* exchange of the public keys (that is taken care of in kanidm for you by the underlying replication/database systems)
* each side does the derivation:
** given entity A and it's private key, use B's public key with ecdh to get shared secret X
** given entity B and it's private key, use A's public key with ecdh to get shared secret X

Then each side does the same set of derivations to get TWO TOTP secrets.

both sides will perform:

* concatenate the bytes of secret X with entity A uuid as bytes. This is TOTP secret A.
* concatenate the bytes of secret X with entity B uuid as bytes. This is TOTP secret B.

Both sides now have the two secrets. 

Entity A is displayed the output of TOTP(totp secret A).
Entity B is displayed the output of TOTP(totp secret B).

Entity B reads out their displayed TOTP B to entity A who types that value in to validate it. 
Entity A reads out their displayed TOTP A to entity B who types that value in to validate it.


This satisfies all our properties:

* The secret X is unique to the combination of A and B validating each other.
* By concatenating X with A/B uuids, we have two seperate secrets for each side to validate the other. 
* TOTP gives us the rolling / time changing of secrets with a grace window to allow some forgiveness in the exchange. We'll likely have to make the TOTP step 10 minutes in this scheme.
* There is no online exchange or communication
* There is no encryption or hmac going on.

So can you update the document to reflect these steps?

Firstyear (changes requested):

book/src/developers/designs/identity-verification.md
Seba-T: I think I should remove the overview section altogether as it fails explains the steps I detail afterwards. By encryption I meant just using the keys for the secret derivation the way you described, no online exchange involved, so I think we're on the same boat here. As in regard for hashing, I thought it could be an alternative to just concatenate the UUID as you suggested. This way, we would hash the shared secret with the user's UUID and use that as a TOTP, hopefully providing better security as it wouldn't be possible to retrive the UUID from the totp.

Seba-T: Update: I realized I never mentioned what the TOTP was going to be, now I hope it's much more clear :)

Seba-T: I was creating the proposal on the Google website: what project size should I pick? They offer just medium or large.

Firstyear: pick large :)

book/src/developers/designs/identity-verification.md
```diff
@@ -0,0 +1,151 @@
...
+- ### Servers belonging to the same domain:
+
+  Since the servers belong to the same domain they will have a replica of the same database, and therefore all the pub/secrets keys of every person will be available on both servers.
+  Let's suppose that Alice and Bob want to mutually authenticate each other, then the following steps will be performed:
+
+  1. Alice shares her SPN with Bob and asks for Bob's SPN
+  1. Bob receives Alice's SPN and replies by sending his SPN
+  1. Both Alice and Bob insert the acquired information in their Kanidm client.
+  1. The servers compute the ECDH shared key by using their user's private key and the other's users public key. DH guarantees that the servers will derive the same shared secret.
+  1. The servers compute a HMAC hash using the derived key as key, the current time and the user's UUID. These will be the TOTPs. Since each server will use its own user's UUID, there will be two different TOTPS. Note that both servers will compute both the TOTPs locally, as they will be needed for verification later.
```
Firstyear, line 92: I know you updated this, but I think you need to update this section ....

book/src/developers/designs/identity-verification.md
Firstyear: TOTP internally already does the HMAC for you, that's why concatenation of the bytes is sufficient.

Seba-T: I see now, so there's no need to mention hmac in the design at all

Firstyear: correct!

Seba-T:
Looks like I forgot to push for 5 days :p 
Btw since in the GSOC website they also require a pdf version of the proposal, can I just upload this last version of _identity-verification.md_  as pdf?

Firstyear:
> Looks like I forgot to push for 5 days :p 

It happens to the best of us :) but at least it wasn't pushing a private key! 

> Btw since in the GSOC website they also require a pdf version of the proposal, can I just upload this last version of _identity-verification.md_ as pdf?

Sounds reasonable to me. You may want to add some gsoc specific bits in the pdf though, but this can absolutely serve as the core of the proposal.

Seba-T:
> Sounds reasonable to me. You may want to add some gsoc specific bits in the pdf though, but this can absolutely serve as the core of the proposal.

As I was outlining the deliverables for the proposal, I wondered if I should also include a front-end implementation for the kanidm client, or maybe the server-side logic is going to take up all my time? For now I just mention that it could be something extra if I have some time to spare. For the time being, these are the deliverables I mention: 

>  The first step will be implementing the personal ECDH keys for kanidm users; then the core server-side logic of the system will follow, and finally, if time will allow it, a frontend interface for the kanidm client. The first step shouldn't take more than 40% of the total time available (350hrs)
while the second one will most likely take all the remaining time.

Seba-T:
> I'd go re-read the gsoc project spec :) [openSUSE/mentoring#196](https://github.com/openSUSE/mentoring/issues/196) specifically to address
> 
> > Assessment: Our intent is to mentor and help you grow as a contributor - and we want to help you succeed! As this is our focus, we have different assessment criteria to many other projects.
> > Has the student:
> > participated actively in the project?
> > written and submitted personal blog/journal or similar about the experience and what they have learnt and attempted?
> > sufficiently attempted one or any of the offered deliverables? (Note - there is no requirement to complete the deliverable. Just evidence of a solid attempt)
> 

Regarding that journal part, I was thinking about writing my bachelor's thesis about the project (for the bachelor we have a requirement of 15-30 pages, so nothing big). I still have to ensure I'm allowed to write it about a software engineering project, but if I were, could that count as the journal? 

> ALso don't stress too much about timelines of the various bits - software projects will always twice as long as you think or half the time. Don't try to force timelines about things.

Sure, I'll get rid of them, they were somewhat made up anyways :p

Firstyear:
> Regarding that journal part, I was thinking about writing my bachelor's thesis about the project (for the bachelor we have a requirement of 15-30 pages, so nothing big). I still have to ensure I'm allowed to write it about a software engineering project, but if I were, could that count as the journal?

That's fine by me - would it be possible to have the content be public post project? I have previously mentored on thesis before so this isn't an unknown to me :) 

> 
> > ALso don't stress too much about timelines of the various bits - software projects will always twice as long as you think or half the time. Don't try to force timelines about things.
> 
> Sure, I'll get rid of them, they were somewhat made up anyways :p

Thanks :)

And remember, the goal of this project is as much to give you the opportunity to be a part of an opensource team to learn what that's like, as much as the ability for you to do some code on some interesting concepts, so expect us to ask you to do code review or other things as part of this to help you learn some skills

Seba-T:
> 
> 
> > Regarding that journal part, I was thinking about writing my bachelor's thesis about the project (for the bachelor we have a requirement of 15-30 pages, so nothing big). I still have to ensure I'm allowed to write it about a software engineering project, but if I were, could that count as the journal?
> 
> 
> 
> That's fine by me - would it be possible to have the content be public post project? I have previously mentored on thesis before so this isn't an unknown to me :) 

Absolutely, it will be public!

> And remember, the goal of this project is as much to give you the opportunity to be a part of an opensource team to learn what that's like, as much as the ability for you to do some code on some interesting concepts, so expect us to ask you to do code review or other things as part of this to help you learn some skills 

Wow, I didn't know about the code review part, I'm really looking forward to it too!

Seba-T:
Btw since on gsco website they mention updating the proposal pdf based on the contributors’ feedback, is there anything I should update or is it good to go? (I don’t know if you can see the pdf I uploaded on the gsoc website)

Firstyear:
Sorry about the delay to respond to this, I was recovering from surgery. Don't worry I won't let that affect the gsoc decision, I know I didn't respond to yours or others questions in the last week so I won't penalise for that :)

Seba-T:
> Sorry about the delay to respond to this, I was recovering from surgery. Don't worry I won't let that affect the gsoc decision, I know I didn't respond to yours or others questions in the last week so I won't penalise for that :)

Oh alright thanks :) And get well soon!

Firstyear (approved):
I think the design is good to go, keen to see you work on this :)

Firstyear:
Did you want to close this @Seba-T ?

Seba-T:
Sure, should I copy the doc in the actual pr I'm working on?

Firstyear:
> Sure, should I copy the doc in the actual pr I'm working on?

Yep, sounds good to me.

Issue #305, opened 2020-08-10
Rewrite kanidmradius to rust


Firstyear:
The initial version of kanidm's radius support was created using rlm python as it exposed a simple way to script inside of freeradius.

However, even now the python code is already quite difficult to work with, and extending it further entails a lot of risk. There are also missing chances for integration with the localhost resolver or similar, and it's frankly hard to test. Experience with other projects shows that python adds non-trivial cognitive complexity due to the lack of a type system, and that faith in the code can only be known through runtime. 

We should rewrite kanidm to a true native rlm module in rust, which will allow better maintenance, we standardise on rust as the only language in the code base, and we are then better able to test it and gain confidence in the module.

cuberoot74088:
This sounds interesting and reasonable. Do we want to tackle that? We could still say it is working for now and there are more urgent or important tasks to do.

For my understanding: this would become a C library used by freeradius? As I have not seen any rlm-rust or so.

As it would involve some http requests, would that be using tide? Probably simply sync as well? Anything else from a "design perspective"?

What are your thoughts on that? Should we start something here?

Firstyear:
This really isn't a trivial issue. It's still a rust library but it just exposes a C compatible abi. So it would involve writing the needed compats and layers. As well we need to investigate if it's single or multi threaded etc. That would affect async or not and how we structure this. 

So for now the reason I've left it alone is exactly that - what we have works, and the rewrite would take a large amount of C/Rust/Radius knowledge to achieve correctly and safely.

PR #1423, opened 2023-03-04
Project: Rewrite Python RLM into Rust (DRAFT)

czzrr:
Reference issue: #305

This is my first draft for the project of rewriting the Python RLM into Rust.
Please let me know if my understanding of the problem is off or if the scope is inappropriate. :-)

### DIFF
--- a/project_plan.md
+++ b/project_plan.md
@@ -1,0 +1,28 @@
+# Project: Rewrite rlm_python in Rust (draft)
+
+## Project description
+Kanidm has support for RADIUS. RADIUS is a service that provides authentication, authorization and management of users that wish to access WLANs or VPNs.
+
+When a user wants to access a WLAN (or VPN) from their device, they submit an username and a password to the NAS. The NAS then sends an access-request with the user's credentials to the RADIUS server. The RADIUS server communicates with the Kanidm server to retrieve the user's RADIUS credentials and other information required for authorization. If the credentials match and the user can be authorized, the RADIUS server sends an access-accept back to the NAS, which finally allows the user's device to access the network.
+
+The RADIUS server is configured to communicate with the Kanidm server through a RADIUS loadable module (RLM), which is a way for the RADIUS server to run custom scripts when it receives requests. RADIUS has support for Python RLMs. Kanidm provides a Docker image for the RADIUS server that includes a RLM written in Python for operating with the Kanidm server.
+
+The goal of the project is to add support for Rust RLMs and to rewrite the current Kanidm Python RLM in Rust. An implementation of RADIUS is freeradius, which is written in C. Therefore, Rust RLM support requires FFIs to the freeradius C source code. The rewrite would allow for writing RLMs in Rust.
+
+Kanidm also provides a Python package (pykanidm) for Kanidm clients, which the Kanidm Python RLM uses. A possibility is to rewrite pykanidm in Rust.
+
+## Project plan
+A RLM defines custom scripts for various RADIUS activities. The Kanidm Python RLM currently defines custom scripts for instantiation of the RADIUS server, authentication of users, and authorization of users. A starting point would then be to add Rust RLM support and to rewrite the Python RLM in Rust with support for customizing scripts for these three activities. Afterwards, support for the remaining RADIUS activities can be considered.
+
+### Project plan steps
+1. Research how RADIUS calls the Kanidm Python RLM and which RADIUS functions need to be interfaced with to support Rust RLMs. Also look into how RLM custom scripts are structured as to be a valid RLM script.
+
+2. Plan a setup which is viable for testing while performing the rewrite.
+
+3. Start implementing support for the three RADIUS activities used by Kanidm as mentioned in the project description.
+
+4. Connect RADIUS to Kanidm such that the RLM funcionality that Kanidm uses can be tested. Perhaps pykanidm should also be rewritten in Rust?
+
+5. Continue implementing support for the remaining RADIUS features?
+
+6. ?


yaleman:
FreeRADIUS calls the Python module using their [rlm_python3](https://freeradius.org/modules/?mod=rlm_python3) plugin which supports Python 3 modules natively. 

If you're going to change this, the path I'd see would be a C-based module that makes FFI calls into Rust... which is pretty cursed, but if you're up for it, then 🥳

Firstyear:
I think the best place to start is to setup kani with freeradius and test it first.

Freeradius is written in C and can "dlopen" dynamic libraries (.so files). Because of how this works, the freeradius binary can call a specific "symbol" (aka function) that is exported as an entry point in that .so to start to use it.

A lot of this is documented here pretty coarsely https://wiki.freeradius.org/contributing/Modules3 

The idea of this would be that with Rust you can create a cdylib (which on linux becomes a .so) that has export/no_mangle symbols that match what freeradius expects when it dlopens the .so. Then the cdylib does all the same work that currently rlm_python does since rlm_python is itself a .so that is dlopened, and spawns a python interpretter to run the python code and pass thorugh the operations.

So this would mean looking at things like https://rust-lang.github.io/rust-bindgen/ which can assist here to automatically generate some of these entries, and then testing the module loading etc. 

Does that help explain? 

But regardless, setting up freeradius and testing it a bit will be absolutely necessary since the current "nice automatic" container method won't work while you're developing!

czzrr:
> The idea of this would be that with Rust you can create a cdylib (which on linux becomes a .so) that has export/no_mangle symbols that match what freeradius expects when it dlopens the .so. Then the cdylib does all the same work that currently rlm_python does since rlm_python is itself a .so that is dlopened, and spawns a python interpretter to run the python code and pass thorugh the operations.

Just to be sure:
[rlm_python](https://github.com/FreeRADIUS/freeradius-server/tree/master/src/modules/rlm_python) is the C code that allows FreeRADIUS to call Python code natively.
Is the idea to rewrite the whole thing in Rust, or keep the base in C (like rlm_python) with calls to "custom scripts" written in Rust?

Firstyear:
> > The idea of this would be that with Rust you can create a cdylib (which on linux becomes a .so) that has export/no_mangle symbols that match what freeradius expects when it dlopens the .so. Then the cdylib does all the same work that currently rlm_python does since rlm_python is itself a .so that is dlopened, and spawns a python interpretter to run the python code and pass thorugh the operations.
> 
> Just to be sure: [rlm_python](https://github.com/FreeRADIUS/freeradius-server/tree/master/src/modules/rlm_python) is the C code that allows FreeRADIUS to call Python code natively. Is the idea to

Correct,

> 
> 1. rewrite the whole thing in Rust with support for Rust custom scripts, interfacing with the FreeRADIUS C source code related to modules through FFI, or

Option 1. The whole thing will be Rust exposing FFI. Don't think about it as "scripts". It's a dedicated and focused module specific to our needs. 

> 2. keep the base in C (like rlm_python), following https://wiki.freeradius.org/contributing/Modules3, with FFI calls to Rust custom scripts?
> 
> **EDIT:** Rereading your comment and https://wiki.freeradius.org/contributing/Modules3, it sounds like 1. So things like the [module_t struct](https://wiki.freeradius.org/contributing/Modules3#module-development_module_t-struct) would be written in Rust in a C-compatible way with the help of rust-bindgen such that FreeRADIUS can properly load the module.

Yep!

Firstyear:
Sounds good. You should be able to take apart the dockerfile for rlm python to set it up initially then you can go from there. FreeRADIUS is *notoriously* difficult and annoying to setup, so please ask for help if needed! In fact FreeRADIUS may even be the *most difficult* piece of opensource software to configure that I've ever seen outside of sendmail.

czzrr:
I'm wondering how to set up a development environment for the Rust module.
Based on https://wiki.freeradius.org/contributing/Modules3, it sounds like we want to build the module and FreeRADIUS together. I can set up a development environment directly on my own machine, of course, but would it make sense to make a Docker image for this purpose, since we want to end up with one as the product anyway?
I suppose the current `rlm_python` Dockerfile can be used, but I would still need to rebuild FreeRADIUS with the module inside the container.

Firstyear:
While you're developing this, you'll likely want to use a VM or similar to actually do your builds and tests allowing you to more quickly rebuild and test against the freeradius install on the VM. Once you make progress here and start to finalise, then you'll be more able to work that into a dockerfile to build the module in the correct way.

czzrr:
Makes sense. Got ahead of myself there a bit.

Anyway, I played around with authenticating Kanidm users to the RADIUS server in the provided Docker image.
I'll try build RADIUS myself and get it working.

Firstyear:
You don't need to build radius from sources, you only need to install the "same" as the docker image as well as freeradius-devel (at a guess) that will let you build/bindgen a module to use on the vm. Does that help?

czzrr:
I think I misunderstood how FreeRADIUS knows about modules.
Is it the case that it loads (dynamically links) *all* modules (.so files) upon startup? If so, what you said makes sense. (FreeRADIUS does not need to be built with modules to know about them.)
The Rust module's .so file could then be placed among the others to be used.

Firstyear:
FreeRADIUS doesn't *link* to the modules. It dlopens them. All FreeRADIUS cares about is that your module has a certain set of known symbols exported that it can call after the dlopen completes. :)

Firstyear:
> I think it would be nice for me to start off by making a minimalistic module that does nothing but print some dummy message or returns a fixed rlm code.

That is exactly the right way to start :) 

> I looked at the freeradius-dev headers, and there's a lot of macros and dependencies for the relevant definitions. Is it ok to just ad-hoc/manually add whatever I need to a common header file that I can run bindgen on?

For sure, that's okay. You could also bind gen and then cherry pick what you need from what it generated. 

> 
> @Firstyear I will also begin preparing my application, which I will share with you elsewhere. 😀

Feel free to email me :)

Firstyear:
@czzrr Reminder to address this part in your gsoc submission too:

> Assessment: Our intent is to mentor and help you grow as a contributor - and we want to help you succeed! As this is our focus, we have different assessment criteria to many other projects.
> 
> Has the student:
> 
> participated actively in the project?
> written and submitted personal blog/journal or similar about the experience and what they have learnt and attempted?
> sufficiently attempted one or any of the offered deliverables? (Note - there is no requirement to complete the deliverable. Just evidence of a solid attempt)

PR #1424, opened 2023-03-05
RADIUS container fixes

yaleman:
Fixes the ongoing issues with container builds.

- Moves the RADIUS container to `freeradius/freeradius` which is based on Ubuntu and is regularly updated.
- Makes an error message in the RADIUS module nicer
- Allows scripted password auth via a secret env var because I'm a bad person and wanted to script server setup.
- Fixes some CLI i/o because I'm a worse person.

- [x] cargo fmt has been run

### DIFF
--- a/rlm_python/Dockerfile
+++ b/rlm_python/Dockerfile
@@ -1,31 +1,20 @@
-ARG BASE_IMAGE=opensuse/tumbleweed:latest
-FROM ${BASE_IMAGE} AS repos
-RUN \
-    --mount=type=cache,id=zypp,target=/var/cache/zypp \
-    zypper mr -k repo-oss && \
-    zypper mr -k repo-update
-
-# ======================
-FROM repos
-
+FROM freeradius/freeradius-server:latest
 EXPOSE 1812 1813
+ARG RADIUS_USER=freerad
+ENV TZ=Etc/UTC
+# These all need to be on one line else the cache ends up in the layers.
+RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone
 
-# These all need to be on one line else the rpm cache ends
-# up in the layers.
-RUN \
-    --mount=type=cache,id=zypp,target=/var/cache/zypp \
-    zypper install -y \
-    freeradius-client \
-    freeradius-server \
-    freeradius-server-python3 \
-    freeradius-server-utils \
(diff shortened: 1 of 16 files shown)


rlm_python/Dockerfile
```diff
@@ -1,33 +1,21 @@
...
-    zypper mr -k repo-oss && \
-    zypper mr -k repo-update
-
-# ======================
-FROM repos
-
+FROM freeradius/freeradius-server:latest
 EXPOSE 1812 1813
+ARG RADIUS_USER=freerad
+ENV TZ=Etc/UTC
```
Firstyear, line 4: I think we shouldn't embed the TZ at build here, but recommend people bind mount it in from the container host `"/etc/localtime:/etc/localtime:ro"`

tools/cli/src/opt/kanidm.rs
```diff
@@ -500,7 +503,11 @@ pub struct LoginOpt {
     #[clap(flatten)]
     copt: CommonOpt,
     #[clap(short, long)]
+    /// Use a passkey for authentication
```
Firstyear, line 506: The webauthn field is never used so I think we can remove it actually ....

tools/cli/src/cli/session.rs
```diff
@@ -228,9 +238,13 @@ impl LoginOpt {
 
     pub async fn exec(&self) {
         let mut client = self.copt.to_unauth_client();
-
-        // TODO: remove this anon, nobody should do default anonymous
-        let username = self.copt.username.as_deref().unwrap_or("anonymous");
+        let username = match self.copt.username.as_deref() {
+            Some(val) => val,
+            None => {
+                error!("Please specify a username to login with!");
```
Firstyear, line 244: Should say "specify a username with `-D` to login"

Firstyear (changes requested):

rlm_python/Dockerfile
yaleman: It's an unavoidable part of the build process - if you install tzdata, you have to select a timezone and there's no other way of fixing it 😡

Firstyear: :(

tools/cli/src/cli/session.rs
Firstyear: Did you push the update with this in it?

yaleman: I have now!

Firstyear (approved):

PR #1425, opened 2023-03-05
trying to fix oauth test failures

yaleman:
References failures in #1350

- [x] cargo fmt has been run

### DIFF
--- a/server/testkit/tests/oauth2_test.rs
+++ b/server/testkit/tests/oauth2_test.rs
@@ -345,13 +345,6 @@ async fn test_oauth2_openid_basic_flow(rsclient: KanidmClient) {
         .validate(&jws_validator, 0)
         .expect("Failed to verify oidc");
 
-    // This is mostly checked inside of idm/oauth2.rs. This is more to check the oidc
-    // token and the userinfo endpoints.
-    assert!(oidc.iss == Url::parse(&format!("{}/oauth2/openid/test_integration", url)).unwrap());
-    eprintln!("{:?}", oidc.s_claims.email);
-    assert!(oidc.s_claims.email.as_deref() == Some("oauth_test@localhost"));
-    assert!(oidc.s_claims.email_verified == Some(true));
-
     let response = client
         .get(format!("{}/oauth2/openid/test_integration/userinfo", url))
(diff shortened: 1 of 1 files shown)


Firstyear:
Just curious how this fixes the issue? It seemed from the screen shot to be more likely an off by one or race condition?

yaleman:
I was basically trying to do the "work" as close as possible, to get that `exp` value the same, since I guessed it was related to timing... doesn't fail on fast machines, does in CI.

Firstyear:
> I was basically trying to do the "work" as close as possible, to get that `exp` value the same, since I guessed it was related to timing... doesn't fail on fast machines, does in CI.

Looking at this it seems more like a race condition OR related to ntp and clock changes.

exp is "issued at time + seconds until expiry". However, internally we don't store seconds until expiry, we store the absolute expiry time. So that means we calc this as:

"issued at time + (expiry abs - current time).as_seconds()".

Generally, this should always result in the same result since iat and exp aren't changing, so that leaves the "shifting" parameter as current time. If between the calls to the userinfo and the oidc endpoint, ntp were to trigger, it's possible that this would cause the value in exp to shift.

So I think the actual fix here is to change the calculation of exp to not involve current time, but to be similar to the expires_in calculation used in access token response.

yaleman:
If `exp` is the expiry time, and we store that, why are we re-calculating that at all?

These two *should* be exactly the same?
- `issued at time + (expiry abs - current time).as_seconds()`.
- `expiry.unix_timestamp()`

... or we just change the test so we're not comparing two complex objects that are *allowed* to have differing fields 😄

Firstyear:
> If `exp` is the expiry time, and we store that, why are we re-calculating that at all?
> 
> These two _should_ be exactly the same?
> 
> * `issued at time + (expiry abs - current time).as_seconds()`.
> * `expiry.unix_timestamp()`
> 
> ... or we just change the test so we're not comparing two complex objects that are _allowed_ to have differing fields 😄

It's because internally we are using offset-date-time for the value, and oidc wants an i64. So I think this conversion that I did was because of that. I think it can be improved for certain.

Firstyear:
Superseded by #1434 which changes how exp works.

Issue #611, opened 2021-11-14
IP access controls for anonymous


Firstyear:
For anonymous searches we should consider limiting these to specific subnet to reduce scope of expose data in unauthenticated contexts.

Issue #345, opened 2021-01-28
hidden name history


Firstyear:
It may be useful to have a "name history" for previous legalname/displaynames/name fields so that username changes and associations can be looked up post-fact. this is a sensitive field so should be protected by default. 

Again, knowing the use cases here would be good. From a security view, to track the history of changes, it may be better to have an audit/event system (which is already a planned feature), where as from a support or service desk view it could be useful for undoing/rolling back changes (but this could also be done via the repl changelog.). 

If this implemented the feature should have a boolean config in the domain config, disabled by default. 

If enabled, another option is to prevent re-use of an already claimed name. IE if someone is named 'a' and renames to 'b', then no other account should be able to claim the name 'a' - it effectively becomes "archived". A question is how to handle this when the account is deleted. If the name should be "forever" archieved. 

Note that an abuse prevention is needed here so that if they name changes "too much" then it is disallowed or requires admin intervention.

MinhPhan8803:
Was the event system ever implemented? If so we might just be able to implement it for database name changes.
Wrt the claimed name, I think we may be able to track names *and* a unique user id. Then we can say that name "a" used to belong to people "c" "b" "d" such and such, without having to freeze the name.
Abuse prevention policy might also be made to be admin bound? They can specify if users can change name an indefinite number of times, or limit it to some specific value.
I'm just gathering ideas here. Also the non empty vec might prove useful here for querying name change events.

Firstyear:
There is an easier way! 

If you have a look at https://github.com/kanidm/kanidm/blob/master/server/lib/src/server/modify.rs you'll see there are calls to "plugins" which are kind of like layered transformers. For example https://github.com/kanidm/kanidm/blob/master/server/lib/src/plugins/cred_import.rs 

On a modify operation the plugin is given the entries before and after state so you can use that to compare the name to see if it changed. If it did, you can just append the old name to a list inside the currently changing entry. You could probably also add the CID (change id) of when the name was changed too into that list so you can see "the account had name X until CID A, and then changed to it's current name".

Does that make sense?

WRT to the name changing policy, that I think could be a part of the plugin in the future which can see the name is changing and then make other decisions. But lets start simple and just have the history to start with.

MinhPhan8803:
So from what I gather, Plugin is essentially a trait that takes in the new and old entries, and a db transaction then let you do whatever with them, is that correct?

Firstyear:
[...]

PR #1456, opened 2023-03-16
RFC 8705 proposal

### LINKED ISSUE #755: Support rfc 8705 - bind oauth2 bearer tokens to tls or browser sessions
Support rfc8705 which allows oauth2 bearer tokens to be bound to mTLS sessions. This can help to prevent API token exfil, and also assists with client validation and security scenarioes.

MinhPhan8803:
Fixes #755 

- [ ] cargo fmt has been run


yaleman:
mTLS can be a challenge, but that's always a challenge with anything cert-based.

I'd suggest having Kanidm do the cert management with a CA stored in the database so it can be used anywhere once replication's a thing. This means we'll have more control over the certificate issuance and a root of trust in the system.

Alternatively, have the admins be able to supply trusted CAs for systems like Hashicorp vault and k8s that generate their certificates for the systems on the fly.

MinhPhan8803:
> mTLS can be a challenge, but that's always a challenge with anything cert-based.
> 
> I'd suggest having Kanidm do the cert management with a CA stored in the database so it can be used anywhere once replication's a thing. This means we'll have more control over the certificate issuance and a root of trust in the system.
> 
> Alternatively, have the admins be able to supply trusted CAs for systems like Hashicorp vault and k8s that generate their certificates for the systems on the fly.

How would a Kanidm CA look like? Do we develop protocols and stuff.
I would lean more towards the second option, since some admins may already have a trusted CA they want to use before adopting/migrating to Kanidm. Or even support both (flexibility at the cost of dev time).

MinhPhan8803:
Now that I think about it, the first option would also have the benefit of reduced latency for acquiring certs, since everything will occur on the Kanidm server.

Firstyear:
Being a CA is a task in and of itself - because to be a CA you need OCSP for CRLs. The hard part of CA's is revocation not just issuing things.

You do seem to have a talent for finding the gnarliest and most difficult tasks in the server and taking a look at them! 

Having a "pluggable" ca api so Kani could contact things and coordinate certificate issuance would be pretty cool that's for certain but I think that would be a really long term kind of goal. 

As well, getting mtls to work with oauth here is actually also really hard because of how we currently use openssl with tide we can't pass through the client cert detail so easily. We are also planning to swap from tide to axum so any work we do here we'd need to redo when we get to axum.

And findally, dpop may actually be the better choice over mtls anyway so I think I'd rather look at dpop first since it's more like to work in the general case.

MinhPhan8803:
> Being a CA is a task in and of itself - because to be a CA you need OCSP for CRLs. The hard part of CA's is revocation not just issuing things.
> 
> You do seem to have a talent for finding the gnarliest and most difficult tasks in the server and taking a look at them!
> 
> Having a "pluggable" ca api so Kani could contact things and coordinate certificate issuance would be pretty cool that's for certain but I think that would be a really long term kind of goal.
> 
> As well, getting mtls to work with oauth here is actually also really hard because of how we currently use openssl with tide we can't pass through the client cert detail so easily. We are also planning to swap from tide to axum so any work we do here we'd need to redo when we get to axum.
> 
> And findally, dpop may actually be the better choice over mtls anyway so I think I'd rather look at dpop first since it's more like to work in the general case.

Hmm do you think a proof of concept for this in axum right now would work? Like it will need integration into the actual server when server gets rewritten but if this is successful we will already kinda have rfc 8705 compliance by that point.

Also I assume that if we use dpop we will also run into the same "tide now but axum later" problem right?

I will also need to look into CAs more closely for the proposal then.

Firstyear:
> 
> Hmm do you think a proof of concept for this in axum right now would work? Like it will need integration into the actual server when server gets rewritten but if this is successful we will already kinda have rfc 8705 compliance by that point.

That's not really the problem in a way. Getting it to work with a PoC and axum is "do-able". It'll be a pain in the ass because of having to do url-specific openssl settings for client authentication though. 

But also the bigger issue is client configuration, documentation, the CA's, issuing certs, testing and more. It's a huge task. That's why we wanted to look at dpop over this because dpop is *way* simpler but solves the same thing in an "anonymous" way without needing a full CA stack. 

> 
> Also I assume that if we use dpop we will also run into the same "tide now but axum later" problem right?

Actually no! dpop is bearer/header based, so it won't hit this.

Where 8705 is mtls based so it relies on that lower level connection authentication. 

> 
> I will also need to look into CAs more closely for the proposal then.

I think writing a CA is a whole proposal in itself let alone 8705. 

So I think you should probably look at dpop instead if you want to pursue this.

MinhPhan8803:
I added a section for DPoP! Can you look through that to see what I'm missing? Also I noticed there's some implementation details shared between DPoP and RFC 8705, which might be helpful at some point.

Firstyear:
Hey there I'm just gonna ask if you've sat down and read both https://www.rfc-editor.org/rfc/rfc6749 and https://datatracker.ietf.org/doc/html/draft-ietf-oauth-dpop ? 

I'm rereading the dpop draft and I have some concerns about the spec and the scope of what this would represent as a project for you, so I am concerned it may be too large - if not impossible for you. 

Currently in my reading I believe it is actually not possible to implement dpop because the specific incorrectly requires the dpop keys on both the authorisation server domain and the resource server domain - this is not possible to achieve due to how webcrypto apis work. I have contacted the draft rfc authors about this to clarify.

Firstyear:
Said a different way - I do not believe *I* could correctly implement dpop today due to limitations of the specification. So I don't want you to apply for something I wouldn't be confident to do yourself.

MinhPhan8803:
Is there active work integrating SAML right now? I see one for SIMC but doesn't look like it's quite complete yet.

Firstyear:
> Is there active work integrating SAML right now? I see one for SIMC but doesn't look like it's quite complete yet.

You have a special talent you know?  🤣 It's always choosing the most complex and difficult issues possible to investigate!

SAML is fucked on many levels because of how it's signatures work, and it's extremely prone to security vulnerabilities. It's why I'm trying to avoid adding it as much as possible. I think to want to tackle SAML would probably be a long task, more than what GSoC provides. 

Part of this could be that you are picking issues you think are "large" to make a "big" project for GSoC but the issue is that those large issues are *really* large. There is nothing wrong with submitting an intent to resolve multiple smaller issues if you want. The challenge is probably that the other "moderate" sized issues are already being looked at by two other prospective students. 

An easier option would likely be hidden name history (  https://github.com/kanidm/kanidm/issues/345 ) and then combine that with other things like possibly ip access controls for anonymous ( https://github.com/kanidm/kanidm/issues/611 ) which actually is two tasks - associate ip metadata on operations / tracing spans, followed by actually having ip bound access rules. 

Does that help?

MinhPhan8803:
Well I can't deny those are exactly my problems :) 
I will look at the issues you pointed to, thank you!

> > Is there active work integrating SAML right now? I see one for SIMC but doesn't look like it's quite complete yet.
> 
> You have a special talent you know? 🤣 It's always choosing the most complex and difficult issues possible to investigate!
> 
> SAML is fucked on many levels because of how it's signatures work, and it's extremely prone to security vulnerabilities. It's why I'm trying to avoid adding it as much as possible. I think to want to tackle SAML would probably be a long task, more than what GSoC provides.
> 
> Part of this could be that you are picking issues you think are "large" to make a "big" project for GSoC but the issue is that those large issues are _really_ large. There is nothing wrong with submitting an intent to resolve multiple smaller issues if you want. The challenge is probably that the other "moderate" sized issues are already being looked at by two other prospective students.
> 
> An easier option would likely be hidden name history ( #345 ) and then combine that with other things like possibly ip access controls for anonymous ( #611 ) which actually is two tasks - associate ip metadata on operations / tracing spans, followed by actually having ip bound access rules.
> 
> Does that help?

PR #1457, opened 2023-03-18
Correcting command typo, works with kanidm/server img ac1fddd16083

jonathanknez:
Fixes # - NA, I am lazy and this is just a doc change.

- [ ] cargo fmt has been run

### DIFF
--- a/book/src/server_configuration.md
+++ b/book/src/server_configuration.md
@@ -48,7 +48,7 @@ text=The server must not be running at this point, as it requires exclusive acce
 
 ```bash
 docker run --rm -i -t -v kanidmd:/data \
-    kanidm/server:latest /sbin/kanidmd recover-account -c /data/server.toml admin
+    kanidm/server:latest /sbin/kanidmd recover_account -c /data/server.toml admin
 # success - recovery of account password for admin: vv...
 ```
 


yaleman:
So... this is awkward. The *current* command is `recover-account`, which changed a while ago...


https://github.com/kanidm/kanidm/blob/d26bef17b679a8a4944d87fef52cdaf3ef46042c/server/daemon/src/opt.rs#L139

jonathanknez:
Thanks for checking. I pulled index.docker.io/kanidm/server:latest on my raspberry pi and got an image from 2022-11-09. Checking your release history, that probably makes it one release old, v1.1.0-alpha.10. Do you know if there is an alpha11 pre-built image available?

Firstyear:
I'm wondering if we botched something with the arches on the aarch images? There has been some weirdness with docker/buildx recently wher they are setting arches from things like "linux/amd64" to "linux/amd64/v2" for example. 

Looking at https://hub.docker.com/repository/docker/kanidm/server/tags?page=1&ordering=last_updated it's showing these as last pushed 4 months ago which means we may have missed a step in release? 

I'll try to fix these images up today @jonathanknez

Firstyear:
Ahh that's right, it's related to https://github.com/docker/buildx/issues/1686

Firstyear:
Okay, I have fixed the published images for latest @jonathanknez really sorry about that.

PR #1475, opened 2023-03-20
Server daemon logging and exit codes

### LINKED ISSUE #1472: Update server daemon logging and exit codes

The main server daemon looks outdated compared to the unixd daemon (use of `eprintln!`'s instead of log macros, and of course force shut down instead of returning exit codes).

A slight rewrite of course.

Leave it be if the daemon is earmarked for rewrite at some point.

**Additional context**

MinhPhan8803:
Fixes #1472 

- [x] cargo fmt has been run

### DIFF
--- a/server/daemon/src/main.rs
+++ b/server/daemon/src/main.rs
@@ -167,18 +167,18 @@ async fn main() {
                 let cfg_meta = read_file_metadata(&(opt.commands.commonopt().config_path));
 
                 if !kanidm_lib_file_permissions::readonly(&cfg_meta) {
-                    eprintln!("WARNING: permissions on {} may not be secure. Should be readonly to running uid. This could be a security risk ...",
+                    warn!("permissions on {} may not be secure. Should be readonly to running uid. This could be a security risk ...",
                     opt.commands.commonopt().config_path.to_str().unwrap_or("invalid file path"));
                 }
 
                 if cfg_meta.mode() & 0o007 != 0 {
-                    eprintln!("WARNING: {} has 'everyone' permission bits in the mode. This could be a security risk ...",
+                    warn!("WARNING: {} has 'everyone' permission bits in the mode. This could be a security risk ...",
                     opt.commands.commonopt().config_path.to_str().unwrap_or("invalid file path")
                     );
                 }
 
                 if cfg_meta.uid() == cuid || cfg_meta.uid() == ceuid {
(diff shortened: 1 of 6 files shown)


MinhPhan8803:
Wait should all the `debug!` be `info!` instead? Considering we don't have an option to actually toggle `RUST_LOG=debug`.

Firstyear:
> Wait should all the `debug!` be `info!` instead? Considering we don't have an option to actually toggle `RUST_LOG=debug`.

Well, there is a way to do it, but I think you should think about this from the view of a user. 

What messages should they always see by default (ie info, warn, error)? What is communicating to them about how the server is running and why it's doing what it's doing? 

So some of these messages should be info, some should stay as debug. I'll let you step back and think about that user experience for now, but if you want I can tell you exactly which messages should change.

MinhPhan8803:
Hmm I will think about it.

MinhPhan8803:
@Firstyear  I changed a few log calls (and did some clean up in the unixd tools).

server/daemon/src/main.rs
```diff
@@ -325,104 +347,104 @@ async fn main() {
                                     }
                                     }
                                 }
-                                eprintln!("Signal received, shutting down");
+                                debug!("Signal received, shutting down");
```
Firstyear, line 350: Should be an info/warn, because the user will want feedback to know the server hasn't hung and is actually responding to their reqest to shutdown :)

Firstyear (changes requested):
Just one thing, other wise looks really good :)

server/daemon/src/main.rs
MinhPhan8803: That should be fixed now

Firstyear:
Just needs a rebase on master and I'll merge it :)

MinhPhan8803:
Wait we still don't yet have a way to toggle `RUST_LOG=debug`

Firstyear:
We do, you just set it in the environment external the the daemon.

MinhPhan8803:
I see

Firstyear:
Remember this is commonly run from docker so setting -e flags is really easy :)

MinhPhan8803:
Yeah I didn't realize that :) only looked at EnvFilter.
Also rebase is done!

Firstyear:
> Yeah I didn't realize that :) only looked at EnvFilter. Also rebase is done!

No stress at all :) It's the challenge of stepping into a large and new project, there is a lot there to learn about how it works.

Firstyear (approved):

PR #1478, opened 2023-03-22
Fix path-relative links in documentation

terrorbyte:
The documentation for [pam and nsswitch integrations was not path relative](https://kanidm.github.io/kanidm/stable/integrations/pam_and_nsswitch.html) and would 404 in all links to documentation. This fixes them by pointing to the parent directory. I also ran [muffet](https://github.com/raviqqe/muffet) to triple check there were no other locations where this was occurring and did not identify any.

Fixes #

- [ ] cargo fmt has been run


Firstyear:
Thank you !

PR #1485, opened 2023-03-25
Proposal for name change history

### LINKED ISSUE #345: hidden name history

MinhPhan8803:
References #345 

- [ ] cargo fmt has been run

Pretty short proposal for now.

### DIFF
--- a/PROPOSAL.md
+++ b/PROPOSAL.md
@@ -1,4 +1,6 @@
 # Proposal for a way to store name history
+## Summary
+This proposal includes a solution for Issue [#345](https://github.com/kanidm/kanidm/issues/345) of this project, the high level implementation details, related tasks and potential further development.
 
 ## The implementation
 The idea is to make use of the currently existing server plugin system. Currently, transaction entries are ran through a series of plugins that can look at them and perform actions on them.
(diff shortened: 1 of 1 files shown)


yaleman:
Updated your comment so it shouldn't auto-close the issue, if this is just a proposal.

MinhPhan8803:
> Updated your comment so it shouldn't auto-close the issue, if this is just a proposal.

Thank you! I also changed the PR name to be more descriptive. The previous one was written midnight and was eh not the best.

Firstyear:
You could probably expand this a bit, wrt to testing and include the link to the issue. Also note that if you finish "early" there are plenty of other things we can work on with you too, so possibly include that in there as well. :)

MinhPhan8803:
I just added a few stuff to the proposal :)

MinhPhan8803:
Should I put project size as medium or large?

Firstyear:
medium I think :)

MinhPhan8803:
Also is there anything I should elaborate on wrt to testing and related stuff?

Firstyear:
That's all expected anyway, test design and inclusion was always part of any featuire IMO so doesn't need special call outs. Just re-review the original task per https://github.com/openSUSE/mentoring/issues/196 and make sure you can do the "assessment tasks"

MinhPhan8803:
Thank you!

MinhPhan8803:
Wait the assessment looks like it would be for *during* the project right?

Firstyear:
What do you mean by "during"?

MinhPhan8803:
I mean like for if we are selected (for GSOC), the assessment will be used for the mentor evaluation part.

Firstyear:
Yeah, I'm saying that in your application you need to understand that during the project you'll need to write about your experiences and contribute generally, so bring up that in your application  and how you want to approach it.

MinhPhan8803:
I see thank you :)

MinhPhan8803:
@Firstyear I just sent an email to your SUSE address. Can you take a look at that?

PR #1498, opened 2023-03-27
WIP: Add Windows Sign On Support

Water261:
Fixes #

- [ ] cargo fmt has been run


micolous:
> I've been talking with @Firstyear and they said you would have some insight about a better way to implement this. This current implementation relies on `static mut` struct which holds all the logic with some wrapper functions.

Hi there! I'm not familiar with the LSA API; so my commentary is based on what I read of the docs, experience working with other Windows APIs, and trying to build stuff for [pGINA](https://github.com/pgina/pgina) a very very long time ago. 😅 

As ugly as it is, a `static mut` is probably the only answer for some of this - the API seems to be built around plug-ins having their own global state, and no "context" or "object" parameter.

The particularly messy part seems to be that for any structures crossing FFI boundaries, LSA assumes that memory was allocated by [`LsaAllocateLsaHeap()`](https://learn.microsoft.com/en-us/windows/win32/api/ntsecpkg/nc-ntsecpkg-lsa_allocate_lsa_heap) and will try to free it when it's done. This will probably need a custom [`GlobalAlloc`](https://doc.rust-lang.org/std/alloc/trait.GlobalAlloc.html); with the caveat it appears that LSA's allocator is a simple `malloc()`-like function with no memory alignment handling.

windows_integration/lsa_mode/src/lib.rs
```diff
@@ -0,0 +1,70 @@
...
+    lsaversion: u32,
+    pkg_ver: *mut u32,
+    pptables: *mut *mut SECPKG_FUNCTION_TABLE,
+    pctables: *mut u32,
+) -> NTSTATUS {
+    unsafe {
+        *pkg_ver = 1u32;
+        *pctables = 39u32;
+
+        let mut tbl_ref = *(*pptables);
```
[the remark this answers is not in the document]
Firstyear: Yeah, I think you'll need a static mut here IMO.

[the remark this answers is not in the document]
Water261: Yeah, fixed it now so the lib creates its own now.

windows_integration/lsa_mode/src/lib.rs
```diff
@@ -0,0 +1,70 @@
...
+#[allow(non_snake_case)]
+pub extern "system" fn SpLsaModeInitialize(
+    lsaversion: u32,
+    pkg_ver: *mut u32,
+    pptables: *mut *mut SECPKG_FUNCTION_TABLE,
+    pctables: *mut u32,
+) -> NTSTATUS {
+    unsafe {
+        *pkg_ver = 1u32;
+        *pctables = 39u32;
```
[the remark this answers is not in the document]
Water261: Fixed

Water261:
Alright, refactor time no. 2. Instead of just being a single lib, there's a daemon which will handle interacting with the kanidm server and caching it correctly. As the SSP/AP will interact with the daemon through windows sockets

windows_integration/src/auth/wrapper.rs
```diff
@@ -0,0 +1,154 @@
...
+        Security::{Authentication::Identity::*, Credentials::STATUS_LOGON_FAILURE},
+        System::Kernel::*,
+    },
+};
+
+use super::package::{AuthError, AuthPackage, AuthInfo};
+
+static mut AUTH_PACKAGE: Lazy<AuthPackage> = Lazy::new(|| AuthPackage::new());
+
+#[tokio::main]
```
Firstyear, line 16: You can wrap tokio::main around a pub async extern like this? This blows my mind.

windows_integration/src/lib.rs
```diff
@@ -0,0 +1,98 @@
...
+    Foundation::{NTSTATUS, STATUS_SUCCESS},
+    Security::Authentication::Identity::SECPKG_FUNCTION_TABLE,
+};
+
+mod auth;
+mod security;
+
+/// # Safety
+/// This should only ever be called by the windows api, and FFI with C++ is always unsafe
+/// So beware of demons I guess :shrug:
```
Firstyear, line 12: Why would this be unsafe? where is this demonic references you mention?

windows_integration/src/lib.rs
```diff
@@ -0,0 +1,98 @@
...
+#[no_mangle]
+#[allow(non_snake_case)]
+pub unsafe extern "system" fn SpLsaModeInitialize(
+    lsa_version: u32,
+    pkg_ver: *mut u32,
+    pptables: *mut *mut SECPKG_FUNCTION_TABLE,
+    pctables: *mut u32,
+) -> NTSTATUS {
+    event!(Level::INFO, "Initialising kanidm Windows client");
```
Firstyear, line 21: Capitalise Kanidm I think

windows_integration/src/lib.rs
```diff
@@ -0,0 +1,98 @@
...
+    event!(
+        Level::INFO,
+        "Local Security Authority Version {}",
+        lsa_version
+    );
+    event!(Level::INFO, "Client Version v{}", env!("CARGO_PKG_VERSION"));
+
+    let package_version_str = format!(
```
Firstyear, line 29: You may need to change these format strings to pad leading zeros since version 1.0.0 would be greater than 0.10.100 but they would format string to 100 and 10100 respectively.

Firstyear (commented):
This is looking really good so far code wise! It would be great to see a README.md in the windows_integration folder that describes "how to test/develop" this post build. Currently while you can do a cargo build to get a dll it's not clear "what" to do with that dll once it exists.

Firstyear:
Also important is "where do the log messages end up" since that'll be needed for debugging :)

windows_integration/src/auth/wrapper.rs
Firstyear, line 16: Forgot to say, you should probably make this "current thlread" to prevent spawning extra threads inside the auth package. Alternately, if you do rt multi-thread, limit it to one created thread.

Water261: Fixed with [e769657](https://github.com/kanidm/kanidm/pull/1498/commits/e7696572064a414b3b1bfd81f5958edecde9cc9a)

windows_integration/README.md
```diff
@@ -0,0 +1,27 @@
...
+
+## Installation
+In order for the client to function, all parts must be installed in the correct place and registry values set. In the future there will be a powershell script and maybe MSI package to automate this.
+
+## SSP/AP
+To install the SSP/AP follow the below steps:
+* Create the directory `C:\Program Files\kanidm`
+* Copy the file `kanidm_win_authlib.dll` to the newly created directory
+* Open the Registry Editor and navigate to `HKEY_LOCAL_MACHINE\System\CurrentControlSet\Control\Lsa`
+* Add the path `C:\Program Files\kandim\kandim_win_authlib\0` (The \0 is a null character)
```
Firstyear, line 15: Should be to the Key "Security Packages".

windows_integration/README.md
```diff
@@ -0,0 +1,27 @@
...
+Currently the client consists of one part, the Security Support Provider & Authentication Package (SSP/AP).
+
+## Building
+The only supported toolchain to build Kanidm's Windows Client is `x86_64-pc-windows-msvc` which requires the MSVC toolchain to be installed. You can get the toolchain with the Visual Studio Workload `Desktop Development with C++`. The toolchain is also available with the same workload in the [Visual Studio Build Tools](https://visualstudio.microsoft.com/downloads/#build-tools-for-visual-studio-2022).
+
+## Installation
+In order for the client to function, all parts must be installed in the correct place and registry values set. In the future there will be a powershell script and maybe MSI package to automate this.
+
+## SSP/AP
+To install the SSP/AP follow the below steps:
```
Firstyear, line 11: you can also do this with powershell:

```
Get-ItemProperty -Path "HKLM:\System\CurrentControlSet\Control\Lsa" -Name "Security Packages"
Set-ItemProperty -Path "HKLM:\System\CurrentControlSet\Control\Lsa" -Name "Security Packages" -Value "C:\Program Files\kanidm\kandim_win_authlib\0"
```

Firstyear (commented):
I've tried this out, but I think I'm missing something - I can't get the AP to load or do anything? Most notably, the log file isn't appearing. Do I need to create that manually?

Water261:
> I've tried this out, but I think I'm missing something - I can't get the AP to load or do anything? Most notably, the log file isn't appearing. Do I need to create that manually? 

The log file should be automatically created. More likely the package failed to initialise. I believe the security support provider functions also have to be implemented for the package to load successfully, so I'll add them in and let you know

windows_integration/README.md
```diff
@@ -0,0 +1,27 @@
...
+* Create the directory `C:\Program Files\kanidm`
+* Copy the file `kanidm_win_authlib.dll` to the newly created directory
+* Open the Registry Editor and navigate to `HKEY_LOCAL_MACHINE\System\CurrentControlSet\Control\Lsa`
+* Add the path `C:\Program Files\kandim\kandim_win_authlib\0\0` (The \0 is a null character)
```
Firstyear, line 15: "Security Packages"

Firstyear: Kan-d-im is incorrect spelling, needs to be kan-i-dm.

Firstyear: Get-EventLog -LogName Security -EntryType Error
Worth putting in the powershell version of the how to get event log to help debug the integration.

Firstyear: Set-ItemProperty -Path "HKLM:\System\CurrentControlSet\Control\Lsa" -Name "Security Packages" -Value "C:\Program Files\kanidm\kanidm_win_authlib\0"

Firstyear: Actually, I think you only need the \0 to seperate multiple dlls - because there is only one here you don't need it and I think that's what's throwing off the reg.

windows_integration/security_package/src/protocol/v1.rs
```diff
@@ -0,0 +1,95 @@
+use kanidm_windows::secpkg::ap_proto::v1::*;
+use tracing::{event, Level};
+
+use crate::package::KANIDM_WINDOWS_CLIENT;
+
+pub async fn handle_request(request: &AuthPkgRequest) -> AuthPkgResponse {
+    return match request {
```
Firstyear, line 7: @Water261 I think that the best way to proceed here will be to simply try running the module like this and attempt a login and see what breaks during usage. We can "guess" what we may need to implement here, but knowing exactly what windows actually calls would be the best way to proceed and implement things. That will also let us know if this is all connected up correctly too. When it comes to SID generation, I have an algorithm in mind too :)

Firstyear:
For future we'll want https://learn.microsoft.com/en-us/windows-server/security/windows-authentication/security-support-provider-interface-architecture

windows_integration/client/src/convert.rs
```diff
@@ -0,0 +1,29 @@
+use tracing::{event, Level};
+use windows::{Win32::Foundation::UNICODE_STRING, core::PWSTR};
+
+pub fn unicode_to_rust(string: UNICODE_STRING) -> Option<String> {
+	match unsafe { string.Buffer.to_string() } {
```
micolous, line 5: This is unsafe, and has buffer overrun problems: https://github.com/microsoft/windows-rs/issues/2566

windows_integration/client/src/convert.rs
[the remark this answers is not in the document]
micolous: Resolving in favour of latter comment

windows_integration/client/install.ps1
```diff
@@ -0,0 +1,21 @@
+$kani_dir = "C:\Program Files\kanidm"
```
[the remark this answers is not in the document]
Water261: 👍 Will convert to registry entry to control location of the config & log

windows_integration/client/src/lib.rs
```diff
@@ -0,0 +1,126 @@
...
+#[tokio::main(flavor = "current_thread")]
+#[no_mangle]
+#[allow(non_snake_case)]
+pub async unsafe extern "system" fn SpLsaModeInitialize(
+    lsa_version: u32,
+    pkg_ver: *mut u32,
+    pptables: *mut *mut SECPKG_FUNCTION_TABLE,
+    pctables: *mut u32,
+) -> NTSTATUS {
+    let file_appender = tracing_appender::rolling::daily(PROGRAM_DIR, "authlib.log");
```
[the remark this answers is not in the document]
Water261: iirc default NT permissions allow for reading by all users, so this does need to be fixed

windows_integration/client/src/lib.rs
```diff
@@ -0,0 +1,126 @@
...
+    let (non_blocking, _guard) = tracing_appender::non_blocking(file_appender);
+
+    tracing_subscriber::fmt().with_writer(non_blocking).init();
+
+    event!(Level::INFO, "Initialising Kanidm Windows client");
```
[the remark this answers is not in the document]
Water261: Probably, might change it at somepoint

windows_integration/client/src/lib.rs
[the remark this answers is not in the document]
yaleman: Can you log this to the application or security log instead?

windows_integration/client/install.ps1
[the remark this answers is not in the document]
Water261: Fixed with 3d42cf6446f43ac7ec2c35231a1a530c317e6a9d

windows_integration/client/src/convert.rs
Water261: @micolous After reading the issue I believe this is fixed with 1954fb05cf7b9217708f86565c3690826f24cdba

windows_integration/client/src/lib.rs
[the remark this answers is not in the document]
Water261: @yaleman With d2307aadb282a001947a2f322006bb398d06aae0 it should log to `Applications and Services Log` now

windows_integration/client/src/convert.rs
micolous: It's not, calling `Buffer.to_string()` is unsafe because it uses `wcslen` to determine the length of the string rather than the `Length` field.

You need to access the raw `*mut u16` pointer from inside the `PWSTR` (`Buffer.0`), and form a slice using the `Length` parameter:

```rust
if string.Length == 0 {
  // MSDN docs state Buffer could be empty for zero length
  return Some(String::new());
}
let result = String::from_utf16(unsafe { std::slice::from_raw_parts(string.Buffer.0, string.Length.into() / 2) });

// error handling for UTF-16 decoding
// ...
```

You should also check:
* `string.Buffer.0` isn't null, and it's properly ali

Water261:
After discussing with a Microsoft employee, we've found out that custom SSP/APs are unsupported and break Microsoft's support requirements (aka no support from MS if custom SSP/APs in use). However this isn't the end of this idea, the recommendation provided was to use Kerberos. This would require further discussion.

PR #1502, opened 2023-03-30
20230330 oauth2 refresh

Firstyear:
This is the design document for oauth2 token refresh. I wanted to implement it today but ended up on a rabbit hole reading the rfcs and specs and considering how it should all work to ensure that we don't expose and security vulnerabilities. I'm mostly happy with the design but I want to think on it for a bit too. Currently I'm thinking the current async/delayed work queue system may need a clean up and rework in the future though to assist with monitoring and general program cleanliness. We may also need to consider a similar refresh or liveness check system for our main session tokens too. 

- [ ] cargo fmt has been run


Firstyear:
Not ready for merge, just for questions/comments.

Firstyear:
I'm thinking to maybe undo the async write for this part, since any code exchange always needs a write, so there is no benefit to putting the write out of band - the delayed/async is for "maybe" you need an update situations like auth where it's hard to know what phase of the auth the user is in so we can't always take write txns to wrap it we want to keep them ro as much as possible and write as needed.

book/src/developers/designs/oauth2_refresh_tokens.md
```diff
@@ -0,0 +1,161 @@
...
+Since the act of refreshing a token is implied activity then we do not require other signaling
+mechanisms.
+
+# Questions
+
+Currently with authorisation code grants and sessions we issue these where the sessions are recorded
+in an async manner. For consistency I believe the same should be true here but is there a concern
+with the refresh being issued but a slight delay before it's recorded? I think given the nature of
+our future replication we already have to consider the async/eventual nature of things, so this
+doesn't impact that further, and may just cause client latency in the update process.
```
yaleman, line 158: That's something to measure when the replication's built I guess - it's the kind of thing you'd want a completion contract internally for these kinds of high-risk write transactions. 

If it's going to take more than x seconds to write a session out locally (ignoring replication delays), then just go into blocking mode because something's going wrong and it's better to hold off on auth attempts than allow race conditions to occur.

book/src/developers/designs/oauth2_refresh_tokens.md
```diff
@@ -0,0 +1,161 @@
...
+
+While this allows a short window where a former access token could be used on the second replica,
+this infrastructure being behind load balancers and outside of an attackers influence significantly
+hinders the ability to attack this for very little gain.
+
+## Attack Detection
+
+[draft oauth security topics section 4.14.2](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-security-topics#section-4.14.2)
+specifically calls out that when refresh token re-use is detected then all tokens of the session
+should be canceled to cause a new authorisation code flow to be initiated.
```
yaleman, line 136: This is definitely good to include, and I'd specifically call out an audit/alarm event (once it's implemented).

yaleman (commented):
Seems sensible to me so far!

book/src/developers/designs/oauth2_refresh_tokens.md
Firstyear: Agreed, I think i'll just put this into the write txn path. Our writes are extremely fast anyway so I think it's not an issue.

book/src/developers/designs/oauth2_refresh_tokens.md
Firstyear: Yep agreed. I'm going to add some tests for this too.

Firstyear:
And yes, this is not only tested in the tests, but also tested in production too :)

yaleman (approved):

server/lib/src/plugins/session.rs
```diff
@@ -67,7 +67,7 @@ impl SessionConsistency {
                 .map(|sessions| {
                     sessions.iter().filter_map(|(session_id, session)| {
                         if !cred_ids.contains(&session.cred_id) {
-                            info!(%session_id, "Removing auth session whose issuing credential no longer exists");
+                            warn!(%session_id, "Removing auth session whose issuing credential no longer exists");
```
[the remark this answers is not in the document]
Firstyear: Reverted.

Issue #267, opened 2020-06-15
Non Empty Vector


Firstyear:
In a lot of places we use a vector, but we have to assume 0 to infinite items. We need a non-empty vec type, probably based on small vec for storage of entry attributes (away from btreeset to improve allocations)

MinhPhan8803:
Is this requirement only concerning parse types (similar to #1452 ) or do we have other use cases where it is desired?
Also how should this vec behave when size reaches 0?

Firstyear:
it should never let you reach 0, by refusing the last "remove" operation from proceeding. 

It's not just for parsers, but some internal types it would be useful for us to be able to prevent and detect cases where structures that must have at least one value.

It would also be useful in our proto structures so that clients can see that values always have at least one. 

It was a while ago that I looked into libraries that provided non-empty vecs but they all had some limitation at some point in the past. However that may have changed now. 

A good way to think about it is numbers 0, 1 and infinite. We have Option<T> which is 0 or 1. We have Vec which is 0 to infinite. But we don't have NonEmptyVec for 1 to infinite.

MinhPhan8803:
Is there some specific example of the requirements our non empty vec that existing crates (in the past) couldn't meet?
Also if I'm not mistaken, refusing the "remove" operation could mean something like returning an `Err()` right? (Or panicking but that'd be awful).

MinhPhan8803:
I'm thinking of testing with this crate first https://docs.rs/nonempty/0.8.1/nonempty/.
It has a Vec like API and serde support. The obscure name is a problem though.

Firstyear:
I don't mind the name? looks like a good crate though, so I'd be keen to use it as a solution to this issue.

MinhPhan8803:
Aight I'm gonna try the crate out.

MinhPhan8803:
Would `DbEntryV1` in this file https://github.com/kanidm/kanidm/blob/master/server/lib/src/be/dbentry.rs be candidate for the the non empty vec? Since the implementation is essentially "this should never happen".

Firstyear:
Yep it would, but we'd need to handle that error properly up through the be code :) 

With a change like this we don't have to do everything at once, we can just do a bit at a time btw.

MinhPhan8803:
Yeah I will probably try to handle this first. Currently I'm just going through all the Vec's in code, see what they look like.

MinhPhan8803:
[...]

PR #1522, opened 2023-04-05
Authsession non empty vec

MinhPhan8803:
References #267 

- [x] cargo fmt has been run

I tried to update `AuthSessionState::Init` to also use a non empty vec since it looks like there's no path for it to be empty. But that got a few tests to panic so ig it doesn't make sense?


Firstyear:
What tests failed? There may be some casess where in auth session init we do actually allow an empty vec to exist to denote the lack of valid credentials.

MinhPhan8803:
One of the failing tests is `test_idm_authsession_webauthn_only_mech()`. There are 3 in total but I forgot what the other 2 are.

Firstyear:
Might be better to investigate in a new PR at some point I think.

MinhPhan8803:
Will do

PR #1543, opened 2023-04-15
AuthSession non empty vec part 2

MinhPhan8803:
References #267 

- [x] cargo fmt has been run

Update `AuthSessionState::Init`.
Tests actually pass now so the last time I probably messed up somehow.
Warning: ugly code ahead.

### DIFF
--- a/server/lib/src/idm/authsession.rs
+++ b/server/lib/src/idm/authsession.rs
@@ -698,7 +698,7 @@ impl CredHandler {
 /// "InProgress", "Success" or "Denied". From there the CredHandler
 /// is interacted with until we move to either "Success" or "Denied".
 enum AuthSessionState {
-    Init(Vec<CredHandler>),
+    Init(NonEmpty<CredHandler>),
     // Stop! Don't make this a vec - make the credhandler able to hold multiple
     // internal copies of it's type and check against them all.
     //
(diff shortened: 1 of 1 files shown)


server/lib/src/idm/authsession.rs
```diff
@@ -757,34 +757,41 @@ impl AuthSession {
...
-
-                if let Ok(ch) = CredHandler::try_from((&account.passkeys, webauthn)) {
-                    handlers.push(ch);
-                };
-
-                if handlers.is_empty() {
-                    security_info!("account has no available credentials");
-                    AuthSessionState::Denied("invalid credential state")
-                } else {
-                    AuthSessionState::Init(handlers)
```
Firstyear, line 787: I think a better idea is to use the traditional vec approach to build handlers, then to convert it to non-empty in this step. It keeps the code cleaner *and* still ends up with the non empty at the end.

Firstyear (changes requested):

server/lib/src/idm/authsession.rs
MinhPhan8803: That approach might have its own ugliness since a non empty can't be straight collected from an iterator. I will try it out and see what it looks like.

MinhPhan8803: Actually we can use slice pattern here.

Firstyear: Yeah, I think there are ways to do this nicely.

MinhPhan8803:
As it turns out, the crate has its own `collect`.

server/lib/src/idm/authsession.rs
```diff
@@ -780,11 +780,11 @@ impl AuthSession {
                     handlers.push(ch);
                 };
 
-                if handlers.is_empty() {
+                if let Some(non_empty_handlers) = NonEmpty::collect(handlers.into_iter()) {
```
Firstyear, line 783: This looks super nice as a solution! Much improved :)

Firstyear (approved):

PR #1554, opened 2023-04-20
Fix incompatible future warnings by removing older crates

Firstyear:
Rust compiler was giving a warning about some of our dependencies being incompatible in future. It turned out that actually, we could have just use regex in both cases allowing us to remove validator and unicode-general-category as deps.


yaleman (approved):
Woot! Looks great

PR #1555, opened 2023-04-20
Idm Group non empty vecs

MinhPhan8803:
References #267 

- [x] cargo fmt has been run

Use a non empty vec for creating groups.

### DIFF
--- a/server/lib/src/idm/group.rs
+++ b/server/lib/src/idm/group.rs
@@ -2,6 +2,7 @@ use std::collections::BTreeSet;
 
 use kanidm_proto::v1::UiHint;
 use kanidm_proto::v1::{Group as ProtoGroup, OperationError};
+use nonempty::{nonempty, NonEmpty};
 use uuid::Uuid;
 
 use crate::entry::{Entry, EntryCommitted, EntryReduced, EntrySealed};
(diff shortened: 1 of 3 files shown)


Firstyear (commented):
Looks reasonable to me. Took me a second to realise the assertion is always true for groups to be non-empty since we always have them.

I think the main thing here is while it's true that this is non-empty, I think with the non-empty it's more for ... expressing state. Here it's not a strict thing, like yes it's a true side effect of the way UPG's are managed that a user will always have a group, but it's also still valid that groups can be 0 to infinite here.

So I think the intent of the non-empty is to allow us to change things from Vec to Option<NonEmpty> for example. Places where we have Vec.is_empty() and that asserts state. The point of the NonEmpty is a type level guarantee that we have 1 or more of something, and I'm not sure this is a really good/valid candidate of that representation change.

So a different way to say it is that we should think about where we target these NonEmpties to places where that state matters. AuthSession was a great example of this - for an auth session to proceed there HAS to be at least 1 credential. 

Does that help with how to look at these things and where to target them?

server/lib/src/idm/group.rs
```diff
@@ -70,30 +71,34 @@ macro_rules! try_from_account_e {
...
             }
         };
-        groups.push(upg);
-        Ok(groups)
+        if let Some(mut non_empty_groups) = NonEmpty::collect(groups.into_iter()) {
+            non_empty_groups.push(upg);
+            Ok(non_empty_groups)
+        } else {
+            Ok(nonempty!(upg))
+        }
```
yaleman, line 79: This seems a bit convoluted, since we always want `upg` to be a member, you could do:

```rust
non_empty_groups = NonEmpty{ head: upg, tail: groups }
```

yaleman (changes requested):

server/lib/src/idm/group.rs
yaleman: heh, closed while I was doing this. 🤷🏻

PR #1577, opened 2023-04-27
Cleanup incorrect log errors of denied entries

Firstyear:
Cleanup access log to prevent incorrect display that internal searches were denied access when empty sets were returned. 

- [ ] cargo fmt has been run


yaleman:
I would have thought that if entries was empty, we could just skip the whole filter bit below the "entries_is_empty" variable being set?

Firstyear:
> I would have thought that if entries was empty, we could just skip the whole filter bit below the "entries_is_empty" variable being set?

We still need to return the empty vec those and this process still allows it to build. We'd need a "return Ok(vec![])" effectively. I just did it like this as a way to avoid the "shortcut" since I want it to "always do the right thing"

yaleman (approved):

PR #1581, opened 2023-04-29
RPM packaging for EL8 and EL9

gvalkov:
Hello,

This PR brings fairly complete EL8 and EL9 (i.e. CentOS, Alma, Rocky) RPM packaging to Kanidm. Since those distributions don't package sufficiently recent Rust toolchains, the approach taken here is to first build the project inside a container and then use rpmbuild to package the results from the working tree (sort of like how it would be done if [FPM](https://github.com/jordansissel/fpm) was used). 

Key points:
- kanidm-server, kanidm-client and kanidm-unixd all include example configuration files.
- Switched to an install-time created user (i.e. `systemd-sysusers) instead of using `DynamicUsers` and `StateDirectory`. 
- Somewhat simpler spec file compared to the OpenSuse one. 

Things that can be improved:
- The dependency to podman/buildah can be dropped - `build-image.sh` can be just a Dockerfile.
- I haven't done much testing on the kanidm-unixd package. 
- Integrate package generation with CI.
- Write a `book/src/packaging_rpms.md`

### DIFF
--- a/book/src/server_configuration.md
+++ b/book/src/server_configuration.md
@@ -27,8 +27,12 @@ text=You MUST set the `domain` name correctly, aligned with your `origin`, else 
 You should test your configuration is valid before you proceed.
 
 ```bash
+# Docker:
 docker run --rm -i -t -v kanidmd:/data \
     kanidm/server:latest /sbin/kanidmd configtest -c /data/server.toml
+
+# OS package:
+sudo -u kanidm -- kanidmd configtest -c /etc/kanidm/server.toml
 ```
 
 ## Default Admin Account
@@ -47,9 +51,12 @@ text=The server must not be running at this point, as it requires exclusive acce
 <!-- deno-fmt-ignore-end -->
 
 ```bash
+# Docker:
 docker run --rm -i -t -v kanidmd:/data \
     kanidm/server:latest /sbin/kanidmd recover-account -c /data/server.toml admin
-# success - recovery of account password for admin: vv...
+
+# OS Package:
+sudo -u kanidm -- kanidmd recover-account -c /etc/kanidm/server.toml admin
 ```
 
 After the recovery is complete the server can be started again.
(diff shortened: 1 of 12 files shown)


book/src/server_configuration.md
```diff
@@ -27,8 +27,12 @@ text=You MUST set the `domain` name correctly, aligned with your `origin`, else
 You should test your configuration is valid before you proceed.
 
 ```bash
+# Docker:
```
yaleman, line 30: Please remove non-container references from the changes here, we only _support_ containerised servers. Folks can run whatever they like, but from a docs/support perspective, that's where we prefer things.

platform/el/build-image.sh
```diff
@@ -0,0 +1,59 @@
+#!/usr/bin/env bash
+# Should have probably been just a Dockerfile ...
```
yaleman, line 2: Given this comment, what would you change? We're already providing containerised versions of the server/client tooling. 

Do you mean a containerised build environment for the EL-specific packages? That'd  be preferred either way...

platform/el/run-image.sh
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env bash
```
yaleman, line 1: I'm not sure what this does - does it run a builder or the final packages etc? A more clear name and docstring is required, especially since `IMAGE_NAME` is required and not specified.

platform/systemd/kanidmd.service
```diff
@@ -1,20 +1,15 @@
...
-#   /usr/lib/systemd/system/kanidmd.service.d/custom.conf
-
 [Unit]
 Description=Kanidm Identity Server
 After=chronyd.service ntpd.service network-online.target
 Before=radiusd.service
 
 [Service]
-Type=simple
-DynamicUser=yes
```
yaleman, line 12: DynamicUser is required, this change is not acceptable.

Ref: https://unix.stackexchange.com/a/635036 / http://0pointer.net/blog/dynamic-users-with-systemd.html

yaleman (changes requested):
Hi, thanks for your contribution! Please attend to the requirements, and add yourself to the contributors file if you wish 😄

book/src/server_configuration.md
Firstyear: Agreed, the server packages are not an 'upstream' thing. We will look to support packages of the *client* components, especially pam and nss, but containers are the preference for the server.

platform/el/kanidm.spec
```diff
@@ -0,0 +1,186 @@
+Name:           kanidm
```
Firstyear, line 1: We don't carry spec files upstream - these are probably better in something like https://build.opensuse.org/project/show/network:idm where we actually would build and publish these.

platform/systemd/kanidmd.service
Firstyear: Agreed.

Firstyear (changes requested):
Mostly as it stands we probably won't accept this - our approach to the server is that you should run it in docker/podman and we already publish images for that. For the cilent tools packaging is acceptable for certain, but I think that's best done in places where we actually publish packages like https://build.opensuse.org/project/show/network:idm 

So a lot of these changes don't really "fit" upstream.

Firstyear:
In otherwords @gvalkov if you want to "fix/improve" el8/el9 builds I think that should be done here: https://build.opensuse.org/project/show/network:idm

Firstyear:
Closing as distro packaging doesn't belong upstream, we have el8/9 packages already available in OBS.

PR #1384, opened 2023-02-16
1355 docker builds

Firstyear:
Fixes #1355 - This improves docker builds with a cache for metadata and packages, and removes steps that aren't needed. There is a bug in zypper which when resolved will improve things again. 

Oh and this cleans up profiles with cpu arches and arch features  :) 

- [ x ] cargo fmt has been run
- [ ] cargo clippy has been run
- [ x ] cargo test has been run and passes
- [ ] book chapter included (if relevant)
- [ ] design document included (if relevant)

PR #1590, opened 2023-05-01
Bring back CpuOptLevel x86_64_v1

bjornfor:
Let's not force people to upgrade to x86_64_v2 yet. Default is unchanged
(x86_64_v2).

Fixes: 0d8d9e1a62652b2c ("1355 docker builds (#1384)") (EDIT: fixed reference)

Fixes #

- [x] cargo fmt has been run
- [ ] cargo clippy has been run -- fails, but must be unrelated to this change

### DIFF
--- a/libs/profiles/src/lib.rs
+++ b/libs/profiles/src/lib.rs
@@ -68,7 +68,7 @@ pub fn apply_profile() {
         CpuOptLevel::neon_v8 => {
             println!("cargo:rustc-env=RUSTFLAGS=-Ctarget-features=+neon,+fp-armv8")
         }
-        CpuOptLevel::x86_64_v1 => println!("cargo:rustc-env=RUSTFLAGS=-Ctarget-feature=+cmov,+cx8,+fxsr,+mmx,+sse,+sse2"),
+        CpuOptLevel::x86_64 => println!("cargo:rustc-env=RUSTFLAGS=-Ctarget-cpu=x86-64"),
         CpuOptLevel::x86_64_v2 => println!("cargo:rustc-env=RUSTFLAGS=-Ctarget-cpu=x86-64-v2"),
         CpuOptLevel::x86_64_v3 => println!("cargo:rustc-env=RUSTFLAGS=-Ctarget-cpu=x86-64-v3"),
     }
--- a/libs/profiles/src/lib.rs
+++ b/libs/profiles/src/lib.rs
@@ -9,6 +9,7 @@ enum CpuOptLevel {
     none,
     native,
     neon_v8,
+    x86_64_v1,
     x86_64_v2,
     x86_64_v3,
 }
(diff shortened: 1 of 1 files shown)


yaleman:
That's not a valid target in rust:
```
➜ rustc --print target-list | xargs -n1 -I {} rustc --target={} --print target-cpus | sort | uniq | grep x86
    x86-64
    x86-64-v2
    x86-64-v3
    x86-64-v4
```

If folks want to use baseline x86-64 CPUs from the dark ages they should probably be using `none`.

Ref: 

- https://rust-lang.github.io/packed_simd/perf-guide/target-feature/rustflags.html#target-cpu
- https://developers.redhat.com/blog/2021/01/05/building-red-hat-enterprise-linux-9-for-the-x86-64-v2-microarchitecture-level

bjornfor:
@yaleman: You're right. Sorry, I was a bit too quick, I even got the commit ref wrong (fixed now). I looked more closely at https://github.com/kanidm/kanidm/commit/0d8d9e1a62652b2ca008154cb90f8e99e2bd4389 now and copied the correct target-features string from it.

I'd have to read more about `none`. This patch was intended to keep the Nixpkgs build of kanidm the same, for the arch level we currently use.

bjornfor:
> I'd have to read more about `none`.

The question is whether `none` maps to "no optimization" or `-march=native`. The latter would be undesirable for Nixpkgs, whereas the former could work (but I would prefer to have *some* optimization).

yaleman:
> > I'd have to read more about `none`.
> 
> The question is whether `none` maps to "no optimization" or `-march=native`. The latter would be undesirable for Nixpkgs, whereas the former could work (but I would prefer to have _some_ optimization).

Looks like `none` sets no flags, so you should be good.

libs/profiles/src/lib.rs
```diff
@@ -66,6 +68,7 @@ pub fn apply_profile() {
         CpuOptLevel::neon_v8 => {
             println!("cargo:rustc-env=RUSTFLAGS=-Ctarget-features=+neon,+fp-armv8")
         }
+        CpuOptLevel::x86_64_v1 => println!("cargo:rustc-env=RUSTFLAGS=-Ctarget-feature=+cmov,+cx8,+fxsr,+mmx,+sse,+sse2"),
```
yaleman, line 71: This should just be `-Ctarget-cpu=x86-64`

libs/profiles/src/lib.rs
```diff
@@ -9,6 +9,7 @@ enum CpuOptLevel {
     none,
     native,
     neon_v8,
+    x86_64_v1,
```
yaleman, line 12: To match the rust target CPU patterns this should just be `x86_64`

yaleman (changes requested):
Shouldn't this just be

libs/profiles/src/lib.rs
Firstyear: I'd rather name it something like `x86_64_obsolete` or `x86_64_legacy` to indicate to users or deployers that this is probably not the right choice for them.

Firstyear (changes requested):

Firstyear:
Hey there, did you still want this? Would you be able to address the comments so it can be merged?

yaleman (approved):

yaleman:
Updated per @Firstyear's comments, should make it work if folks are digging up hardware out of the refuse station and want the specific optimisations.

bjornfor:
> @bjornfor why not use the "none" profile and it should work fine?

Yes, that's the way https://github.com/NixOS/nixpkgs/pull/229309 is heading, I think.

Firstyear (approved):

Firstyear:
If nixos is heading to none do we need this at all then?

yaleman:
it's in now, PRs to remove it welcome 🙄

PR #1592, opened 2023-05-02
Document that you need to tell Nginx or other proxies to use HTTP 1.1…

### LINKED ISSUE #1591: Document that you need to tell Nginx or other proxies to use HTTP 1.1 or later.

A problem I've seen twice now is that Nginx will send HTTP 1.0 requests by default. This will fail as tide will immediately drop the connection without logging it.

Well Tide could probably do something about logging dropped requests.

For kanidm I'd suggest a reverse proxy article in the docs.

Change upstream so it just treats 1.0 as http 1.1 (the differences aren't huge)

**Additional context**

My comment I made several years ago that mentions it. (Kinda)

https://github.com/kanidm/kanidm/issues/191#issuecomment-895955297

Support comment of a user running into the same issue.

https://matrix.to/#/!fzbFDgtoaOQXSSxOZc:gitter.im/$mHxO_cupAfdJVIv39S0nz6ny91ThNEJIS7rN8KlOdLU?via=gitter.im&via=matrix.org&via=tako.mx

yaleman:
… or later.

Fixes #1591

### DIFF
--- a/.github/workflows/kanidm_book.yml
+++ b/.github/workflows/kanidm_book.yml
@@ -3,6 +3,9 @@ name: GitHub Pages
 
 "on":
   push:
+    branches:
+      - "master"
+  pull_request:
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
@@ -67,3 +70,11 @@ jobs:
           ln -s ${{ needs.generate_tags.outputs.latest}}/ stable && \
           cd ..
           ls -la docs/
+
+      - name: Deploy
+        uses: peaceiris/actions-gh-pages@v3
+        if: ${{ github.ref == 'refs/heads/master' }}
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          publish_dir: ./docs
+          destination_dir: ./docs
--- a/book/src/troubleshooting.md
+++ b/book/src/troubleshooting.md
@@ -67,6 +67,20 @@ related to the failing operation.
 
 Try running commands with `RUST_LOG=debug` to get more information:
 
-```
+```shell
 RUST_LOG=debug kanidm login --name anonymous
+```
+
+## Reverse Proxies not sending HTTP/1.1 requests
+
+NGINX (and probably other proxies) send HTTP/1.0 requests by default. This'll lead to errors like this in your proxy logs:
+
+```text
+*17 upstream prematurely closed connection while reading response header from upstream, client: 172.19.0.1, server: example.com, request: "GET / HTTP/1.1", upstream: "https://172.19.0.3:8443/", host: "example.com:8443"
+```
+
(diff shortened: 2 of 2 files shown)


book/src/troubleshooting.md
```diff
@@ -67,6 +67,20 @@ related to the failing operation.
...
 Try running commands with `RUST_LOG=debug` to get more information:
 
-```
+```shell
 RUST_LOG=debug kanidm login --name anonymous
 ```
+
+## Reverse Proxies not sending HTTP/1.1 requests
+
+NGINX (and probably other proxies) send HTTP/1.0 requests by default. This'll lead to errors like this in your proxy logs:
```
Bert-Proesmans, line 76: A good addition would be "[..] requests, to the upstream server, by default.
This will/might improve search result quality for keywords "nginx kanidm upstream".

book/src/troubleshooting.md
```diff
@@ -67,6 +67,20 @@ related to the failing operation.
...
+
+```text
+*17 upstream prematurely closed connection while reading response header from upstream, client: 172.19.0.1, server: example.com, request: "GET / HTTP/1.1", upstream: "https://172.19.0.3:8443/", host: "example.com:8443"
+```
+
+The fix for NGINX is to set the [proxy_http_version](https://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_http_version) to `1.1`.
```
trissylegs, line 82: I'd add "This goes in the same block as the `proxy_pass` option."

.github/workflows/kanidm_book.yml
```diff
@@ -67,3 +70,11 @@ jobs:
           ln -s ${{ needs.generate_tags.outputs.latest}}/ stable && \
           cd ..
           ls -la docs/
+
+      - name: Deploy
```
Firstyear, line 74: What does this do?

book/src/troubleshooting.md
Firstyear, line 82: It may be better to divide this into sections like:

```
### Nginix

Change the value of ... 

### SomethingElse


```

That way it feels like nginix specific.

Firstyear (changes requested):

book/src/troubleshooting.md
yaleman: Good catch, have added the words in case search engines filter out the text below it for matching.

book/src/troubleshooting.md
yaleman: Funnily enough, I looked around and NGINX seems to be the *only* one that does this. Apache's config options seem to indicate that you have to *force* it to use 1.0 instead of the other way around.

.github/workflows/kanidm_book.yml
yaleman: actually pushes it to github page, somehow that didn't get re-added

PR #1595, opened 2023-05-04
more testing cleanup

yaleman:
Fixes issues where the containers try and get pushed on forks.


Firstyear:
@micolous should probably look at this :)

PR #1606, opened 2023-05-07
20230505 replication groundwork

Firstyear:
This is more work on replication - this cleans up some of the RUV maintenance code so that it's consistent after the initial content refresh. To achieve this some changes to memberof and dyngroups were needed. However now that the two RUV's on the replicas are in sync, we can now start to develop incremental replication! 

### DIFF
--- a/server/lib/src/entry.rs
+++ b/server/lib/src/entry.rs
@@ -576,7 +576,7 @@ impl Entry<EntryInit, EntryNew> {
     // event for the attribute.
     /// Add an attribute-value-assertion to this Entry.
     pub fn add_ava(&mut self, attr: &str, value: Value) {
-        self.add_ava_int(attr, value)
+        self.add_ava_int(attr, value);
     }
 
     /// Replace the existing content of an attribute set of this Entry, with a new set of Values.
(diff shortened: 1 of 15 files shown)


server/lib/src/entry.rs
```diff
@@ -2401,7 +2403,15 @@ where
             .eclog
             .add_ava_iter(&self.valid.cid, attr, std::iter::once(value.clone()));
         */
-        self.add_ava_int(attr, value)
+        self.add_ava_int(attr, value);
+    }
+
+    pub fn add_ava_if_not_exist(&mut self, attr: &str, value: Value) {
+        // This returns true if the value WAS changed!
```
yaleman, line 2410: can we add/move this return value docstring to the ada_ava_int function? 🤔 going by the implementation it returns false if it didn't exist before, but was added which is surprising

yaleman (approved):

yaleman:
Ruh roh, tests be failin'

server/lib/src/entry.rs
Firstyear: It's the same as the official rust btreeset/hashset, that's why it follows the same pattern. Yes we can move the doc string. :)

Firstyear:
> Ruh roh, tests be failin'

The failing line of code doesn't exist and is pointing at a test case that doesn't exist? What the fiuck?

Issue #1622, opened 2023-05-12
Identifiable Tokens


yaleman:
Kanidm tokens should have a unique pattern, making them easy to recognize. This is crucial for security systems that aim to prevent incorrect credential storage. Without a distinct pattern, like with bare JWTs that look like any base64-encoded data, we risk false alarms.

I've written a design document with more references in #1623

**Describe the solution you'd like**

A standard format for tokens issued from Kanidm.

```text
kanidm_<CREDENTIAL>
```
Where:

- `<CREDENTIAL>` is the actual credential.

PR #1623, opened 2023-05-12
Identifiable tokens

yaleman:
References #1622

- [ ] cargo fmt has been run

### DIFF
--- a/book/src/developers/designs/identifiable_secrets.md
+++ b/book/src/developers/designs/identifiable_secrets.md
@@ -1,0 +1,37 @@
+# Identifiable Secrets
+
+It should be possible to identify a token issued by Kanidm from its pattern, so that humans and systems can tell what they're dealing with.
+
+This is especially important with security systems that attempt to stop folks from saving credentials where they shouldn't be. Bare JWTs largely look like any other base64-encoded noise, so matching on them by default is prone to false-positives.
+
+## The Kanidm pattern
+
+> **RFC Discussion here**
+>
+> An alternative pattern could be just `kanidm_<CREDENTIAL>`, and folks can work it out themselves what they've got.
+
+```text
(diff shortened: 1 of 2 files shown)


book/src/developers/designs/identifiable_secrets.md
```diff
@@ -0,0 +1,34 @@
...
+Kanidm tokens should have a unique pattern, making them easy to recognize. This is crucial for security systems that aim to prevent incorrect credential storage. Without a distinct pattern, like with bare JWTs that look like any base64-encoded data, we risk false alarms.
+
+## The Kanidm pattern
+
+> **RFC Discussion here**
+>
+> Another option could be to use the pattern `kanidm_<CREDENTIAL>`. This way, users can figure out what kind of token they have on their own.
+
+```text
+kanidm_<TYPE>_<CREDENTIAL>
```
Firstyear, line 12: What benefit does the type serve here do you think? Does it make it easier for people to identify what the token belonged to?  Couldn't we just have a "reflector" command line tool that shows what's in the variety of tokens to help people ID what token it was? 

Otherwise  Ithink most of this work is needed in compact_jwt, because I think being able to generically have a way to prefix tokens there would be best for reuse in other applications too

yaleman: Yeah, I'm easy either way - if we can easily do the reflector thing and cover all the various tokens we generate - that'd work too 😄

Firstyear: Well the reflector is more so an admin can look at a token and say "what is it". 

So I think kanidm_<credential> would be sufficient for us? What do you think?

yaleman: Seems legit to me!

Firstyear (approved):

PR #1639, opened 2023-05-16
Fix use-reset-token command hint

cuberoot74088:
Running `kanidm person credential create-reset-token` prints out information on how to do the credential reset.

This fixes the provided cli command, the book, and struct comment .


yaleman (approved):
Good catch, thanks for the fixes!

PR #1648, opened 2023-05-22
Time travelling

### LINKED ISSUE #1640: WASM builds failing due to stdweb v0.4.20
### I did this

`cd server/web_ui && ./build_wasm.sh`

### I expected the following

It to build

Current commit, not sure which one broke it, but was recent.

### Any other comments

See https://github.com/rustwasm/wasm-bindgen/issues/3415 for details, looks like `stdweb` isn't maintained anymore, and was removed from `time` - which pulls it in - in 0.3.0.

So, we need to update ... something. I originally said `time` but that doesn't work. 😄

yaleman:
Fixes #1640

- [x] cargo fmt has been run
- [x] cargo clippy has been run (one thing that's unrelated to this change)


yaleman:
This obviously can't work until v0.2.0 of `scim` is released 😄

Firstyear:
Doing it now :)

Cargo.toml
```diff
@@ -134,9 +134,7 @@ tide = "^0.16.0"
 # Including brotli *very* slow, so don't do that. Including the "default" feature pulls a mime-type list from the internet on build, which isn't used.
 tide-compress = { version="0.10.6", default-features = false, features = [ "gzip", "regex-check" ] }
 tide-openssl = "^0.1.1"
-# Unable to increase version due to removing ability to detect
-# local platform time.
-time = "=0.2.27"
+time = { version = "^0.3.0", features = ["formatting", "local-offset"] }
```
Firstyear, line 137: Needs to be 0.3.19 else a hidden cfg param is needed

server/core/src/actors/v1_read.rs
```diff
@@ -199,8 +199,16 @@ impl QueryServerReadV1 {
         trace!(eventid = ?msg.eventid, "Begin online backup event");
 
         #[allow(deprecated)]
```
Firstyear, line 201: Should we remove the allow deprecated?

server/lib/src/valueset/session.rs
```diff
@@ -360,11 +378,16 @@ impl ValueSetT for ValueSetSession {
...
+                        #[allow(clippy::expect_used)]
+                        odt.format(&time::format_description::well_known::Rfc3339)
+                            .expect("Failed to format timestamp into RFC3339!")
                     }),
                     issued_at: {
                         debug_assert!(m.issued_at.offset() == time::UtcOffset::UTC);
-                        m.issued_at.format(time::Format::Rfc3339)
+                        #[allow(clippy::expect_used)]
+                        m.issued_at
+                            .format(&time::format_description::well_known::Rfc3339)
```
Firstyear, line 389: Should we just make Rfc3339 a "pub use" in lib.rs preludo to save using it like this so much?

Cargo.toml
yaleman: Nice of them to be incredibly vague about this in the changelogs :rage:

Done!

server/core/src/actors/v1_read.rs
yaleman: Done

server/lib/src/valueset/session.rs
yaleman: Done, with use statements elsewhere.

Firstyear (approved):

PR #1656, opened 2023-05-25
Add more replication tests, improve some handling of tombstones.

Firstyear:
More improvements to incremental replication, mainly around tombstone handling. Most of this change is more tests however. 

### DIFF
--- a/server/lib/src/plugins/mod.rs
+++ b/server/lib/src/plugins/mod.rs
@@ -345,7 +345,6 @@ impl Plugins {
         domain::Domain::post_repl_incremental(qs, pre_cand, cand)?;
         spn::Spn::post_repl_incremental(qs, pre_cand, cand)?;
         refint::ReferentialIntegrity::post_repl_incremental(qs, pre_cand, cand)?;
-        spn::Spn::post_repl_incremental(qs, pre_cand, cand)?;
         memberof::MemberOf::post_repl_incremental(qs, pre_cand, cand)
     }
 
--- a/server/lib/src/plugins/mod.rs
+++ b/server/lib/src/plugins/mod.rs
@@ -154,6 +154,7 @@ trait Plugin {
             "plugin {} has an unimplemented pre_repl_incremental!",
             Self::id()
         );
+        // debug_assert!(false);
         // Err(OperationError::InvalidState)
         Ok(())
     }
@@ -167,6 +168,7 @@ trait Plugin {
             "plugin {} has an unimplemented post_repl_incremental!",
             Self::id()
         );
+        // debug_assert!(false);
         // Err(OperationError::InvalidState)
         Ok(())
     }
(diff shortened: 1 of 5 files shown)


server/lib/src/repl/consumer.rs
```diff
@@ -237,14 +204,16 @@ impl<'a> QueryServerWriteTransaction<'a> {
...
         match ctx {
             ReplIncrementalContext::NoChangesAvailable => {
                 info!("no changes are available");
-                Ok(())
+                Ok(ConsumerState::Ok)
             }
             ReplIncrementalContext::RefreshRequired => {
-                todo!();
+                error!("Unable to proceed with consumer incremental - the supplier has indicated that our RUV is outdated, and replication would introduce data corruption.");
+                error!("This servers content must be refreshed to proceed. If you have nominated for automatic refresh, this will occur shortly.");
```
yaleman, line 215:
```suggestion
                error!("This server's content must be refreshed to proceed. If you have configured automatic refresh, this will occur shortly.");
```

server/lib/src/plugins/mod.rs
```diff
@@ -343,9 +342,11 @@ impl Plugins {
         pre_cand: &[Arc<EntrySealedCommitted>],
         cand: &[EntrySealedCommitted],
     ) -> Result<(), OperationError> {
-        refint::ReferentialIntegrity::post_repl_incremental(qs, pre_cand, cand)
-            .and_then(|_| spn::Spn::post_repl_incremental(qs, pre_cand, cand))
-            .and_then(|_| memberof::MemberOf::post_repl_incremental(qs, pre_cand, cand))
+        domain::Domain::post_repl_incremental(qs, pre_cand, cand)?;
+        spn::Spn::post_repl_incremental(qs, pre_cand, cand)?;
+        refint::ReferentialIntegrity::post_repl_incremental(qs, pre_cand, cand)?;
+        spn::Spn::post_repl_incremental(qs, pre_cand, cand)?;
```
yaleman, line 348: out of curiosity, why's the spn one get run twice?

yaleman (approved):
Minor nitpick on wording and a question, but otherwise, woo!

server/lib/src/plugins/mod.rs
Seba-T: I second this. And also, won't they all just throw "plugin {} has an unimplemented post_repl_incremental!" since they don't implement post_repls_incremental?

Firstyear: Oh yeah, why is that run twice? That's a mistake.

The reason we allow it to "pass" doing nothing for now, is that after I finish the object state machine tests, I'm going to flip these to err() states and then implement them and their plugin tests on replication because there are also a bunch of states to handle there too.

Seba-T (approved):

PR #1661, opened 2023-05-27
SELinux support for kanidm-unixd-tasks daemon

### LINKED ISSUE #1655: kanidm-unixd-tasks should relabel new user home dirs on SELinux systems

I'm working on integrating `kanidm-unixd` onto my infrastructure which is all SELinux-enabled, but I've hit a bit of a snag. Policy-writing aside, I notice that after `kanidm_unixd_tasks` creates a new user's home directory, there doesn't seem to be a way to also set the new home directory's SELinux user. On the system I am testing with, I see these contexts on the home directory after its created:
```
# ls -lZd /home/14f4a2c1-4541-4018-9e7d-9b81010ece6c/
drwxr-x---. 1 concord concord system_u:object_r:user_home_dir_t:s0 64 May 24 22:17 /home/14f4a2c1-4541-4018-9e7d-9b81010ece6c/
```
The problem here is that the home dir's contexts should match the SELinux user of the user which we are creating the home directory for, e.g. `staff_u` or `unconfined_u`:
```
# ls -lZd /home/14f4a2c1-4541-4018-9e7d-9b81010ece6c/
drwxr-x---. 1 concord concord staff_u:object_r:user_home_dir_t:s0 64 May 24 22:17 /home/14f4a2c1-4541-4018-9e7d-9b81010ece6c/
```

This is problematic especially on systems with UBAC enforced, as the user would not be able to access their home directory at all under these conditions.

Ideally in the future kanidm could gain full support for SELinux, but in the interim the ability to specify the path to a script file or command (with arguments) in the config file would suffice. Such an option may allow for more advanced configurations, although I cannot think of any that would fit my use case.

As far as my use case goes, a simple `restorecon -RF` on the newly created home directory by this config option would get the job done.

There might be a way to do this same effect in PAM (`pam_exec` maybe?).

**Additional context**

N/A

0xC0ncord:
Add initial SELinux support for `kanidm-unixd-tasks`. This allows the daemon to ensure that newly created home directories are labeled correctly according to what the SELinux policy says they should be. This is done by doing a lookup for the default context on each object to be created (home directory itself and the contents from `/etc/skel`), then ensuring that the newly created object will have that label.

Fixes #1655

I'm opening this as a draft for comments (I'm still new to Rust and this is my first contribution to a Rust project other than my own) and to discuss potential improvements, including style and idiomacy. For one I would like to instead have the SELinux functionality as an optional compile-time feature, but I wasn't able to get this to work for reasons I'm not yet sure of (please see https://github.com/0xC0ncord/kanidm/tree/unixd-tasks-selinux-optional).

- [x] cargo fmt has been run


unix_integration/src/constants.rs
```diff
@@ -13,3 +13,4 @@ pub const DEFAULT_HOME_ALIAS: Option<HomeAttr> = Some(HomeAttr::Spn);
 pub const DEFAULT_USE_ETC_SKEL: bool = false;
 pub const DEFAULT_UID_ATTR_MAP: UidAttr = UidAttr::Spn;
 pub const DEFAULT_GID_ATTR_MAP: UidAttr = UidAttr::Spn;
+pub const DEFAULT_SELINUX: bool = true;
```
yaleman, line 16: Probably should default to false for now, at least until we have a wider test base. @Firstyear ?

yaleman (changes requested):

unix_integration/src/tasks_daemon.rs
```diff
@@ -111,10 +133,34 @@ fn create_home_directory(
...
+    // Construct a path for SELinux context lookups
+    let sel_lookup_path_raw = match selinux {
+        true => match get_user_by_uid(info.gid) {
+            Some(v) => format!("{}{}", home_prefix, v.name().to_str().unwrap()),
+            None => {
+                return Err("Failed looking up username by uid for SELinux relabeling".to_string());
+            }
+        },
+        false => String::new(),
+    };
```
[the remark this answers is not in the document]
0xC0ncord: `hd_path_raw` may be the UUID of the user (default) or spn instead of their username. The SELinux policy will have contexts defined for `/home/<username>` which is why we ask the policy about the context for `/home/<username>`, otherwise the context will always be `user_u:object_r:user_home_dir_t` instead of `staff_u:object_r:user_home_dir_t` for example.

An alternative to this is to have the daemon create equivalence rules in the policy (so that e.g. `/home/<uuid>` is the same as `/home/<username>` for all contexts), but from a policy perspective that means allowing a process to modify the running policy which is generally a bad idea.

unix_integration/src/tasks_daemon.rs
```diff
@@ -135,6 +181,22 @@ fn create_home_directory(
...
+                            .path()
+                            .strip_prefix(skel_dir)
+                            .map_err(|e| e.to_string())?,
+                    ),
+                    false => Path::new("").to_path_buf(),
+                };
+                do_selinux_fscreatecon_for_path(
+                    &sel_lookup_path.to_str().unwrap().to_string(),
+                    &labeler,
+                )?;
```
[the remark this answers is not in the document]
0xC0ncord: You're right -- I missed wrapping this call in that check from an earlier revision.

unix_integration/src/constants.rs
0xC0ncord: I'm fine with leaving it disabled as the default, but users running SELinux will need to know to enable it during installation or there will be problems.

yaleman: Yep, that can and should be called out in the documentation.

unix_integration/src/tasks_daemon.rs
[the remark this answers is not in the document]
yaleman: Ok, that makes sense from the path perspective, thanks.

Please double check the gid/uid bit :)

unix_integration/src/constants.rs
Firstyear: I think better would be to actually try and detect at run time / start up if selinux is present on the system. We can probably use the selinux::boot_mode / selinux::current_mode calls to work this out. That way selinux support is enabled automatically when we are on a system that uses selinux.

Firstyear: Actually I think this is already the case given how the default config is constructed. So I think default to true here is okay since it will automatically disable when selinux is not present.

unix_integration/src/tasks_daemon.rs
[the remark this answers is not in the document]
Firstyear: I think the problem here is that if `/home/<spn>` has a different label to `/home/<username>` if we label `/home/<spn>` with the policy of `/home/<username>` while that will work, on the next restorecon or relable of the FS it will break. So wouldn't that imply an issue with the policy that is on the system?

Firstyear (commented):
Looks good, there are some comments and questions, but also please add yourself to the contributors.md file too :)

0xC0ncord:
> also please add yourself to the contributors.md file too :)

I shall! :)

> Actually I think this is already the case given how the default config is constructed. So I think default to true here is okay since it will automatically disable when selinux is not present.

That's right. The end result is that users who aren't using SELinux can completely ignore the setting in the config, as if it's enabled it will be disabled at runtime if SELinux isn't present anyway. My intent was so that only SELinux users should worry about the setting, and have the option to disable it if they really want (testing, etc.).

> So wouldn't that imply an issue with the policy that is on the system?

That's a good point. The underlying problem is that the SELinux policy has no knowledge of which user `/home/<spn>` belongs to, since it relies on the directory name being a username in order to know what SELinux user should "own" this home directory. Indeed without a policy change, relabeling the user's home directory would cause it to be relabeled to the wrong SELinux user.

I'll dig a bit deeper into this and see if there are other ways we could solve this, but the most robust solution seems to be to have `kanidm-unixd-tasks` add an equivalence rule to the policy. If anything, the policy can make this access conditional behind a boolean.

0xC0ncord:
Also, I just realized we will more than likely need to make SELinux support a compile-time feature, since an SELinux build will result in binaries linked against `libselinux.so` which won't exist on systems without SELinux userspace. I'm still not sure why https://github.com/0xC0ncord/kanidm/commit/47dcacbdd21332f67e80767cdd9446ae9679c368 doesn't seem to work, as flipping the SELinux feature on/off has no effect (it's always disabled).

Firstyear:
> > So wouldn't that imply an issue with the policy that is on the system?
> 
> That's a good point. The underlying problem is that the SELinux policy has no knowledge of which user `/home/<spn>` belongs to, since it relies on the directory name being a username in order to know what SELinux user should "own" this home directory. Indeed without a policy change, relabeling the user's home directory would cause it to be relabeled to the wrong SELinux user.

How does it check the username? Is it resolving it via nsswitch? Or is selinux just straight copying the text somehow?

On a system the spn, uuid and name will all resolve to the user via nss, so if selinux were to look this up it "should work". 

> 
> I'll dig a bit deeper into this and see if there are other ways we could solve this, but the most robust solution seems to be to have `kanidm-unixd-tasks` add an equivalence rule to the policy. If anything, the policy can make this access conditional behind a boolean.

If there isn't a simpler or more direct solution, I think this is the best way to go so that the labels survive subsequent restorecons. 

> Also, I just realized we will more than likely need to make SELinux support a compile-time feature, since an SELinux build will result in binaries linked against `libselinux.so` which won't exist on systems without SELinux userspace. I'm still not sure why [0xC0ncord@47dcacb](https://github.com/0xC0ncord/kanidm/commit/47dcacbdd21332f67e80767cdd9446ae9679c368) doesn't seem to work, as flipping the SELinux feature on/off has no effect (it's always disabled).

`#[cfg(all(target_family = "unix", selinux))]`, selinux isn't a cfg flag, it's a feature. So you need `#[cfg(all(target_family = "unix", feature = "selinux"))]` instead.

0xC0ncord:
> How does it check the username? Is it resolving it via nsswitch? Or is selinux just straight copying the text somehow?

For home directories, the policy looks up the SELinux user for the directory name. For example:
```
$ matchpathcon /home/concord
/home/concord	staff_u:object_r:user_home_dir_t:s0
```
Here, the SELinux user `staff_u` is derived from the SELinux logins configured on the system:
```
# semanage login -l

Login Name           SELinux User         MLS/MCS Range        Service

__default__          user_u               s0-s0                *
concord              staff_u              s0-s0                *
root                 root                 s0-s0                *
```

If my user was not present here (I added it manually when I installed this machine), the SELinux user would default to `user_u`. This default value is what is used for any user (or Unix group) not explicitly defined. In our case, a user's UUID or spn is not a "user" that appears in these logins, so the context falls back to the default of `user_u`.

SELinux DOES use nsswitch for looking up these users and groups for logins, but the policy does not know that `/home/<uuid>` or `/home/<spn>` is actually an alias for `/home/concord`.

Firstyear:
If you also add `<uuid>` and `<spn>` to the semanage user list as staff_u, does that change the results? Alternately, if you add *only* the uuid to the semanage user list how does that work? Part of the reason for all the spn/name to uuid maps is to allonw users to change their names at any time without breaking things, but I suspect if selinux policy is embedding this name into the policy rules that will likely be problematic.

0xC0ncord:
> If you also add `<uuid>` and `<spn>` to the semanage user list as staff_u, does that change the results?

That does work, but this means that the local admin will need to do this for each user or `kanidm-unixd-tasks` will need to do it (or add the aforementioned equivalence rule) for each new user.

> Alternately, if you add only the uuid to the semanage user list how does that work?

This works too.

> Part of the reason for all the spn/name to uuid maps is to allonw users to change their names at any time without breaking things, but I suspect if selinux policy is embedding this name into the policy rules that will likely be problematic.

Right -- I don't disagree with the logic behind that. It just means we will need to account for it somehow.

I think ultimately the best course of action would be to create the necessary equivalence rules in the policy automatically for each new user.

Firstyear:
> I think ultimately the best course of action would be to create the necessary equivalence rules in the policy automatically for each new user.

Yep, I agree with this.

0xC0ncord:
> #[cfg(all(target_family = "unix", selinux))], selinux isn't a cfg flag, it's a feature. So you need #[cfg(all(target_family = "unix", feature = "selinux"))] instead.

That fixed it! :)
I'll have SELinux as a compile-time feature in the next revision.

unix_integration/src/tasks_daemon.rs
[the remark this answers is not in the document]
0xC0ncord: > Please double check the gid/uid bit :)

Since Kanidm uses the GID number for the UID as well, this is doing a Unix user lookup by the same GID number. It admittedly looks confusing, so I can add a comment here to clarify.

0xC0ncord:
I've made the daemon create the needed equivalance rules when it creates home directories and made SELinux into a compile-time feature. Lastly I've updated the relevant docs to address the new `selinux` config option.

Unless there are any further outstanding comments I think this is ready. :)

examples/unixd
```diff
@@ -7,4 +7,5 @@
 # home_alias = "spn"
 # use_etc_skel = false
 # uid_attr_map = "spn"
-# gid_attr_map = "spn"
\ No newline at end of file
+# gid_attr_map = "spn"
+# selinux = true
```
Firstyear, line 11: Could be worth a comment here for admins who will read this and assume "oh selinux is always enabled when I'm not using it". when the comment should be "If selinux is detected, relabel created home directories. If you are not using selinux then no action is taken".
Alternately, just remove it from this example since the default is sane.

unix_integration/src/tasks_daemon.rs
```diff
@@ -135,16 +169,48 @@ fn create_home_directory(
...
                     fs::create_dir_all(dest).map_err(|e| e.to_string())?;
                 } else {
                     fs::copy(entry.path(), dest).map_err(|e| e.to_string())?;
                 }
                 chown(dest, info.gid)?;
+
+                // Create equivalence rule in the SELinux policy
+                #[cfg(all(target_family = "unix", feature = "selinux"))]
+                if Command::new("semanage")
+                    .args(["fcontext", "-ae", &sel_lookup_path_raw, &hd_path_raw])
```
Firstyear, line 198: I imagine that the reason for the command here is that the selinux crate doesn't support this?

Firstyear (changes requested):

Firstyear:
Also if not already done, add yourself to contributors.md in this PR please :)

examples/unixd
0xC0ncord: I think it's best to remove it so there's no confusion.

unix_integration/src/tasks_daemon.rs
0xC0ncord: Correct -- I didn't see any way to do it with the crate, and in any case this is better from a policy perspective because it allows writing a rule to allow `kanidm_unixd_tasks` to execute the `semanage` binary instead of having all of the privileges needed to do the policy changes itself.

0xC0ncord:
> Also if not already done, add yourself to contributors.md in this PR please :)

D'oh!

Firstyear (approved):

Firstyear:
@yaleman Waiting on your review :)

yaleman (approved):

yaleman:
All good here, I'm one of those "selinux? dissssaaaaable" jerks, so it looks fine to me 😄

PR #1704, opened 2023-06-05
Also check running SELinux mode during kanidm_unixd_tasks startup

0xC0ncord:
For `kanidm_unixd_tasks`, check the current SELinux mode in addition to kernel support. If SELinux is disabled at runtime, any attempts to query the policy will fail, so also disable SELinux features if this is the case.

- [x] cargo fmt has been run

### DIFF
--- a/unix_integration/src/selinux_util.rs
+++ b/unix_integration/src/selinux_util.rs
@@ -1,9 +1,19 @@
 use std::ffi::CString;
 
-use selinux::{kernel_support, label::back_end::File, label::Labeler, KernelSupport};
+use selinux::{
+    current_mode, kernel_support, label::back_end::File, label::Labeler, KernelSupport, SELinuxMode,
+};
 
 pub fn supported() -> bool {
-    return !matches!(kernel_support(), KernelSupport::Unsupported);
+    // check if the running kernel has SELinux support
+    if matches!(kernel_support(), KernelSupport::Unsupported) {
+        return false;
+    }
+    // check if SELinux is actually running
+    match current_mode() {
+        SELinuxMode::Permissive | SELinuxMode::Enforcing => true,
+        _ => false,
+    }
 }
 
 pub fn get_labeler() -> Result<Labeler<File>, String> {


unix_integration/src/selinux_util.rs
```diff
@@ -1,9 +1,12 @@
 use std::ffi::CString;
 
-use selinux::{kernel_support, label::back_end::File, label::Labeler, KernelSupport};
+use selinux::{
+    current_mode, kernel_support, label::back_end::File, label::Labeler, KernelSupport, SELinuxMode,
+};
 
 pub fn supported() -> bool {
-    return !matches!(kernel_support(), KernelSupport::Unsupported);
+    return !matches!(kernel_support(), KernelSupport::Unsupported)
```
Firstyear, line 8: As pedantic as it seems, it may be better to list the conditions we *do* need for support instead. Perhaps this should be `Supported && ( Permissive || Enforcing )` or similar so that the logic reads as "this is supported if kernel support is on and the mode is permissive or enforcing".

Firstyear (changes requested):

unix_integration/src/selinux_util.rs
0xC0ncord: Sure, though the `KernelSupport` enum can consist of `SELinux` and `SELinuxMLS` which may be confusing by itself I think. Maybe we should just expand the `SELinuxMode` check? If anything we can also add a comment to clarify what exactly we are checking here.

Firstyear: Yeah, the comments would help. Alternately we could do a match or something?

0xC0ncord: > Alternately we could do a match or something?

That's another option. I will experiment and see what I can come up with. :)

Firstyear: Thanks, I think it would be really good to be clear and explicit here is all :) just in case we expand the modes we support later or have to change it.

0xC0ncord: I think I'm happy with this. I would have liked to just use match arms to check `NotRunning`, `Permissive`, and `Enforcing`, but the `SELinuxMode` enum is `non_exhaustive` for some reason which means we must account for `_`.

Firstyear (approved):

Firstyear:
@0xC0ncord Thanks for your patience, I'm really happy with this too now. Appreciate your time as always!

PR #1705, opened 2023-06-05
Draft of a possible implementation of the identity verification system in kanidm 

Seba-T:
- [ ] cargo fmt has been run

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2459,6 +2459,7 @@ dependencies = [
  "kanidmd_lib",
  "ldap3_proto",
  "libc",
+ "once_cell",
  "openssl",
  "profiles",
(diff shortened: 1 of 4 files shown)


Firstyear (changes requested):
Very nice! Looks like it all works. So now the question is how to transform it to something we can use inside Kani. It's probably best to make something similar to Totp here, where we have a "IdentityVerifier" struct that we init from a private key / public key, and then it does the operations on the server. We can make that up and get it tested/committed before we proceed with wiring in the db storage of keys etc. 

So would that seem like the next best place to go?

Firstyear:
Did you want to close this @Seba-T ?

Seba-T:
this one def yes

PR #1727, opened 2023-06-12
Name change history

### LINKED ISSUE #345: hidden name history

Seba-T:
Fixes #345
It's a very draftish implementation of #345, it's missing most of the features, such as the ability to claim the name and the option to turn it off
- [ ] cargo fmt has been run

- [ ] cargo fmt has been run

### DIFF
--- a/server/lib/src/constants/schema.rs
+++ b/server/lib/src/constants/schema.rs
@@ -171,6 +171,41 @@ pub const JSON_SCHEMA_ATTR_LEGALNAME: &str = r#"{
       ]
     }
 }"#;
+
+pub const JSON_SCHEMA_ATTR_NAME_HISTORY: &str = r#"{
+    "attrs": {
+      "class": [
+        "object",
+        "system",
+        "attributetype"
+      ],
+      "description": [
+        "The history of names that a person has had"
+      ],
+      "index": [
+        "EQUALITY"
+      ],
+      "unique": [
+        "false"
+      ],
+      "multivalue": [
+        "true"
+      ],
+      "sync_allowed": [
+        "true"
+      ],
+      "attributename": [
+        "name_history"
+      ],
+      "syntax": [
+        "AUDIT_LOG_STRING"
+      ],
+      "uuid": [
+        "00000000-0000-0000-0000-ffff00000133"
+      ]
+    }
+}"#;
+
 pub const JSON_SCHEMA_ATTR_RADIUS_SECRET: &str = r#"{
     "attrs": {
       "class": [
@@ -1578,7 +1613,8 @@ pub const JSON_SCHEMA_CLASS_ACCOUNT: &str = r#"
         "oauth2_consent_scope_map",
         "user_auth_token_session",
         "oauth2_session",
-        "description"
+        "description",
+        "name_history"
       ],
       "systemmust": [
         "displayname",
--- a/server/lib/src/constants/uuids.rs
+++ b/server/lib/src/constants/uuids.rs
@@ -227,6 +227,7 @@ pub const UUID_SCHEMA_ATTR_PRIVATE_COOKIE_KEY: Uuid = uuid!("00000000-0000-0000-
 pub const _UUID_SCHEMA_ATTR_DOMAIN_LDAP_BASEDN: Uuid =
     uuid!("00000000-0000-0000-0000-ffff00000131");
 pub const UUID_SCHEMA_ATTR_DYNMEMBER: Uuid = uuid!("00000000-0000-0000-0000-ffff00000132");
+pub const UUID_SCHEMA_ATTR_NAME_HISTORY: Uuid = uuid!("00000000-0000-0000-0000-ffff00000133");
 
 // System and domain infos
 // I'd like to strongly criticise william of the past for making poor choices about these allocations.
--- a/server/lib/src/plugins/mod.rs
+++ b/server/lib/src/plugins/mod.rs
@@ -19,6 +19,7 @@ pub(crate) mod dyngroup;
 mod gidnumber;
 mod jwskeygen;
 mod memberof;
+mod namehistory;
 mod protected;
 mod refint;
 mod session;
@@ -207,6 +208,7 @@ impl Plugins {
             .and_then(|_| gidnumber::GidNumber::pre_create_transform(qs, cand, ce))
             .and_then(|_| domain::Domain::pre_create_transform(qs, cand, ce))
             .and_then(|_| spn::Spn::pre_create_transform(qs, cand, ce))
+            .and_then(|_| namehistory::NameHistory::pre_create_transform(qs, cand, ce))
             // Should always be last
             .and_then(|_| attrunique::AttrUnique::pre_create_transform(qs, cand, ce))
     }
@@ -245,6 +247,7 @@ impl Plugins {
             .and_then(|_| domain::Domain::pre_modify(qs, pre_cand, cand, me))
             .and_then(|_| spn::Spn::pre_modify(qs, pre_cand, cand, me))
             .and_then(|_| session::SessionConsistency::pre_modify(qs, pre_cand, cand, me))
+            .and_then(|_| namehistory::NameHistory::pre_modify(qs, pre_cand, cand, me))
             // attr unique should always be last
             .and_then(|_| attrunique::AttrUnique::pre_modify(qs, pre_cand, cand, me))
     }
@@ -276,6 +279,7 @@ impl Plugins {
             .and_then(|_| domain::Domain::pre_batch_modify(qs, pre_cand, cand, me))
             .and_then(|_| spn::Spn::pre_batch_modify(qs, pre_cand, cand, me))
             .and_then(|_| session::SessionConsistency::pre_batch_modify(qs, pre_cand, cand, me))
+            .and_then(|_| namehistory::NameHistory::pre_batch_modify(qs, pre_cand, cand, me))
             // attr unique should always be last
             .and_then(|_| attrunique::AttrUnique::pre_batch_modify(qs, pre_cand, cand, me))
     }
--- a/server/lib/src/plugins/namehistory.rs
+++ b/server/lib/src/plugins/namehistory.rs
@@ -1,0 +1,267 @@
+use std::sync::Arc;
+
+use kanidm_proto::v1::OperationError;
+
+use crate::entry::{EntryInvalidCommitted, EntrySealedCommitted};
+use crate::event::ModifyEvent;
+use crate::plugins::Plugin;
+use crate::prelude::*;
+use crate::prelude::{BatchModifyEvent, QueryServerWriteTransaction};
+use crate::repl::cid::Cid;
+use crate::value::PartialValue;
+
+pub struct NameHistory {}
+
+lazy_static! {
+    // it contains all the partialvalues used to match against an Entry's class,
+    // we just need a partialvalue to match in order to target the entry
+    static ref CLASSES_TO_UPDATE: [PartialValue; 1] = [PartialValue::new_iutf8("account")];
+    static ref HISTORY_ATTRIBUTES: [&'static str;1] = ["name"];
+}
+
+impl NameHistory {
+    fn is_entry_to_update<VALUE, STATE>(entry: &mut Entry<VALUE, STATE>) -> bool {
+        CLASSES_TO_UPDATE
+            .iter()
+            .any(|pv| entry.attribute_equality("class", pv))
+    }
+
+    fn get_ava_name(history_attr: &str) -> String {
+        format!("{}_history", history_attr)
+    }
+
+    fn handle_name_updates(
+        pre_cand: &[Arc<EntrySealedCommitted>],
+        cand: &mut Vec<EntryInvalidCommitted>,
+        cid: &Cid,
+    ) -> Result<(), OperationError> {
+        for (pre, post) in pre_cand.iter().zip(cand) {
+            // here we check if the current entry has at least one of the classes we intend to target
+            if Self::is_entry_to_update(post) {
+                for history_attr in HISTORY_ATTRIBUTES.iter() {
+                    let pre_name_option = pre.get_ava_single(history_attr);
+                    let post_name_option = post.get_ava_single(history_attr);
+                    if let (Some(pre_name), Some(post_name)) = (pre_name_option, post_name_option) {
+                        if pre_name != post_name {
+                            let ava_name = Self::get_ava_name(history_attr);
+                            //// WARNING!!! this match will have to be adjusted based on what kind of attribute
+                            //// we are matching on, for example for displayname we would have to use Value::utf8 instead!!
+                            // as of now we're interested just in the name so we use Iname
+                            match post_name {
+                                Value::Iname(n) => post.add_ava_if_not_exist(
+                                    &ava_name,
+                                    Value::AuditLogString(cid.clone(), n),
+                                ),
+                                _ => return Err(OperationError::InvalidValueState),
+                            }
+                        }
+                    }
+                }
+            }
+        }
+        Ok(())
+    }
+
+    fn handle_name_creation(
+        cands: &mut Vec<EntryInvalidNew>,
+        cid: &Cid,
+    ) -> Result<(), OperationError> {
+        for cand in cands.iter_mut() {
+            if Self::is_entry_to_update(cand) {
+                for history_attr in HISTORY_ATTRIBUTES.iter() {
+                    if let Some(name) = cand.get_ava_single(history_attr) {
+                        let ava_name = Self::get_ava_name(history_attr);
+                        match name {
+                            Value::Iname(n) => cand.add_ava_if_not_exist(
+                                &ava_name,
+                                Value::AuditLogString(cid.clone(), n),
+                            ),
+                            _ => return Err(OperationError::InvalidValueState),
+                        }
+                    }
+                }
+            }
+        }
+
+        Ok(())
+    }
+}
+
+impl Plugin for NameHistory {
+    fn id() -> &'static str {
+        "plugin_name_history"
+    }
+
+    fn pre_create_transform(
+        qs: &mut QueryServerWriteTransaction,
+        cand: &mut Vec<EntryInvalidNew>,
+        _ce: &CreateEvent,
+    ) -> Result<(), OperationError> {
+        Self::handle_name_creation(cand, qs.get_txn_cid())
+    }
+
+    fn pre_modify(
+        qs: &mut QueryServerWriteTransaction,
+        pre_cand: &[Arc<EntrySealedCommitted>],
+        cand: &mut Vec<EntryInvalidCommitted>,
+        _me: &ModifyEvent,
+    ) -> Result<(), OperationError> {
+        Self::handle_name_updates(pre_cand, cand, qs.get_txn_cid())
+    }
+
+    fn pre_batch_modify(
+        qs: &mut QueryServerWriteTransaction,
+        pre_cand: &[Arc<EntrySealedCommitted>],
+        cand: &mut Vec<EntryInvalidCommitted>,
+        _me: &BatchModifyEvent,
+    ) -> Result<(), OperationError> {
+        Self::handle_name_updates(pre_cand, cand, qs.get_txn_cid())
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use std::time::Duration;
+
+    use crate::entry::{Entry, EntryInit, EntryNew};
+    use crate::prelude::uuid;
+    use crate::repl::cid::Cid;
+    use crate::value::Value;
+
+    #[test]
+    fn name_purge_and_set() {
+        // Add another uuid to a type
+        let cid = Cid::new(
+            uuid!("d2b496bd-8493-47b7-8142-f568b5cf47ee"),
+            Duration::new(20, 2),
+        );
+        let ea = entry_init!(
+            ("class", Value::new_class("account")),
+            ("class", Value::new_class("posixaccount")),
+            ("name", Value::new_iname("old_name")),
+            (
+                "uuid",
+                Value::Uuid(uuid!("d2b496bd-8493-47b7-8142-f568b5cf47ee"))
+            ),
+            (
+                "name_history",
+                Value::new_audit_log_string((cid.clone(), "old_name".to_string())).unwrap()
+            ),
+            ("description", Value::new_utf8s("testperson")),
+            ("displayname", Value::new_utf8s("old name person"))
+        );
+        let preload = vec![ea];
+        run_modify_test!(
+            Ok(()),
+            preload,
+            filter!(f_eq("name", PartialValue::new_iname("old_name"))),
+            modlist!([
+                m_purge("name"),
+                m_pres("name", &Value::new_iname("new_name_1"))
+            ]),
+            None,
+            |_| {},
+            |qs: &mut QueryServerWriteTransaction| {
+                let e = qs
+                    .internal_search_uuid(uuid!("d2b496bd-8493-47b7-8142-f568b5cf47ee"))
+                    .expect("failed to get entry");
+                let c = e
+                    .get_ava_set("name_history")
+                    .expect("failed to get primary cred.");
+                dbg!(c.clone());
+                return assert!(
+                    c.contains(&PartialValue::new_utf8s("old_name"))
+                        && c.contains(&PartialValue::new_utf8s("new_name_1"))
+                );
+            }
+        );
+    }
+
+    #[test]
+    fn name_creation() {
+        // Add another uuid to a type
+        let ea = entry_init!(
+            ("class", Value::new_class("account")),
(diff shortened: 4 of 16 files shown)


Firstyear:
> Btw how the hell do I change an `EntrySealedNew`?

That's the trick - you don't.  The entries have a state machine in them to force you to go through a specific lifecycle. To change an EntrySealed you need to *invalidate* it and then reprocess it via schema validation and value syntax validation. 

So generally, if you are hitting a barrier like "how do I change entry sealed new" there is a 95% chance that something about your approach is in the wrong place because a lot of time and thought has been placed into the current code and the writeable phases. There is a 5% chance you encountered something we actually have to change and adapt.

Seba-T:
> > Btw how the hell do I change an `EntrySealedNew`?
> 
> 
> 
> That's the trick - you don't.  The entries have a state machine in them to force you to go through a specific lifecycle. To change an EntrySealed you need to *invalidate* it and then reprocess it via schema validation and value syntax validation. 
> 

Yeah I was messing around with that but I got stuck at ‘invalidate’ as I couldn’t figure out where to get the cid and then how to subsequently validate the thing again :p

Firstyear:
> Yeah I was messing around with that but I got stuck at ‘invalidate’ as I couldn’t figure out where to get the cid and then how to subsequently validate the thing again :p

You shouldn't nede to invalidate or revalidate that though. The plugins are layered at various points in the stack so if you have an entrysealednew you are using the wrong function hook. There are tooks for EntryInvalidNew/Commited that you should be using instead.

server/lib/src/constants/uuids.rs
```diff
@@ -118,7 +118,7 @@ pub const _UUID_SCHEMA_ATTR_PRIMARY_CREDENTIAL: Uuid =
 pub const _UUID_SCHEMA_CLASS_PERSON: Uuid = uuid!("00000000-0000-0000-0000-ffff00000044");
 pub const _UUID_SCHEMA_CLASS_GROUP: Uuid = uuid!("00000000-0000-0000-0000-ffff00000045");
 pub const _UUID_SCHEMA_CLASS_ACCOUNT: Uuid = uuid!("00000000-0000-0000-0000-ffff00000046");
-// GAP - 47
+pub const UUID_SCHEMA_ATTR_NAMEHISTORY: Uuid = uuid!("00000000-0000-0000-0000-ffff00000133");
```
Firstyear, line 121: Should be in numerical order in the file, so move this down to the bottom of the list IMO.

server/lib/src/plugins/mod.rs
```diff
@@ -23,6 +23,7 @@ mod protected;
 mod refint;
 mod session;
 mod spn;
+mod nameupdate;
```
Firstyear, line 26: Shouldn't this be "name history" not "name update"?

server/lib/src/plugins/nameupdate.rs
```diff
@@ -0,0 +1,176 @@
...
+        }
+        Ok(())
+    }
+}
+
+impl Plugin for NameUpdate {
+    fn id() -> &'static str {
+        "plugin_name_update"
+    }
+    fn pre_create(_qs: &mut QueryServerWriteTransaction, _cand: &[EntrySealedNew], _ce: &CreateEvent) -> Result<(), OperationError> {
```
Firstyear, line 118: I don't think you need a pre_create hook here.

server/lib/src/plugins/nameupdate.rs
```diff
@@ -0,0 +1,176 @@
...
+        }).collect();
+        return target_cand;
+    }
+    fn name_was_updated(modification: &ModifyList<ModifyValid>) -> bool {
+        modification.into_iter().fold(false, |acc, prec| acc || match prec {
+            Modify::Present(val, _) => val.eq("name"),
+            Modify::Assert(_, _) | Modify::Removed(_, _) | Modify::Purged(_) => false
+        })
+    }
+    fn retrieve_previous_name(_cand: &Vec<&mut EntryInvalidCommitted>) -> Option<value::Value> {
```
Firstyear, line 50: What is the intent of this function?

server/lib/src/plugins/nameupdate.rs
```diff
@@ -0,0 +1,176 @@
...
+        let previous_names = _cand.iter().find_map(|ava| {
+            ava.get_ava().get("name")
+        });
+        if let Some(names) = previous_names {
+            names.to_value_single()
+        } else {
+            None
+        }
+    }
```
Firstyear, line 59: formatting, spaces between fn and }.

Firstyear (changes requested):
I think it'll be easier to just focus on the name history first, and get that working before you touch legal name or displayname. So remove those IMO. I think you only need modify hooks too. 

You also shouldn't try to check if the name that was changed previously already is in the name history - you want to set it anyway via the add_ava_if_not_exist, because that way it touches in the replication changestate to ensure that the name change *and* the history are both updated during replication. Otherwise there is a possibility of losing a name history when two servers update.

server/lib/src/constants/uuids.rs
Firstyear: Can you move this?

server/lib/src/plugins/nameupdate.rs
Firstyear: formatting ?

server/lib/src/plugins/nameupdate.rs
```diff
@@ -0,0 +1,240 @@
...
+    ) -> Result<(), OperationError> {
+        Self::handle_name_updates(pre_cand, cand, me)
+    }
+    fn pre_batch_modify(
+        _qs: &mut QueryServerWriteTransaction,
+        pre_cand: &[Arc<EntrySealedCommitted>],
+        cand: &mut Vec<EntryInvalidCommitted>,
+        me: &BatchModifyEvent,
+    ) -> Result<(), OperationError> {
+        // Still needs work
```
Firstyear, line 149: What needs to happen here?

server/lib/src/plugins/nameupdate.rs
```diff
@@ -0,0 +1,240 @@
...
+            |_| {},
+            |qs: &mut QueryServerWriteTransaction| {
+                let e = qs
+                    .internal_search_uuid(uuid!("d2b496bd-8493-47b7-8142-f568b5cf47ee"))
+                    .expect("failed to get entry");
+                dbg!(e.get_ava());
+                let c = e
+                    .get_ava_set("namehistory")
+                    .expect("failed to get primary cred.");
+                return assert!(c.contains(&PartialValue::new_iname(&"old_name")));
```
Firstyear, line 199: Should also check that "old_old_name" still is present.

server/lib/src/plugins/nameupdate.rs
```diff
@@ -0,0 +1,240 @@
...
+            }
+        );
+    }
```
Firstyear, line 239: Should there be another test to check what happens if the user changes their name to the *same* value as it currently is?

Firstyear (changes requested):
Still review comments that needed to be addressed :)

server/lib/src/constants/schema.rs
```diff
@@ -171,6 +171,107 @@ pub const JSON_SCHEMA_ATTR_LEGALNAME: &str = r#"{
...
+        "name_history"
+      ],
+      "syntax": [
+        "UTF8STRING_INAME"
+      ],
+      "uuid": [
+        "00000000-0000-0000-0000-ffff00000133"
+      ]
+    }
+}"#;
```
Firstyear, line 207: space

server/lib/src/plugins/namehistory.rs
```diff
@@ -0,0 +1,344 @@
...
+}
+
+lazy_static! {
+    // the naming can be confusing, here it means: 'the set of the uuids of the entries we want to target'
+    static ref entries_uuids_to_update: HashSet<PartialValue> = {
+        let mut set = HashSet::new();
+        set.insert(PartialValue::new_iutf8("account"));
+        set
+    };
+}
```
Firstyear, line 75: space

server/lib/src/plugins/namehistory.rs
```diff
@@ -0,0 +1,344 @@
...
+                });
+                return if is_target_entry {
+                    Some((i, elm))
+                } else {
+                    None
+                };
+            })
+            .collect::<Vec<(usize, &mut Entry<VALID, STATE>)>>();
+        return target_cand;
+    }
```
Firstyear, line 98: space

server/lib/src/plugins/namehistory.rs
```diff
@@ -0,0 +1,344 @@
...
+        modification.into_iter().for_each(|elm| {
+            match elm {
+                Modify::Present(val, _) => add_name_if_valid(val.as_str(), &mut modified_names),
+                Modify::Removed(val, _) => add_name_if_valid(val.as_str(), &mut modified_names),
+                Modify::Purged(val) => add_name_if_valid(val.as_str(), &mut modified_names),
+                Modify::Assert(_, _) => {}
+            };
+        });
+        return modified_names;
+    }
```
Firstyear, line 116: space

server/lib/src/plugins/namehistory.rs
```diff
@@ -0,0 +1,344 @@
...
+    //this function is stupid and it doesn't make sense
+    fn retrieve_previous_name(pre_cand: &[Arc<EntrySealedCommitted>]) -> Option<value::Value> {
+        let previous_names = pre_cand.iter().find_map(|ava| ava.get_ava().get("name"));
+        if let Some(names) = previous_names {
+            names.to_value_single()
+        } else {
+            None
+        }
+    }
```
Firstyear, line 125: space

server/lib/src/plugins/namehistory.rs
```diff
@@ -0,0 +1,344 @@
...
+            for name_type in updated_names.iter().cloned() {
+                let current_name = pre_cand[*i].get_ava_single(&name_type.to_string());
+                if let Some(name) = current_name {
+                    let ava_name = format!("{}_history", name_type.to_string());
+                    cand.add_ava_if_not_exist(&ava_name, name);
+                }
+            }
+        }
+        Ok(())
+    }
```
Firstyear, line 149: space

server/lib/src/plugins/namehistory.rs
```diff
@@ -0,0 +1,344 @@
...
+}
+
+impl Plugin for NameHistory {
+    fn id() -> &'static str {
+        "plugin_name_update"
+    }
```
Firstyear, line 188: space

server/lib/src/plugins/namehistory.rs
```diff
@@ -0,0 +1,344 @@
...
+    fn pre_modify(
+        _qs: &mut QueryServerWriteTransaction,
+        pre_cand: &[Arc<EntrySealedCommitted>],
+        cand: &mut Vec<EntryInvalidCommitted>,
+        me: &ModifyEvent,
+    ) -> Result<(), OperationError> {
+        Self::handle_name_updates(pre_cand, cand, me)
+    }
```
Firstyear, line 196: space

Firstyear (changes requested):
Please fix the spacing

```
// BAD
fn a() {
}
fn b() {
}
```

```
// GOOD
fn a() {
}

fn b() {
}
```

server/lib/src/constants/schema.rs
Seba-T: fixed!

server/lib/src/plugins/namehistory.rs
Seba-T: fixed!

server/lib/src/plugins/namehistory.rs
Seba-T: fixed :)

server/lib/src/plugins/namehistory.rs
Seba-T: fixed

server/lib/src/plugins/namehistory.rs
Seba-T: fixed

server/lib/src/plugins/namehistory.rs
Seba-T: fixed

server/lib/src/plugins/namehistory.rs
Seba-T: fixed

server/lib/src/plugins/namehistory.rs
Seba-T: fixed

Seba-T:
I think I got them all :p

server/lib/src/plugins/namehistory.rs
```diff
@@ -0,0 +1,351 @@
...
+                Modify::Present(val, _) => add_name_if_valid(val.as_str(), &mut modified_names),
+                Modify::Removed(val, _) => add_name_if_valid(val.as_str(), &mut modified_names),
+                Modify::Purged(val) => add_name_if_valid(val.as_str(), &mut modified_names),
+                Modify::Assert(_, _) => {}
+            };
+        });
+        return modified_names;
+    }
+
+    //this function is stupid and it doesn't make sense
```
Firstyear, line 120: Why do you think it's stupid?

server/lib/src/plugins/namehistory.rs
```diff
@@ -0,0 +1,351 @@
...
+        }
+        let mut cand_as_hashmap = entries_to_update
+            .into_iter()
+            .filter_map(|entry| match entry.1.get_uuid() {
+                None => None,
+                Some(uuid) => Some((uuid, entry)),
+            })
+            .collect::<HashMap<Uuid, (usize, &mut EntryInvalidCommitted)>>();
+        let modset = me.modset.clone();
+        for (uuid, modlist) in modset {
```
Firstyear, line 172: I don't think you need to clone here.

server/lib/src/plugins/namehistory.rs
```diff
@@ -0,0 +1,351 @@
...
+                } else {
+                    None
+                };
+            })
+            .collect::<Vec<(usize, &mut Entry<VALID, STATE>)>>();
+        return target_cand;
+    }
+
+    fn updated_names(modification: &ModifyList<ModifyValid>) -> HashSet<name_type> {
+        let mut modified_names: HashSet<name_type> = HashSet::new();
```
Firstyear, line 102: There may be an easier way to do this that avoids allocating a hashset.

Also worth noting that while hashsets are cool and all, they are *expensive* to allocate and free. It's better to simply have a btreeset for small sets. 

BUT - IMO there is likely a way to achieve this without needing a hashset at all :)

Firstyear (changes requested):
okay this is getting there! 

Why not have the "for (i, cand) in entry" be something like:

```
for (pre, post) in pre_cand.zip(post) {
    for history_attr in ["name", "legalname", "displayname"] {
        if pre.get_ava(history_attr) != post.get_ava(history_attr) {
             post.add_ava_if_not_exist( <state of pre attr> )
        }
    }
}
```

This saves a bunch of allocations, you don't have to check the modlists, and you can re-use it between mod/mod batch because they both take a pre/post set.

server/lib/src/plugins/namehistory.rs
Seba-T: because I think it returns the first name it finds, whereas there could be multiple previous names if we had multiple updates

server/lib/src/plugins/namehistory.rs
Seba-T: I nearly rewrote everything following what you wrote above so the hashset just disappeared :p

server/lib/src/server/mod.rs
```diff
@@ -149,6 +149,8 @@ pub trait QueryServerTransaction<'a> {
 
     fn get_resolve_filter_cache(&mut self) -> &mut ResolveFilterCacheReadTxn<'a>;
 
+    fn get_txn_cid(&self) -> &Cid;
```
Firstyear, line 152: This shouldn't be part of this trait. Should be removed.

server/lib/src/server/mod.rs
```diff
@@ -889,6 +893,10 @@ impl<'a> QueryServerTransaction<'a> for QueryServerReadTransaction<'a> {
     fn get_domain_display_name(&self) -> &str {
         &self.d_info.d_display
     }
+
+    fn get_txn_cid(&self) -> &Cid {
```
Firstyear, line 897: No todos in production - plus this isn't needed on the *read* transaction anyway.

server/lib/src/server/mod.rs
```diff
@@ -1003,6 +1011,10 @@ impl<'a> QueryServerTransaction<'a> for QueryServerWriteTransaction<'a> {
...
```
Firstyear, line 1015: Should be in impl queryServerWrite NOT in impl QueryServerTransaction for QueryServerWrite.

server/lib/src/valueset/auditlogstring.rs
```diff
@@ -0,0 +1,162 @@
...
+use crate::repl::cid::Cid;
+use crate::repl::proto::ReplAttrV1;
+use crate::schema::SchemaAttribute;
+use crate::valueset::{DbValueSetV2, ValueSet};
+
+type AuditLogStringType = (Cid, String);
+
+#[derive(Debug, Clone)]
+pub struct ValueSetAuditLogString {
+    set: SmolSet<[AuditLogStringType; 1]>,
```
Firstyear, line 15: Read up on how SmolSet works :)

server/lib/src/valueset/auditlogstring.rs
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+    pub fn new(s: AuditLogStringType) -> Box<Self> {
+        let mut set = SmolSet::new();
+        set.insert(s);
+        Box::new(ValueSetAuditLogString { set })
+    }
+
+    pub fn push(&mut self, s: AuditLogStringType) -> bool {
+        if self.set.len() >= 8 {
```
Firstyear, line 33: What about in insert checked?

Firstyear (changes requested):

server/lib/src/plugins/namehistory.rs
```diff
@@ -0,0 +1,223 @@
...
+use crate::repl::cid::Cid;
+use crate::server::QueryServerTransaction;
+use crate::value::PartialValue;
+
+pub struct NameHistory {}
+
+lazy_static! {
+    // it contains all the partialvalues used to match against an Entry's class,
+    // we just need a partialvalue to match in order to target the entry
+    static ref CLASSES_TO_UPDATE: HashSet<PartialValue> = {
```
Firstyear, line 20: Why does this need a hashset?

server/lib/src/plugins/namehistory.rs
```diff
@@ -0,0 +1,223 @@
...
+
+#[cfg(test)]
+mod tests {
+    use std::time::Duration;
+
+    use crate::entry::{Entry, EntryInit, EntryNew};
+    use crate::prelude::uuid;
+    use crate::repl::cid::Cid;
+    use crate::value::Value;
+
```
Firstyear, line 98: I think the behaviour we want is that we want the name history to show the time when a name was *added* not the time when it was *removed* from the account. If you do it as removed, you'll hit replication issues.

So how would you re-model this part of the code so that when you create the entry, the cid of the create and name is added? When the modify shows a new name, add the new name to the history rather than the old one?

Firstyear (changes requested):

server/lib/src/valueset/auditlogstring.rs
Seba-T: wopsies :p

Seba-T:
I think I fixed everything :)

server/lib/src/plugins/namehistory.rs
```diff
@@ -0,0 +1,268 @@
...
+    ) -> Result<(), OperationError> {
+        for (pre, post) in pre_cand.iter().zip(cand) {
+            // here we check if the current entry has at least one of the classes we intend to target
+            // maybe using a filter would be more appropriate?
+            if Self::is_entry_to_update(post) {
+                for history_attr in HISTORY_ATTRIBUTES.iter() {
+                    let pre_name_option = pre.get_ava_single(history_attr);
+                    let post_name_option = post.get_ava_single(history_attr);
+                    if let (Some(pre_name), Some(post_name)) = (pre_name_option, post_name_option) {
+                        let ava_name = Self::get_ava_name(history_attr);
```
Firstyear, line 46: move this into the if condition one line down, since it's only needed in that part.

Firstyear (changes requested):
One *super* minor comment, otherwise good to go!

server/lib/src/valueset/auditlogstring.rs
```diff
@@ -0,0 +1,164 @@
...
+
+    fn generate_idx_eq_keys(&self) -> Vec<String> {
+        self.set.iter().map(|(d, s)| format!("{d}-{s}")).collect()
+    }
+
+    fn syntax(&self) -> SyntaxType {
+        SyntaxType::AuditLogString
+    }
+
+    fn validate(&self, _schema_attr: &SchemaAttribute) -> bool {
```
Firstyear, line 112: Add a check for len > 8 here

server/lib/src/valueset/auditlogstring.rs
```diff
@@ -0,0 +1,164 @@
...
+        let set = iter.into_iter().collect();
+        Some(Box::new(ValueSetAuditLogString { set }))
+    }
+}
+
+impl ValueSetT for ValueSetAuditLogString {
+    fn insert_checked(&mut self, value: Value) -> Result<bool, OperationError> {
+        match value {
+            Value::AuditLogString(c, s) => {
+                if self.set.len() >= 8 {
```
Firstyear, line 62: should be > 8, not >=.

Firstyear (changes requested):

server/lib/src/valueset/auditlogstring.rs
Seba-T: wait but wouldn't this allow to have 9 elements in the set? Suppose the current size is 9, using your condition we would remove the 9th element to then add a new 9th right back, bringing the size back to 9. Whereas if we have 8 elements, we remove the 8th and we add a new one, the max size will still be 8

server/lib/src/valueset/auditlogstring.rs
Seba-T: done!

server/lib/src/valueset/auditlogstring.rs
Seba-T: Just tried with > instead of >= and it's not passing the history full test 
`thread 'plugins::namehistory::tests::name_purge_and_set_with_filled_history' panicked at 'assertion failed: !c.contains(&PartialValue::new_utf8s(&\"old_name1\")) &&\n    c.contains(&PartialValue::new_utf8s(&\"new_name\"))', server/lib/src/plugins/namehistory.rs:261:24`

server/lib/src/plugins/namehistory.rs
```diff
@@ -0,0 +1,268 @@
...
+        format!("{}_history", history_attr)
+    }
+
+    fn handle_name_updates(
+        pre_cand: &[Arc<EntrySealedCommitted>],
+        cand: &mut Vec<EntryInvalidCommitted>,
+        cid: &Cid,
+    ) -> Result<(), OperationError> {
+        for (pre, post) in pre_cand.iter().zip(cand) {
+            // here we check if the current entry has at least one of the classes we intend to target
```
Firstyear, line 39: Nah, this is the right way to do it :) feel free to remove this comment.

Firstyear (approved):

server/lib/src/plugins/namehistory.rs
Seba-T: Sure!

Firstyear: Also resolve these github conversations once youve actioned them, it makes it easier to see what you've done/havent :)

Seba-T: Wow I never noticed the resolve button :p

PR #1736, opened 2023-06-15
Add support for argon2id

### LINKED ISSUE #1734: LDAP password import failing (argon2)
### I did this

Trying to use the LDAP importer.

```
ERROR    │     ┝━ 🚨 [error]: password_import was unable to convert hash format | hint: {ARG | entry_id: mmmmmmm
ERROR    │     ┝━ 🚨 [error]: Pre-Modify operation failed (plugin), Plugin(CredImport("password_import was unable to convert hash format")) | event_tag_id: 1
ERROR    │     ┕━ 🚨 [error]: Unable to apply modifications to sync entries.
ERROR    ┕━ 🚨 [error]: Internal error -> Response sent | event_tag_id: 4 | status: 500 - Internal Server Error
```

The hash uses argon2id.

### I expected the following

Successful import of users and groups, reusing the given hashes, since they are compatible with `crypt()`.

[`780c82f` (#1733)](https://github.com/kanidm/kanidm/pull/1733/commits/780c82f8ec1b40919aa16aec8524af070fb89d47)

Linux sso2 5.15.113 #1-NixOS SMP Wed May 24 16:36:55 UTC 2023 x86_64 GNU/Linux

Firstyear:
Fixes #1734 - This adds support for argon2id. It also allows auto-tuning it's parameters based on system performance. I am still unsure yet if to change to it by default, as I want to consult some members of the SUSE product security team first. 

### DIFF
--- a/libs/crypto/src/lib.rs
+++ b/libs/crypto/src/lib.rs
@@ -46,6 +46,7 @@ const DS_SSHA512_HASH_LEN: usize = 64;
 
 const ARGON2_SALT_LEN: usize = 24;
 const ARGON2_KEY_LEN: usize = 32;
+const ARGON2_MAX_RAM_KIB: u32 = 32 * 1024;
 
 #[derive(Serialize, Deserialize)]
 #[allow(non_camel_case_types)]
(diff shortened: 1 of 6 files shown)


Firstyear:
cc @mweinelt

Firstyear:
@mweinelt I'm assuming this worked for you?

libs/crypto/src/lib.rs
```diff
@@ -84,31 +116,37 @@ impl fmt::Debug for DbPasswordV1 {
...
         CryptoPolicy {
             pbkdf2_cost: PBKDF2_MIN_NIST_COST,
+            argon2id_params: Params::default(),
         }
     }
 
-    pub fn time_target(t: Duration) -> Self {
-        let r = match Password::bench_pbkdf2(PBKDF2_MIN_NIST_COST * 10) {
+    pub fn time_target(target_time: Duration) -> Self {
+        const PBKDF2_BENCH_FACTOR: usize = 10;
```
yaleman, line 133: Wouldn't this const be better up with the other things? 😄

yaleman (approved):
More algos moar better.

libs/crypto/src/lib.rs
Firstyear: It could be, but it's also only ever used in this one function so I think tho locality is nicer

mweinelt:
> @mweinelt I'm assuming this worked for you? 

Yes, it did. Thanks again

PR #1742, opened 2023-06-16
Fix debian packaging

leoleoasd:
Fixes #

- [ ] cargo fmt has been run

1. Add prerm according to [document](https://wiki.ubuntu.com/PAMConfigFrameworkSpec):
> Packages implementing PAM profiles under this spec should call pam-auth-update with the --package argument in their postinst script. In the prerm script, these packages should call pam-auth-update --package --remove only when the prerm has been called with remove as its first argument; e.g.:
> ```
> if [ "$1" = remove ]; then
>         pam-auth-update --package --remove unix
> fi
> ```
2. Install pam config manually. I had a misunderstanding on `dh_installpam`. It will install pam configs for the application (such as `/etc/pam.d/sudo`) rather than the system (`/usr/share/pam-configs`).

Decision to be made: Should we automatically enable pam configs during installation? According to the document:
> Packages implementing PAM profiles under this spec should call pam-auth-update with the --package argument in their postinst script.
Will we do anything bad for the system? pam config with a stopped `kanidm-unixd` (since during installation we enable it but there are no working configuration) wouldn't block anything right?


yaleman:
I always feel a little icky auto-enabling a PAM module like that on install 🤔

Firstyear:
@micolous You know debian policy better than anyone - Is it acceptable to add a pam/nss module on install? my immediate thinking is "no" from an opensuse background.

leoleoasd:
It seems that `libpam-sss` (https://debian.pkgs.org/11/debian-main-amd64/sssd-common_2.4.1-2_amd64.deb.html) will automatically enable it's pam in their postinst:

image


`pam-auth-update --package` will enable the config if it specifies `Default: yes`:
image

micolous:
> @micolous You know debian policy better than anyone - Is it acceptable to add a pam/nss module on install? my immediate thinking is "no" from an opensuse background.

Most Debian packages attempt to build a minimal working config with `dpkg` questions, and start daemons by default on installation.

It looks like the `pam-auth-update` tool and `sssd-common` configs were imported from Ubuntu: https://salsa.debian.org/sssd-team/sssd/-/commit/89e72a82a14d63cdde9ed3a9b87515935962f518

There appears to be an expectation for PAM modules to register themselves _by default_: 

* https://bugs.launchpad.net/ubuntu/+source/pam-pkcs11/+bug/1650366
* https://bugs.launchpad.net/ubuntu/+source/sssd/+bug/557398 
* https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=568577

> Will we do anything bad for the system? pam config with a stopped kanidm-unixd (since during installation we enable it but there are no working configuration) wouldn't block anything right?

That's one of the questions to answer. It looks like this is all optional, so kanidm not running shouldn't block someone from logging in, or changing their password... and the other thing to answer is to make sure it doesn't allow authenticating with an invalid credential.

Firstyear:
Provided that other packages have a precedent of doing this, I'm okay with it then :)

Firstyear:
@yaleman Want to doubl check this?

platform/debian/kanidm-unixd/postinst
```diff
@@ -8,13 +8,13 @@ set -e
 
 case "$1" in
     configure)
+        pan-auth-update --package
```
yaleman, line 11:
```suggestion
        pam-auth-update --package
```

yaleman (changes requested):
Command typo, but otherwise let's see how this goes!

yaleman (approved):

PR #1756, opened 2023-06-20
Mention client configuration in tool installation guide

mre:
While going through the documentation and testing out the kanidm client, I noticed that the commands to test the client didn't work out of the box. It requires a config to connect to the server, which is thankfully already documented [here](https://github.com/kanidm/kanidm/blob/master/examples/config), but I thought it would still be handy to show a minimal configuration so that users don't have to move away from the documentation during that step.

- [ ] cargo fmt has been run

### DIFF
--- a/book/src/installing_client_tools.md
+++ b/book/src/installing_client_tools.md
@@ -126,6 +126,19 @@ If you have a ca.pem you may need to bind mount this in as required as well.
 alias kanidm="docker run ..."
 ```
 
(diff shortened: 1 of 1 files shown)


Firstyear (approved):

yaleman:
This'll be in the repo, but until I (or someone smarter than me, which isn't hard) fixes the docs-build, the docs won't update. Thanks! ❤️

PR #1764, opened 2023-06-22
Ux improvements - Allow enrolling other devices

### LINKED ISSUE #164: Multi device enrollment workflow

### LINKED ISSUE #1763: Kanidm client thinks CA Cert is not 0644 when it is
### I did this
I installed Kanidm on my nixos system with:
```nix
services.kanidm = {
    enableClient = true;
    clientSettings.uri = "https://idm.example.com";
    clientSettings.enablePam = true;
    clientSettings.verify_ca = true;
    clientSettings.verify_hostnames = true;
    # puts the CA cert in /run/secrets/caCert (yes I know this doesn't need to be a secret please don't judge me)
    clientSettings.ca_path = config.sops.secrets.caCert.path;
};
```
I then ran `kanidm system domain show` and got the following warning:
```
$ kanidm system domain show
2023-06-21T23:33:39.539053Z  WARN kanidm_client: permissions on /run/secrets/caCert may not be secure. Should be set to 0644. This could be a security risk ...
2023-06-21T23:33:39.568662Z ERROR kanidm_cli::common: No valid authentication tokens found. Please login with the 'login' subcommand.
```
(error was expected, I hadn't set up the server yet)
but my file *was* set to 644:
```
$ ls -l /run/secrets/caCert
-rw-r--r-- 1 root root 671 Jun 21 19:33 /run/secrets/caCert
```
### I expected the following
Not to get the warning

kanidm 1.1.0-alpha.12

`Linux nixos 6.3.5 #1-NixOS SMP PREEMPT_DYNAMIC Tue May 30 13:17:29 UTC 2023 x86_64 GNU/Linux`

### Any other comments
I think I tracked down the issue, forgive me for not using the proper rust tools, I haven't taken the time to put together a serious rust workflow/dev environement, and I'm not very familiar with the language:
Looking through github for a similar warning, it appears to come from [this line](https://github.com/kanidm/kanidm/blob/1.1.0-alpha.12/libs/client/src/lib.rs#L160). Searching for the function in GDB, I found  `_ZN13kanidm_client19KanidmClientBuilder17parse_certificate17hefc40d75aacc759eE.llvm.990996743465532897`, which is probably the best I'll get. I converted 0o644 to hex (0x1a4) and stepped through (stepping out of other functions) the assembly until reaching a matching comparison:
```gdb
0x555555b5b62f <...parse_certificate...+1087>   cmp    $0x1a4,%ebx
```
Running `info registers ebx` returns 0x81a4, which is not 0x01a4. This makes me believe that  `ca_meta.mode()` is returning 0x81a4 and failing the aforementioned check on [line 159](https://github.com/kanidm/kanidm/blob/1.1.0-alpha.12/libs/client/src/lib.rs#L159). This can be confirmed with `stat`:
```
$ stat -t /run/secrets/caCert
/run/secrets/caCert 671 8 81a4 0 0 1b 135993 1 0 0 1687391445 1687390410 1687391436 0 4096
```
What I think is going on, is that Linux's `mode_t` type actually contains a fair bit (pardon the pun) more than just the 3-4 permission octals. The 0x81a4 from `stat` and `gdb` translates to 0o0100644, according to [the inode(7) manpage](https://www.man7.org/linux/man-pages/man7/inode.7.html), that 0b1000 hanging out at bits 15:12 is telling us that it's a regular file, and it's not a socket or a symlink. I think the least intrusive fix would be to just mask the permission bits with `ca_meta.mode() & 0o007777 != 0o0644`. However, I think we could do better by considering less permissive modes secure as well with `ca_meta.mode() & 0o007133`, which is truthy if any of setuid, setgid, sticky, any execute, or group/user write are true. That way permissions like 0o0444 or 0o0440 would still be considered secure.
And hopefully this is reproducible because if it isn't then I'll look very silly with all these wrong explanations for what's happening :stuck_out_tongue:

Firstyear:
Fixes #164 Fixes #1763 This fixes client cert checking and adds support for self-generating credential update qr codes.


yaleman (approved):
Woot 😄

Firstyear:
I'm going to tweak one thig in how a variable is passed in wasm :)

PR #1768, opened 2023-06-25
Started chasing noise, found some code to delete...

yaleman:
Fixes ... stuff.

 - added a generic function in the webui for making web calls
 - tomllint things
 - fixed a content-security-policy header thing for bootstrap inline images
 - moved the webui static content into one place so the build script isn't chasing files
 - added a script to set up a dev environment with a bunch of "default" things

Checklist:
- [x] cargo fmt has been run


server/web_ui/src/lib.rs
```diff
@@ -35,3 +39,52 @@ pub fn run_app() -> Result<(), JsValue> {
...
+/// Build and send a request to the backend, with some standard headers and pull back
+/// (kopid, status, json, headers)
+pub async fn do_request(
+    uri: &str,
+    method: RequestMethod,
+    body: Option<JsValue>,
+) -> Result<(Option<String>, u16, JsValue, Headers), FetchError> {
+    let mut opts = RequestInit::new();
+    opts.method(&method.to_string());
+    opts.mode(RequestMode::SameOrigin);
```
Firstyear, line 70: I think we want opts.credentials(SameOrigin here?)

Firstyear (changes requested):

server/web_ui/src/lib.rs
yaleman: negatory, from my PR comment 😄 

> where you see me remove opts.credentials(RequestCredentials::SameOrigin); it's because [that's the default](https://developer.mozilla.org/en-US/docs/Web/API/Request/credentials)

Firstyear: Ahhh. Even if it's the default, is there a harm in being explicit?

yaleman: Sure, no harm at all

Firstyear (approved):

PR #1786, opened 2023-06-27
Improve cli to support multi-domain handling.

### LINKED ISSUE #1783: cached tokens show username not matching the target domain when accessing multiple systems
### I did this

Logged into one domain (user@foo.com), then logged into a second one (user@bar.com).

### I expected the following

Auth to fail and not be prompted to log in with user@foo.com

Current HEAD

Any 😄 

### Any other comments

Yeah, I know it's unlikely but it's a kooky bug that causes dev fail.

Firstyear:
Fixes #1783 - this references tokens by spn instead of name, and now allows a username to act as a "filter" for current tokens. 

### DIFF
--- a/tools/cli/src/cli/common.rs
+++ b/tools/cli/src/cli/common.rs
@@ -129,7 +129,7 @@ impl CommonOpt {
 
                 // Is it in the store?
                 match possible_token {
-                    Some(t) => t.clone(),
+                    Some(t) => t,
                     None => {
                         error!(
                             "No valid authentication tokens found for {}.",
--- a/tools/cli/src/cli/common.rs
+++ b/tools/cli/src/cli/common.rs
@@ -110,10 +110,11 @@ impl CommonOpt {
                         .get(filter_username)
                         .map(|t| (filter_username.clone(), t.clone()))
                 } else {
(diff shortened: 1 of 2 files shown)


tools/cli/src/cli/common.rs
```diff
@@ -101,19 +101,41 @@ impl CommonOpt {
...
+                let possible_token = if filter_username.contains('@') {
+                    // If there is an @, it's an spn so just get the token directly.
+                    tokens
+                        .get(filter_username)
+                        .map(|t| (filter_username.clone(), t.clone()))
+                } else {
+                    // First, filter for tokens that match.
+                    let mut token_refs: Vec<_> = tokens
+                        .iter()
+                        .filter(|(t, _)| t.starts_with(filter_username))
```
yaleman, line 116: Shouldn't this have an @ appended to the end? If you do `-D foo` and the token `foobar@example.com` exists, that'll match.

```suggestion
                        .filter(|(t, _)| t.starts_with(&format!("{}@", filter_username)))
```

yaleman (changes requested):

yaleman (approved):
Wewt

Issue #1220, opened 2022-11-22
Add healthcheck to docker container


Firstyear:

yaleman:
As discussed we can implement this as a Kanidmd command which loads the config, figures out the URL and checks the status endpoint.

cuberoot74088:
After having the server capable of probing it's own health, is the remaining work to "just" add some line into the Dockerfile like documented here?

https://docs.docker.com/engine/reference/builder/#healthcheck

Happy to support here.

Btw, why not just a simple curl to localhost?

yaleman:
> Btw, why not just a simple curl to localhost?

Because folks can change their ports and things, and it was a tiny change to the existing binary without adding any packages to the build 😄 

Also yep, just needs a `Dockerfile` change.

cuberoot74088:
OK, thanks for clarification. I will propose something.

svde:
Not sure if this is related, the containers health is "unhealthy", but everything works. The health error is:

📜 Using config file: "/data/server.toml" 00000000-0000-0000-0000-000000000000 ERROR 🚨 [error]: CRITICAL: Connection failed: error sending request for url (https://[::]:8443/status) Logging pipeline completed shutdown 

server.toml contains:
bindaddress = "[::]:8443"

Firstyear:
@yaleman I think if the bind addr is [::] then it needs to map to ::1? Similar for 0.0.0.0 to 127.0.0.1.

yaleman:
@svde can you please open this as a new issue, as the tooling's there but your particular config's making it sad - if you can include your config and if you're triggering the healthcheck another way that'd be good to hear

PR #1789, opened 2023-06-28
Fixing the kanidmd healthcheck

yaleman:
Refers to #1220 

- [x] cargo fmt has been run

### DIFF
--- a/server/daemon/src/main.rs
+++ b/server/daemon/src/main.rs
@@ -17,6 +17,7 @@ static ALLOC: tikv_jemallocator::Jemalloc = tikv_jemallocator::Jemalloc;
 use std::fs::{metadata, File};
 // This works on both unix and windows.
 use fs2::FileExt;
+use kanidm_proto::messages::ConsoleOutputMode;
 #[cfg(target_family = "unix")]
 use std::os::unix::fs::MetadataExt;
 use std::path::PathBuf;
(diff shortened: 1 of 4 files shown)


server/daemon/src/main.rs
```diff
@@ -523,36 +530,41 @@ async fn main() -> ExitCode {
...
-                    //             Err(err) => {
-                    //                 error!("Failed to read TLS chain: {err:?}");
-                    //                 client
-                    //             }
-                    //         }
-
-                    //     },
-                    //     None => client,
-                    // };
+                    client = match &sopt.ca_cert {
```
Firstyear, line 547: IMO you don't need a ca_cert param. Chain files are a series of pem, so just passing the chain to from_pem should read the first pem only which is the leaf cert.

Firstyear (changes requested):

server/daemon/src/main.rs
yaleman: You'd think that... `reqwest` was *not* impressed.

a) it dies when you pass multiple files
b) it won't trust it based on the leafs
c) I can't remember but there was something else

Firstyear: Okay I'm pretty against adding ca_cert as an option too here. We already have the ca available via the chain file.

So I think we either:

* Split the chain file to make reqwest happy.
* Don't validate CA's since we are (presumably) going via localhost anyway ...

server/daemon/src/opt.rs
```diff
@@ -76,6 +76,12 @@ struct HealthCheckArgs {
     /// Disable TLS verification
     #[clap(short, long, action)]
     no_verify_tls: bool,
+    /// Specify the CA cert file path to use for TLS verification
+    #[clap(short='C', long,env="KANIDM_CA_CERT")]
+    ca_cert: Option<String>,
+    /// Check the 'origin' URL from the server configuration file, instead of the 'address'
+    #[clap(short='O', long, action)]
+    check_origin: bool,
```
Firstyear, line 84: If orgin goes via a load balancer this could be a huge footgun because if something happens then every servers healthcheck will fail and it'll spiral. It could also mean instances arle healthchecking partner instances not themself.

Firstyear (changes requested):

Firstyear:
I think I'd rather skip cert validation and guarantee we are contacting localhost/same instance, rather than attempt to validate the certs/tls/domain name. This also removes the need for the new config options.

server/daemon/src/opt.rs
yaleman: But that depends on if one *intends* to go via the origin or not - the default is to use the `address` field, but in my use case I could use it to identify if something in the path was down/funky, including the cert having expired.

As of the current update I just pushed, if the user just supplies the config file, it'll go to the `https://address:port` and turn off TLS validation (since the bind address is rarely the hostname in a cert). If one wants to do a full end-to-end check on a system then they can healthcheck with extra options.

server/daemon/src/main.rs
yaleman: I'll see if I can get the splitting going, I really don't want to remove the validation option if I can avoid it - see the other comment thread.

server/daemon/src/opt.rs
Firstyear: But this isn't intended to be a full end to end check, it's just meant to be for "is my container still running"

I think we should limit it to local host only IMO. For end to end, we should have a different approach than this.

server/daemon/src/main.rs
Firstyear: Could you use https://docs.rs/openssl/latest/openssl/ssl/struct.SslContextBuilder.html#method.set_certificate_chain_file somehow?

server/daemon/src/opt.rs
Firstyear: I think some of these ideas are good but bigger than this issue perhaps especially the cert validation etc.

server/daemon/src/main.rs
yaleman: Splittin' works, just grabbed the last one 😄

server/daemon/src/opt.rs
yaleman: My original intent for it was an "is this both **responding** _and_ **healthy**" which I'd expect to include:

- cert validation, especially since we're requiring TLS end-to-end and certs have a habit of expiring.
- (eventually) the introspection to be able to deal with "replication is good", the system's queue is not overloaded etc.

The origin check is kinda handy for prechecks in dev etc too, making sure the cert's valid.

Firstyear: Also I think that maybe we are looking at this as different things. You're looking at healthcheck as a "lets check the whole kanidm fleet" I'm thinking about it as a local "is this one instance okay". That could be part of the thing here about the origin thing.

yaleman: Yeah, I can see both sides, hopefully folks are smart enough to know the difference - and it's ~5 lines difference between the two, with a default for the "container" check 😄

PR #1790, opened 2023-06-28
20230628 tpm minor issue with key regen

Firstyear:
Minor issue with key regen - if a key is unloadable, we need to zero it and proceed rather than error. 

### DIFF
--- a/unix_integration/src/db.rs
+++ b/unix_integration/src/db.rs
@@ -844,7 +844,7 @@ pub(crate) mod tpm {
                             })
                             .ok()
                             // It loaded, so sub in our context.
-                            .map(|()| maybe_ctx)
+                            .map(|_handle: TpmHandle| maybe_ctx)
                     })
             } else {
                 None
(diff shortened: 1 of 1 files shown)


yaleman:
If we've specified use of a TPM and it's unavailable, shouldn't it fail shut?

Firstyear:
> If we've specified use of a TPM and it's unavailable, shouldn't it fail shut?

This is at a different phase of the setup - This is attempting to connect to the tpm and validate our tpm key is operational. If we have required, we fail shut at a later step.

In this case, if we succeed to connect to the tpm and then find our tpm key is invalid, we have to regenerate the tpm hmac key. this effectively "fails shut" all previous password hashes since we'll never be able to validate them as the hmac keys have rolled. 

So far I noticed this because the libvirt soft tpm appears to reset the context on a reboot which triggered me to find this behaviour to need to roll the keys on a reboot. I think I need to rework a bit of key management to make this more durable anyway, but in this specific case we have a secure forward path at least on the client cache.

yaleman (approved):

PR #1797, opened 2023-07-02
Converting from tide to axum

### LINKED ISSUE #1787: Replace the web server framework

`tide` is unmaintained, time to find a new HTTPS service.

Migrate from `tide` to `axum`

- Rocket ("easy, but can be slow")
- Actix (seems good, has some curious ways, happy to look at it if Axum doesn't work)
- writing something entirely from scratch (NO, WILLIAM!)

**Additional context**

- https://www.arewewebyet.org
- https://blog.logrocket.com/top-rust-web-frameworks/

** Steps **

- [ ] Ensure *all* the current routes are tested in some way
- [ ] replace tide
- [ ] ???
- [ ] 🦀 🍰 🥳 
- [ ] ensure HTTP/2 works properly.
- [ ] ensure all endpoints work the same.

yaleman:
Tasks: 
- [x] the "startup" phase is done, supports testkit etc, uses openssl
- [x] first run through converting all the routes
- [x] tests passing 
- [x] compression validation
- [x] CSP headers
- [x] Caching headers
- [ ] X-Forwarded-For handling

Fixes #1787

- [x] cargo fmt has been run


server/core/src/https/mod.rs
```diff
@@ -429,568 +110,475 @@ pub async fn create_https_server(
...
+                    "Couldn't find Web UI package path: ({}), quitting.",
+                    env!("KANIDM_WEB_UI_PKG_PATH")
                 );
-            })?;
+                std::process::exit(1);
+            }
+            Router::new()
+                .route("/", get(|| async { Redirect::temporary("/ui/login") }))
+                .route("/ui", get(crate::https::ui::ui_handler))
+                .route("/ui", get(crate::https::ui::ui_handler))
```
cuberoot74088, line 170: should this second ui route have a trailing slash like this: `"/ui/"` ?

Cargo.toml
```diff
@@ -42,6 +42,8 @@ repository = "https://github.com/kanidm/kanidm/"
 argon2 = { version = "0.5.0", features = ["alloc"] }
 async-recursion = "1.0.4"
 async-trait = "^0.1.68"
+axum = {version = "0.6.18", features = ["json", "http2", "macros", "tracing", "headers", "original-uri", "query", "form"]}
+axum-csp = { path="/Users/yaleman/Projects/goatns/axum-csp" }
```
cuberoot74088, line 46: Is it okay to use your published crate as well? Like this I was not able to build it.

server/core/src/https/middleware/mod.rs
```diff
@@ -0,0 +1,83 @@
...
+/// For holding onto the event ID
+pub struct KOpId {
+    pub eventid: Uuid,
+    pub value: String,
+    pub uat: Option<String>,
+}
+
+/// This runs at the start of the request, adding an extension with the OperationID
+pub async fn kopid_start<B>(
+    headers: HeaderMap,
```
cuberoot74088, line 37: Maybe use a TypedHeader for auth token extraction?

https://docs.rs/axum/0.6.18/axum/struct.TypedHeader.html#as-extractor

with this

https://docs.rs/headers/0.3.8/headers/authorization/struct.Bearer.html

Would that work or have I missed something?

Cargo.toml
yaleman: of course - I've released the latest version as 0.0.4 now

server/core/src/https/middleware/mod.rs
yaleman: yep, it's on my list of things to go back and do :smile: added it as an explicit todo now. I can't just use an extractor because that *enforces* the header, and I just tried to do it as an optional but then axum got grumpy

Firstyear:
@yaleman > work out how to make sessions work (the axum-compatible session handlers all seem to use... not Uuid's as IDs)

What do you mean by this? I think we put uuids in where we needed them?

server/core/src/https/mod.rs
yaleman: Good catch - turns out I needed to add *another* route to cover that, because `/ui` and `/ui/*key` weren't enough.

yaleman:
> @yaleman > work out how to make sessions work (the axum-compatible session handlers all seem to use... not Uuid's as IDs)
> 
> What do you mean by this? I think we put uuids in where we needed them?

I think I worked it out in [87b1613](https://github.com/kanidm/kanidm/pull/1797/commits/87b161387c76f8862703f6e7bd94ba64a0991fff) - IIRC I stubbed something out while working through the many, many compile errors early on and didn't go back for it.

server/core/src/https/middleware/mod.rs
cuberoot74088: Hi @yaleman,
looks like I messed up the "demo" PR for you.
I did an optional auth header extraction here https://github.com/kanidm/kanidm/commit/c090310f2df1f9dd4afc72b6aee0eb908adaced8

The build works as some tests are still failing it is hard for me to reason why. Hopefully, this is still useful to you.

yaleman: Thanks for that, I'll give it a go - I wouldn't say you've messed anything up at all! 😄

yaleman: Woo, that seems to work, thanks!

libs/sketching/src/middleware.rs
```diff
@@ -1,107 +1,107 @@
-use tide::{self, Middleware, Next, Request};
-use tracing::{self, instrument};
+// use tide::{self, Middleware, Next, Request};
```
Firstyear, line 1: Should we just RM this file?

server/core/src/config.rs
```diff
@@ -64,16 +64,22 @@ pub struct ServerConfig {
 }
 
 impl ServerConfig {
-    pub fn new<P: AsRef<Path>>(config_path: P) -> Result<Self, ()> {
+    pub fn new<P: AsRef<Path>>(config_path: P) -> Result<Self, &'static str> {
         let mut f = File::open(config_path).map_err(|e| {
             eprintln!("Unable to open config file [{:?}] 🥺", e);
+            "you aren't going to read this anyway"
```
Firstyear, line 70: Why not just place holders then?

server/core/src/https/generic.rs
```diff
@@ -0,0 +1,35 @@
...
+        .status_ref
+        .handle_request(StatusRequestEvent {
+            eventid: kopid.eventid,
+        })
+        .await;
+    Response::new(format!("{}", r))
+}
+
+pub async fn robots_txt() -> impl IntoResponse {
+    #[allow(clippy::expect_used)]
```
Firstyear, line 25: I think we can use:

```
(
    [(CONTENT_TYPE, "text/plain;charset=utf-8")],
    Html(r#" ... "#)
)
```

This avoids the expect.

server/core/src/https/javascript.rs
```diff
@@ -0,0 +1,57 @@
...
+                Ok(value) => value,
+                Err(error) => {
+                    return Err(format!(
+                        "Failed to read {:?}, skipping: {:?}",
+                        wasm_filepath, error
+                    ));
+                }
+            };
+            #[allow(clippy::expect_used)]
+            let shasum = openssl::hash::hash(openssl::hash::MessageDigest::sha384(), &filecontents)
```
Firstyear, line 22: Function returns Result, why not .map_err()?;

server/core/src/https/manifest.rs
```diff
@@ -127,32 +128,48 @@ pub async fn manifest(req: tide::Request<AppState>) -> tide::Result {
...
+            Err(_) => String::from(""),
+        };
+    let mut res = Response::new(manifest_string);
 
-    res.set_content_type(MIME_TYPE_MANIFEST);
-    res.set_body(manifest_string);
+    #[allow(clippy::unwrap_used)]
+    res.headers_mut().insert(
+        "Content-Type",
+        HeaderValue::from_str(MIME_TYPE_MANIFEST).unwrap(),
```
Firstyear, line 171: from_static() ?

server/core/src/https/middleware/csp_headers.rs
```diff
@@ -0,0 +1,33 @@
...
+
+    // add the header
+    let headers = response.headers_mut();
+    headers.insert("Content-Security-Policy", state.csp_header);
+
+    response
+}
+
+/// Removes the CSP headers from the response
+pub async fn strip_csp_headers<B>(request: Request<B>, next: Next<B>) -> Response {
```
Firstyear, line 24: Why do we have a "strip" csp header? Shouldn't we just "not add them" ?

server/core/src/https/middleware/mod.rs
```diff
@@ -0,0 +1,93 @@
...
+const KANIDM_VERSION: &str = env!("CARGO_PKG_VERSION");
+
+/// Injects a header into the response with "X-KANIDM-VERSION" matching the version of the package.
+pub async fn version_middleware<B>(request: Request<B>, next: Next<B>) -> Response {
+    let mut response = next.run(request).await;
+    let headers = response.headers_mut();
+    #[allow(clippy::unwrap_used)]
+    headers.insert(
+        "X-KANIDM-VERSION",
+        HeaderValue::from_str(KANIDM_VERSION).unwrap(),
```
Firstyear, line 26: from_static() ?

server/core/src/https/mod.rs
```diff
@@ -429,568 +128,306 @@ pub async fn create_https_server(
...
+        .add(CspDirectiveType::WorkerSource, vec![CspValue::None])
+        // frame-ancestors 'none'
+        .add(CspDirectiveType::FrameAncestors, vec![CspValue::None])
+        .add(CspDirectiveType::ScriptSource, js_directives)
+        .add(
+            CspDirectiveType::ImgSrc,
+            vec![CspValue::SelfSite, CspValue::SchemeData],
+        );
+
+    // TODO this whole session store is kinda cursed and doesn't work the way we need, I think?
```
Firstyear, line 188: No, it's the one we want, it puts the whole state into the cookie - it's the memory store we can't use because then we can't transfer cookies between instances/restarts of the server.

server/core/src/https/mod.rs
```diff
@@ -429,568 +128,306 @@ pub async fn create_https_server(
...
+    let store = async_session::CookieStore::new();
+    // let secret = b"..."; // MUST be at least 64 bytes!
```
Firstyear, line 190: Is it 64 bytes or ... ?

server/core/src/https/mod.rs
```diff
@@ -429,568 +128,306 @@ pub async fn create_https_server(
...
-            eprintln!(
-                "Couldn't find Web UI package path: ({}), quitting.",
-                env!("KANIDM_WEB_UI_PKG_PATH")
-            );
-            std::process::exit(1);
+    let static_routes = match config.role {
+        ServerRole::WriteReplica | ServerRole::ReadOnlyReplica => {
+            Router::new()
+                .route("/", get(crate::https::ui::ui_handler))
+                .route("/*ui", get(crate::https::ui::ui_handler))
```
Firstyear, line 214: This is matching "/*ui" which also is /helloui and /tui. So I don't think this is the right url here. Are you trying to do the spa url capture? Because if you are you need:

```
    // This isn't clear but it means "serve this dir" followed by
    // "if nothing is found in the dir, serve pkg/index.html", it's not relative to /pkg.
    // So this will serve /hello too.
    let serve_dir = ServeDir::new("pkg")
        .not_found_service(ServeFile::new("pkg/index.html"));

    let spa_serve = Router::new()
        .nest_service("/pkg", serve_dir.clone())
        .fallback_service(serve_dir)
        .with_state(app_state.clone());
```

server/core/src/https/mod.rs
```diff
@@ -429,568 +128,306 @@ pub async fn create_https_server(
...
+    acceptor: SslAcceptor,
+    stream: AddrStream,
+    svc: ResponseFuture<Router, SocketAddr>,
+    protocol: Arc<Http>,
+) -> Result<(), hyper::Error> {
+    #[allow(clippy::expect_used)]
+    let mut tls_stream = SslStream::new(
+        #[allow(clippy::expect_used)]
+        // because if this fails well, then we're going to bail!
+        Ssl::new(acceptor.context()).expect("Failed to build SSL session"),
```
Firstyear, line 356: Then return an error rather then the panic?

Firstyear: I think this will panic if a plaintext conn comes in to a tls port ....

server/core/src/https/mod.rs
```diff
@@ -429,568 +128,306 @@ pub async fn create_https_server(
...
+}
 
-            if let Err(e) = listener.bind(tserver).await {
-                error!(
-                    "Failed to start server listener on address {:?} -> {:?}",
-                    &address, e
-                );
-                return Err(());
-            }
+// /// Silly placeholder response for unimplemented routes
```
Firstyear, line 375: remove

Firstyear (changes requested):

server/core/src/config.rs
yaleman: because I was mad on power and trying to make clippy happy - have replaced these with `std::io::Error`

server/core/src/https/mod.rs
yaleman: I've replaced these with `std::io::Error` handlers and that should sort it

libs/sketching/src/middleware.rs
yaleman: bye bye tideypie, done

server/core/src/https/generic.rs
yaleman: wow, that's.. a bunch of spooky magic. thanks :)

server/core/src/https/manifest.rs
yaleman: indeed!

server/core/src/https/middleware/csp_headers.rs
yaleman: That was another failed experiment, I just rolled with "CSP all the things".

server/core/src/https/mod.rs
yaleman: From the docs:
> `SessionLayer::new` will panic if the secret is less than 64 bytes.

And testing it, it did:

```
thread 'main' panicked at '`secret` must be at least 64 bytes.', /Users/yaleman/.cargo/registry/src/index.crates.io-6f17d22bba15001f/axum-sessions-0.5.0/src/session.rs:107:13
```

server/core/src/https/mod.rs
yaleman: This does the equivalent of a regex of `\/(?P<ui>.*)` and lets you add a "ui" `Path` extractor in `ui_handler` to catch the part of the path - ref [Wildcards](https://docs.rs/axum/0.6.18/axum/routing/struct.Router.html#wildcards).

Adding a `ServeFile` under `/ui/` isn't needed since all the static things are under `/pkg` and that's already there later based on ServerRole. `crate::https::ui::ui_handler` responds with the HTML which loads the WASM frontend.

I'm updating the routes to match /ui/* better - to `/ui/*key`, and a redirector from `/` to `/ui/login`.

yaleman: ... and that explains why some of the tests were passing - it was responding with a 200 to /*

server/core/src/https/mod.rs
Firstyear: For now, I'd just make a vec and use the 32 bytes twice in a row. We'll fix it after.

yaleman: The current secret is 32 bytes, so we ... need to change it?

Firstyear (approved):
You are a wizard. Well done <3

PR #1819, opened 2023-07-06
Identity verification feature

### LINKED ISSUE #337: Feature request: verify identity functionality

Seba-T:
Fixes #337

- [x] cargo fmt has been run


server/core/src/https/v1.rs
```diff
@@ -971,6 +971,59 @@ pub async fn account_delete_id_unix_credential(
...
 }
 
+pub async fn identify_user(
+    State(_state): State<ServerState>,
+    Extension(_kopid): Extension<KOpId>,
+    Path(id): Path<String>,
+    Json(user_state): Json<IdentifyUserState>,
+) -> impl IntoResponse {
+    let _filter = filter_all!(f_id(id.as_str()));
+    dbg!(user_state.clone());
```
Firstyear, line 981: trace!()

server/core/src/https/v1.rs
```diff
@@ -4,27 +4,27 @@ use axum::extract::{Path, Query, State};
...
 use axum::{Extension, Json, Router};
 use axum_macros::debug_handler;
 use axum_sessions::extractors::{ReadableSession, WritableSession};
 use compact_jwt::Jws;
 use http::{HeaderMap, HeaderValue, StatusCode};
 use hyper::Body;
+use serde::{Deserialize, Serialize};
+use uuid::Uuid;
+
+use kanidm_lib_identity_verification_types::{IdentifyUserState, IdentifyUserTransition};
```
Firstyear, line 17: Why not use kanidm_proto?

Firstyear (changes requested):
As previously stated - try writing tests first in server/testkit/tests/ - this will show you the protocol, test everything and work it out. As slick as the webui is, I think this will help you get some of these parts ironed out :)

server/core/src/https/v1.rs
Seba-T: Rightt

server/core/src/https/v1.rs
Seba-T: Cause I didn’t think of that :p

Seba-T:
Sure! So should I now get the server working before moving on to the cli? Do you think the webui needs any tweaking?

Firstyear:
I think you should write client api tests for server/testkit/tests/ that poke the expected API end points. For now, they should all fail  and 404. But writing the will give you a framework for "what do I need to fill in to make the server do what the UI's expect?"

server/testkit/tests/identity_verification_tests.rs
```diff
@@ -0,0 +1,184 @@
...
+}
+
+async fn create_other_user_valid(rsclient: &KanidmClient) {
+    let e: Entry = serde_json::from_str(
+        r#"{
+            "attrs": {
+                "class": ["account", "service_account"],
+                "name": ["testaccount_2"],
+                "displayname": ["testaccount"],
+                "ECDHKey": [45,45,45,45,45,66,69,71,73,78,32,80,82,73,86,65,84,69,32,75,69,89,45,45,45,45,45,10,77,73,71,72,65,103,69,65,77,66,77,71,66,121,113,71,83,77,52,57,65,103,69,71,67,67,113,71,83,77,52,57,65,119,69,72,66,71,48,119,97,119,73,66,65,81,81,103,99,53,67,106,106,68,89,74,101,75,79,51,43,52,52,115,10,102,82,122,97,79,52,52,53,53,116,86,80,116,105,75,114,47,106,70,122,102,115,81,106,53,108,117,104,82,65,78,67,65,65,83,87,86,104,86,82,48,117,51,107,90,103,104,65,74,109,106,68,51,82,52,104,68,89,97,81,118,102,75,121,10,113,87,100,49,50,66,99,57,67,48,74,108,122,99,66,73,75,86,81,120,108,87,43,65,120,115,98,82,108,88,118,108,65,74,75,108,74,115,51,47,53,54,81,122,71,120,90,85,65,111,107,104,115,97,56,51,10,45,45,45,45,45,69,78,68,32,80,82,73,86,65,84,69,32,75,69,89,45,45,45,45,45,10],
```
Firstyear, line 84: You'll never need to externally create the ec keys, the server will generate these internally.

Firstyear: Also you should be making persons, not accounts. service accounts don't need to verify each other! humans do :)

server/web_ui/src/components/totpdisplay.rs
```diff
@@ -0,0 +1,322 @@
...
+use yew::prelude::*;
+
+use crate::constants::{CLASS_DIV_LOGIN_BUTTON, ID_IDENTITY_VERIFICATION_SYSTEM_TOTP_MODAL};
+use crate::utils::document;
+use crate::views::identityverification::{
+    IdentifyUserState, IdentifyUserTransition, CORRUPT_STATE_ERROR,
+};
+use crate::{do_request, utils, RequestMethod};
+
+static TIME_STEP: u128 = 30;
```
Firstyear, line 22: This should be returned from the server when you get the TOTP so we can just the TOTP window server side and then the ui reflects that.

server/web_ui/src/components/totpdisplay.rs
```diff
@@ -0,0 +1,322 @@
...
+    secret: TotpStatus,
+    main_timer: Option<Timeout>,
+    ticks_timer: Option<Interval>,
+    sync_timer: Option<Timeout>,
+    ticks_left: u8,
+}
+
+pub enum Msg {
+    FetchTotpAndResetTimer,
+    NewTotp { totp: String },
```
Firstyear, line 45: totp's are ints.

server/web_ui/src/components/totpdisplay.rs
```diff
@@ -0,0 +1,322 @@
...
+            Ok((_, _, response, _)) => response,
+            Err(_) => return Msg::Cancel,
+        };
+        console!(response.clone());
+        match serde_wasm_bindgen::from_value(response) {
+            // TODO: check how the backend encodes the totp when sending it
+            Ok(IdentifyUserResponse::ProvideCode { totp }) => Msg::NewTotp { totp },
+            _ => Msg::Cancel,
+        }
+    }
```
Firstyear, line 86: spaces?

server/web_ui/src/views/mod.rs
```diff
@@ -249,11 +254,20 @@ impl ViewsApp {
                     </li>
 
                     <li class="mb-1">
+                      <Link<ViewRoute> classes="nav-link" to={ViewRoute::IdentityVerification}>
```
Firstyear, line 257: Wrap in experimental features for now IMO.

server/web_ui/src/views/identityverification.rs
```diff
@@ -0,0 +1,533 @@
...
+            }
+        };
+        let response = match Self::make_request(request, self.other_id).await {
+            Ok(res) => res,
+            Err(s) => return IdentifyUserTransition::Error { msg: s.to_string() },
+        };
+        return IdentifyUserTransition::from(response);
+    }
+
+    async fn make_request(
```
Firstyear, line 487: I think there is a generic version of this in lib.rs (do_request).

server/web_ui/src/views/identityverification.rs
```diff
@@ -0,0 +1,533 @@
...
+          </>
+        }
+    }
+
+    // the purpose of the following functions is to get what to do next in the state machine.
+    // each main view has its own function, that is the start view (even tho it's displayed for few ms), the id_display_and_submit view
+    // and the submit_code
+
+    async fn get_transition_from_start(state: IdentifyUserState) -> IdentifyUserTransition {
+        let request = match state {
```
Firstyear, line 416: This might be better as a debug assert since the only time this is called is during create?

Firstyear (changes requested):

server/core/src/https/v1.rs
```diff
@@ -971,6 +971,46 @@ pub async fn account_delete_id_unix_credential(
     to_axum_response(res)
 }
 
+pub async fn identify_user(
+    State(_state): State<ServerState>,
+    Extension(_kopid): Extension<KOpId>,
+    Path(id): Path<String>,
+    Json(user_state): Json<IdentifyUserRequest>,
+) -> impl IntoResponse {
+    let _filter = filter_all!(f_id(id.as_str()));
```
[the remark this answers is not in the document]
Seba-T: alright, but I'm not entirely sure I get the difference

server/testkit/tests/identity_verification_tests.rs
Seba-T: sure!

server/web_ui/src/views/identityverification.rs
Seba-T: yep, I was using it inside the `make_request` function, but now I reckon that namewise it doesn't make much sense. I think I should rather call it something like "do_typed_request" to differentiate from the already existing do_request

Cargo.toml
```diff
@@ -82,6 +82,7 @@ kanidmd_lib = { path = "./server/lib" }
 kanidmd_lib_macros = { path = "./server/lib-macros" }
 kanidm_lib_crypto = { path = "./libs/crypto" }
 kanidm_lib_file_permissions = { path = "./libs/file_permissions" }
+kanidm_lib_identity_verification_types = { path = "./libs/identity_verification_types" }
```
Firstyear, line 85: I'm not sure we need another crate here, can't we use proto? Or do these types need to be buried into their respective parts? I think a whole crate here is the wrong way.

server/core/src/https/v1.rs
```diff
@@ -971,6 +971,44 @@ pub async fn account_delete_id_unix_credential(
...
+    State(_state): State<ServerState>,
+    Extension(_kopid): Extension<KOpId>,
+    Path(id): Path<String>,
+    Json(user_state): Json<IdentifyUserRequest>,
+) -> impl IntoResponse {
+    let _filter = filter!(f_id(id.as_str()));
+    trace!("{:?}", &user_state);
+    match user_state {
+        IdentifyUserRequest::Start => {
+            // here we need to check if the user can use the identity verification feature
```
Firstyear, line 984: So here you'll need to make a new type of "event". You can see this in server/lib/src/idm/serviceaccount.rs where an GenerateApiTokenEvent is used. We want to wrap every "request" to an "event" type which completely encapsulates the request, this way we can always forge events in testing to simulate what was sent to us via the event layers. So have a look an trace something like that to undestand the layers here.

Firstyear (changes requested):

server/web_ui/src/components/totpdisplay.rs
Seba-T: fixed!

Cargo.toml
Seba-T: yeah, as a matter of fact I thought I was using proto :p

server/core/src/https/v1.rs
```diff
@@ -971,6 +971,44 @@ pub async fn account_delete_id_unix_credential(
...
+            // RETURNS EITHER IdentifyUserResponse::IdentityVerificationAvailable or IdentifyUserResponse::IdentityVerificationAvailable
+            //TODO!
+            return if id.is_empty() {
```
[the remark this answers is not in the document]
Seba-T: I thought that I could achieve the same by passing the user own uuid: if the server sees that the user sent its own uuid it just tells him if the feature is available or not on his account

server/core/src/https/v1.rs
Seba-T: Done!

server/core/src/actors/v1_read.rs
```diff
@@ -35,6 +38,8 @@ use kanidmd_lib::{
     idm::serviceaccount::ListApiTokenEvent,
 };
 
+use crate::config::LogLevel;
```
Firstyear, line 41: Why are you importing this? I don't see you using it?

server/core/src/actors/v1_read.rs
```diff
@@ -877,6 +882,48 @@ impl QueryServerReadV1 {
...
+        let target = idms_prox_read
+            .qs_read
+            .name_to_uuid(other_id.as_str())
+            .map_err(|e| {
+                admin_error!(err = ?e, "Error resolving id to target");
+                e
+            })?;
+
+        trace!("{:?}", &user_request);
+        match user_request {
```
Firstyear, line 914: This looks good!

server/core/src/https/v1.rs
```diff
@@ -971,6 +971,19 @@ pub async fn account_delete_id_unix_credential(
     to_axum_response(res)
 }
 
+pub async fn identify_user(
```
Firstyear, line 974: A more verbose fn name here would be good IMO. Note that (most) other fn's define the http verb type so this could be something like "person_post_identify_user".

server/testkit/tests/identity_verification_tests.rs
```diff
@@ -0,0 +1,448 @@
...
+    let res: Result<IdentifyUserResponse, Error> = make_request(
+        &rsclient,
+        IdentifyUserRequest::Start,
+        Some(VALID_USER_A_NAME.to_string()),
+    )
+    .await;
+    assert!(res.is_err());
+    if let Err(err) = res {
+        assert_eq!(err.status().unwrap(), StatusCode::FORBIDDEN)
+    }
```
Firstyear, line 59: Spaces between the different  asserts, you can also do a single assert with:
`assert!(matches!(Err(StatusCode::FORBIDDEN), res));`

server/testkit/tests/identity_verification_tests.rs
```diff
@@ -0,0 +1,448 @@
...
+        &rsclient,
+        IdentifyUserRequest::Start,
+        Some(VALID_USER_A_NAME.to_string()),
+    )
+    .await;
+    assert!(res.is_ok());
+    assert_eq!(
+        res.unwrap(),
+        IdentifyUserResponse::IdentityVerificationUnavailable
+    );
```
Firstyear, line 108: space :) lets the codes breaaaathe. 

It helps the reader deliniate one case from the next else it's hard to know if it's all part of one blob or separate.

server/testkit/tests/identity_verification_tests.rs
```diff
@@ -0,0 +1,448 @@
...
+
+    assert!(matches!(
+        response.unwrap(),
+        IdentifyUserResponse::ProvideCode { .. } | IdentifyUserResponse::CodeFailure
+    ));
+}
+
+// here we actually test the full idm flow by duplicating the server
+#[kanidmd_testkit::test]
+
```
Firstyear, line 250: For once, too much space 🤣

server/testkit/tests/identity_verification_tests.rs
```diff
@@ -0,0 +1,448 @@
...
+    .unwrap();
+    assert_eq!(user_b_req_2_wrong, IdentifyUserResponse::CodeFailure);
+    // now we do it using the right totp
+    let user_b_req_2_right = make_request(
+        &valid_user_B_client,
+        IdentifyUserRequest::SubmitCode { other_totp: totp },
+        Some(VALID_USER_A_NAME.to_string()),
+    )
+    .await;
+    // since user B has already provided their code this is their last action and they must get a Success if
```
Firstyear, line 323: Perfect! This looks really good!

Firstyear (changes requested):
This is really coming together, I think you've done great here to take on the feedback and improve the tests, and start to wire in the parts you need. I assume if you do a cargo test right now you have a bunch of expected failures?

server/core/src/actors/v1_read.rs
Seba-T: Yeah dumb autoimports stuff

server/core/src/actors/v1_read.rs
Seba-T: Thanks! :)

server/testkit/tests/identity_verification_tests.rs
Seba-T: I like the single patter more tbh

Seba-T:
Yeah it’s all over the place at the moment

server/testkit/tests/identity_verification_tests.rs
Seba-T: thanks :)

server/lib/src/value.rs
```diff
@@ -380,7 +383,7 @@ pub enum PartialValue {
     UiHint(UiHint),
     Passkey(Uuid),
     DeviceKey(Uuid),
-    // The label, if any.
+    ECDHKey, // The label, if any.
```
Firstyear, line 386: I don't think it needs a partialvalue? If there are no query selectors rn, then we don't need this.

server/lib/src/value.rs
```diff
@@ -900,6 +905,7 @@ pub enum Value {
 
     TotpSecret(String, Totp),
     AuditLogString(Cid, String),
+    ECDHKey(Vec<u8>),
```
Firstyear, line 908: This should be the validated and deserialised key state, not the DER bytes here.

server/lib/src/server/mod.rs
```diff
@@ -544,6 +544,7 @@ pub trait QueryServerTransaction<'a> {
                         .map_err(|()| OperationError::InvalidAttribute("Invalid uihint syntax".to_string())),
                     SyntaxType::TotpSecret => Err(OperationError::InvalidAttribute("TotpSecret Values can not be supplied through modification".to_string())),
                     SyntaxType::AuditLogString => Err(OperationError::InvalidAttribute("Audit logs are generated and not able to be set.".to_string())),
+                    SyntaxType::ECDHKey => Err(OperationError::InvalidAttribute("Personal ecdh keys are generated and not able to be set.".to_string())),
```
Firstyear, line 547: I don't think the word "Personal" is needed here.

server/lib/src/plugins/ecdhkeygen.rs
```diff
@@ -0,0 +1,142 @@
...
+        Ok(())
+    }
+
+    // fn verify(
+    //     _qs: &mut crate::server::QueryServerReadTransaction,
+    // ) -> Vec<Result<(), kanidm_proto::v1::ConsistencyError>> {
+    //     admin_error!("plugin {} has an unimplemented verify!", Self::id());
+    //     vec![Err(ConsistencyError::Unknown)]
+    // }
+}
```
Firstyear, line 142: Tests?

server/lib/src/plugins/ecdhkeygen.rs
```diff
@@ -0,0 +1,142 @@
...
+    };
+}
+pub struct EcdhKeyGen {}
+
+impl EcdhKeyGen {
+    fn is_entry_to_update<VALUE, STATE>(entry: &mut Entry<VALUE, STATE>) -> bool {
+        CLASSES_TO_UPDATE
+            .iter()
+            .any(|pv| entry.attribute_equality("class", pv))
+    }
```
Firstyear, line 29: You know what I'm about to say here ... :)

server/lib/src/plugins/ecdhkeygen.rs
```diff
@@ -0,0 +1,142 @@
...
+    fn pre_create_transform(
+        _qs: &mut QueryServerWriteTransaction,
+        cand: &mut Vec<EntryInvalidNew>,
+        _ce: &CreateEvent,
+    ) -> Result<(), OperationError> {
+        Self::handle_user_account_creation(cand)
+    }
+
+    fn pre_modify(
```
Firstyear, line 61: This needs to handle mods too. Because else how do we generate keys for users that already exist?

server/lib/src/plugins/ecdhkeygen.rs
```diff
@@ -0,0 +1,142 @@
...
+use crate::prelude::{Entry, EntryInvalidNew, OperationError};
+use crate::server::QueryServerWriteTransaction;
+use crate::value::PartialValue;
+use sketching::tagged_event;
+use sketching::EventTag;
+
+lazy_static! {
+    // it contains all the partialvalues used to match against an Entry's class,
+    // we just need a partialvalue to match in order to target the entry
+    static ref CLASSES_TO_UPDATE: [PartialValue; 1] = [PartialValue::new_iutf8("account")];
```
Firstyear, line 16: Wrong class type. Only *humans* need to verify each other, not accounts. Accounts also include machine/service accounts.

server/lib/src/idm/identityverification.rs
```diff
@@ -0,0 +1,278 @@
...
+            IdentifyUserResponse::ProvideCode {
+                step: TOTP_STEP as u32,
+                totp,
+            }
+        } else {
+            IdentifyUserResponse::Success
+        };
+        Ok(res)
+    }
+}
```
Firstyear, line 278: Tests? You probably should get into the habit of writing the tests first before you impl a wall of feature. :)

server/lib/src/entry.rs
```diff
@@ -2471,6 +2473,16 @@ impl<VALID, STATE> Entry<VALID, STATE> {
             .and_then(|vs| vs.to_jws_key_es256_single())
     }
 
+    pub fn get_ava_single_eckey_private(&self, attr: &str) -> Option<EcKey<Private>> {
+        self.attrs
+            .get(attr)
+            .and_then(|vs| vs.to_eckey_private_single())
+    }
```
Firstyear, line 2480: You know what :)

server/lib/src/constants/schema.rs
```diff
@@ -1636,7 +1670,8 @@ pub const JSON_SCHEMA_CLASS_ACCOUNT: &str = r#"
         "user_auth_token_session",
         "oauth2_session",
         "description",
-        "name_history"
+        "name_history",
+        "id_verification_eckey"
```
Firstyear, line 1674: See below - wrong class.

server/lib/src/constants/schema.rs
```diff
@@ -211,6 +211,40 @@ pub const JSON_SCHEMA_ATTR_NAME_HISTORY: &str = r#"{
...
 
+pub const JSON_SCHEMA_ATTR_ECDH_KEY: &str = r#"{
+    "attrs": {
+      "class": [
+        "object",
+        "system",
+        "attributetype"
+      ],
+      "description": [
+        "Account verification public and optionally private key "
```
Firstyear, line 222: Should probably just comment that it's an EC key because it could be used for other things.

Firstyear: Also given it's an ECKey, it's not limited to ECDH, so name this ECKEY not ECDH.

Firstyear (changes requested):

server/lib/src/value.rs
Seba-T: actually I started with that, but then I got into much trouble when I tried to insert something like `EcKey<Public>` in a `SmolSet`, as it was complaining that the key didn't have `Eq` or `PartialEq`. That's why I went for the bytes

server/lib/src/entry.rs
Seba-T: dang I'll develop a tool that checks for spaces after functions :p

server/lib/src/plugins/ecdhkeygen.rs
Seba-T: yeah I wrote the plugin on the fly just to see how the main tests were doing, but I still need to thoroughly think about what the plugin is supposed to do

server/lib/src/plugins/ecdhkeygen.rs
Seba-T: yep ..

server/lib/src/valueset/eckey.rs
```diff
@@ -99,7 +99,12 @@ impl ValueSetT for ValueSetEcKey {
     }
 
     fn generate_idx_eq_keys(&self) -> Vec<String> {
-        Vec::with_capacity(0)
+        let key_der = match self {
```
Firstyear, line 102: Do we actually even need idx keys here? I don't think we will be able to query this type, I think we return an empty vec.

server/lib/src/schema.rs
```diff
@@ -214,7 +214,7 @@ impl SchemaAttribute {
             // Comparing on the label.
             SyntaxType::TotpSecret => matches!(v, PartialValue::Utf8(_)),
             SyntaxType::AuditLogString => matches!(v, PartialValue::Utf8(_)),
-            SyntaxType::EcKey => matches!(v, PartialValue::Iutf8(_)),
+            SyntaxType::EcKey => matches!(v, PartialValue::PrivateBinary),
```
Firstyear, line 217: I think this is a good idea, it allows us to check presence this way.

server/lib/src/valueset/eckey.rs
Seba-T: yeah I added it at 2am because it was throwing some `InvalidSyntax` error on startup so I started doing random shit

server/lib/src/schema.rs
Seba-T: Yeah but as I wrote it I got entangled like A LOT in ensuring I used the right `PartialValue`. Since it has such a different name from `EcKey` or something it's hard to spot where you're using the wrong PartialValue. Maybe we should introduce another variant with no data associated to it just for clarity?

server/lib/src/idm/identityverification.rs
Seba-T: I was trusting the api tests to handle everything, yet after thinking about it I realized they depend on too many things, so more specific tests could help

Seba-T:
Alright I think I did everything that was suggested and then some, I hope I didn't miss out on any bits. All the code so far in this PR  fully implements the identity verification feature and the respective webui; the cli is missing though. At this point everything is tested and passes, although I would bet I missed some edge cases in the tests.

server/lib/src/constants/schema.rs
```diff
@@ -211,6 +211,40 @@ pub const JSON_SCHEMA_ATTR_NAME_HISTORY: &str = r#"{
...
+      "class": [
+        "object",
+        "system",
+        "attributetype"
+      ],
+      "description": [
+        "Account verification public and optionally private key."
+      ],
+      "index": [
+        "EQUALITY"
```
Firstyear, line 225: No indexes for this attr.

server/lib/src/constants/schema.rs
```diff
@@ -211,6 +211,40 @@ pub const JSON_SCHEMA_ATTR_NAME_HISTORY: &str = r#"{
...
 
+pub const JSON_SCHEMA_ATTR_EC_KEY: &str = r#"{
+    "attrs": {
+      "class": [
+        "object",
+        "system",
+        "attributetype"
+      ],
+      "description": [
+        "Account verification public and optionally private key."
```
Firstyear, line 222: I think we should actually just make this a private key only. I think public will have to eventually be a seperate type to allow for splitting the values and ensuring at a type/attr level we don't leak across trust boundaries. 

How angy will you be if I ask you to make this "EC_KEY_PRIVATE" rather than EC_KEY?

server/lib/src/constants/schema.rs
```diff
@@ -211,6 +211,40 @@ pub const JSON_SCHEMA_ATTR_NAME_HISTORY: &str = r#"{
...
+      ],
+      "index": [
+        "EQUALITY"
+      ],
+      "unique": [
+        "true"
```
Firstyear, line 228: don't need unique here, I think it just adds cost to the update of these values, and it requires indexing.

server/lib/src/constants/schema.rs
```diff
@@ -211,6 +211,40 @@ pub const JSON_SCHEMA_ATTR_NAME_HISTORY: &str = r#"{
...
+        "false"
+      ],
+      "sync_allowed": [
+        "true"
+      ],
+      "attributename": [
+        "id_verification_eckey"
+      ],
+      "syntax": [
+        "EC_KEY"
```
Firstyear, line 240: As above, I think we go EC_KEY_PRIVATE.

server/lib/src/repl/proto.rs
```diff
@@ -390,6 +390,9 @@ pub enum ReplAttrV1 {
     AuditLogString {
         set: Vec<(Cid, String)>,
     },
+    ECDHReplKey {
```
Firstyear, line 393: Just because it's in repl proto, doesn't mean it needs repl in the name.

server/lib/src/server/mod.rs
```diff
@@ -672,6 +673,7 @@ pub trait QueryServerTransaction<'a> {
                             OperationError::InvalidAttribute("Invalid uihint syntax".to_string())
                         }),
                     SyntaxType::AuditLogString => Ok(PartialValue::new_utf8s(value)),
+                    SyntaxType::EcKey => Ok(PartialValue::Iutf8(value.to_string())),
```
Firstyear, line 676: I don't think we need to be able to query these keys. Why would we need to look up a key by it's ID? We only need to lookup the users spn/name, and the key is associated.

server/lib/src/value.rs
```diff
@@ -113,6 +116,39 @@ pub struct Address {
     pub country: String,
 }
 
+#[derive(Debug, Clone)]
+pub enum PubOrPrivEcKey {
```
Firstyear, line 120: I think this should be private only.

server/lib/src/valueset/eckey.rs
```diff
@@ -0,0 +1,207 @@
...
+use kanidm_proto::v1::OperationError;
+use openssl::ec::EcKey;
+use openssl::pkey::{Private, Public};
+use openssl::sha::sha256;
+
+use super::ValueSet;
+
+#[derive(Debug, Clone, PartialEq)]
+pub struct ValueSetEcKey {
+    key: PubOrPrivEcKey,
```
Firstyear, line 16: I think given the above, this could be change to rehydrate the vaue from the db/repl as a private key, but then the struct should have

```
priv: EcKey<Private>
public: EcKey<Public>
```
This way we only have to do the private -> public extract once on load from disk or update, and then we can rely on the entries being Arc counted and we can return references etc.

server/testkit/tests/identity_verification_tests.rs
```diff
@@ -0,0 +1,364 @@
...
+        .await
+        .unwrap();
+
+    let user_b_req_1 = valid_user_b_client
+        .idm_person_identify_user(USER_A_NAME, IdentifyUserRequest::Start)
+        .await
+        .unwrap();
+
+    // now we have to consider the two separate cases: first we address the case a has the lowest uuid
+    if user_a_uuid < user_b_uuid {
```
Firstyear, line 187: I wonder if there is a way to do this neater, like having a tuple that swaps the clients like:

```
let (client_left, client right) = if user_a_uuid < user_b_uuid {
    client_a, client_b
} else {
    client_b, client_a
};
```

server/lib/src/idm/identityverification.rs
```diff
@@ -0,0 +1,611 @@
...
+
+            let res_a_correct = idms_prox_read.handle_identify_user_submit_code(
+                &IdentifyUserSubmitCodeEvent::new(user_b_uuid, ident_a.clone(), totp),
+            );
+
+            assert!(matches!(
+                res_a_correct,
+                Ok(IdentifyUserResponse::ProvideCode { .. })
+            ));
+
```
Firstyear, line 515: Add a test that asserts that the totps on both sides are never the same. The totp I give you shouldn't be the same as the totp you give me. I think you've missed this in your cryptographic code, I can't see where you invert the secret based on uuid order.

server/lib/src/constants/schema.rs
Seba-T: very mildly :p, but actually makes sense

server/testkit/tests/identity_verification_tests.rs
Seba-T: niiice way better than how I was doing it before

server/lib/src/idm/identityverification.rs
Seba-T: Basically from the document we wrote back in march we ruled that each user secret would be composed by the shared ecdh secret and the user own uuid, so `user a` will have `totp_secret_a = shared_ecdh_secret + user_a_uuid` and `user b` will have `totp_secret_b = shared_ecdh_secret + user_b_uuid`

In `get_self_totp_secret` I have
> let Some(self_uuid) = ident.get_uuid() else {
            return Err(OperationError::NotAuthenticated)
        };
        shared_key.extend_from_slice(self_uuid.as_bytes());


while in `get_other_user_totp_secret`
>  shared_key.extend_from_slice(target.as_bytes());

Seba-T:
I had to rebase because something was failing with what I had before :/

server/lib/src/valueset/eckey.rs
```diff
@@ -0,0 +1,184 @@
...
+use crate::prelude::ValueSetT;
+use crate::repl::proto::ReplAttrV1;
+use crate::value::{PartialValue, PubOrPrivEcKey, SyntaxType, Value};
+use kanidm_proto::v1::OperationError;
+use openssl::ec::EcKey;
+use openssl::pkey::{Private, Public};
+
+use super::ValueSet;
+
+pub type ValueSetEcKey = PubOrPrivEcKey;
```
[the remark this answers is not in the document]
Seba-T: Yep :)

server/lib/src/valueset/eckey.rs
```diff
@@ -0,0 +1,184 @@
...
+            Value::EcKey(k) => Ok(self.push(k)),
+            _ => {
+                debug_assert!(false);
+                Err(OperationError::InvalidValueState)
+            }
+        }
+    }
+
+    fn clear(&mut self) {
+        drop(self)
```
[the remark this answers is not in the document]
Seba-T: Yeah clippy was screaming at me for that

server/lib/src/constants/schema.rs
```diff
@@ -211,6 +211,40 @@ pub const JSON_SCHEMA_ATTR_NAME_HISTORY: &str = r#"{
...
+        "PRESENCE"
+      ],
+      "unique": [
+        "false"
+      ],
+      "multivalue": [
+        "false"
+      ],
+      "sync_allowed": [
+        "true"
```
Firstyear, line 234: sync_allowed should be false.

server/lib/src/idm/identityverification.rs
Firstyear: Still good to have a test case proving that the totps aren't the same :)

Firstyear (changes requested):
We are so close! One small test change and a schema comment and I think it's done!

server/lib/src/constants/schema.rs
Seba-T: rightt

server/lib/src/idm/identityverification.rs
```diff
@@ -0,0 +1,641 @@
...
+        assert!(matches!(
+            res_higher_user_2_correct,
+            Ok(IdentifyUserResponse::Success)
+        ));
+    }
+
+    #[idm_test]
+
+    async fn test_codes_differ(idms: &IdmServer, _idms_delayed: &IdmServerDelayed) {
```
Seba-T, line 550: Isn't this checking that? Or do you mean in some other way?

Firstyear: Oh how did I miss this.

Firstyear (approved):
Needs a rebase and it's done 🎉

server/web_ui/pkg/kanidmd_web_ui.d.ts
yaleman, line 1: *.d.ts files being here show you're building in debug mode, which is fine while working on things - but you need to run `build_wasm_release.sh` to get the optimized build please.

`server/web_ui/pkg/kanidmd_web_ui_bg.wasm Binary file modified BIN +12.2 MB (810%)`  <+=- this is why 😄

yaleman (changes requested):

server/lib/src/idm/identityverification.rs
```diff
@@ -0,0 +1,641 @@
...
+        if let Some(early_response) = self.check_for_early_return_conditions(ident, target)? {
+            return Ok(early_response);
+        }
+
+        let totp_secret = self.get_other_user_totp_secret(target, ident)?;
+        let other_user_totp = self.compute_totp(totp_secret)?;
+        if other_user_totp != *code {
+            return Ok(IdentifyUserResponse::CodeFailure);
+        }
+        // if we are the fist it means now it's time to go for ProvideCode, otherwise we just confirm that the code is correct
```
yaleman, line 111:
```suggestion
        // if we are the first it means now it's time to go for ProvideCode, otherwise we just confirm that the code is correct
```

server/lib/src/idm/identityverification.rs
```diff
@@ -0,0 +1,641 @@
...
+        };
+
+        self.qs_read
+            .search(&search)
+            .and_then(|mut entries| entries.pop().ok_or(OperationError::NoMatchingEntries))
+            .and_then(
+                |entry| match entry.get_ava_single_eckey_private("id_verification_eckey") {
+                    Some(key) => Ok(key.clone()),
+                    None => Err(OperationError::InvalidAccountState(
+                        "The user private key is missing!".to_string(),
```
yaleman, line 211: This should identify who it's missing for

server/lib/src/idm/identityverification.rs
```diff
@@ -0,0 +1,641 @@
...
+            Some(key) => Ok(key.check_key().is_ok()),
+            None => Ok(false),
+        }
+    }
+
+    fn get_user_own_key(&mut self, ident: &Identity) -> Result<EcKey<Private>, OperationError> {
+        let search = match SearchEvent::from_whoami_request(ident.clone(), &self.qs_read) {
+            Ok(s) => s,
+            Err(e) => {
+                admin_error!("Failed to retrieve user with the given UUID: {:?}", e);
```
yaleman, line 199: The UUID should be included in the log, especially since it says "with the given UUID" 😄

server/lib/src/idm/identityverification.rs
```diff
@@ -0,0 +1,641 @@
...
+                }
+            };
+        self.qs_read
+            .search(&search)
+            .and_then(|mut entries| entries.pop().ok_or(OperationError::NoMatchingEntries))
+            .and_then(
+                |entry| match entry.get_ava_single_eckey_public("id_verification_eckey") {
+                    Some(key) => Ok(key.clone()),
+                    None => Err(OperationError::InvalidAccountState(
+                        "The requested public key is missing!".to_string(),
```
yaleman, line 237: This should identify who it's missing for

server/lib/src/server/migrations.rs
```diff
@@ -458,6 +458,7 @@ impl<'a> QueryServerWriteTransaction<'a> {
             SCHEMA_ATTR_CREDENTIAL_UPDATE_INTENT_TOKEN.clone().into(),
             SCHEMA_ATTR_DEVICEKEYS.clone().into(),
             SCHEMA_ATTR_DISPLAYNAME.clone().into(),
+            SCHEMA_ATTR_EC_KEY_PRIVATE.clone().into(),
```
yaleman, line 461: I alphabetised these! (ok, except for mail but it was being weird.) Please put it in the right place 😃

server/lib/src/valueset/eckey.rs
```diff
@@ -0,0 +1,220 @@
...
+            })
+        })
+    }
+
+    fn merge(&mut self, other: &super::ValueSet) -> Result<(), kanidm_proto::v1::OperationError> {
+        if let Some(other_key) = other.as_ec_key_private() {
+            let priv_key = other_key.clone();
+            #[allow(clippy::expect_used)]
+            let pub_key = Self::private_key_to_public_key(&priv_key).expect(
+                "Unable to retrieve public key from private key, likely corrupted. You must restore from backup.",
```
yaleman, line 190: This should log an error-level message and return an operation error, not panic.

server/web_ui/src/components/totpdisplay.rs
```diff
@@ -0,0 +1,343 @@
...
+            .map(|s| JsValue::from(&s))
+             else {
+            return Msg::Cancel
+        };
+        let response = match do_request(&uri, RequestMethod::POST, Some(state_as_jsvalue)).await {
+            Ok((_, _, response, _)) => response,
+            Err(_) => return Msg::Cancel,
+        };
+        match serde_wasm_bindgen::from_value(response) {
+            // TODO: check how the backend encodes the totp when sending it
```
yaleman, line 69: Is this TODO-do'd? 😄

server/web_ui/src/components/totpdisplay.rs
```diff
@@ -0,0 +1,343 @@
...
+        } else if time_remaining <= (ALERT_THRESHOLD * self.step as f32) as u32 {
+            "red"
+        } else if time_remaining <= (WARNING_THRESHOLD * self.step as f32) as u32 {
+            "orange"
+        } else {
+            "green"
+        })
+    }
+
+    // TODO! It's not remotely doing what I wanted to do, damned CSS
```
yaleman, line 109: Does it work or not? 🤔

server/web_ui/src/views/identityverification.rs
```diff
@@ -0,0 +1,529 @@
...
+            </div>
+            <p class="text-center">
+              <a href="/"><button href="/" class="btn btn-secondary" aria-label="Return home">{"Return to the home page"}</button></a>
+            </p>
+          </>
+        }
+    }
+
+    // the purpose of the following functions is to get what to do next in the state machine.
+    // each main view has its own function, that is the start view (even tho it's displayed for few ms), the id_display_and_submit view
```
yaleman, line 409:
```suggestion
    // each main view has its own function, that is the start view (even though it's displayed for few ms), the id_display_and_submit view
```

yaleman (changes requested):

server/web_ui/src/views/identityverification.rs
```diff
@@ -0,0 +1,529 @@
...
+                        totp_valid,
+                    }
+                    | IdentifyUserState::SubmitCodeSecond {
+                        other_totp,
+                        totp_valid,
+                    } => {
+                        *totp_valid = VALIDATE_TOTP_RE.is_match(&input);
+                        #[cfg(debug_assertions)]
+                        console::debug!(input.clone());
+                        *other_totp = input.parse::<u32>().ok(); //this should never fail as we set the input type to "number"
```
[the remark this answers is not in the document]
Seba-T: Yeah but even if it did when I unwrap the option I use `unwrap_or_default`, here with `.ok` I'm just transforming the Result into Option so I don't think it will panic even if it's `Err`. But yeah prob I should remove the comment then

server/lib/src/idm/identityverification.rs
Seba-T: sure :)

server/lib/src/server/migrations.rs
Seba-T: wopsies

server/lib/src/valueset/eckey.rs
Seba-T: I agree but currently the ValueSet trait expects a direct result, so I went for the same approach as in `Jwt` for example.
There we have something like
```
 fn to_repl_v1(&self) -> ReplAttrV1 {
        ReplAttrV1::JwsKeyEs256 { set: self.set.iter()
            .map(|k| {
                #[allow(clippy::expect_used)]
                k.private_key_to_der()
                    .expect("Unable to process private key to der, likely corrupted. You must restore from backup.")
            })
            .map(|b| b.into())
            .collect()
        }
    }
```
I could open an issue for this tho

server/web_ui/src/components/totpdisplay.rs
Seba-T: yep :)

server/web_ui/src/components/totpdisplay.rs
Seba-T: Errrr let's say it's not harming anyone for the time being. I mean it's a very hacky solution to get the transition to work the way I'd want it, but it's very unreliable and most of the times it's as if it didn't exist. If we remove it everything else should just work fine

server/lib/src/valueset/eckey.rs
yaleman: but this one has a `Result<>` response, which is why I mentioned it 😄 

```rust
fn merge(&mut self, other: &super::ValueSet) -> Result<(), kanidm_proto::v1::OperationError> {
```

server/web_ui/src/views/identityverification.rs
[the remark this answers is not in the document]
yaleman: That's what I get for reviewing code at midnight, you're right the `.ok` coerces it into a `None`, my bad 😊 

The comment's misleading however 👍🏻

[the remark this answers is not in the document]
Seba-T: fixed it :)

server/lib/src/valueset/eckey.rs
Seba-T: Yep, it's a bummer that some do and some other don't. I wonder if soon we'll have to pay the tech debt again on this one :p

Seba-T: If I remember correctly William also mentioned that the valueset trait could be redesigned, since it's based on the assumption that every value will be stored in a set, and that's not always true, as it's the case for the EcdhKey.
If this ever happens it could also fix the return values of the trait functions.

Firstyear: Yeah, I think it needs a redesign with schema to have the multivalue type removed from schema, and value being single/multi is a property of the type instead.

Firstyear:
@yaleman Once you're happy can you approve and merge :)

server/lib/src/valueset/eckey.rs
yaleman: Either way, less panics, more errors in the meantime please 😃 Then this one's ready to go!

Seba-T: Wopsies I must be blind I thought you wanted me to remove the expect on the method that didn't return the result. 
Anyway I fixed it!

yaleman (approved):

PR #1844, opened 2023-07-09
headless webdriver testing, starting on brotli feature

yaleman:
Fixes #

- [x] cargo fmt has been run


server/lib/src/idm/oauth2.rs
```diff
@@ -1408,9 +1408,10 @@ impl<'a> IdmServerProxyReadTransaction<'a> {
             };
 
         if consent_previously_granted {
+            let pretty_scopes: Vec<String> = granted_scopes.iter().map(|s| s.to_owned()).collect();
```
Firstyear, line 1411: Wont this break log line formatting because pretty is over multiple lines?

Firstyear (changes requested):

server/lib/src/idm/oauth2.rs
yaleman: It's just me to_string'ing the keys of the hashmap so I can do the join ',' later

Firstyear: Oh I misread this sorry, I had a cat in my face

Firstyear (approved):

PR #1874, opened 2023-07-17
Add env var KANIDM_TOKEN_PATH

tumbl3w33d:
that allows overriding the path where the CLI client searches for its token store.

That's especially useful when you run multiple environments and want to switch on a per-directory base, e.g., using tools like direnv.

Fixes #

Checklist

- [ x] This pr contains no AI generated code
- [x ] cargo fmt has been run
- [x ] cargo clippy has been run
- [ x] cargo test has been run and passes

### DIFF
--- a/tools/cli/src/cli/session.rs
+++ b/tools/cli/src/cli/session.rs
@@ -18,16 +18,22 @@ use crate::common::prompt_for_username_get_username;
 use crate::webauthn::get_authenticator;
 use crate::{LoginOpt, LogoutOpt, ReauthOpt, SessionOpt};
 
-static TOKEN_DIR: &str = "~/.cache";
 static TOKEN_PATH: &str = "~/.cache/kanidm_tokens";
+
+fn get_env_aware_token_path() -> PathBuf {
+    let env_aware_token_path =
+        std::env::var("KANIDM_TOKEN_PATH").unwrap_or_else(|_| TOKEN_PATH.to_owned());
+
+    return PathBuf::from(shellexpand::tilde(&env_aware_token_path).into_owned());
+}
 
 #[allow(clippy::result_unit_err)]
 pub fn read_tokens() -> Result<BTreeMap<String, String>, ()> {
-    let token_path = PathBuf::from(shellexpand::tilde(TOKEN_PATH).into_owned());
+    let token_path = get_env_aware_token_path();
     if !token_path.exists() {
         debug!(
             "Token cache file path {:?} does not exist, returning an empty token store.",
-            TOKEN_PATH
+            token_path
         );
         return Ok(BTreeMap::new());
     }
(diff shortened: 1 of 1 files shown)


yaleman:
I'd definitely rather this be a hidden option in the `clap` parser than something buried in that part of the code.

tumbl3w33d:
Then you would need to pass it explicitly on every execution, which is quite inconvenient. I'd see this being used with something like direenv to automatically pick up the env per execution environment.  It could be done in clap in addition, though. What is your motivation of hiding it?

I'm fine with picking up the vars in a different place than this here if you want to suggest one.

yaleman:
> Then you would need to pass it explicitly on every execution, which is quite inconvenient. 

Not at all, see similar example here for passing passwords in with an env var: https://github.com/kanidm/kanidm/blob/41aa315f23a4ae7705b06ed71013bc5014f0c8ee/tools/cli/src/opt/kanidm.rs#L535

Firstyear:
I think my question here is "what are you trying to achieve" @tumbl3w33d. I think this relates to your desire to do ansible-ish state based operations on the server. Are you trying to use the cli tool for scripting?

Firstyear:
Also as a gentle reminder to everyone, per https://github.com/kanidm/kanidm/blob/master/project_docs/RELEASE_AND_SUPPORT.md#api-stability "The CLI is not an API and can change with the interest of human interaction during any release." 

So I think we need to seriously consider a seperate tool for scripting rather than overloading the current cli tool, and likely we need to do that with scim in place since that's the best / closest representation to what people here want to achieve.

Firstyear:
@tumbl3w33d Also note, since you are syncing users from an external source you *should* consider a sync driver instead https://kanidm.github.io/kanidm/stable/sync/concepts.html so that the entries have attribute protection and reflection that they are externally controlled.

tumbl3w33d:
Nope, not this time. :)

I already have 3 environments (production, a test environment and now a dev setup) and I want to switch context with the CLI when working with them.

Firstyear:
> Nope, not this time. :)
> 
> I already have 3 environments (production, a test environment and now a dev setup) and I want to switch context with the CLI when working with them.

In git master there is an update so that the tokens are bound to the spn not name so you can have multiple environments here. 

Second, @yaleman has an "environment config" switcher, though I think we should consider having it in mainline so that people can have a config like:

```
[default]
uri = ...

[dev]
uri = ...

[stage]
uri = ...
```

Then you can select the env with kanidm -e dev <commands>.

I think that's a far better solution to your problem than allowing tokens in env.

tumbl3w33d:
Yeah, your proposal in combination with the clap-induced env var to select it would be great. So I am closing this.

I will go down the python module route for now and thanks for your hint to the sync functionality advantage. Didn't think that's relevant for my use case, but I understood now.

yaleman:
I use https://github.com/yaleman/kanidm-profiles as my profile-switcher, for reference.

Firstyear:
I think we should have a profile switcher internally in the kanidm tool, should we leave this open to achieve that? Can we open an issue for it?

PR #1876, opened 2023-07-17
Unix gid duplicate fix

### LINKED ISSUE #1875: When changing a unix group gid the old gid isn't removed, resulting in the group having 2 gids
### On a newly create dev server run the following: 
- kanidm group create demo_group --name idm_admin
- kanidm group posix set --name idm_admin demo_group
- kanidm group posix set --name idm_admin demo_group --gidnumber 2001

### I expected the following

Success adding POSIX configuration for group demo_group

version 1.1.0-beta.13-dev (1.1.0-beta.13-dev)

[kanidm(d) --version]
kanidmd version: kanidmd 1.1.0-beta.13-dev

5.19.0-46-generic #47~22.04.1-Ubuntu SMP PREEMPT_DYNAMIC Wed Jun 21 15:35:31 UTC 2 x86_64 x86_64 x86_64 GNU/Linux

Seba-T:
Fixes #1875

Checklist

- [x] cargo fmt has been run

### DIFF
--- a/server/core/src/actors/v1_write.rs
+++ b/server/core/src/actors/v1_write.rs
@@ -1108,6 +1108,7 @@ impl QueryServerWriteV1 {
             "class".into(),
             Value::new_class("posixgroup"),
         )))
+        .chain(iter::once(Some(Modify::Purged("gidnumber".into()))))
         .chain(iter::once(gx.gidnumber.map(|n| {
             Modify::Present("gidnumber".into(), Value::new_uint32(n))
         })))
--- a/server/core/src/actors/v1_write.rs
+++ b/server/core/src/actors/v1_write.rs
@@ -1,6 +1,5 @@
-use std::iter;
-use std::sync::Arc;
 use std::time::Duration;
+use std::{iter, sync::Arc};
 
 use kanidm_proto::v1::{
     AccountUnixExtend, CUIntentToken, CUSessionToken, CUStatus, CreateRequest, DeleteRequest,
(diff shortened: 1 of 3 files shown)


server/core/src/actors/v1_write.rs
```diff
@@ -1108,6 +1108,7 @@ impl QueryServerWriteV1 {
             "class".into(),
             Value::new_class("posixgroup"),
         )))
+        .chain(iter::once(Some(Modify::Purged("gidnumber".into()))))
```
Firstyear, line 1111: What happens if the user changes another attribute but not the gid here now? I think this purge needs to be associated only if a gid is actually set too.

server/testkit/tests/proto_v1_test.rs
```diff
@@ -532,6 +532,12 @@ async fn test_server_rest_posix_lifecycle(rsclient: KanidmClient) {
         .idm_group_unix_extend("posix_group", None)
         .await
         .unwrap();
+    // here we check that we can successfully change the gid without breaking anything
```
Firstyear, line 535: Change another attribute and check the gid stays the same ;)

Firstyear (changes requested):

server/core/src/actors/v1_write.rs
Seba-T: I think I should move it inside the map then

Firstyear: Or just restructure the code to make it clearer what the intent is.

server/testkit/tests/proto_v1_test.rs
Seba-T: what other attributes can I change besides the gid? Because in `GroupUnixExtend` I just see `pub gidnumber: Option<u32>,`

Firstyear: 🤔 good point. But if the person runs the command with a "None" in the gid it should be idempotent and not change the gid.

Seba-T: agreed, so I'll ensure that won't happen

Firstyear (approved):

PR #1880, opened 2023-07-19
1788 admin unix socket

### LINKED ISSUE #1788: Administrative unix domain socket
To assist with use cases like #1747 we need a way to access the kanidmd binary *while* it's running. Currently we have some limited provisioning from within the kanidmd binary but it's also disruptive since we need to stop the server, take the action, and then start it again.

Since root or same user access to kani is what is used to restart and perform these actions, we can also use this as unix domain socket auth to authorise system commands. This will allow us to make a localhost specific admin tool which can pull out secrets, reset admin credentials or similar as required for automations and administration. It's likely this will also make the setup process for new users easier too.

Firstyear:
Fixes #1788 - This adds an administrative unix socket to the main kanidmd daemon allowing tasks to be performed while the server is online. Currently this is limited to the recover-account workflow. The socket requires the calling user to be the same uid as the kanidmd process or root. 

Checklist


Firstyear:
@yaleman Should we add default config path to profilse to make the commands easier to use here?

tools/orca/src/ipa.rs
```diff
@@ -1,5 +1,5 @@
+use hashbrown::{HashMap, HashSet};
```
yaleman, line 1: Isn't this just a cheeky workaround for the `std::collections::HashMap` ban thing? The `hashbrown` crate says it's identical?

> Since Rust 1.36, this is now the HashMap implementation for the Rust standard library.

unix_integration/src/daemon.rs
```diff
@@ -729,12 +729,12 @@ async fn main() -> ExitCode {
                                             // all good!
                                         } else {
```
yaleman, line 730: Shouldn't this just be `if ucred.uid() != 0 { warn... continue}`

yaleman (approved):

yaleman:
> @yaleman Should we add default config path to profilse to make the commands easier to use here?

Yeah, I think there should be a default with the option to modify it, say:

- dev `../../examples/insecure_server.toml` (or `<the package root>/examples/insecure_server.toml` if we can pick that up)
- non-container `/etc/kanidm/kanidmd.toml`
- container `/data/kanidmd.toml`

tools/orca/src/ipa.rs
Firstyear: It uses a different hasher though in std lib. hashbrown uses ahash, but std uses siphash which is significantly slower. That's why hashbrown is preferred.

Firstyear:
> > @yaleman Should we add default config path to profilse to make the commands easier to use here?
> 
> Yeah, I think there should be a default with the option to modify it, say:
> 
> * dev `../../examples/insecure_server.toml` (or `<the package root>/examples/insecure_server.toml` if we can pick that up)
> * non-container `/etc/kanidm/kanidmd.toml`
> * container `/data/kanidmd.toml`

Done, added it to profiles, -c always overrides, and we fall back to the default. Added the defaults in line with what the book suggests for config paths.

yaleman (approved):

PR #1881, opened 2023-07-20
20230720 unix int modular

Firstyear:
This makes the Db cache slightly more modular, getting ready for it to be a dyn trait or more portable/exposed. cc @dmulder since this is of interest to you :) 

Next will probably be to make the client it's own trait too.

Checklist

### DIFF
--- a/unix_integration/src/daemon.rs
+++ b/unix_integration/src/daemon.rs
@@ -25,8 +25,9 @@ use clap::{Arg, ArgAction, Command};
 use futures::{SinkExt, StreamExt};
 use kanidm_client::KanidmClientBuilder;
 use kanidm_proto::constants::DEFAULT_CLIENT_CONFIG_PATH;
-use kanidm_unix_common::resolver::Resolver;
 use kanidm_unix_common::constants::DEFAULT_CONFIG_PATH;
+use kanidm_unix_common::db::Db;
+use kanidm_unix_common::resolver::Resolver;
 use kanidm_unix_common::unix_config::KanidmUnixdConfig;
 use kanidm_unix_common::unix_passwd::{parse_etc_group, parse_etc_passwd};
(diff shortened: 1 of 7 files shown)


unix_integration/src/daemon.rs
```diff
@@ -660,8 +661,16 @@ async fn main() -> ExitCode {
...
-                cfg.db_path.as_str(), // The sqlite db path
+            let db = match Db::new(cfg.db_path.as_str(), &cfg.tpm_policy) {
+                Ok(db) => db,
+                Err(_e) => {
+                    error!("Failed to create database");
+                    return ExitCode::FAILURE
+                }
+            };
+
+            let cl_inner = match Resolver::new(
```
yaleman, line 672: At some point we should probably just pass `db, rsclient, cfg.clone()` and deal with it internally 😄

yaleman (commented):
There's a *lot* of `.map_err(|_| ())?` with no logging/feedback in the calls for admins to find out where/why things are failing, this feels bad - but by the looks of it, it's not making it more of an issue than the current situation?

unix_integration/src/daemon.rs
Firstyear: Yes, that's the goal. Sorry there were some emails that @dmulder and I had that you weren't cc'ed on. I'll forward you the goals - docs. But that's the plan :)

Firstyear:
> There's a _lot_ of `.map_err(|_| ())?` with no logging/feedback in the calls for admins to find out where/why things are failing, this feels bad - but by the looks of it, it's not making it more of an issue than the current situation?

Yes, this is going to be improved here too. Here we actually log in the lower level so the map errs here are just to allow bubbling up but the errors are logged at a lower level.

So this is on the todo :)

Firstyear:
Updated cargo.lock, rebased to master.

PR #1928, opened 2023-08-01
added compression layer for the pkg route

### LINKED ISSUE #1827: precompute + deliver brotli compressed WASM package
- **Is your feature request related to a problem? Please describe.**

A bit of a skunkworks idea - but we can precompress the WASM package and save 1/3 of the size, delivering it if they send an `Accept-Encoding` header 
which includes `br`.

- Include a build step that uses `brotli` to compress the WASM package (testing shows this takes ~2 seconds and reduces the release package from 1.4MB to 
400KB)
- For requests to the WASM package, check for the header and return a different file

Enabling the `brotli` feature on `axum-compress`, but this introduces *significant* response time delays as it's doing it on the fly.

**Additional context**

2/3 less bytes, 50% more snappy. :crab:

Seba-T:
Fixes #1827 

Checklist

- [x] cargo fmt has been run

### DIFF
--- a/server/core/src/https/mod.rs
+++ b/server/core/src/https/mod.rs
@@ -207,7 +207,6 @@ pub async fn create_https_server(
         }
         ServerRole::WriteReplicaNoUI => Router::new(),
     };
-
     let app = Router::new()
         .route("/robots.txt", get(robots_txt))
         .route("/status", get(status))
@@ -226,7 +225,10 @@ pub async fn create_https_server(
                 );
                 std::process::exit(1);
             }
-            app.nest_service("/pkg", ServeDir::new(pkg_path))
+            let pkg_router = Router::new()
+                .nest_service("/pkg", ServeDir::new(pkg_path).precompressed_br())
+                .layer(middleware::compression::new());
+            app.merge(pkg_router)
         }
     };
 


Seba-T:
Basically this works only because of this: 'Responses that are already compressed (ie have a content-encoding header) will never be recompressed, regardless what they predicate says.' (https://docs.rs/tower-http/latest/tower_http/compression/struct.Compression.html#method.compress_when). So for each request in the '/pkg' route first we try to find an already compressed file, if that's successful then the content-encoding header is set and no other middleware will try to compress that request. Otherwise if no pre compressed file is found the service will return the plain file and the subsequent compression middleware will actually compress it.

server/core/src/https/mod.rs
```diff
@@ -226,7 +225,10 @@ pub async fn create_https_server(
                 );
                 std::process::exit(1);
             }
-            app.nest_service("/pkg", ServeDir::new(pkg_path))
+            let pkg_router = Router::new()
+                .nest_service("/pkg", ServeDir::new(pkg_path).precompressed_br())
```
yaleman, line 229: where's `precompressed_br()` coming from? 🤔

PR #1930, opened 2023-08-01
fixed serialization of oauth2 token scope

### LINKED ISSUE #1929: Refresh token scope list format
### I did this

I send a `POST` to `/oauth2/token` to refresh the access token. The payload looks like this:
```python
{'grant_type': 'refresh_token', 'scope': 'admin email profile displayname preferred_username', 'refresh_token': 'someverylongtoken=='}
```

### I expected the following

A refreshed token to be returned, but I got a 400 and the server log
```
[error]: atr parse error - invalid type: string "admin email profile displayname preferred_username", expected a sequence
```

Am I using the wrong endpoint? I didn't find any other in the documentation. This format seems correct according to spec.

kanidmd 1.1.0-alpha.12

official container

Seba-T:
Fixes #1929 

Checklist

- [x] cargo fmt has been run

### DIFF
--- a/proto/src/oauth2.rs
+++ b/proto/src/oauth2.rs
@@ -2,6 +2,8 @@ use std::collections::{BTreeMap, BTreeSet};
 
 use base64urlsafedata::Base64UrlSafeData;
 use serde::{Deserialize, Serialize};
+use serde_with::formats::SpaceSeparator;
+use serde_with::{serde_as, skip_serializing_none, StringWithSeparator};
 use url::Url;
 
 #[derive(Serialize, Deserialize, Debug, PartialEq, Eq, Clone, Copy)]
@@ -77,6 +79,9 @@ pub enum AuthorisationResponse {
     Permitted,
 }
 
+#[serde_as]
+#[skip_serializing_none]
+// this is the equivalent of serde(skip_serializing_if = "Option::is_none") applied to ALL the options
 #[derive(Serialize, Deserialize, Debug)]
 #[serde(tag = "grant_type", rename_all = "snake_case")]
 pub enum GrantTypeReq {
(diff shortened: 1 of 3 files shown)


Seba-T:
@Firstyear I'm 99% sure we were in the wrong with the syntax we were using, but this is the 'only' thing I found to back it up: https://datatracker.ietf.org/doc/html/rfc6749#section-3.3. The reason I'm not 100% sure is that it feels a bit unlikely that no one stumbled across this issue until now, both the devs and anyone using kani in general.

tumbl3w33d:
In the introspection response you do it [right](https://datatracker.ietf.org/doc/html/rfc7662#section-2.2), btw. That's where I found the space separation first and was confused as well, because I did not expect that. I was talking to kani's API using python's [oauthlib](https://pypi.org/project/oauthlib/) (via flask-dance) and they send the field space separated for refresh_token().

Seba-T:
Yep, I'm seeing now that for the introspection we're doing  `str_join(&scopes)` which indeed joins all the different scopes in a single space separated string. Then I believe no one ever thought about this formatting issue for the refresh token response so this is a legit fix.

proto/src/oauth2.rs
```diff
@@ -77,6 +79,9 @@ pub enum AuthorisationResponse {
     Permitted,
 }
 
+#[serde_as]
+#[skip_serializing_none]
+// this is the equivalent of serde(skip_serializing_if = "Option::is_none") applied to ALL the options
```
Firstyear, line 84: That's actually cool.

Firstyear (approved):

proto/src/oauth2.rs
Seba-T: Yeah I thought about updating all our existing `serde(skip_serialize_if = "Option::is_none)`, but I think that would be better places in its own pr.

PR #1941, opened 2023-08-05
Allow one-character usernames

jcrawfordor:
Previously, usernames had to be at least two characters because of this regex. I'm not aware of any issues around one-character usernames and happen to have one in my environment. This seems fine, e.g. adduser uses a validation regex that allows them (but I'd be happy to hear if there's a reason not to allow them). This just changes the regex to make a single letter a valid username.

Checklist

- [X] cargo fmt has been run

### DIFF
--- a/server/lib/src/value.rs
+++ b/server/lib/src/value.rs
@@ -53,7 +53,7 @@ lazy_static! {
     /// Only lowercase+numbers, with limited chars.
     pub static ref INAME_RE: Regex = {
         #[allow(clippy::expect_used)]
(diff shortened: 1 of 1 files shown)


yaleman:
I don't know of a historical reason for a two-char limit, I'll keep this one pending @Firstyear's double-check as he's got a lot more IDP/auth history.

yaleman:
Checked with @Firstyear, it was partially related to indexing being slower for single-graphemes, but there's no (current) requirement for it, so I'll merge this.

PR #1957, opened 2023-08-08
Add tests for X-Forwarded-For header (kinda)

### LINKED ISSUE #1818: Add tests for X-Forwarded-For header things
Integration tests, need to test for the following:

- trust_xff is on, sending header
- trust_xff is on, don't send header
- the above two, but trust_xff off

Seba-T:
Fixes #1818 

This one confused me quite a bit as I couldn't figure out what the expected behavior should be 

Checklist

- [x] cargo fmt has been run
- [x] cargo test has been run and passes (kind of)

### DIFF
--- a/proto/src/v1.rs
+++ b/proto/src/v1.rs
@@ -6,7 +6,6 @@ use serde_with::skip_serializing_none;
 use std::cmp::Ordering;
 use std::collections::{BTreeMap, BTreeSet};
 use std::fmt;
-use std::net::IpAddr;
 use std::str::FromStr;
 use time::OffsetDateTime;
 use url::Url;
@@ -1196,18 +1195,6 @@ impl SingleStringRequest {
     }
 }
 // Use OperationResponse here ...
-
-#[cfg(debug_assertions)]
-#[derive(Serialize, Deserialize, Debug)]
-pub struct DebugIpInfoResponse {
-    pub your_ip: IpAddr,
-}
-#[cfg(debug_assertions)]
-impl DebugIpInfoResponse {
-    pub fn new(your_ip: IpAddr) -> Self {
-        DebugIpInfoResponse { your_ip }
-    }
-}
 
 #[cfg(test)]
 mod tests {
--- a/proto/src/v1.rs
+++ b/proto/src/v1.rs
@@ -6,6 +6,7 @@ use serde_with::skip_serializing_none;
 use std::cmp::Ordering;
 use std::collections::{BTreeMap, BTreeSet};
 use std::fmt;
+use std::net::IpAddr;
 use std::str::FromStr;
 use time::OffsetDateTime;
 use url::Url;
@@ -1195,6 +1196,18 @@ impl SingleStringRequest {
     }
 }
 // Use OperationResponse here ...
+
+#[cfg(debug_assertions)]
+#[derive(Serialize, Deserialize, Debug)]
+pub struct DebugIpInfoResponse {
+    pub your_ip: IpAddr,
+}
+#[cfg(debug_assertions)]
+impl DebugIpInfoResponse {
+    pub fn new(your_ip: IpAddr) -> Self {
+        DebugIpInfoResponse { your_ip }
+    }
+}
 
 #[cfg(test)]
 mod tests {
--- a/server/core/src/https/v1.rs
+++ b/server/core/src/https/v1.rs
@@ -1425,9 +1425,18 @@ pub async fn auth_valid(
     to_axum_response(res)
 }
 
+#[cfg(debug_assertions)]
+pub async fn debug_ipinfo(
+    State(state): State<ServerState>,
+    TrustedClientIp(ip_addr): TrustedClientIp,
+) -> impl IntoResponse {
+    let res = state.qe_r_ref.handle_debug_ipinfo(ip_addr);
+    to_axum_response(Ok(res))
+}
+
 #[instrument(skip(state))]
 pub fn router(state: ServerState) -> Router<ServerState> {
-    Router::new()
+    let mut router = Router::new()
         .route("/v1/oauth2", get(super::oauth2::oauth2_get))
         .route("/v1/oauth2/_basic", post(super::oauth2::oauth2_basic_post))
         .route(
@@ -1728,5 +1737,9 @@ pub fn router(state: ServerState) -> Router<ServerState> {
             post(sync_account_token_post).delete(sync_account_token_delete),
         )
         .with_state(state)
-        .layer(from_fn(dont_cache_me))
+        .layer(from_fn(dont_cache_me));
+    if cfg!(debug_assertions) {
+        router = router.route("/v1/debug/ipinfo", get(debug_ipinfo));
+    }
+    router
 }
--- a/server/core/src/https/v1.rs
+++ b/server/core/src/https/v1.rs
@@ -1427,11 +1427,10 @@ pub async fn auth_valid(
 
 #[cfg(debug_assertions)]
 pub async fn debug_ipinfo(
-    State(state): State<ServerState>,
+    State(_state): State<ServerState>,
     TrustedClientIp(ip_addr): TrustedClientIp,
 ) -> impl IntoResponse {
-    let res = state.qe_r_ref.handle_debug_ipinfo(ip_addr);
-    to_axum_response(Ok(res))
+    to_axum_response(Ok(vec![ip_addr]))
 }
 
 #[instrument(skip(state))]
--- a/server/testkit/tests/https_extractors.rs
+++ b/server/testkit/tests/https_extractors.rs
@@ -2,8 +2,6 @@ use kanidm_client::{KanidmClient, APPLICATION_JSON};
 use kanidm_proto::v1::{AuthIssueSession, AuthRequest, AuthStep};
 use reqwest::header::CONTENT_TYPE;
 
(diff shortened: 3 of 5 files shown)


Seba-T:
I think I broke git or something

Seba-T:
Never mind I just didn't staged the deleted file :smiling_face_with_tear:

server/testkit/tests/https_extractors.rs
```diff
@@ -0,0 +1,91 @@
...
+        .header(CONTENT_TYPE, APPLICATION_JSON)
+        .body(serde_json::to_string(&auth_init).unwrap())
+        .send()
+        .await
+        .unwrap();
+    assert!(dbg!(response).status() == 200);
+}
+
+#[kanidmd_testkit::test(trust_x_forward_for = true)]
+async fn dont_trust_xff_dont_send_header(rsclient: KanidmClient) {
```
yaleman, line 35: the naming here is weird, you've got `trust_x_forward_for = true` and called it `dont_trust_xff` - shouldn't it be `trust_xff`?

server/testkit/tests/https_extractors.rs
```diff
@@ -0,0 +1,91 @@
...
+        .header(
+            "X-Forwarded-For",
+            "a VERY much invalid header that WON'T get through!!",
+        )
+        .header(CONTENT_TYPE, APPLICATION_JSON)
+        .body(serde_json::to_string(&auth_init).unwrap())
+        .send()
+        .await
+        .unwrap();
+    assert!(dbg!(response).status() == 400);
```
yaleman, line 75: we should also be testing the "trust header, send valid header" cases with
- single IPv4 and IPv6 addresses
- combined lists of the above to test the "leftmost" address. ie, (IPv4, IPv4), (IPv4, IPv6), (IPv6, IPv6), (IPv6, IPv4)

yaleman (changes requested):
Looks good, a naming weirdness and suggestion for additional tests - I'm happy to write the tests if you don't want to :)

server/testkit/tests/https_extractors.rs
Seba-T: yep you're right :)

server/testkit/tests/https_extractors.rs
Seba-T: Sounds good :). btw I've noticed we reject requests only if the leftmost address is invalid, is this expected?

yaleman: Since we're only trusting the left-most address, then ... it seems OK? We probably *should* check all of them to ensure it's a "valid" statement.

Firstyear: We might have to end up logging the other right side addresses too, so probably we should check these :)

Firstyear (commented):
I wonder if what we should have here is when cfg(debug_assertions) are present, we have an extra route that reflects the result of xff so that these tests can actually see/test what was decoded by the server. Heck it might even be good to have an ip reflector in general when the user is authorised? What do you think @yaleman ?

yaleman:
> I wonder if what we should have here is when cfg(debug_assertions) are present, we have an extra route that reflects the result of xff so that these tests can actually see/test what was decoded by the server. Heck it might even be good to have an ip reflector in general when the user is authorised? What do you think @yaleman ?

A set of debug route paths isn't a bad idea, something like `/v1/debug/ipinfo`

We could probably do the IP reflector thing on the `whoami` endpoint, either just adding the data or requiring a [Prefer](https://httpwg.org/specs/rfc7240.html) header with the value `ipinfo=true` or similar?

Firstyear:
Less if/else is betterer. I'd just return the info always in the whoami response.

Seba-T:
> Less if/else is betterer. I'd just return the info always in the whoami response.

Should I make a separate pr to include this info?

yaleman:
> > Less if/else is betterer. I'd just return the info always in the whoami response.
> 
> Should I make a separate pr to include this info?

I'd say so - keep this PR to what it's doing now

Seba-T:
> > > Less if/else is betterer. I'd just return the info always in the whoami response.
> > 
> > 
> > Should I make a separate pr to include this info?
> 
> I'd say so - keep this PR to what it's doing now

Sounds good :) I think I got somewhat lost, should I add anything else to this pr right now? Or should I get to work on the one that provides the ip and then introduce new tests here after that one gets merged?

Firstyear:
Add the ip reflector then come back to this :)

Seba-T:
> Add the ip reflector then come back to this :)
> 
> EDIT: Actually I changed my mind. The ip reflector is really bound to this PR/issue since we need the reflector for these tests, and we need these tests to test the reflector. So maybe they sohuld be the same PR ?

Yeah that's what I was also thinking, but I didn't want to mix up too many things. I can do something like getting the reflector working just for whoami in this pr, and then in another pr do the debug endpoint

Firstyear:
I think if whoami has a reflector then we don't need a debug endpoint?

yaleman:
> I think if whoami has a reflector then we don't need a debug endpoint?

Debug's handy for debug things and doesn't need auth etc, but .... maybe? 😄

Seba-T:
I was trying to get this to work, but I quickly realized that whoami just returns an entry in the response body, so I figured I should provide the ip in some sort of response header, not in the response body, right?

yaleman:
> I was trying to get this to work, but I quickly realized that whoami just returns an entry in the response body, so I figured I should provide the ip in some sort of response header, not in the response body, right?

You can modify the entry and add it to the response, we should be doing better response typing as it is 😄

Firstyear:
Or we rewrite the whoami response type to actually be structured rather than an Entry ... which sounds like the better plan.

Seba-T:
> Or we rewrite the whoami response type to actually be structured rather than an Entry ... which sounds like the better plan.

Yeah I agree, but on a side note getting the reflector + tests to work on the whoami endpoint is getting exponentially harder, both because now I'd have to change the whoami endpoint itself, and also because including the X-Forwarded-For header in the requests without having to rewrite the entire client lib is far from trivial. As far as this pr is concerned I would implement a quick debug endpoint that requires no authentication so it's way easier to test it, and then since the whoami would use the same ip extractor I don't think we would specifically need tests on that extractor too

Firstyear:
I think changing the output of the whoami struct is less work than you think it will be :)

Firstyear:
Regardless, for now lets add the debug ip reflector then I guess?

Seba-T:
> I think changing the output of the whoami struct is less work than you think it will be :)

No absolutely it’s just that I thought it would need some back and forth feedback and with the time zone difference I wasn’t sure I had the time, but probably I kinda overthought it

Firstyear:
And remember if you start and can't finish it, I can take over the PR too :)

Seba-T:
I mean thinking about it it should probably be just a struct right?

Seba-T:
> And remember if you start and can't finish it, I can take over the PR too :)

I mean I don’t mean to leave it hanging, it was just about doing it before the gsoc ends, but I’d be fine even if we didn’t include it

Seba-T:
After fighting against the whoami stuff for a bit I gave up and did this

proto/src/v1.rs
```diff
@@ -1196,6 +1197,18 @@ impl SingleStringRequest {
 }
 // Use OperationResponse here ...
 
+#[cfg(debug_assertions)]
+#[derive(Serialize, Deserialize, Debug)]
+pub struct DebugIpInfoResponse {
+    pub your_ip: IpAddr,
+}
```
Firstyear, line 1204: Spaces :)

server/core/src/https/v1.rs
```diff
@@ -1425,9 +1425,18 @@ pub async fn auth_valid(
     to_axum_response(res)
 }
 
+#[cfg(debug_assertions)]
+pub async fn debug_ipinfo(
+    State(state): State<ServerState>,
+    TrustedClientIp(ip_addr): TrustedClientIp,
+) -> impl IntoResponse {
+    let res = state.qe_r_ref.handle_debug_ipinfo(ip_addr);
```
Firstyear, line 1433: Could just skip the qe_r_ref here since this doesn't do anything to the query server, it's axum-level only. Makes the code simpler too.

proto/src/v1.rs
```diff
@@ -1196,6 +1197,18 @@ impl SingleStringRequest {
...
+#[cfg(debug_assertions)]
+impl DebugIpInfoResponse {
+    pub fn new(your_ip: IpAddr) -> Self {
+        DebugIpInfoResponse { your_ip }
```
Firstyear, line 1208: I'd probably make this simple a Vec of IpAddr rather than a custom struct here. Saves some extra struct defs.

Firstyear (changes requested):

server/core/src/https/v1.rs
Seba-T: wopsies :)

proto/src/v1.rs
Seba-T: I mean since it's debugging stuff I couldn't agree more

Firstyear: Well also because xff is a vector of addresses :P

Firstyear (approved):

Firstyear:
Up to you @yaleman :)

PR #1964, opened 2023-08-09
Fighting with zypper, tagging our images

yaleman:
* Zypper loves to break on ARM builds, so I'm adding a script to do more checking and fixing.
* Tagging containers with 
  * `com.kanidm.git-commit` to show when they were built.
  * `com.kanidm.version` to show the current version as far as the build's  concerned. 

Checklist

- [x] cargo fmt has been run

### DIFF
--- a/scripts/zypper_fixing.sh
+++ b/scripts/zypper_fixing.sh
@@ -1,0 +1,20 @@
+#!/bin/bash
+
+# makes sure the repos are configured because the containers are derpy sometimes
+
+#disable the openh264 repo
+if [ "$(zypper lr | grep -ci 'repo-openh264')" -eq 1 ]; then
+    zypper mr -d -f -n 'repo-openh264'
+fi
+
+# add the non-oss repo if it doesn't exist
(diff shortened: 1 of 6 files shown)


scripts/zypper_fixing.sh
```diff
@@ -0,0 +1,20 @@
...
+if [ "$(zypper lr | grep -c 'repo-non-oss')" -eq 0 ]; then
+    zypper ar -f -n 'Non-OSS' http://download.opensuse.org/tumbleweed/repo/non-oss/ repo-non-oss
+fi
+
+# update the repos and make sure the ones we want are enabled
+zypper mr -k repo-oss
+zypper mr -k repo-non-oss
+zypper mr -k repo-update
+zypper ref --force
+zypper -v dup -y
```
Firstyear, line 20: comment - force the refresh because zypper is too silly to work out it needs to do it itself.
zypper -v on the dup to show which mirror is failing if an error occurs (otherwise zypper shows the wrong mirror url)

Firstyear (approved):
Minor request to add some extra comments but looks great otherwise.

PR #1967, opened 2023-08-11
Are we JSON yet? Kinda. But we're closer.

### LINKED ISSUE #1966: Missing content-type header in token response
### I did this

Tried to connect an application to kanidm via OIDC.

### I expected the following

To be able to login to the application.

### What actually happened

The application threw an error and some digging turned out that a nullpointer exception happened which was caused by a missing content-type header in the token response from kanidm. [It wants to check](https://github.com/elastic/elasticsearch/blob/main/x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/oidc/OpenIdConnectAuthenticator.java#L621) that to refuse working with anything that's not `application/json`.

`kanidmd 1.1.0-beta.13`

official container

### Any other comments

I checked which headers are included in the kanidm response:
```
access-control-allow-origin, content-length, cache-control, pragma, content-security-policy, x-content-type-options, permissions-policy, referrer-policy, x-kanidm-version, strict-transport-security, x-kanidm-opid, date
```

From Matrix I know that @yaleman also has an issue with an OIDC connection with a different app and this could be related.

yaleman:
Fixes #1966

- adds checks when running in debug mode (and returning a success response) to ensure we're sending a JSON content-type header
- (generally) sends the header as part of `to_axum_response`
- some added attempts to use more typed responses instead of just serializing the body and making a response object

Checklist

- [x] cargo fmt has been run


tumbl3w33d (approved):
👏

It's a shame clients don't come with is_success, is_client_error, is_server_error on the response object and make users do such complicated math. 🤓 I once used such client, but don't even remember what language it was in, nor what lib it was.

yaleman:
> 👏
> It's a shame clients don't come with is_success, is_client_error, is_server_error on the response object and make users do such complicated math. 🤓 I once used such client, but don't even remember what language it was in, nor what lib it was.

Python's requests has a `response.raise_for_status()` which throws an exception if it's not a 200.. I wouldn't want an is_xxx for all of the ... MANY status codes there are - and that's ignoring the whole other class of errors that is a failed HTTP conversation 😄 

You *can* do the following if you wanted to go through the variations on codes:

```rust
match response.status() as http::status::StatusCode {
    http::status::StatusCode::OK => {}
    http::status::StatusCode::TEMPORARY_REDIRECT => {} 
...etc etc
}
```

It's pretty simple code, "if it's a 2xx status, then check for the JSON header". I'm only limiting the check to that because I haven't verified that all API responses should be JSON - otherwise I'd have just done it based on URI.

Firstyear:
I'm extremely keen to review this on monday, y'all been busy :)

yaleman:
> You are usually not interested if it's a 200, a 201 or what kind of special redirect the SEO specialist decided on this week, so you can leave your code untouched if all you want to know is, whether the operation was a success, your client is broken or the server currently has a problem.

That's what `reqwest` does with [error_for_status](https://docs.rs/reqwest/0.11.18/reqwest/struct.Response.html#method.error_for_status). I can't say I've used it, because if I'm checking status', I care what the specifics are. 😄

tumbl3w33d:
Ah right, [that's](https://docs.rs/reqwest/0.11.18/reqwest/struct.StatusCode.html#method.is_success) where I know it from!

libs/client/src/lib.rs
```diff
@@ -24,6 +24,7 @@ use std::os::unix::fs::MetadataExt;
 use std::path::Path;
 use std::time::Duration;
 
+use kanidm_proto::constants::APPLICATION_JSON;
```
Firstyear, line 27: Is there not a version of this header in reqwest?

libs/client/src/lib.rs
```diff
@@ -621,7 +621,7 @@ impl KanidmClient {
             .map_err(|e| ClientError::JsonDecode(e, opid))
     }
 
-    async fn perform_post_request<R: Serialize, T: DeserializeOwned>(
+    pub async fn perform_post_request<R: Serialize, T: DeserializeOwned>(
```
Firstyear, line 624: Why make this public? Is there a client that wants to consume the wrapper directly?

server/core/src/https/middleware/mod.rs
```diff
@@ -31,6 +34,41 @@ pub struct KOpId {
...
+#[test]
+fn test_from_200_to_299() {
+    assert!(from_200_to_299(http::StatusCode::OK));
+    assert!(from_200_to_299(http::StatusCode::IM_USED));
+    assert!(!from_200_to_299(http::StatusCode::BAD_REQUEST));
+    assert!(!from_200_to_299(http::StatusCode::INTERNAL_SERVER_ERROR));
+}
+
+#[cfg(debug_assertions)]
+/// This is a debug middleware to ensure that /v1/ endpoints only return JSON
```
Firstyear, line 52: Oooo I really like this idea too. Good stuff !

server/core/src/https/middleware/security_headers.rs
```diff
@@ -29,29 +29,24 @@ pub async fn security_headers_layer<B>(
     #[allow(clippy::expect_used)]
```
Firstyear, line 29: can drop the expect used now.

Firstyear (approved):

libs/client/src/lib.rs
yaleman: Nope, I looked around trying to find it 

* there's one in `mime` since it's a header value rather than a header key, but we don't currently use this lib, and it's a bit weird from memory
* I checked we were already building `kanidm_proto` in the libs where I added this, so  I'm pretty sure it's not adding to the mess much

server/core/src/https/middleware/mod.rs
yaleman: I was like "how do I write tests for this... oh, let's just make it part of the routing layer" :smile:

server/core/src/https/middleware/security_headers.rs
yaleman: Done!

libs/client/src/lib.rs
yaleman: Yeah, I'm using it to test sending bad things (with embedded auth) over here: https://github.com/yaleman/kanidm/blob/6a8361d8d520e73753af2fe944af7393a88b42d5/server/testkit/tests/proto_v1_test.rs#L644 (and similar tests).

PR #1970, opened 2023-08-11
Himmelblau requires a mechanism to send pam_prompt

dmulder:
In Himmelblau I need a mechanism for sending an additional pam prompt (for performing 2FA via a browser window on another device). I've implemented that here by providing a pam continue feature, which sends a message with some optional data back to pam for an additional prompt. The purpose of the additional data is simply for continuing the authentication later (in my case, it's a json string).

Fixes #

Checklist

- [ ] cargo fmt has been run

### DIFF
--- a/unix_integration/src/idprovider/kanidm.rs
+++ b/unix_integration/src/idprovider/kanidm.rs
@@ -201,6 +201,22 @@ impl IdProvider for KanidmProvider {
         }
     }
 
+    async fn unix_user_authenticate_continue(
+        &self,
+        id: &Id,
+        resp: &str,
+        data: Option<String>,
+    ) -> Result<Option<UserToken>, IdpError> {
+        Err(IdpError::BadRequest)
+    }
+
+    async fn unix_user_authenticate_initiate_continue(
+        &self,
+        id: &Id,
+    ) -> Result<(Option<String>, Option<String>), IdpError> {
+        Err(IdpError::BadRequest)
+    }
+
     async fn unix_group_get(&self, id: &Id) -> Result<GroupToken, IdpError> {
         match self
             .client
--- a/unix_integration/pam_kanidm/src/pam/mod.rs
+++ b/unix_integration/pam_kanidm/src/pam/mod.rs
@@ -241,7 +241,7 @@ impl PamHooks for PamKanidm {
             Ok(cfg) => cfg,
             Err(e) => return e,
         };
-        let req = ClientRequest::PamAuthenticate(account_id, authtok);
+        let req = ClientRequest::PamAuthenticate(account_id.clone(), authtok);
 
         match call_daemon_blocking(cfg.sock_path.as_str(), &req, cfg.unix_sock_timeout) {
             Ok(r) => match r {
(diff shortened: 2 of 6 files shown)


dmulder:
FYI, I've updated my distribution, done a cargo clean, even tried a fresh check out of kanidm, but I still get this build error:
```
error[E0599]: no associated item named `NO_CHECK_TIME` found for struct `X509VerifyFlags` in the current scope
    --> /home/dmulder/.cargo/registry/src/index.crates.io-6f17d22bba15001f/webauthn-rs-core-0.4.9/src/attestation.rs:1277:49
     |
1277 |             .set_flags(verify::X509VerifyFlags::NO_CHECK_TIME)
     |                                                 ^^^^^^^^^^^^^
     |                                                 |
     |                                                 associated item not found in `X509VerifyFlags`
     |                                                 help: there is an associated constant with a similar name: `USE_CHECK_TIME`
```
I wonder perhaps if this flag has changed in a recent version of openssl (I have 3.1.2)?

yaleman:
Looks like the build issue relates to the version of openssl on your machine, older versions of the `openssl-sys` crate (which is a dependency of `webauthn-rs` don't support openssl3.x - I'll see if I can update that without major changes.

yaleman:
I've updated webauthn-rs' deps in https://github.com/kanidm/webauthn-rs/pull/340, which will solve that issue up/side-stream 😄

unix_integration/src/idprovider/kanidm.rs
```diff
@@ -201,6 +201,22 @@ impl IdProvider for KanidmProvider {
         }
     }
 
+    async fn unix_user_authenticate_continue(
+        &self,
+        id: &Id,
+        resp: &str,
+        data: Option<String>,
```
yaleman, line 208: If they're unused you have to mark them as such with underscores - build blocker.
```suggestion
        _id: &Id,
        _resp: &str,
        _data: Option<String>,
```

unix_integration/src/idprovider/kanidm.rs
```diff
@@ -201,6 +201,22 @@ impl IdProvider for KanidmProvider {
...
+    ) -> Result<Option<UserToken>, IdpError> {
+        Err(IdpError::BadRequest)
+    }
+
+    async fn unix_user_authenticate_initiate_continue(
+        &self,
+        id: &Id,
```
yaleman, line 215:
```suggestion
        _id: &Id,
```

yaleman (changes requested):
Ooh, additional prompts from PAM? I like this 😈

Firstyear (changes requested):
Generally I think this is going to need a bit of a redesign to be more flexible. We need to have a clientAuthSession associated to the pam-unix socket that is created after a pam-auth-init which contains the context of the auth, and if there are optionally cache auth mechanisms we can proceed with. We need this for future stepped / challenge response auth to be more flexible and then we can just have the auth state machine drive all the steps here. This way we can also have the pam client send the pam-auth-init and then have an inner event loop that calls on the daemon and waits for responses and does what it needs.

dmulder:
> Generally I think this is going to need a bit of a redesign to be more flexible. We need to have a clientAuthSession associated to the pam-unix socket that is created after a pam-auth-init which contains the context of the auth, and if there are optionally cache auth mechanisms we can proceed with. We need this for future stepped / challenge response auth to be more flexible and then we can just have the auth state machine drive all the steps here. This way we can also have the pam client send the pam-auth-init and then have an inner event loop that calls on the daemon and waits for responses and does what it needs.

Sounds good to me. I figured the best way to start a conversation was to hack something up :-)
Perhaps we can talk more about this in our Wednesday meeting.

dmulder:
I also realized I need to support other message types (specifically PAM_TEXT_INFO), so this MR needs work.

dmulder:
I also need to be able to modify the timeout for the call to the daemon. My code is displaying a message requesting the user authenticate with MFA via a different device (such as a cell). The daemon is polling for the response. The default 5 second timeout isn't sufficient (in fact, the user has 15 minutes to complete the auth).

Firstyear:
Sounds good. I think that we do need some testing/planning and thought about this, because in future when we add ctap support we'll need to be able to handle this too. So lets chat about it tomorrow and get a clearer design and go from there :)

PR #1971, opened 2023-08-13
Struct-ifying schema things

yaleman:
I can't remember how I started doing this but I'm sure I was avoiding something more important...

* Rebuilt most of the schema things as structs instead of JSON string blobs
* Added build-time validation that we're not re-using UUIDs

Checklist

- [x] cargo fmt has been run


server/lib/build.rs
```diff
@@ -12,4 +14,27 @@ fn main() {
     }
 
     profiles::apply_profile();
+
+    // check we don't have duplicate UUIDs
```
Firstyear, line 18: This is clever, I really like this!

Firstyear (approved):
This is fantastic! Now how I would have approached it but still an excellent improvement and I really like what you've done here :)

yaleman:
> This is fantastic! Now how I would have approached it but still an excellent improvement and I really like what you've done here :)

I was pondering last night making `Schema` like this:

```rust
enum SchemaType {
    SchemaAttribute,
    SchemaClass,
}


struct Schema {
    uuid: Uuid,
    name: &'static str,
    description: &'static str
    schema_type: SchemaType,
    attrs: SchemaAttrs,
}
```

Which would allow us to automagically do the `uuid/name/description/class` fields and set things up... how would you have done it? I'm happy to rewrite it again, this was just a mad-lad first rewrite.

Seba-T:
Wow this looks marvelous! it feels like living 10 yrs into the future :p

Firstyear:
> Which would allow us to automagically do the `uuid/name/description/class` fields and set things up... how would you have done it? I'm happy to rewrite it again, this was just a mad-lad first rewrite.

I probably would have just used the entry_init! macro, but I think this is a bit nicer/structured. :)

Firstyear:
> I was pondering last night making `Schema` like this:

This is a good idea, then you can have the enum for attr/class have the specific bits for those types.

yaleman:
> > I was pondering last night making `Schema` like this:
> 
> This is a good idea, then you can have the enum for attr/class have the specific bits for those types.

Y'know what's even smarter... USING THE STRUCTS WE ALREADY HAD! (done)

PR #1988, opened 2023-08-16
sqlite - where IN for id entry

### LINKED ISSUE #258: sqlite where IN for id entry
In get identry raw we use a loop on the ids, which does a lot of selects. It would be better to use the "where in" functionality, but it's not clear how to use a tuple of ints in rusqlite.

yaleman:
Fixes #258

Checklist

- [x] cargo fmt has been run

### DIFF
--- a/server/lib/src/be/idl_sqlite.rs
+++ b/server/lib/src/be/idl_sqlite.rs
@@ -8,6 +8,7 @@ use std::time::Duration;
 use hashbrown::HashMap;
 use idlset::v2::IDLBitRange;
 use kanidm_proto::v1::{ConsistencyError, OperationError};
+use rusqlite::vtab::array::Array;
 use rusqlite::{Connection, OpenFlags, OptionalExtension};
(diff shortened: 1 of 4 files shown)


server/lib/src/be/idl_sqlite.rs
```diff
@@ -157,47 +158,31 @@ pub trait IdlSqliteTransaction {
                     .collect()
             }
             IdList::Partial(idli) | IdList::PartialThreshold(idli) | IdList::Indexed(idli) => {
+                let mut id_list: Vec<rusqlite::types::Value> = vec![];
+                for item in idli.into_iter() {
+                    id_list.push(rusqlite::types::Value::Integer(
+                        i64::try_from(item).map_err(|_| OperationError::InvalidEntryId)?,
```
Firstyear, line 164: I think we just need to make entry id i64 lol.

Firstyear (approved):
Looks really good, thanks for solving a super long standing mystery :)

PR #1989, opened 2023-08-16
Less human strings more enums

### LINKED ISSUE #1987: replace all string-class-references with consts
Raised by #1985 - it's easy to misspell things and easy to fix, replace things like "service_account" with a single central const. :)

yaleman:
Refers to #1987

This is one of MANY cycles this is going to take to get through it. Sorry it's so big, future ones will be in smol'er chunks.

- Replaces a *lot* of copy-pasta'd strings with central `const`s, starting to use `Enum`'s where it was easy to get them in, and directly referencing the const str's where it was trickier on this first run to string-replace
- There's Struct's for SchemaAttribute, BuiltinGroup, BuiltinUser, BuiltinAcp which have `from/into` for EntryInitNew.
  - I've made at least one of each of the new builtins as a test, and in one case I was able to replace the string name of the group in a test with a reference to its builtin name! 🥳 
- Changed `f_eq` to use the new Enum, which removed one place a lot of stringy things was being used - the plan should be to do this across more of the functions to remove that point of failure. 😄 

Checklist

- [x] cargo fmt has been run


Seba-T:
Damnn this provides such an ergonomic developer experience! What was the rationale behind using human strings to begin with tho?

Firstyear:
> Damnn this provides such an ergonomic developer experience! What was the rationale behind using human strings to begin with tho?

Lack of foresight. Regrets. Mistakes.

Firstyear:
There might be a way to avoid the to_str() calls, I think you can use as_ref() possibly inside the Value constructors so that the callsite just needs ValueAttribute::Class with no as_str calls

Firstyear:
Is it worth looking into AsRef and AsStr on fn's to make it easier to pass these around with less as_str()?

Also where are ValueAttribute defined?

yaleman:
> Is it worth looking into AsRef and AsStr on fn's to make it easier to pass these around with less as_str()?

the `as_str`'s are interim workarounds until I clean up some other things 😄 
 
> Also where are ValueAttribute defined?

it's there now, probably wasn't visible until I pushed the "working" bit tonight

yaleman:
[IDM_ADMINS_ACP_RECYCLE_SEARCH_V1](https://github.com/yaleman/kanidm/blob/c15b6ed333bdda12b9f227dbca99175110772d5f/server/lib/src/constants/acp.rs#L51-L70) is an example of where I'm starting to head with defining ACPs - there's two strings total (name, description) and the rest is strongly typed, repeatable elements.

If I can remove the `String`s from ProtoFilter we could get to removing `lazy_static` entirely, but that's a whole other thing.

I think this is the end of the current iteration of this PR, it's already HUGE and I don't want to get too far from `master`. 

It passes all the current tests including my manual by running the "setup dev environment" script, logging in and changing the password, so I'm *pretty* sure it's safe to merge.

yaleman:
At the moment the `.as_ref()`'s are there, but once we move to using the structs for  defining the objects instead of the `entry_init!()` macro, we can much easier remove/tweak them and the related signatures.

proto/src/constants.rs
```diff
@@ -6,3 +6,144 @@ pub const APPLICATION_JSON: &str = "application/json";
...
+pub const ATTR_UIDNUMBER: &str = "uidnumber";
+pub const ATTR_UNIQUE: &str = "unique";
+pub const ATTR_UNIX_PASSWORD: &str = "unix_password";
+pub const ATTR_USER_AUTH_TOKEN_SESSION: &str = "user_auth_token_session";
+pub const ATTR_USERID: &str = "userid";
+pub const ATTR_USERPASSWORD: &str = "userpassword";
+pub const ATTR_UUID: &str = "uuid";
+pub const ATTR_VERSION: &str = "version";
+
+pub const TEST_ATTR_NON_EXIST: &str = "non-exist";
```
Firstyear, line 123: Could these be cfg!(test) if they are test only?

proto/src/constants.rs
```diff
@@ -6,3 +6,144 @@ pub const APPLICATION_JSON: &str = "application/json";
...
+pub const TEST_ATTR_TEST_ATTR: &str = "testattr";
+pub const TEST_ATTR_EXTRA: &str = "extra";
+
+pub const CLASS_SOURCEUUID: &str = "source_uuid";
```
Firstyear, line 127: ATTR not CLASS

proto/src/constants.rs
```diff
@@ -6,3 +6,144 @@ pub const APPLICATION_JSON: &str = "application/json";
...
+pub const OAUTH2_SCOPE_GROUPS: &str = "groups";
+pub const OAUTH2_SCOPE_OPENID: &str = "openid";
+pub const OAUTH2_SCOPE_READ: &str = "read";
+pub const OAUTH2_SCOPE_SUPPLEMENT: &str = "supplement";
+
+pub const LDAP_ATTR_OU: &str = "ou";
+pub const LDAP_ATTR_CN: &str = "cn";
+pub const LDAP_ATTR_OBJECT_CLASS: &str = "objectClass";
+pub const LDAP_ENTRYDN: &str = "entrydn";
+pub const LDAP_MAIL: &str = "mail";
```
Firstyear, line 139: LDAP_ATTR_MAIL ? for consistency?

server/core/src/actors/v1_read.rs
```diff
@@ -744,7 +744,7 @@ impl QueryServerReadV1 {
                     // get the first entry
                     .and_then(|e| {
                         // From the entry, turn it into the value
-                        e.get_ava_iter_sshpubkeys("ssh_publickey")
+                        e.get_ava_iter_sshpubkeys(Attribute::SshUnderscorePublicKey.into())
```
Firstyear, line 747: Why spell out the Underscore here?

server/lib/src/constants/acp.rs
```diff
@@ -2,336 +2,475 @@
...
-            "acp_receiver_group",
+            ATTR_ACP_RECEIVER_GROUP,
             Value::Refer(UUID_IDM_PEOPLE_MANAGE_PRIV)
         ),
         (
-            "acp_targetscope",
+            ATTR_ACP_TARGET_SCOPE,
             Value::new_json_filter_s(
-                "{\"and\": [{\"eq\": [\"class\",\"account\"]}, {\"eq\": [\"class\",\"person\"]}, {\"andnot\": {\"or\": [{\"eq\": [\"memberof\",\"00000000-0000-0000-0000-000000001000\"]}, {\"eq\": [\"class\", \"tombstone\"]}, {\"eq\": [\"class\", \"recycled\"]}]}}]}"
+                r#"{"and": [
```
Firstyear, line 441: I'm a fan of this multi-line expansion.

server/lib/src/entry.rs
```diff
@@ -354,35 +354,42 @@ impl Entry<EntryInit, EntryNew> {
                 } else {
                 let attr = AttrString::from(k.to_lowercase());
                 let vv: ValueSet = match attr.as_str() {
-                    "attributename" | "classname" | "domain" => {
+                    kanidm_proto::constants::ATTR_ATTRIBUTENAME | kanidm_proto::constants::ATTR_CLASSNAME | kanidm_proto::constants::ATTR_DOMAIN => {
```
Firstyear, line 357: hey, I want to remove this function too, this one is awful.

server/core/src/actors/v1_read.rs
yaleman: because we have both "sshpublickey" and "ssh_publickey" and at one point I broke everything

server/lib/src/constants/acp.rs
yaleman: You'll be a bigger fan when the JSON is gone 😄

server/lib/src/entry.rs
yaleman: me2me2

proto/src/constants.rs
yaleman: derp, fixed

proto/src/constants.rs
yaleman: There's something weird with constants - if I set any of the various `#[cfg(test)]` / `#[cfg(any(test, debug_assertions))]` etc it won't find it when running `cargo test`

Firstyear: Hmmm okay. Maybe we TODO flag this then.

server/core/src/actors/v1_read.rs
Firstyear: OHH yeah we do, because of ldap. Couldn't we just have the non-unsdorced version exist in LDAP_ATTR only?

proto/src/constants.rs
yaleman: makes sense :D

server/core/src/actors/v1_read.rs
yaleman: Nope, 'cause someone made it a virtual Schema attribute here:

https://github.com/yaleman/kanidm/blob/4371640519dda239a5c6c21468569b846105be28/server/lib/src/schema.rs#L1667

yaleman: We could rename it to LdapSshPublicKey or something?

Firstyear: Yeah, because it's there to simulate / prevent people creating it later since it's an ldap attr. So maybe we name it Attribute::SshPublicKey and Attribute::LdapSshPublicKey. Doesn't match the text of the attrs, but describes their use.

Firstyear (approved):

PR #1993, opened 2023-08-17
Authentication shortcut to get a RW session

### LINKED ISSUE #1940: Shortcut for authenticating with the API?
I am using the kanidm API to programatically modify the assets in there. With the latest release, the initial authentication got more complicated. Before, I just created a session by POSTing to /v1/auth (init: username). Then a second POST initiated the authentication (begin: password) and a third POST (cred: password) actually logged me in and I could start operating.

In the newest version this leaves me with a read-only session. I found out that I need to use the /v1/reauth endpoint to escalate privileges, but this means I need 2 more requests. First I POST (token: null, which I found weird). That gives me a new session ID, so the one I received a second ago is already obsolete and I continue with this one. Now I POST once more to /v1/auth (cred: password) using the new session as header.

It's the same behavior with the kanidm CLI. You login and if you want to make a write operation right after, you need to reauth, so you repeat your auth steps you did 5 seconds ago.

I understand the idea behind it and you are the experts on authentication. My question is: Is there no safe shortcut, like when initiating a session to tell the server "I'd like to start with a write session"?

Oh, and by the way: If I programmatically fire these 5 requests to initially auth, I always run into a timing error: `Ident session is not present in entry. Perhaps replication is delayed?`

I worked around that with a "sleep" statement in my script before talking to the reauth endpoint, but well…

scabrero:
What do you think about this approach? I am not convinced about the `privileged` name.

Do we also want to expose this functionality in the client library? E.g.:
`$ kanidm login --privileged`

Fixes #1940 

Checklist

- [X] cargo fmt has been run

### DIFF
--- a/proto/src/v1.rs
+++ b/proto/src/v1.rs
@@ -919,6 +919,7 @@ pub enum AuthStep {
     Init2 {
         username: String,
         issue: AuthIssueSession,
+        #[serde(default)]
         privileged: bool,
     },
     // We want to talk to you like this.
--- a/proto/src/v1.rs
+++ b/proto/src/v1.rs
@@ -919,6 +919,7 @@ pub enum AuthStep {
     Init2 {
         username: String,
         issue: AuthIssueSession,
+        privileged: bool,
     },
     // We want to talk to you like this.
     Begin(AuthMech),
--- a/server/lib/src/idm/authsession.rs
+++ b/server/lib/src/idm/authsession.rs
@@ -1355,9 +1355,7 @@ mod tests {
         }};
     }
 
-    #[test]
-    fn test_idm_authsession_simple_password_mech() {
-        sketching::test_init();
+    fn start_session_simple_password_mech(privileged: bool) -> () {
         let webauthn = create_webauthn();
         // create the ent
         let mut account = entry_to_account!(E_ADMIN_V1.clone());
@@ -1396,7 +1394,7 @@ mod tests {
         // === Now begin a new session, and use a good pw.
 
         let (mut session, pw_badlist_cache) =
-            start_password_session!(&mut audit, account, &webauthn, false);
+            start_password_session!(&mut audit, account, &webauthn, privileged);
 
         let attempt = AuthCredential::Password("test_password".to_string());
         match session.validate_creds(
@@ -1421,6 +1419,12 @@ mod tests {
         assert!(async_rx.blocking_recv().is_none());
         drop(audit_tx);
         assert!(audit_rx.blocking_recv().is_none());
+    }
+
+    #[test]
+    fn test_idm_authsession_simple_password_mech() {
+        sketching::test_init();
+        start_session_simple_password_mech(false);
     }
 
     #[test]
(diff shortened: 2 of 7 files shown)


proto/src/v1.rs
```diff
@@ -917,6 +917,8 @@ pub enum AuthStep {
     Init2 {
         username: String,
         issue: AuthIssueSession,
+        #[serde(skip_serializing_if = "Option::is_none")]
+        privileged: Option<bool>,
```
Firstyear, line 921: I think you can make this just `bool` because bool impls default, and bool defaults to false. That way you can have `#[serde(default)]` here instead. That then makes some of your later code a bit easier.

server/lib/src/idm/authsession.rs
```diff
@@ -1327,72 +1341,98 @@ mod tests {
...
-        let cred = Credential::new_password_only(&p, "test_password").unwrap();
-        account.primary = Some(cred);
-
-        let (async_tx, mut async_rx) = unbounded();
-        let (audit_tx, mut audit_rx) = unbounded();
-
-        // now check
-        let (mut session, pw_badlist_cache) =
-            start_password_session!(&mut audit, account, &webauthn);
+    macro_rules! start_session_simple_password_mech {
```
Firstyear, line 1344: Why make this a macro? Why not leave it as a function?

server/lib/src/idm/authsession.rs
```diff
@@ -1327,72 +1341,98 @@ mod tests {
...
+            assert!(audit_rx.blocking_recv().is_none());
+        }};
+    }
 
-        drop(async_tx);
-        assert!(async_rx.blocking_recv().is_none());
-        drop(audit_tx);
-        assert!(audit_rx.blocking_recv().is_none());
+    #[test]
+    fn test_idm_authsession_simple_password_mech() {
```
Firstyear, line 1433: Should be two seperate tests rather than the sequential calls here like this.

server/testkit/tests/proto_v1_test.rs
```diff
@@ -1460,3 +1462,178 @@ async fn test_server_user_auth_reauthentication(rsclient: KanidmClient) {
     eprintln!("{:?} {:?}", now, uat.purpose);
     assert!(uat.purpose_readwrite_active(now));
 }
+
+macro_rules! start_password_session {
```
Firstyear, line 1466: Generally we should avoid macros unless really needed for small ergonomic tasks. Better to use functions.

Firstyear (changes requested):

Firstyear:
> Do we also want to expose this functionality in the client library? E.g.:
> `$ kanidm login --privileged`

To start with, no, I think we should leave this as api only. The cli is a bit more targetted to human users and already has some elements to assist with this. 


Generally though, looks really good for the first PR. I think that macros should be avoided a bit more here as they add a chunk to compile times and generally make it a bit harder to trace some things. We used a lot of macros in the past and have been trying to "undo it" for functions. Generally macros are good for small ergonomic wrappers (like vec![] or btreemap![]) or allowing some variable arg situations in some cases.

proto/src/v1.rs
scabrero: Done.

Firstyear (approved):
This looks really good!

PR #1994, opened 2023-08-17
Himmelblau requires a mechanism to send pam_prompt

dmulder:
In Himmelblau I need a mechanism for sending an additional pam prompt (for performing 2FA via a browser window on another device). I've implemented that here by causing pam to operate like a state machine, with the daemon requesting prompts only as required.
This replaces MR #1970.

Fixes #

Checklist

- [x] cargo fmt has been run

### DIFF
--- a/unix_integration/pam_kanidm/src/pam/mod.rs
+++ b/unix_integration/pam_kanidm/src/pam/mod.rs
@@ -286,10 +286,16 @@ impl PamHooks for PamKanidm {
                     };
                     match prompt.cred_type {
                         Some(CredType::Password) => {
-                            req = ClientRequest::PamAuthenticateStep(Some(PamCred::Password(resp)));
+                            req = ClientRequest::PamAuthenticateStep(
+                                Some(PamCred::Password(resp)),
+                                prompt.data,
+                            );
                         }
                         _ => {
-                            req = ClientRequest::PamAuthenticateStep(Some(PamCred::MFACode(resp)));
+                            req = ClientRequest::PamAuthenticateStep(
+                                Some(PamCred::MFACode(resp)),
+                                prompt.data,
+                            );
                         }
                     }
                     // We've consumed any authtok, delete it
@@ -305,7 +311,7 @@ impl PamHooks for PamKanidm {
                             return err;
                         }
                     }
-                    req = ClientRequest::PamAuthenticateStep(None);
+                    req = ClientRequest::PamAuthenticateStep(None, prompt.data);
                 }
                 _ => {
                     return PamResultCode::PAM_IGNORE;
--- a/unix_integration/pam_kanidm/src/pam/mod.rs
+++ b/unix_integration/pam_kanidm/src/pam/mod.rs
@@ -38,7 +38,7 @@ use std::ffi::CStr;
 use kanidm_unix_common::client_sync::call_daemon_blocking;
 use kanidm_unix_common::constants::DEFAULT_CONFIG_PATH;
 use kanidm_unix_common::unix_config::KanidmUnixdConfig;
-use kanidm_unix_common::unix_proto::{ClientRequest, ClientResponse};
+use kanidm_unix_common::unix_proto::{ClientRequest, ClientResponse, CredType, PamCred, PamPrompt};
 
 use crate::pam::constants::*;
 use crate::pam::conv::PamConv;
@@ -188,8 +188,24 @@ impl PamHooks for PamKanidm {
             }
         };
 
-        let authtok = match pamh.get_authtok() {
-            Ok(atok) => atok,
+        let cfg = match get_cfg() {
+            Ok(cfg) => cfg,
+            Err(e) => return e,
+        };
+        let mut req = ClientRequest::PamAuthenticateInit(account_id);
+        let mut prompt: PamPrompt = Default::default();
+
+        let mut authtok = match pamh.get_authtok() {
+            Ok(Some(v)) => Some(v),
+            Ok(None) => {
+                if opts.use_first_pass {
+                    if opts.debug {
+                        println!("Don't have an authtok, returning PAM_AUTH_ERR");
+                    }
+                    return PamResultCode::PAM_AUTH_ERR;
+                }
+                None
+            }
             Err(e) => {
                 if opts.debug {
                     println!("Error get_authtok -> {:?}", e);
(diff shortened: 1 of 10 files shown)


dmulder:
Getting better! This MR includes having built and run `cargo fmt` and `cargo clippy` :)
I'm getting an odd linker error during `cargo test` though, which I'll need to investigate.

dmulder:
@Firstyear this is my implementation of the pam state machine, based on our discussion the other day. Feel welcome to criticize and attack as you please! :-)

unix_integration/src/daemon.rs
```diff
@@ -193,6 +195,7 @@ async fn handle_client(
     };
 
     let mut reqs = Framed::new(sock, ClientCodec);
+    let mut pam_state = PamState::Uninitialized;
```
dmulder, line 198: The pam state machine only has 2 states here; Uninitialized or Step (meaning it's performing auth step operations). While the id provider returns PamPrompt requests, the machine will remain in the Step state (which holds the auth id). As soon as any other response is returned (either Success or failure, etc) from `pam_account_authenticate_step`, the state returns to Uninitialized.

unix_integration/src/resolver.rs
```diff
@@ -733,27 +735,48 @@ where
...
                     self.delete_cache_usertoken(n_tok.uuid).await?;
-                    Ok(None)
+                    Ok(ClientResponse::PamStatus(None))
                 } else {
                     debug!("online password check success.");
                     self.set_cache_usertoken(&mut n_tok).await?;
-                    self.set_cache_userpassword(n_tok.uuid, cred).await?;
-                    Ok(Some(true))
+                    match cred {
+                        Some(PamCred::Password(cred)) => {
```
dmulder, line 762: Here we've kept track of whether the id provider indicated that it was requesting a Password from the pam prompt. If it was a password, then we cache it for offline auth. If it was anything else (such as an MFA token), we ignore caching (otherwise we'll muck up the offline cache).

unix_integration/src/resolver.rs
```diff
@@ -788,18 +823,29 @@ where
...
         token: &Option<UserToken>,
-        cred: &str,
-    ) -> Result<Option<bool>, ()> {
+        cred: Option<PamCred>,
+    ) -> Result<ClientResponse, ()> {
+        let cred = match cred {
+            Some(cred) => match cred {
+                PamCred::Password(v) => v.clone(),
+                // We can only authenticate using a password
+                _ => return Ok(ClientResponse::PamStatus(Some(false))),
```
dmulder, line 832: I wasn't certain how to handle this case exactly. If we were in the middle of a authentication attempt, and suddenly went offline, then we could be stuck with an MFA token and no way to perform an offline logon. I've chosen to simply reject the auth attempt, but we could instead return a request for a password prompt here (such as you see on line 835). The only downside is the user could be prompted multiple times for the password. Also, if we come back online while waiting for that additional password, we'll have to start all over again. This could technically go on indefinitely, if there is some network instability.

unix_integration/src/tool.rs
```diff
@@ -58,37 +61,99 @@ async fn main() -> ExitCode {
             return ExitCode::FAILURE
         };
 
-            let password = match rpassword::prompt_password("Enter Unix password: ") {
```
dmulder, line 61: The password prompt here will change for auth test, since we rely on the id provider to specify the prompt now (it is currently defaulting to "Password: ").

unix_integration/src/tool.rs
```diff
@@ -58,37 +61,99 @@ async fn main() -> ExitCode {
...
+                            }
+                        }
+                        passcode = passcode.trim_end_matches('\n').to_string();
+                        match prompt.cred_type {
+                            Some(CredType::Password) => {
+                                req = ClientRequest::PamAuthenticateStep(Some(PamCred::Password(
+                                    passcode,
+                                )));
+                            }
+                            _ => {
```
dmulder, line 140: You really should never get None here, since you're requesting a PamPromptEchoOn, but I handled it anyway as an MFA prompt. The point is just to prevent caching the response unless we are certain it is a password.
Same goes for line 117 (and the similar lines in the pam module).

unix_integration/src/unix_proto.rs
```diff
@@ -16,6 +17,59 @@ pub struct NssGroup {
...
+pub enum PamMessageStyle {
+    #[default]
+    PamPromptEchoOff,
+    PamPromptEchoOn,
+    PamErrorMsg,
+    PamTextInfo,
+}
+
+impl PamMessageStyle {
+    pub fn value(&self) -> i32 {
```
dmulder, line 30: This ugly mess is to allow the PamMessageStyle enum to map correctly to a c_int in pam (i32). Perhaps this won't work everywhere?

unix_integration/src/unix_proto.rs
```diff
@@ -16,6 +17,59 @@ pub struct NssGroup {
...
+impl PamPrompt {
+    // Produce a typical password prompt
+    pub fn passwd_prompt() -> Self {
+        PamPrompt {
+            style: PamMessageStyle::PamPromptEchoOff,
+            msg: "Password: ".to_string(),
+            timeout: None,
+            cred_type: Some(CredType::Password),
+        }
+    }
```
dmulder, line 64: We could perhaps provide a default MFA prompt here, but nothing would use it at the moment.

yaleman:
If this is the replacement and #1970's not needed, please close it 😄

unix_integration/src/daemon.rs
```diff
@@ -281,15 +281,15 @@ async fn handle_client(
                 debug!("pam authenticate init");
                 pam_state = PamState::Step(account_id.clone());
                 cachelayer
-                    .pam_account_authenticate_step(account_id.as_str(), None)
+                    .pam_account_authenticate_step(account_id.as_str(), None, None)
```
Firstyear, line 284: Shouldn't this be a new call to pam_account_authenticate_init, which then creates an auth session we bind here into pam_state? I feel like pam state should be created by and managed by the cache layer, not the caller. That way the "pam_state" can contain things like the user token (if any) and our knowledge of the credentials and what's needed. 

Then in pamAuthenticateStep you pass that pam_state back into cachelayer, and mostly, the pam_state actually has everything needed to allow the auth to proceed already and we only actually access the db or the id provider if something online needs checking.

unix_integration/src/pam_data.rs
```diff
@@ -0,0 +1,7 @@
+use serde::{Deserialize, Serialize};
+
+/* This is the definition for extra data to be sent along with a pam_prompt
+ * request. It will be sent back to the idprovider to continue an auth attempt.
+ */
+#[derive(Serialize, Deserialize, Debug)]
+pub struct PamData {}
```
Firstyear, line 7: Rather than a struct, could this be an enum that has say :

```
enum PamData {
    None,
    MFACred( String ),
}
```

Or something similar? Really even with this, I think PamData and PamCred are very "overlapping" here in what they represent being a enum of possible inputs to a pam challenge.

Firstyear: In other words, could PamCred::MFACred { value: ..., data }, actually be the enum variant to take the extra data you need in that step only?

unix_integration/pam_kanidm/src/pam/mod.rs
```diff
@@ -286,10 +286,16 @@ impl PamHooks for PamKanidm {
                     };
                     match prompt.cred_type {
```
Firstyear, line 287: Where are we creating this cred_type to pass back up to the pam module here?

unix_integration/src/daemon.rs
```diff
@@ -281,15 +281,15 @@ async fn handle_client(
...
                     .await
                     .unwrap_or(ClientResponse::Error)
             }
-            ClientRequest::PamAuthenticateStep(cred) => {
+            ClientRequest::PamAuthenticateStep(cred, data) => {
                 debug!("pam authenticate step");
                 match &pam_state {
                     PamState::Step(account_id) => match cachelayer
-                        .pam_account_authenticate_step(account_id.as_str(), cred)
+                        .pam_account_authenticate_step(account_id.as_str(), cred, data)
```
Firstyear, line 292: Need a way to see whant the result from _step is here so we can potentially clear the auth session on completion so that a client can do a sequence of pam auths if required without dropping/opening the unix socket.

Firstyear (changes requested):

unix_integration/src/daemon.rs
dmulder: Ah, I like that better. Perhaps we could eliminate passing the PamData down to the pam layer this way, since it's really only needed by the daemon, and we can just keep it within the pam_state?

unix_integration/pam_kanidm/src/pam/mod.rs
```diff
@@ -188,94 +188,136 @@ impl PamHooks for PamKanidm {
...
-            None => {
-                if opts.use_first_pass {
-                    if opts.debug {
-                        println!("Don't have an authtok, returning PAM_AUTH_ERR");
+        loop {
+            let timeout = match prompt.timeout {
+                Some(timeout) => timeout,
+                None => cfg.unix_sock_timeout,
+            };
+            match call_daemon_blocking(cfg.sock_path.as_str(), &req, timeout) {
```
dmulder, line 231: I just discovered in testing that `call_daemon_blocking` disconnects and starts a new session with the daemon each time it's called, destroying our pam state data. This causes us to fail calling `PamAuthenticateStep` because the the state is uninitialized again.

unix_integration/pam_kanidm/src/pam/mod.rs
dmulder: That's happening in the idprovider. The idprovider is calling `passwd_prompt` at unix_integration/src/unix_proto.rs:59.

unix_integration/src/pam_data.rs
dmulder: Except in Himmelblau PamData contains one of these:
```rust
pub struct DeviceAuthorizationFlow {
    pub device_code: String,
    pub user_code: String,
    pub verification_uri: String,
    /* MS doesn't implement verification_uri_complete yet, but our
     * authentication will be simpler once they do, so assume it works and fall
     * back to verification_uri if it doesn't.
     */
    pub verification_uri_complete: Option<String>,
    pub expires_in: u32,
    pub interval: Option<u32>,
    pub message: Option<String>,
}
```

dmulder: What I'm trying to do is ensure this information is kept secret, and only available to that auth attempt. I'm thinking now it makes more sense to keep this in the pam state instead.

dmulder: > In other words, could PamCred::MFACred { value: ..., data }, actually be the enum variant to take the extra data you need in that step only?

It wasn't meant to be extra data received. It's supposed to represent data _maintained_ from one call to step function to the next.

unix_integration/src/daemon.rs
dmulder: That's what I was getting at in unix_integration/src/daemon.rs:300. If we sent a pam_prompt request, we're not done. Anything else (some other client response), and we know the auth request has completed. Maybe not the most elegant, but I did handle this.

unix_integration/src/pam_data.rs
dmulder: Regardless, I do think it makes more sense for this to be an enum, but it should be in the PamState, not in the PamPrompt.

dmulder: It should look something like this, I think:
```rust
enum PamData {
    AwaitingPassword,
    AwaitingMFA(DeviceAuthorizationFlow),
}
```
This way if the idprovider can check if it receives an appropriate response from the pam module (it shouldn't get a password response when trying to do MFA, for example).

unix_integration/src/daemon.rs
dmulder: The PamState would look like this?
```rust
#[derive(Debug)]
pub enum ProviderResult {
    PamPrompt(PamPrompt),
    UserToken(Option<UserToken>),
}

pub enum PamState {
    Uninitialized,
    Init(String, PamData),
    Step(ProviderResult, PamData),
}
```

Firstyear:
See https://github.com/kanidm/kanidm/pull/2022 - I'm reworking it for you since you're away for a few weeks :)

PR #2000, opened 2023-08-19
patching serde_derive dependency source

yaleman:
This causes a warning on build, but given the lack of consideration for software sourcing in [serde-rs/serde/#2538](https://github.com/serde-rs/serde/issues/2538) I think it's warranted to remove a non-auditable binary dependency from builds on an IDP.

Checklist

- [x] cargo fmt has been run

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -41,9 +41,9 @@ dependencies = [
 
 [[package]]
 name = "aho-corasick"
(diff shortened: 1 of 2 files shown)


Firstyear:
I actually think we don't need to action this. The time crate already does serde pinning, and since we now use time >= 3, we will automatically inherit this.

PR #1960, opened 2023-08-09
error handling and web server logging fixes

yaleman:
* Fixes #1958
* adds some tests relating to #1958
* Cleans up some URL-generation in the client code
* chases some clippy lints
* fixes the setup_dev_environment script to support the new UNIX socket system

Checklist

- [x] This pr contains no AI generated code
- [x] cargo fmt has been run
- [x] cargo clippy has been run
- [x] cargo test has been run and passes
- [ ] book chapter included (if relevant)
- [ ] design document included (if relevant)

PR #2001, opened 2023-08-20
idv cli

Seba-T:
Fixes ..nothing :p 

Checklist

- [x] cargo fmt has been run


Seba-T:
I marked this as draft because I included the pdf file but I just wanted to get some feedback before turning it into a markdown so it can be merged. 
And also the various todos I left here and there, mainly the one about the regex, the others I can figure it out on my own.

yaleman (changes requested):
This is looking good so far, all the error-id-TODO's are for #1960 

If you can't merge them directly and want to, hit me up - there should be a button in the GitHub UI

yaleman:
Also - can you please change the `cursive` backend to use `crossterm-backend` to avoid having to include ncurses, and so it'll work on Windows? - https://github.com/gyscos/cursive/wiki/Backends#available-backends

server/web_ui/src/views/identityverification.rs
```diff
@@ -60,7 +60,7 @@ static INVALID_USERID_ERROR: &str = "The provided UserID is invalid!";
 lazy_static::lazy_static! {
     pub static ref VALIDATE_TOTP_RE: Regex = {
         #[allow(clippy::expect_used)]
-        Regex::new(r"^\d{6}$").expect("Invalid singleline regex found")
+        Regex::new(r"^\d{5}$|^\d{6}$").expect("Invalid singleline regex found")
```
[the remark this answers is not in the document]
Seba-T: Should I duplicate them also for the cli :thinking: ? Should I move the regex to a shared place?

tools/cli/src/cli/lib.rs
```diff
@@ -139,3 +179,571 @@ pub(crate) fn password_prompt(prompt: &str) -> Option<String> {
     }
     None
 }
+
+// TODO: this regex is also used in the webui (https://github.com/kanidm/kanidm/blob/003234c2d0a52146683628156e2a106bf61fe9f4/server/web_ui/src/views/identityverification.rs#L60) should we move it to proto?
```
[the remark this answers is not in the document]
Seba-T: My only issue with moving it to proto is that it kinda felt misplaced, like everything inside proto is some sort of message, this would be quite different from everything else.

Seba-T:
> Also - can you please change the `cursive` backend to use `crossterm-backend` to avoid having to include ncurses, and so it'll work on Windows? - https://github.com/gyscos/cursive/wiki/Backends#available-backends

Sure :smile:

tools/cli/src/cli/lib.rs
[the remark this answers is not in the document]
yaleman: You could probably dump the regex entirely, and just do a length check since you're already validating it's a number with the u32 parse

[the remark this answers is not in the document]
Seba-T: yep that actually makes lots of sense. Initially I didn't do it because I thought "what if in the future we want to parse something that's not a number and we'll have to write the regex from scratch", but actually if that ever happens then we'll write the regex as needed.

tools/cli/Cargo.toml
```diff
@@ -50,6 +50,9 @@ tokio = { workspace = true, features = ["rt", "macros"] }
 url = { workspace = true, features = ["serde"] }
 uuid = { workspace=true }
 zxcvbn = { workspace=true }
+regex.workspace = true
+lazy_static.workspace = true
+cursive = "0.20.0"
```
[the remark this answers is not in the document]
Seba-T: yess :)

tools/cli/src/cli/lib.rs
```diff
@@ -10,12 +10,23 @@
 #![deny(clippy::trivially_copy_pass_by_ref)]
 // We allow expect since it forces good error messages at the least.
 #![allow(clippy::expect_used)]
-
 #[macro_use]
 extern crate tracing;
 
 use crate::common::OpType;
-use std::path::PathBuf;
+use cursive::{
```
Firstyear, line 17: For now I'd like this to be a separate feature and in it's own file. 

What about a simple text-only version like the credential reset dialogue? Not every terminal can do tui.

Firstyear (changes requested):
I think the feature should be a md file not pdf. Any reason to make it a pdf?

Seba-T:
> I think the feature should be a md file not pdf. Any reason to make it a pdf?

Currently it’s only about having more colors and shapes for the diagram which makes it easier to read. Maybe I can export the diagram to a png and include it in the md?

yaleman:
> > I think the feature should be a md file not pdf. Any reason to make it a pdf?
> 
> Currently it’s only about having more colors and shapes for the diagram which makes it easier to read. Maybe I can export the diagram to a png and include it in the md?

That'd be good, or use [mermaid charts](https://github.com/badboy/mdbook-mermaid) which I'm *pretty sure* are part of the default packages of most things. (or should be)

tools/cli/src/cli/lib.rs
Seba-T: I actually started doing it that way, but i ended up with many rough edges, like no closable pop ups, no countdown, and something else I can’t remember now. Still doable though. I still have that version and it’s like 80% complete, should I add it somewhere?

yaleman: As nice as all the features are, the user just needs to get prompted with the token and told they've got x minutes to complete it, and success/errors if it fails 😄

The prettiness can happen in the web UI which has *much* better native handling for such things.

Seba-T: Alright then :) I think I got everything working, I'll push shortly.

tools/cli/src/cli/lib.rs
```diff
@@ -139,3 +179,298 @@ pub(crate) fn password_prompt(prompt: &str) -> Option<String> {
...
+        DisplayCodeSecond { self_totp: u32, step: u32 },
+    }
+
+    fn server_error(e: ClientError) {
+        eprintln!("Server error!"); // TODO: add an error ID (internal error, restart)
+        eprintln!("{:?}", e);
+        println!("Exiting...");
+    }
+
+    #[async_recursion] // dear lord have mercy for I have sinned writing this function
```
Firstyear, line 228: Yes, you have. Have you tried just having a loop and updating your state each loop iteration?

tools/cli/src/cli/lib.rs
```diff
@@ -139,3 +179,298 @@ pub(crate) fn password_prompt(prompt: &str) -> Option<String> {
...
+                    _ => {
+                        eprintln!("{INVALID_STATE_ERROR_MESSAGE}");
+                        return;
+                    }
+                }
+            }
+            IdentifyUserState::DisplayCodeFirst { self_totp, step } => {
+                println!("Provide the following code when asked: {}", self_totp);
+                let seconds_left = get_ms_left_from_now(step as u128) / 1000;
+                println!("This codes expires in {seconds_left} seconds!");
```
Firstyear, line 387: Don't add the exclaimation here, it adds too much panic.

Firstyear (changes requested):

tools/cli/src/cli/lib.rs
Seba-T: I considered it but I couldn't overcome my feelings for purely functional stuff. Using `mut` in a function argument was too much for my little heart to handle :crying_cat_face:. I'll begrudgingly fix it though

tools/cli/src/cli/lib.rs
Seba-T: sure :)

tools/cli/src/cli/lib.rs
Firstyear: Rust isn't functional. And if you treat it like it is, you'll hit stack depth and crash. Sometimes the simple way, is the right way.

Seba-T: Yes btw that was actually happening at one point lol. I was doing the totp refetch in the dumbest of ways by recursively calling the function and since I had some very poor logic going on it would make batch of 100 calls to itself .. let's say it didn't last more than 20 seconds. Lesson learned

Seba-T: fixed!

Firstyear:
Up to you now @yaleman :)

yaleman (approved):
LGTM

yaleman:
@Seba-T this just needs another rebase 🥳 well done!

Seba-T:
done ! :)

Firstyear:
Once again, well done :)

Issue #240, opened 2020-05-28
Sudo rules and other authorisation interactions


Firstyear:
We may consider supporting sudo rules as part of the kanidm module. There are some dynamic ways to achieve this, such as:

https://blog.sudo.ws/posts/2020/05/sudo-1.9-using-the-group-plugin-from-python/

More design and thought needed in this space!

PR #2015, opened 2023-08-21
Kanidm Sudo Plugin (beginnings)

yaleman:
Refers to #240

Pretty sure this'll be a challenge to *actually* use with sudo, whose interface is ... embedded Python. Ick. But it's a start. Oh, and I've added some more functionality/testing to pykanidm.

Checklist

- [x] cargo fmt has been run


Firstyear (changes requested):
I like that you've added this, but I don't want any part of the system installed code to be python. I've been burnt really hard before in other projects trying to package/distribute/test python, and I want to avoid it here. I'm okay with the radius python since it's contained to a container, and I'm okay with the python api wrappers, but I'm not okay with unix_integration having python. 

In addition this should be a module that's part of the daemon. This lets us read from the unix daemon cache and work in offline mode too. It also lets us have sudo rules in kani if we want (but that starts to get into some weird sudo distribution stuff .... for now we might want to configure it just in unixd)

Would you also mind elaborating on what you want to achieve/solve by having sudo rules like this? Or is it just to resolve a ticket?

Firstyear:
https://sssd.io/design-pages/sudo_support.html

Prior art

Firstyear:
https://linux.die.net/man/8/sudo_plugin

yaleman:
Fair enough, I was just implementing the python module example linked to in #240 😄

Firstyear:
Yeah I think that was more about "look into how to do this later" rather than "lets do it in python". But yeah, sudo rules will need a lot of thought in general I think :)

PR #2017, opened 2023-08-21
feat: add json output modes for `person list` and `system oauth2 list`

oddlama:
This implements json output for `kanidm person list` and  `kanidm system oauth2 list`, so that the argument `--output=json` is no longer ignored.

Checklist

- [x] cargo fmt has been run

### DIFF
--- a/tools/cli/src/cli/oauth2.rs
+++ b/tools/cli/src/cli/oauth2.rs
@@ -34,7 +34,16 @@ impl Oauth2Opt {
             Oauth2Opt::List(copt) => {
                 let client = copt.to_client(OpType::Read).await;
                 match client.idm_oauth2_rs_list().await {
-                    Ok(r) => r.iter().for_each(|ent| println!("{}", ent)),
+                    Ok(r) => match copt.output_mode {
+                        OutputMode::Json => {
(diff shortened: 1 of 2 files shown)


yaleman (approved):
Sweet, thanks

Firstyear:
I'm getting a bit wary about all these json modes in the cli, it's getting dangerously close to breaking peoples setups because the cli is not and will never be a stable api target per https://github.com/kanidm/kanidm/blob/master/project_docs/RELEASE_AND_SUPPORT.md#api-stability 

So I think we need to consider how we approach this - either a seperate json interaction tool, or improvements to the api in general to avoid cli call outs.

yaleman:
A *lot* of CLI tools (kubectl/massscan/nmap/pytest/terraform, just thinking of recent uses) support a variation of "output=xxx" as an expectation these days, with no guarantee of stability. I think the "CLI is not a stable API" rule is something we can point to if it breaks people's workflows - see also the `apt` command's explicit statements on the CLI for the longest time).

There's no way you could replicate the functionality of the CLI in the API and make it as usable (consider Webauthn auth on the CLI, as one key example), and a separate tool with JSON output would just be duplication of the exact functionality that `kanidm -o json` would provide, so I'm not sure how that'd make things better?

PR #2021, opened 2023-08-22
Trying to fix the (current) container build failures

yaleman:
* `busybox-adduser` fails to install on amd64 because `shadow` is installed, which means it fails in GHA builds and there's I can't find an automated alternative for just creating a user in the container 😡 - removing this dependency from the `kanidm` container and reverting to using the default (root) user. 
* python changed how salty they get about installing system-level packages, so forcing that to work.

Checklist

- [ ] cargo fmt has been run

### DIFF
--- a/rlm_python/Dockerfile
+++ b/rlm_python/Dockerfile
@@ -1,12 +1,13 @@
 ARG BASE_IMAGE=opensuse/tumbleweed:latest
-FROM ${BASE_IMAGE} AS repos
+ARG RADIUS_USER=radiusd
+FROM ${BASE_IMAGE}
+
+EXPOSE 1812 1813
+ENV KANIDM_CONFIG_FILE="/data/kanidm"
+
 ADD ../scripts/zypper_fixing.sh /zypper_fixing.sh
 RUN --mount=type=cache,id=zypp,target=/var/cache/zypp /zypper_fixing.sh
 
-# ======================
-FROM repos
-EXPOSE 1812 1813
-ARG RADIUS_USER=radiusd
 
 RUN \
     --mount=type=cache,id=zypp,target=/var/cache/zypp \
@@ -60,7 +61,6 @@ RUN python3 -m pip install \
 
 COPY rlm_python/radius_entrypoint.py /radius_entrypoint.py
 
-ENV KANIDM_CONFIG_FILE="/data/kanidm"
 
 RUN chmod a+r /etc/raddb/certs/ -R
 USER $RADIUS_USER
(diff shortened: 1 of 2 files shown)


yaleman:
@Firstyear OpenSUSE is just... broken again

yaleman:
... all I had to do was shit-talk it and it worked? 😡

rlm_python/Dockerfile
```diff
@@ -1,12 +1,13 @@
...
+FROM ${BASE_IMAGE}
+
+EXPOSE 1812 1813
+ENV KANIDM_CONFIG_FILE="/data/kanidm"
+
 ADD ../scripts/zypper_fixing.sh /zypper_fixing.sh
 RUN --mount=type=cache,id=zypp,target=/var/cache/zypp /zypper_fixing.sh
 
-# ======================
-FROM repos
```
Firstyear, line 7: Is there a reason to remove this layering? Since all the containers have a "similar" repos setup, they should all use the same base?

Firstyear (approved):

rlm_python/Dockerfile
yaleman: it kept breaking in dumb ways, and I'm not sure the "same base" thing actually works unless we're actually building that base as another image, then loading it

yaleman: it's ... weird.

https://github.com/kanidm/kanidm/actions/runs/5946810823/job/16127922982#step:5:218

```
#9 9.304 Retrieving repository 'openSUSE-Tumbleweed-Update' metadata [....
#9 10.58 Warning: File 'repomd.xml' from repository 'openSUSE-Tumbleweed-Update' is unsigned.
#9 10.58 
#9 10.58     Note: Signing data enables the recipient to verify that no modifications occurred after the data
#9 10.58     were signed. Accepting data with no, wrong or unknown signature can lead to a corrupted system
#9 10.58     and in extreme cases even to a system compromise.
#9 10.58 
#9 10.58     Note: File 'repomd.xml' is the repositories master index file. It ensures the integrity of the
#9 10.58     whole repo.
#9 10.58 
#9 10.58     Warning: We can't verify that no one meddled with this file, so it might not be trustworthy
#9 10.58     anymore! You should not continue unless you know it's safe.
#9 10.58 
#9 10.58 File 'repomd.xml' from repository 'openSUSE-Tumbleweed-Update' is unsigned, continue? [yes/no] (no): Cannot read input: bad stream or EOF.
```

yaleman: Same issue as https://github.com/openSUSE/zypper/issues/478 again

PR #2071, opened 2023-09-04
Schema dooby doo part two

yaleman:
Refers to #1987 

Checklist

- [x] cargo fmt has been run


server/lib/src/server/delete.rs
```diff
@@ -261,8 +261,8 @@ mod tests {
         assert!(cr.is_ok());
 
         // Delete filter is syntax invalid
-        let de_inv = DeleteEvent::new_internal_invalid(filter!(f_pres("nhtoaunaoehtnu")));
-        assert!(server_txn.delete(&de_inv).is_err());
+        // let de_inv = DeleteEvent::new_internal_invalid(filter!(f_pres("nhtoaunaoehtnu")));
```
yaleman, line 264: Given that f_pres doesn't accept a raw string now, should we be testing this with one of the "test" attributes that this object doesn't have?

Firstyear: The problem here is that we need to actually reject the invalid attribute name. So provided there is a test attribute name that "will never exist" in schema then yes we can do that.

server/lib/src/schema.rs
```diff
@@ -421,41 +421,43 @@ impl SchemaClass {
...
-            .get_ava_iter_iutf8("supplements")
-            .map(|i| i.map(AttrString::from).collect())
+            .get_ava_iter_iutf8(Attribute::Supplements.as_ref())
+            .map(|i| i.map(AttrString::from).collect()) // TODO: this should probably just be an into?
             .unwrap_or_else(Vec::new);
         let systemexcludes = value
-            .get_ava_iter_iutf8("systemexcludes")
-            .map(|i| i.map(AttrString::from).collect())
+            .get_ava_iter_iutf8(Attribute::SystemExcludes.as_ref())
+            .map(|i| i.map(AttrString::from).collect()) // TODO: this should probably just be an into?
```
Firstyear, line 456: Yeah but then it has to be i.map(|v| v.into()).collect().

server/lib/src/server/delete.rs
yaleman: ok I'll work out a way of rejecting it at build time 😄

Firstyear: Shouldn't we just have an Attribute that is "CanNeverExist" that yields the gibberish, and then we cfg(test, debug assert) it out? Seems easier ....

yaleman: Weeeellll the Attribute can't exist if it's not in the enum 😄

Firstyear: You could also bypass the macro and just create the FC directly.

Firstyear (approved):
Looks great <3

PR #2077, opened 2023-09-05
minor tweaks to Orca

yaleman:
The clap derives were wrong, and I've updated the script to do some more stuff.

Checklist

- [x] cargo fmt has been run

### DIFF
--- a/tools/orca/src/kani.rs
+++ b/tools/orca/src/kani.rs
@@ -298,11 +298,12 @@ impl KaniHttpServer {
     }
 
     pub async fn close_connection(&self) {
-        self.client
+        assert!(self
+            .client
             .logout()
             .await
             .map_err(|e| error!("close_connection {:?}", e))
-            .ok();
+            .is_ok());
     }
 
     pub async fn search(
--- a/tools/orca/src/kani.rs
+++ b/tools/orca/src/kani.rs
@@ -99,9 +99,16 @@ impl KaniHttpServer {
         all_entities: &HashMap<Uuid, Entity>,
     ) -> Result<(), ()> {
         // Create all the accounts and groups
-        for u in targets {
-            let e = all_entities.get(u).unwrap();
-            match e {
+        let num_uuids = targets.len();
+        let mut current_slice = 1;
+        info!("Have to do {} uuids", num_uuids);
+        for (index, uuid) in targets.iter().enumerate() {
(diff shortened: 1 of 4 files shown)


tools/orca/src/kani.rs
```diff
@@ -99,9 +99,16 @@ impl KaniHttpServer {
         all_entities: &HashMap<Uuid, Entity>,
     ) -> Result<(), ()> {
         // Create all the accounts and groups
-        for u in targets {
-            let e = all_entities.get(u).unwrap();
-            match e {
+        let num_uuids = targets.len();
+        let mut current_slice = 1;
+        info!("Have to do {} uuids", num_uuids);
+        for (index, uuid) in targets.iter().enumerate() {
```
Firstyear, line 105: I wonder if it might be easier to do:

```
for (pct, chunk) in targets.chunks(targets.len() / 10).enumerate() {
    for uuid in chunk {
    }
    info!("{}% complete" pct * 10);
}
```

tools/orca/src/kani.rs
```diff
@@ -291,12 +298,11 @@ impl KaniHttpServer {
...
 
     pub async fn close_connection(&self) {
-        assert!(self
-            .client
+        self.client
             .logout()
             .await
             .map_err(|e| error!("close_connection {:?}", e))
-            .is_ok());
+            .ok();
```
Firstyear, line 305: .ok() makes this an option, then the `;` causes it to be ignored since close_connection() returns unit. So I think the assert was correct? Were you seeing assertion failures?

yaleman: Yeah it was yelling a lot about 404's and crashing

tools/orca/src/kani.rs
yaleman: That'd mean having to pull the targets out into a new vec as well, because targets is a HashBrown::hash_set, and it just gets messier from there

tools/orca/src/kani.rs
Firstyear: Shouldn't we try and resolve those rather than hiding the error?

yaleman: yeah, I just haven't gotten that far into working out what orca was doing yet 😃 I'll revert that bit for now

Firstyear (approved):

PR #2079, opened 2023-09-06
Implement DeviceAuthorizationGrant for MFA

dmulder:
Himmelblau will use the DeviceAuthorizationGrant
(defined in RFC8628) to perform MFA. This commit
adds the bits to Kanidm to make that possible,
using the new pam state machine code.

Fixes #

Checklist

- [x] cargo fmt has been run

### DIFF
--- a/unix_integration/pam_kanidm/src/pam/mod.rs
+++ b/unix_integration/pam_kanidm/src/pam/mod.rs
@@ -135,16 +135,15 @@ impl PamHooks for PamKanidm {
         let req = ClientRequest::PamAccountAllowed(account_id);
         // PamResultCode::PAM_IGNORE
 
-        let mut daemon_client =
-            match DaemonClientBlocking::new(cfg.sock_path.as_str(), cfg.unix_sock_timeout) {
-                Ok(dc) => dc,
-                Err(e) => {
-                    error!(err = ?e, "Error DaemonClientBlocking::new()");
-                    return PamResultCode::PAM_SERVICE_ERR;
-                }
-            };
+        let mut daemon_client = match DaemonClientBlocking::new(cfg.sock_path.as_str()) {
+            Ok(dc) => dc,
+            Err(e) => {
+                error!(err = ?e, "Error DaemonClientBlocking::new()");
+                return PamResultCode::PAM_SERVICE_ERR;
+            }
+        };
 
-        match daemon_client.call_and_wait(&req) {
+        match daemon_client.call_and_wait(&req, cfg.unix_sock_timeout) {
             Ok(r) => match r {
                 ClientResponse::PamStatus(Some(true)) => {
                     debug!("PamResultCode::PAM_SUCCESS");
@@ -203,14 +202,14 @@ impl PamHooks for PamKanidm {
             Err(e) => return e,
         };
 
-        let mut daemon_client =
-            match DaemonClientBlocking::new(cfg.sock_path.as_str(), cfg.unix_sock_timeout) {
-                Ok(dc) => dc,
-                Err(e) => {
-                    error!(err = ?e, "Error DaemonClientBlocking::new()");
-                    return PamResultCode::PAM_SERVICE_ERR;
-                }
-            };
+        let mut timeout = cfg.unix_sock_timeout;
+        let mut daemon_client = match DaemonClientBlocking::new(cfg.sock_path.as_str()) {
+            Ok(dc) => dc,
+            Err(e) => {
+                error!(err = ?e, "Error DaemonClientBlocking::new()");
+                return PamResultCode::PAM_SERVICE_ERR;
+            }
+        };
 
         // Later we may need to move this to a function and call it as a oneshot for auth methods
         // that don't require any authtoks at all. For example, imagine a user authed and they
@@ -242,7 +241,7 @@ impl PamHooks for PamKanidm {
         let mut req = ClientRequest::PamAuthenticateInit(account_id);
 
         loop {
-            match daemon_client.call_and_wait(&req) {
+            match daemon_client.call_and_wait(&req, timeout) {
                 Ok(r) => match r {
                     ClientResponse::PamAuthenticateStepResponse(PamAuthResponse::Success) => {
                         return PamResultCode::PAM_SUCCESS;
@@ -282,7 +281,32 @@ impl PamHooks for PamKanidm {
                         };
 
                         // Now setup the request for the next loop.
+                        timeout = cfg.unix_sock_timeout;
                         req = ClientRequest::PamAuthenticateStep(PamAuthRequest::Password { cred });
+                        continue;
+                    }
+                    ClientResponse::PamAuthenticateStepResponse(
+                        PamAuthResponse::DeviceAuthorizationGrant { data },
+                    ) => {
+                        let msg = match &data.message {
+                            Some(msg) => msg.clone(),
+                            None => format!("Using a browser on another device, visit:\n{}\nAnd enter the code:\n{}",
+                                            data.verification_uri, data.user_code)
+                        };
+                        match conv.send(PAM_TEXT_INFO, &msg) {
+                            Ok(_) => {}
+                            Err(err) => {
+                                if opts.debug {
+                                    println!("Message prompt failed");
+                                }
+                                return err;
+                            }
+                        }
+
+                        timeout = u64::from(data.expires_in);
+                        req = ClientRequest::PamAuthenticateStep(
+                            PamAuthRequest::DeviceAuthorizationGrant { data },
+                        );
                         continue;
                     }
                     _ => {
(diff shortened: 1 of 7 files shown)


unix_integration/pam_kanidm/src/pam/mod.rs
```diff
@@ -203,14 +203,14 @@ impl PamHooks for PamKanidm {
...
 
-        let mut daemon_client =
-            match DaemonClientBlocking::new(cfg.sock_path.as_str(), cfg.unix_sock_timeout) {
-                Ok(dc) => dc,
-                Err(e) => {
-                    error!(err = ?e, "Error DaemonClientBlocking::new()");
-                    return PamResultCode::PAM_SERVICE_ERR;
-                }
-            };
+        let mut timeout = cfg.unix_sock_timeout;
```
dmulder, line 206: I had to add `call_and_wait_timeout` because of the new DaemonClientBlocking struct, which stores the timeout. This might have seemed reasonable at the time, but we need to be able to modify the timeout to do a DeviceAuthorizationGrant (because the daemon may take up to `DeviceAuthorizationResponse.expires_in` seconds to respond (which defaults to 900 sec/15 min).

dmulder: This workaround fails, and it still times out after the default timeout. If I set the config timeout to something larger, than that works. The `DaemonClientBlocking` is ignoring my attempt to change the timeout.

dmulder: Resolved with the latest push.

unix_integration/src/client_sync.rs
```diff
@@ -104,4 +104,21 @@ impl DaemonClientBlocking {
 
         Ok(cr)
     }
+
+    pub fn call_and_wait_timeout(
+        &mut self,
+        req: &ClientRequest,
+        timeout: u64,
+    ) -> Result<ClientResponse, Box<dyn Error>> {
+        self.timeout = Duration::from_secs(timeout);
```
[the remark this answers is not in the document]
Firstyear: In otherwords, I think setting self.timeout here isn't quite right.

Firstyear (changes requested):

unix_integration/src/idprovider/kanidm.rs
```diff
@@ -223,13 +223,22 @@ impl IdProvider for KanidmProvider {
...
-              */
+            }
+            (
+                AuthCredHandler::DeviceAuthorizationGrant,
+                PamAuthRequest::DeviceAuthorizationGrant { .. },
+            ) => {
+                error!("DeviceAuthorizationGrant not implemented!");
+                Err(IdpError::BadRequest)
+            }
+            _ => Err(IdpError::BadRequest),
```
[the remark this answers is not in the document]
dmulder: Lol, I wasn't paying attention to the commented code.

unix_integration/src/resolver.rs
```diff
@@ -1047,6 +1062,10 @@ where
                             }
                         }
                     }
+                    _ => {
```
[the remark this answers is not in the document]
dmulder: We end up with weird matches here, such as (AuthCredHandler::Password, PamAuthRequest::DeviceAuthorizationGrant { .. }), (AuthCredHandler::DeviceAuthorizationGrant, PamAuthRequest::Password { cred }), which don't make any sense. That's why I did the wild card match. Should we always list all permutations? This could get rather large and ugly.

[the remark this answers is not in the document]
Firstyear: You can just do:

```
(AuthCredHandler::Password, PamAuthRequest::DeviceAuthorisationGrant { .. } ) |
(AuthCredHandler::DeviceGrant { .. }, PamAuthRequest::Password { .. } ) => {
    ...
}
```

This allows multiple matches to asingle output here, and again, the point is to help guide future developers when we do make these changes. I'd rather it be correct and exhaustive, even if ugly, then have us forget to update something and consider possibilities.

unix_integration/src/client_sync.rs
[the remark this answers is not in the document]
Firstyear: We always need a default timeout though. We just need to set a default timeout on the DaemonClientBlocking and have two options for call_and_wait, one with a timeout override, one without. Then inside the client we have a private fn that actually does the call and takes "timeout_override: Option<Duration>".

unix_integration/src/resolver.rs
```diff
@@ -993,6 +993,21 @@ where
                         self.set_cache_userpassword(token.uuid, &cred).await?;
                         Ok(AuthResult::Success { token })
                     }
+                    Ok((AuthResult::Next(req), AuthCacheAction::PasswordHashUpdate { cred })) => {
```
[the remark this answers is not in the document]
Firstyear: Mostly true - it depends on the type of the MFA provider. In this case with device grant, we can never do offline auth because of the device code requirement. But with something like webauthn we *could* do offline auth since we could potentially cache the public key of the authenticator. But yes, in most cases we can't do offline auth with  MFA here since most MFA's rely on a secret/seed that we don't have access to.

unix_integration/src/idprovider/kanidm.rs
```diff
@@ -223,13 +223,18 @@ impl IdProvider for KanidmProvider {
...
-                    error!("invalid authentication request state");
-                    Err(IdpError::BadRequest)
-                }
-              */
+            }
+            (
+                AuthCredHandler::DeviceAuthorizationGrant,
+                PamAuthRequest::DeviceAuthorizationGrant { .. },
+            ) => {
+                error!("DeviceAuthorizationGrant not implemented!");
```
Firstyear, line 231: Shouldn't this be calling something like self.client.read().idm_account_device_auth_grant_verify()? Or are you going to add that to the interface in a subsequent PR?

dmulder: I did that for Himmelblau. I wasn't planning to implement this for Kanidm. So yes, this would be done in a subsequent PR, but not necessarily by me.

Firstyear: I'm confused how you did this in himmelblau? The structure of this code currently has the resolver passing the requests back-forth between the pam module and the idp client. How did you get this to work in himmelblau without this being implemented to make a call into the Idp client?

dmulder: Why would Himmelblau use `KanidmClient`? The 'client' backend of Himmelblau is a `PublicClientApplication` from the Microsoft Authentication Library (msal). I'm in the process of switching this backend to instead use a `ConfidentialClientApplication` to support device enrollment.
Himmelblau doesn't implement all the features of `KanidmClient` because it doesn't need them. Take a look at the source in https://github.com/openSUSE/himmelblau/blob/main/src/common/src/idprovider/himmelblau.rs and see for yourself.

dmulder: We've got a meeting scheduled for this week. If you can make that, then perhaps we can hash this out in that meeting?

Firstyear: It's not using the kani client - this is the resolver calling the *trait* for a client implementation. Kanidm happens to implement that trait, as would himmelblau. So you need to update https://github.com/kanidm/kanidm/blob/master/unix_integration/src/idprovider/interface.rs with the DAG interaction traits, then each implementation can choose to use or ignore it.

dmulder: You said I should implement something like `idm_account_device_auth_grant_verify`, which I assumed meant something similar to functions like `idm_account_unix_cred_verify` implemented in the **struct KanidmClient**. Your initial comment was not clear that you meant I should implement something in the **IdProvider trait**.
So do I understand now, that you mean we should have a function for handling DAG in the IdProvider trait?

dmulder: The way I've implemented this in Himmelblau is that `unix_user_online_auth_step` currently both begins the DAG and completes the DAG, depending on the conditions. In Himmelblau, we can auth via username and password, and then we get a error response if the user requires MFA. If `unix_user_online_auth_step` receives this error response (REQUIRES_MFA), it returns:
```rust
                    return Ok(AuthResult::Next(AuthRequest::DeviceAuthorizationGrant {
                        data: resp.into(),
                    }));
```
Where `data` is the response from an `initiate_device_flow` to the client.
This request is forwarded back to pam, which sends the requested prompt and then calls back to the daemon and eventually leads us back to `unix_user_online_auth_step` again. `unix_user_online_auth_step` matches against the received `(cred_handler, pam_next_req)` and checks for a `DeviceAuthorizationGrant`, at which point it begins polling for a response from the server.

Firstyear: > So do I understand now, that you mean we should have a function for handling DAG in the IdProvider trait? 

You know, I think what happened is I had confused myself. I had *completely* misread that you implemented this section in the Kanidm.rs provider, rather than the trait. I think I need to re-read this whole PR to work out why I confused myself so badly 🤣

Firstyear: Okay I see the mistake I have made - I mistakenly thought this was in the resolver code so I was commenting on a problem that didn't exist. I'm so sorry. Your PR as it currently is is all good.

Firstyear (approved):

PR #2080, opened 2023-09-07
CLI and test things

yaleman:
Started poking around at some more test things for end-to-end testing, and had issues with a lack of response codes - added `handle_client_error` to the CLI which throws a non-zero exit code when we get a 500 back from the server.

Checklist

- [x] cargo fmt has been run

### DIFF
--- a/scripts/test_run_release_server.sh
+++ b/scripts/test_run_release_server.sh
@@ -33,7 +33,7 @@ cargo run --bin kanidmd server --config ../../examples/insecure_server.toml &
 
 sleep 5
 
-cd server/daemon/ && ../../scripts/setup_dev_environment.sh
+../../scripts/setup_dev_environment.sh
 
 fg
 
--- a/scripts/test_run_release_server.sh
+++ b/scripts/test_run_release_server.sh
@@ -2,9 +2,11 @@
 
 set -e
 
+WAIT_TIMER=30
+
 terminate_crab () {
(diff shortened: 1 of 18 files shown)


scripts/test_run_release_server.sh
```diff
@@ -0,0 +1,48 @@
...
+    echo "I'm not sure where you are, please run this from the root of the repository or the server/daemon directory"
+    exit 1
+fi
+
+mkdir -p /tmp/kanidm/
+
+echo "Generating certificates..."
+cargo run --bin kanidmd --release cert-generate --config ../../examples/insecure_server.toml
+echo "Running the server..."
+cargo run --bin kanidmd --release server --config ../../examples/insecure_server.toml &
```
Firstyear, line 31: What's the error condition here? Shouldn't we try an admin pw recovery and use the success/failure of that to determine is this is a pass/fail?

yaleman: It starts the server so the setup dev environment script can work so it tests "live" database and web things... which we've missed in testing before 😨

Firstyear: Yeah, but my point here is that if the server fails to start do we detect or error here in this script?

yaleman: Ah... I'll double check that.

yaleman: Sorted... I think!

Firstyear (approved):

PR #2082, opened 2023-09-08
Schema-dooby-doo-part-trois

yaleman:
Relates to that other issue I keep forgetting the number for.

MORE TECH DEBT WRANGLING!

Checklist

- [x] cargo fmt has been run

### DIFF
--- a/server/lib/src/constants/acp.rs
+++ b/server/lib/src/constants/acp.rs
@@ -114,7 +114,8 @@ lazy_static! {
 }
 
 lazy_static! {
-    pub static ref E_IDM_SELF_ACP_READ_V1: EntryInitNew = entry_init!(
+    pub static ref E_IDM_SELF_ACP_READ_V1: EntryInitNew =
+        entry_init!(
         (ATTR_CLASS, EntryClass::Object.to_value()),
         (ATTR_CLASS, EntryClass::AccessControlProfile.to_value()),
         (ATTR_CLASS, EntryClass::AccessControlSearch.to_value()),
@@ -129,7 +130,6 @@ lazy_static! {
         (ATTR_ACP_RECEIVER_GROUP, Value::Refer(UUID_IDM_ALL_ACCOUNTS)),
         (
             ATTR_ACP_TARGET_SCOPE,
-            // Value::new_json_filter_s("\"self\"").expect("Invalid JSON filter")
             Value::JsonFilt(ProtoFilter::SelfUuid)
         ),
         (ATTR_ACP_SEARCH_ATTR, Attribute::Class.to_value()),
--- a/server/lib/src/constants/acp.rs
+++ b/server/lib/src/constants/acp.rs
@@ -17,6 +17,24 @@ pub struct BuiltinAcp {
     receiver_group: Uuid,
     target_scope: ProtoFilter,
     search_attrs: Vec<Attribute>,
+    modify_removed_attrs: Vec<Attribute>,
+    modify_classes: Vec<EntryClass>,
+}
+
+impl Default for BuiltinAcp {
+    fn default() -> Self {
+        Self {
+            classes: Default::default(),
+            name: Default::default(),
+            uuid: Default::default(),
+            description: Default::default(),
+            receiver_group: Default::default(),
+            search_attrs: Default::default(),
+            modify_removed_attrs: Default::default(),
+            modify_classes: Default::default(),
+            target_scope: ProtoFilter::SelfUuid,
+        }
+    }
 }
 
 impl From<BuiltinAcp> for EntryInitNew {
@@ -44,6 +62,12 @@ impl From<BuiltinAcp> for EntryInitNew {
                 .map(|sa| sa.to_value())
                 .collect::<Vec<Value>>(),
         );
+        value.modify_removed_attrs.into_iter().for_each(|attr| {
+            entry.add_ava(Attribute::AcpModifyRemovedAttr.as_ref(), attr.to_value());
+        });
+        value.modify_classes.into_iter().for_each(|class| {
+            entry.add_ava(Attribute::AcpModifyClass.as_ref(), class.to_value());
+        });
         entry
     }
 }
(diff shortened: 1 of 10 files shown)


server/lib/src/constants/acp.rs
```diff
@@ -17,6 +17,24 @@ pub struct BuiltinAcp {
...
+        Self {
+            classes: Default::default(),
+            name: Default::default(),
+            uuid: Default::default(),
+            description: Default::default(),
+            receiver_group: Default::default(),
+            search_attrs: Default::default(),
+            modify_removed_attrs: Default::default(),
+            modify_classes: Default::default(),
+            target_scope: ProtoFilter::SelfUuid,
```
Firstyear, line 35: Not sure how I feel about this. Maybe it should be an empty filter instead so nothing matches?

server/lib/src/constants/acp.rs
```diff
@@ -110,8 +129,8 @@ lazy_static! {
         (ATTR_ACP_RECEIVER_GROUP, Value::Refer(UUID_IDM_ALL_ACCOUNTS)),
         (
             ATTR_ACP_TARGET_SCOPE,
-            Value::new_json_filter_s("\"self\"").expect("Invalid JSON filter")
-            // Value::JsonFilt(ProtoFilter::SelfUuid)
+            // Value::new_json_filter_s("\"self\"").expect("Invalid JSON filter")
```
Firstyear, line 132: Delete the comment?

server/lib/src/constants/entries.rs
```diff
@@ -642,623 +643,526 @@ impl EntryClass {
...
+#[derive(Clone, Debug, Default)]
 /// Built-in group definitions
 pub struct BuiltinGroup {
     pub name: &'static str,
     description: &'static str,
-    classes: Vec<EntryClass>,
     uuid: uuid::Uuid,
-    member: uuid::Uuid,
+    members: Vec<uuid::Uuid>,
+    dyngroup: bool,
```
Firstyear, line 653: I think dyngroup should be seperate to BuiltinGroup since a built in dyngroup should have *no* members hardcoded.

server/lib/src/constants/entries.rs
```diff
@@ -642,623 +643,526 @@ impl EntryClass {
...
+    dyngroup_filter: Option<Filter>,
+    // TODO: additional attributes (for things like the uihint group)
+    extra_attributes: Vec<(Attribute, Value)>,
```
Firstyear, line 656: Do we have other extra attrs beside uihints? Could we just list uihints as an option here?

server/lib/src/macros.rs
```diff
@@ -60,28 +60,28 @@ macro_rules! setup_test {
...
-            .map(|s| Value::new_spn_str(s, "example.com"))
-            .expect("Failed to munge spn from name!");
-        e.set_ava(Attribute::Spn.as_ref(), once(spn));
-
-        let e = e.into_sealed_committed();
-
-        Account::try_from_entry_no_groups(&e).expect("Account conversion failure")
-    }};
-}
+// #[cfg(test)]
```
Firstyear, line 63: Woooo hoo lets delete this hack.

server/lib/src/server/modify.rs
```diff
@@ -556,18 +556,19 @@ mod tests {
...
-            ModifyList::new_list(vec![Modify::Present(
-                Attribute::Description.into(),
-                Value::from("anusaosu"),
-            )]),
-        );
-        assert!(server_txn.modify(&me_nochg) == Err(OperationError::NoMatchingEntries));
+        // TODO: @yaleman fix this because we don't have a way to do this anymore
+        // let me_nochg = ModifyEvent::new_impersonate_entry_ser(
+        //     JSON_ADMIN_V1 ,
+        //     filter!(f_eq(
```
Firstyear, line 562: Same as before? Manually create the FC to bypass this?

Firstyear (changes requested):
Small changes but looks amazing :)

server/lib/src/constants/acp.rs
yaleman: Ah, I *knew* there were some things I'd forgotten to check with you on :D

How do we define an "empty" filter in this context?

server/lib/src/constants/acp.rs
yaleman: done!

server/lib/src/constants/acp.rs
Firstyear: Try `And(vec::with_capacity(0))`. An And filter with no elements should never match anything per entry_match_no_index_inner in server/lib/src/entry.rs. Specifically if you have a rust iter with .all() and no elements contained then it returns false https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.any

Firstyear: image

I was there when the deep magic was written.

server/lib/src/constants/entries.rs
yaleman: I don't know (yet) but I figured it would be easier at this stage of the yak-shaving

server/lib/src/constants/entries.rs
yaleman: I was trying to limit the number of new types to define, especially since later we're doing a Vec<BuiltinGroup> to build the schema, which could lead to some messy dependencies around which order things get added to the database - I've added a runtime check creating these groups with members defined.

server/lib/src/macros.rs
yaleman: less hax moar betters

server/lib/src/server/modify.rs
yaleman: it was more the JSON_ADMIN_V1 bit because I hadn't worked out the account objects yet :smile: All good

server/lib/src/constants/entries.rs
Firstyear: I think that's a reasonable compromise.

server/lib/src/constants/entries.rs
Firstyear: Fair :)

Firstyear (approved):

PR #2108, opened 2023-09-13
Cinco de yakko

yaleman:
Refers to #1987

Checklist

- [x] cargo fmt has been run


server/lib/src/entry.rs
```diff
@@ -159,16 +159,22 @@ pub struct EntryReduced {
     uuid: Uuid,
 }
 
+// One day this is going to be Map<Attribute, ValueSet> - @yaleman
+// pub type Eattrs = Map<Attribute, ValueSet>;
```
Firstyear, line 163: How will that work with custom attributes that users define? Will we need a way for them to make and assign there or wrap these somehow?

Firstyear (approved):

server/lib/src/entry.rs
yaleman: Yeah, I figure we'll have a `Attribute::CustomAttribute(String)` or something like that?

Probably good to auto-prepend "custom_" on the name when they're created, and maybe hide it when displaying, or something?

yaleman:
I'll hold this one and rebase it once you're done with the attrs thing you mentioned.

Firstyear:
That's merged now, so you should be able to rebase :)

PR #2110, opened 2023-09-15
68 20230912 session consistency

Firstyear:
Relates #68 - This adds support for special-casing sessions in replication to allow them to internally trim and merge so that session revocations and creations are not lost between replicas. This is important for security and user experience. As usual this adds extensive tests of all the states and trimming related to how it will work under replication. Since this also tidies up some internal state about sessions it also allows better handling of logouts and revocations in some edge cases (login with immediate logout would fall in the grace window meaning the session had a tiny valid window. This is now closed). 

Checklist

### DIFF
--- a/server/lib/src/idm/account.rs
+++ b/server/lib/src/idm/account.rs
@@ -230,6 +230,8 @@ impl Account {
         // ns value which breaks some checks.
         let ct = ct - Duration::from_nanos(ct.subsec_nanos() as u64);
         let issued_at = OffsetDateTime::UNIX_EPOCH + ct;
+        // Note that currently the auth_session time comes from policy, but the already-privileged
+        // session bound is hardcoded.
         let expiry =
             Some(OffsetDateTime::UNIX_EPOCH + ct + Duration::from_secs(auth_session_expiry as u64));
         let limited_expiry = Some(
@@ -251,13 +253,7 @@ impl Account {
                 // These sessions are always rw, and so have limited life.
                 (UatPurpose::ReadWrite { expiry }, limited_expiry)
             }
-            SessionScope::PrivilegeCapable =>
-            // Return a rw capable session with the expiry currently invalid.
-            // These sessions COULD live forever since they can re-auth properly.
-            // Today I'm setting this to 24hr though.
-            {
-                (UatPurpose::ReadWrite { expiry: None }, expiry)
-            }
+            SessionScope::PrivilegeCapable => (UatPurpose::ReadWrite { expiry: None }, expiry),
         };
 
         Some(UserAuthToken {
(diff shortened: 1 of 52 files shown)


Firstyear:
Oh and as always - even though we change some DB types, it's a transparent forward migration :)

server/lib/src/macros.rs
```diff
@@ -486,9 +486,11 @@ macro_rules! mergemaps {
...
     ) => {{
         $b.iter().for_each(|(k, v)| {
-            if !$a.contains_key(k) {
-                $a.insert(k.clone(), v.clone());
-            }
+            // I think to be consistent, we need the content of b to always
+            // the content of a
+            // if !$a.contains_key(k) {
+            $a.insert(k.clone(), v.clone());
+            // }
```
yaleman, line 493: This doesn't merge them in the way I'd expect, since if a contains the key, and the value's a `Vec` it *overwrites* it rather than extending, etc...?

more a `mv -R b/* a/*` than `rsync -a b/* a/`

server/lib/src/idm/account.rs
```diff
@@ -243,17 +249,14 @@ impl Account {
...
             SessionScope::ReadWrite => {
                 // These sessions are always rw, and so have limited life.
-                (UatPurpose::ReadWrite { expiry }, expiry)
+                (UatPurpose::ReadWrite { expiry }, limited_expiry)
             }
             SessionScope::PrivilegeCapable =>
             // Return a rw capable session with the expiry currently invalid.
             // These sessions COULD live forever since they can re-auth properly.
             // Today I'm setting this to 24hr though.
             {
```
yaleman, line 258: This comment probably needs an update for clarity

yaleman (approved):

yaleman (approved):

server/lib/src/macros.rs
Firstyear: Yeah, think the argument order may not match command line expectations, but from the view of rust/api you can think of it like the "extend" function, which is a.extend(b), where the content of b that overlaps replaces the content of b. The same is true here, and it's how we use it, for example https://github.com/kanidm/kanidm/blob/master/server/lib/src/valueset/ssh.rs#L161 where the intent is that the content of "other" should be merged ontop of the content of self.

Merge generally actually is only used in a few places though btw. It's used for merging memberof reference sets (which are all just uuids anyway), and it's used in set_ava in entry where the existing attribute state is to be extended from an iterator/valueset.

Firstyear:
This adds support for special-casing sessions in replication to allow them to internally trim and merge so that session revocations and creations are not lost between replicas.

Issue #2057, opened 2023-09-03
In-system file storage


yaleman:
**Is your feature request related to a problem? Please describe.**

For #1966, #2056 (and other uses) we need to be able to store images and serve them to users via the HTTPS API.

**Describe the solution you'd like**

Images stored in the database as a class type attached to the relevant entry. 

I'd say a max of 256kb? Technically there's no real limit on cell sizes in SQLite but... urgh. 

This requires:

- [x] New EntryClass (Image?)
- [ ] CRUD API endpoints updates for modifying these entries
- [x] Validation of supported file formats (PNG, JPG, WEBP, SVG?) and max filesize/image size
- [ ] CLI management for 
  - [x] oauth2 app images
  - [ ] user images

- [ ] WebUI management
  - [ ] oauth2 app images
  - [ ] user images 
  - [ ] Crop/preview functionality

Access to images can be controlled with the ACP on the entry.

If they're changing regularly (as in the case of user profile images) that could turn into an issue requiring more regular vacuums due to the blob size, but it'll just take some testing.

**Describe alternatives you've considered**

- s3-compatible object storage with [presigned URLs](https://docs.aws.amazon.com/AmazonS3/latest/userguide/using-presigned-url.html) on retrieval.
- File system storage/serving with a modified replication backend (permissions? Access controls? nooooope!)

**Additional context**

Firstyear:
I actually think we should NOT use sqlite blobs here. We don't have any caching mechanism for these at all, so they'd need to always be served from disk.

I'd rather have these blobs stored in entries with a dedicated class type, then on things like the oauth2 applink they can reference the resource id.

This would also allow us to enforce referential integrity that you can't delete a blob if it's still in use somewhere! It also means these blob-entries would be in our in-memory cache and able to be served much quicker, and they automatically gain replicaiton. It also lets admins manage them in the same way as anything else, and we can define acess controls etc. So lots of nice benefits.

Firstyear:
It also means on the blob-serving path we can have a special handler to either allow-all reads of blobs, or we can actually still enforce access controls. They can also be marked as cached then too.

Firstyear:
@yaleman Is this finished now? What's outstanding?

yaleman:
At the moment it's only hacked together to upload things via the API, and no return path for the data to users, so we need to do all the un-ticked things

PR #2112, opened 2023-09-17
In-system image storage

yaleman:
Refers to #2057, starting to store images for OAuth2 RSen.

So far you can upload, delete and retrieve the image data, with tests covering validating image formats.

Checklist

- [x] cargo fmt has been run


server/lib/src/be/dbvalue.rs
```diff
@@ -651,6 +652,8 @@ pub enum DbValueSetV2 {
     AuditLogString(Vec<(Cid, String)>),
     #[serde(rename = "EK")]
     EcKeyPrivate(Vec<u8>),
+    #[serde(rename = "IM")]
+    Image(Vec<ImageValue>),
```
Firstyear, line 656: Normally we keep the DB values as seperate enums to proto or anything else, so that if we have to change the proto representation, we keep the DB version stable. This lets us create forward migrations. So I think you need a "DB version" of the ImageValue here.

server/lib/src/repl/proto.rs
```diff
@@ -6,6 +6,7 @@ use crate::prelude::*;
 use crate::schema::{SchemaReadTransaction, SchemaTransaction};
 use crate::valueset;
 use base64urlsafedata::Base64UrlSafeData;
+use kanidm_proto::internal::ImageValue;
```
Firstyear, line 9: Could also be good to split the internal representation from the proto one in case we end up with multiple proto versions that our internal rep has to satisfy in future.

server/lib/src/schema.rs
```diff
@@ -1763,6 +1765,21 @@ impl<'a> SchemaWriteTransaction<'a> {
                 syntax: SyntaxType::Uint32,
             },
         );
+        self.attributes.insert(
```
Firstyear, line 1768: Attrs in schema.rs are intended to be "critical" to boot the server, and everything else is loaded via migrations. So does this need to be here? Does a "critical" element in the schema depend on this?

server/lib/src/server/mod.rs
```diff
@@ -681,6 +683,7 @@ pub trait QueryServerTransaction<'a> {
                         }),
                     SyntaxType::AuditLogString => Ok(PartialValue::new_utf8s(value)),
                     SyntaxType::EcKeyPrivate => Ok(PartialValue::SecretValue),
+                    SyntaxType::Image => Ok(PartialValue::new_image(value)),
```
Firstyear, line 686: Should probably actually PartialValue to utf8 to match on the tags. PartialValue doesn't have ta equate to the Value itself, since it's just a selector. Look at schema.rs validate for PartialValue. You'll see that Oauth2 accepts Refer as PartialValue for example.

server/lib/src/value.rs
```diff
@@ -387,6 +391,8 @@ pub enum PartialValue {
     UiHint(UiHint),
     Passkey(Uuid),
     DeviceKey(Uuid),
+    /// We compare on the value hash
+    Image(String),
```
Firstyear, line 395: Why the hash? Could there be other ways to compare?

server/lib/src/valueset/mod.rs
```diff
@@ -83,6 +86,7 @@ pub type ValueSet = Box<dyn ValueSetT + Send + Sync + 'static>;
 dyn_clone::clone_trait_object!(ValueSetT);
 
 pub trait ValueSetT: std::fmt::Debug + DynClone {
+    // TODO: uuhhhh.... is the bool a "yeah, it worked?" or a "yeah, it was already there?"
```
Firstyear, line 89: Follows btreeset in std lib. 

"""
Returns whether the value was newly inserted. That is:

    If the set did not previously contain an equal value, true is returned.
    If the set already contained an equal value, false is returned, and the entry is not updated.

""'

server/lib/src/valueset/mod.rs
```diff
@@ -562,6 +566,12 @@ pub trait ValueSetT: std::fmt::Debug + DynClone {
         None
     }
 
+    // TODO: not sure if I really need this
+    fn as_imageset(&self) -> Option<&HashSet<ImageValue>> {
```
Firstyear, line 570: It's useful to have in a few places tbh.

server/lib/src/valueset/image.rs
```diff
@@ -0,0 +1,466 @@
...
+    fn validate_is_gif(&self) -> Result<(), ImageValidationError>;
+    fn validate_is_jpg(&self) -> Result<(), ImageValidationError>;
+    fn validate_is_webp(&self) -> Result<(), ImageValidationError>;
+    fn validate_is_svg(&self) -> Result<(), ImageValidationError>;
+
+    /// A sha256 of the filename/type/contents
+    fn hash_imagevalue(&self) -> String;
+
+    fn get_limits(&self) -> image::io::Limits {
+        let mut limits = image::io::Limits::default();
```
Firstyear, line 38: We likeuly need a plan on how to have these configurable in future.

server/lib/src/be/dbvalue.rs
yaleman: Can do!

server/lib/src/value.rs
yaleman: Yeah, I was trying to avoid collisions of filename/type/contents, since you and I might upload `logo.png` with different contents for two different things... I didn't know if this was an in-db representation or something else?

Firstyear: Well I think it depends how we end up structuring this internally too. Is this going to "inline" the image attribute on the entry itself? For example:

```
oauth2_rs: name
image: <logo>
```

Or is it going to be a generic blob store and it references?

```
oauth2_rs: name
image: <uuid> (or should it be a hash to the image? Would this be a new type of reference?)

image: <hash>
uuid: <uuid>
```

I feel the blob store might be the better approach here ... 🤔 And then yeah, we reference it by hash of the blob contents. Or we just reference it by asset uuid. 

I think the blob store also is better wrt to access controls? That any authenticated user can "fetch" a blob if you know it's hash, but you have to know the hash? Is there any reason we'd want to "hide" logos? I think querying all logos by any user shouldn't be allowed though ....

server/lib/src/schema.rs
yaleman: Weeeelllll it could be once we add a system-wide logo that'd probably be tied into domain_info?

server/lib/src/valueset/image.rs
yaleman: Yeah, I'd say it'll be yet another replicable system setting of some kind.

Firstyear: In that case it should be in the fn validate of the valueset because I think we have access to the query server at that point?

Firstyear: Or we can make the limits come from the schema for example. I dunno. Thought needed. For now leave it like this.

server/lib/src/value.rs
Firstyear: did we ever make a design doc for this?

yaleman: I'd consider: 

* logos of oauth2 RSen to be a secret, the same way we hide OAuth apps to users who don't have access to them, 'cause ... sekrets.
* system logo - which is public
* User profile pics I'm guessing we'll end up with "opt-in public" or maybe a system default with opt out/in public options?

It kinda makes sense to make images separate somehow in the DB/storage, but then that gets into its own complexity journey with "who owns what, and what permissions do we apply to CRUD them?". It'd be handy if they were a foreign reference to save on memory, but inextricably linked so one doesn't end up with unowned blobs in the system storage.

Firstyear: Thats true. I think we inline them on entries then. If we ever need to truly save memory, we can de-dupe them another way. I think this inlining also means we can re-use our existing oauth2 permission controls we have where users who are elligble to an rs get read access to limited attrs of it.

Even though this would be authenticated, would it still be cacheable 🤔

server/lib/src/repl/proto.rs
yaleman: Created `ReplImageV1`

Firstyear: I think in future the repl proto and the db proto are really similar, since they both are just about serialisation. So I might end up deduplicating-merging them into a serialisable format crate.

yaleman: Seems sensible to me, I can make them the same in this instance?

server/lib/src/server/mod.rs
yaleman: I'm not sure what you mean by "match on the tags" here :)

server/lib/src/repl/proto.rs
Firstyear: Yeah, lets do that for now. I'll have to commit to the repl proto cleanup before we stamp as good for production anyway.

server/lib/src/value.rs
yaleman: The size is mostly a `Vec<u8>` so my understanding is that it's shoved off into the heap anyhow, which saves memory things?

We can set the image to be (quickly) cacheable on the client side to save requests for the big object, I'm not sure how the internal cache works for Kanidm :D

server/lib/src/valueset/image.rs
yaleman: Yeah, as a first-run this'll work :D

server/lib/src/value.rs
Firstyear: Yes and no. It's just where the allocation exists. What actually saves memory here is that entries in the kanidm db cache are wrapped with Arc<> and shared by read transactions, so there is "one" allocation of the image in the heap, that many threads can all be sharing and accessing. It's only when the entry is written to that we need to duplicate the content of the image in memory for the new write (copy-on-write afterall). 

So given that oauth2-rs are a "low write" target, then mostly there will just be one copy of the image in memory at a time, associated to any oauth2-rs. Similar for the domain.

The exception here is persons/accounts because they ARE a high frequency write target, because of credential updates and session updates as they login/out. So that may cause some impacts on memory/cloning, but honestly probably nothing too wild. 

I think we do it the simple way like this first, and if it becomes a problem, then we find a way to transparently cache these blobs in a separate manner / storage pool. But that can be added later with migrations etc.

server/lib/src/valueset/image.rs
```diff
@@ -101,14 +107,80 @@ enum ChunkStatus {
...
+    #[cfg(any(debug_assertions, test))]
+    trace!(
+        "length_bytes: {:?} length: {} chunk_type: {:?} buflen: {}",
+        length_bytes,
+        &length,
+        &chunk_type,
+        &buf.len()
+    );
+
+    let (_, buf) = buf.split_at(length as usize);
```
Firstyear, line 139: length here is input controlled, not controlled by us. So as a result, we need to validate at this point that length <= buf.len else split_at will panic.

server/lib/src/valueset/image.rs
```diff
@@ -101,14 +107,80 @@ enum ChunkStatus {
...
+    #[cfg(any(debug_assertions, test))]
+    trace!("new buflen: {}", &buf.len());
+
+    let (_checksum, buf) = buf.split_at(4);
```
Firstyear, line 143: Similar here, since we now advanced past the end of buf, we don't know if we have 4 bytes, so we need to check that buf.len() >= 4 here.

Firstyear (changes requested):

server/lib/src/valueset/image.rs
yaleman: yep, un-pushed change has these checks 😄

server/lib/src/valueset/image.rs
yaleman: yep, un-pushed change has these checks 😄

proto/src/internal.rs
```diff
@@ -45,3 +48,91 @@ pub enum IdentifyUserResponse {
     CodeFailure,
     InvalidUserId,
 }
+
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize, Hash, Ord, PartialOrd)]
+#[serde(rename_all = "lowercase")]
+pub enum ImageType {
+    Png,
+    Jpg,
+    Gif,
```
Firstyear, line 57: I am so keen for animated gifs.

proto/src/internal.rs
```diff
@@ -45,3 +48,91 @@ pub enum IdentifyUserResponse {
...
+
+impl TryFrom<&str> for ImageValue {
+    type Error = String;
+    fn try_from(s: &str) -> Result<Self, String> {
+        serde_json::from_str(s)
+            .map_err(|e| format!("Failed to decode ImageValue from {} - {:?}", s, e))
+    }
+}
+
+impl core::hash::Hash for ImageValue {
```
Firstyear, line 122: std::hash::Hash.

Also you can derive Hash on ImageValue instead :)

server/core/src/https/mod.rs
```diff
@@ -208,7 +208,8 @@ pub async fn create_https_server(
                 .route("/", get(|| async { Redirect::temporary("/ui") }))
                 .route("/manifest.webmanifest", get(manifest::manifest))
                 .nest("/ui", spa_router)
-                .layer(middleware::compression::new()) // TODO: this needs to be configured properly
+                .layer(middleware::compression::new())
+                .route("/ui/images/oauth2/:rs_name", get(oauth2::oauth2_image_get))
```
Firstyear, line 212: Should this route be *above* compression? Or does it need to be after?

Firstyear, line 212: Should this be in ui? Or should it just be a get on /v1/oauth2/:rs_name/image?

Firstyear: Is this route cacheable?

server/lib/src/idm/oauth2.rs
```diff
@@ -238,6 +239,8 @@ pub struct Oauth2RS {
     scopes_supported: BTreeSet<String>,
     prefer_short_username: bool,
     type_: OauthRSType,
+    /// the display image for the OAuth2 Resource
+    display_image: Option<ImageValue>,
```
Firstyear, line 243: Does this need to be here in memory? I think we are serving the image from the entry directly, rather than the in-memory oauth2 resource here, so I think this should just be a bool for if the image is set or not?

server/lib/src/schema.rs
```diff
@@ -229,6 +229,10 @@ impl SchemaAttribute {
             // Comparing on the label.
             SyntaxType::TotpSecret => matches!(v, PartialValue::Utf8(_)),
             SyntaxType::AuditLogString => matches!(v, PartialValue::Utf8(_)),
+            SyntaxType::Image => {
+                dbg!("validate_partialvalue_image {}", v);
```
Firstyear, line 233: trace not dbg

server/lib/src/valueset/image.rs
```diff
@@ -0,0 +1,663 @@
...
+        }
+    }
+}
+
+#[derive(Debug)]
+enum ChunkStatus {
+    SeenEnd { has_trailer: bool },
+    MoreChunks,
+}
+#[test]
```
Firstyear, line 108: Should this test be at the foot of the file? Seems messy up here ...

server/lib/src/valueset/image.rs
```diff
@@ -0,0 +1,663 @@
...
+
+fn png_has_trailer(png_data: &Vec<u8>) -> Result<bool, ImageValidationError> {
+    let buf = png_data.as_slice();
+    // let magic = buf.split_off(PNG_PRELUDE.len());
+    let (magic, buf) = buf.split_at(PNG_PRELUDE.len());
+
+    let buf = buf.to_owned();
+    let mut buf = buf.as_slice();
+
+    dbg!(&magic, &PNG_PRELUDE);
```
Firstyear, line 206: trace

server/lib/src/valueset/image.rs
```diff
@@ -0,0 +1,663 @@
...
+        hasher.update(self.filename.as_bytes());
+        hasher.update(&filetype_repr);
+        hasher.update(&self.contents);
+        hex::encode(hasher.finish())
+    }
+}
+
+#[test]
+/// tests that we can load a bunch of test images and it'll throw errors in a way we expect
+fn test_imagevalue_things() {
```
Firstyear, line 360: Very on board for these tests 🎉

server/lib/src/valueset/image.rs
```diff
@@ -0,0 +1,663 @@
...
+    }
+}
+
+impl ValueSetT for ValueSetImage {
+    fn insert_checked(&mut self, value: Value) -> Result<bool, OperationError> {
+        match value {
+            Value::Image(image) => match image.validate_image() {
+                Ok(_) => Ok(self.set.insert(image)),
+                Err(err) => {
+                    admin_error!(
```
Firstyear, line 498: There is a separate validation phase though, so is this needed here?

server/testkit/tests/proto_v1_test.rs
```diff
@@ -958,6 +960,75 @@ async fn test_server_rest_oauth2_basic_lifecycle(rsclient: KanidmClient) {
 
     assert!(oauth2_config_updated2 != oauth2_config_updated3);
 
+    // Check we can upload an image
+    let image_path = Path::new("../../server/lib/src/valueset/test_images/ok.png");
+    dbg!(&image_path.canonicalize());
```
Firstyear, line 965: trace?

Firstyear (changes requested):
The only missing thing here is the acp rule to allow users to read the image. You can look at server/lib/src/server/access/search.rs to see how we synthesise the read access for members of an oauth2 rs that have the scopes granted to them.

server/lib/src/idm/oauth2.rs
yaleman: Done!

server/core/src/https/mod.rs
yaleman: Yeah I was tossing up between the load of attempting to compress the image again or not, since ... they should already be compressed - below the middleware removes the compression attempt which is probably better for now (testing to come later, but ... finger in the wind it's not beneficial.)

yaleman: I... don't think we do the caching thing properly at the moment >.>

And it's a UI-related thing more than an API thing because it doesn't respond with a JSON blob - and it's only user-facing-UI related.

server/lib/src/schema.rs
yaleman: Yeet not keep :D

server/lib/src/valueset/image.rs
yaleman: 'tis now, whoops. :smile:

server/testkit/tests/proto_v1_test.rs
yaleman: yeet

server/lib/src/valueset/image.rs
yaleman: I'm pretty sure this could just call self.set.clear - it's no-mv, so if you want to remove anything then it's only going to be one thing

Firstyear: Well, we need to assume for now it could be mv. I want to change syntax of attrs in future that the attr type defines if it's mv or not, but for now we're stuck wuth everything being possible mv.

yaleman: ezpz, can sort that out

yaleman:
I ... think this is ready to go now 😄

server/core/src/actors/v1_read.rs
```diff
@@ -402,6 +403,53 @@ impl QueryServerReadV1 {
...
+
+        let filter_orig = filter_all!(f_self())
+            .validate(idms_prox_read.qs_read.get_schema())
+            .map_err(OperationError::SchemaViolation)?;
+        let filter = rs
+            .clone()
+            .validate(idms_prox_read.qs_read.get_schema())
+            .map_err(OperationError::SchemaViolation)?;
+
+        let search = SearchEvent {
```
Firstyear, line 433: Avoid building this manually. I think you'll break access control here becaus filter_orig and filter have extremely precise and subtle meanings. 

There are searchEvent builders that can help you here, and you should use one of them.

The way this works is filter_orig is the "filter as given by the user". then filter is "the filter as we are going to actually execute it, with extra conditionals provided that MAY bypass the users access controls. 

This is that in filter we can wrap with class exclusions for tombstone/recycled, even if the account *can not* read "class". But then we evaluate access based on "filter_orig" which is what the user actually asked for. 

So in this case you bypass access controls because you do filter_orig (f_self()) which is allowed, but then filter as executed is completely different.

Firstyear: And yes, the searchEvent builders methods do this all for you correctly, so you proably want `SearchEvent::from_internal_message(ident, rs, attrs, &idms_prox_read.qs_read)` here.

Firstyear (changes requested):
Almost there, just need to fix that SearchEvent builder.

Firstyear (approved):

Firstyear:
Well done! Thanks so much for your hardwork on this, it's going to make Kanidm look so much more professional :)

server/core/src/actors/v1_read.rs
yaleman: Done!

PR #2127, opened 2023-09-19
CLI and kanidm_client changes to handle errors and TLS validation changes

yaleman:
Changes

* Adds new `ClientError` value `UntrustedCertificate(String)` - the String is because hyper comes back with all sorts of random things based on the underlying libraries, which can't be relied upon.
* Adds the option to explicitly skip hostname verification on the command line instead of in-config. 
* Removes `exitcode` as a dependency because it's a waste of code.

Checklist

- [x] cargo fmt has been run

### DIFF
--- a/tools/cli/src/opt/kanidm.rs
+++ b/tools/cli/src/opt/kanidm.rs
@@ -49,6 +49,13 @@ pub struct CommonOpt {
     /// Log format (still in very early development)
     #[clap(short, long = "output", env = "KANIDM_OUTPUT", default_value = "text")]
     output_mode: OutputMode,
+    /// Skip hostname verification
+    #[clap(
+        long = "skip-hostname-verification",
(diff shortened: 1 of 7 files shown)


tools/cli/src/opt/kanidm.rs
```diff
@@ -49,6 +49,13 @@ pub struct CommonOpt {
     /// Log format (still in very early development)
     #[clap(short, long = "output", env = "KANIDM_OUTPUT", default_value = "text")]
     output_mode: OutputMode,
+    /// Skip hostname verification
+    #[clap(
```
Firstyear, line 53: Should this be a hidden flag?

Firstyear (approved):

tools/cli/src/opt/kanidm.rs
yaleman: 🤷🏻 it's a configuration option, I don't see a reason to hide it.

PR #2128, opened 2023-09-19
[DRAFT] Restrict posix passwords on ldap bind with config

### LINKED ISSUE #2027: Prevent posix password from operating with LDAP without a config option
Once we have #41, I think that we should guard posix password checking via LDAP with a config/domain config flag. This is because app passwords are a better solution to legacy application integration, and we don't want people to reach for these in common cases. They should be a "last resort".

NavinShrinivas:
draft fix #2027 
A draft PR, tests fmt clippy have not been confirmed yet.
Checklist

- [ ] cargo fmt has been run


server/lib/src/constants/entries.rs
```diff
@@ -361,6 +363,18 @@ impl From<Attribute> for &'static str {
     fn from(val: Attribute) -> Self {
         match val {
             Attribute::Account => ATTR_ACCOUNT,
+            Attribute::SystemMay => ATTR_SYSTEMMAY,
```
Firstyear, line 366: @yaleman ?

server/lib/src/constants/schema.rs
```diff
@@ -110,6 +110,16 @@ pub static ref SCHEMA_ATTR_DOMAIN_NAME: SchemaAttribute = SchemaAttribute {
     syntax: SyntaxType::Utf8StringIname,
     ..Default::default()
 };
+pub static ref SCHEMA_ATTR_DOMAIN_LDAP_ALLOW_UNIX_PW_BIND: SchemaAttribute = SchemaAttribute {
+    uuid: UUID_SCHEMA_ATTR_DOMAIN_LDAP_ALLOW_UNIX_PW_BIND,
+    name: Attribute::DomainLdapAllowUnixPwBind.into(),
+    description: "Configuration to allow binds to LDAP objects using UNIX passwords.".to_string(),
+
+    index: vec![IndexType::Equality, IndexType::Presence],
```
Firstyear, line 118: I don't think this needs indexing at all, just unset this.

server/lib/src/constants/schema.rs
```diff
@@ -110,6 +110,16 @@ pub static ref SCHEMA_ATTR_DOMAIN_NAME: SchemaAttribute = SchemaAttribute {
...
+    unique: true,
```
Firstyear, line 119: can't be unique, unset this.

server/lib/src/server/mod.rs
```diff
@@ -793,6 +794,14 @@ pub trait QueryServerTransaction<'a> {
             })
     }
 
+    fn get_domain_ldap_allow_unix_pw_bind(&mut self) -> Result<bool, OperationError> {
+        let res = self.internal_search_uuid(UUID_DOMAIN_INFO);
+        match res.unwrap().get_ava_single("domain_ldap_allow_unix_pw_bind"){
```
yaleman, line 799: needs to be the const, not a string

server/lib/src/idm/ldap.rs
```diff
@@ -1,4 +1,4 @@
-//! LDAP specific operations handling components. This is where LDAP operations
+//! LDAP specific operations handling components. This is where LDAP operationsldap.rs
```
yaleman, line 1: Not sure where this came from? :)

Firstyear, line 1: typo?

NavinShrinivas:
Ignore the random prints, the code is very rough.....only till I get the functionality down

server/lib/src/idm/ldap.rs
NavinShrinivas: me spamming keybinds on nvim xD

server/lib/src/server/mod.rs
NavinShrinivas: you mean the attribute in `get_ava_single()`? Should I be passing the in const that represents the attrribute?

server/lib/src/idm/ldap.rs
```diff
@@ -445,18 +445,36 @@ impl LdapServer {
...
+            debug!(
+                "here {:?}",
+                idms.proxy_read()
+                    .await
+                    .qs_read
+                    .d_info
+                    .d_ldap_allow_unix_pw_bind
+            );
+            let lae = LdapAuthEvent::from_parts(target_uuid, pw.to_string())?;
+            idm_auth.auth_ldap(&lae, ct).await.and_then(|r| {
```
Firstyear, line 464: Why not read this setting for allowing the unique bind here inside of auth_ldap? I think because youi have the proxy read open above that's causing a double-read transaction which is wrong. There should only be *one* transaction per operation. So I'd revert the changes just here, and look into checking this flag in auth_ldap.

server/lib/src/idm/ldap.rs
```diff
@@ -643,15 +661,19 @@ mod tests {
...
         assert!(idms_prox_write.commit().is_ok());
-
         let anon_t = ldaps.do_bind(idms, "", "").await.unwrap().unwrap();
         assert!(anon_t.effective_session == LdapSession::UnixBind(UUID_ANONYMOUS));
         assert!(
             ldaps.do_bind(idms, "", "test").await.unwrap_err() == OperationError::NotAuthenticated
         );
 
         // Now test the admin and various DN's
+        // Writing test for optional use of POSIX password for bind
```
Firstyear, line 676: Probably should make a new test case here, that also tests when the condition is *false*.

server/lib/src/server/mod.rs
```diff
@@ -1500,8 +1510,20 @@ impl<'a> QueryServerWriteTransaction<'a> {
     pub(crate) fn reload_domain_info(&mut self) -> Result<(), OperationError> {
         let domain_name = self.get_db_domain_name()?;
         let display_name = self.get_db_domain_display_name()?;
+        let domain_ldap_allow_unix_pw_bind = match self.get_domain_ldap_allow_unix_pw_bind(){
+            Ok(v) => v,
+            _ => {
+                admin_warn!("Defaulting ldap_allow_unix_pw_bind to false");
+                self.set_domain_ldap_allow_unix_pw_bind(false)?;
+                false
```
Firstyear, line 1518: For now we have to default it to true here. This is because if we defaulted it to false right now, we'd break ldap for a lot of people.

server/lib/src/server/mod.rs
```diff
@@ -1500,8 +1510,20 @@ impl<'a> QueryServerWriteTransaction<'a> {
...
+            }
+        };
         let domain_uuid = self.be_txn.get_db_d_uuid();
         let mut_d_info = self.d_info.get_mut();
+        if mut_d_info.d_ldap_allow_unix_pw_bind != domain_ldap_allow_unix_pw_bind{
+            admin_warn!("Using of unix pw for ldap bind is now allowed : {}", domain_ldap_allow_unix_pw_bind);
```
Firstyear, line 1524: I don't understand this message and why it was added?

server/lib/src/server/mod.rs
yaleman: You probably need to rebase, `get_ava_single` should only take an `Attribute`

server/lib/src/server/mod.rs
```diff
@@ -1537,7 +1559,15 @@ impl<'a> QueryServerWriteTransaction<'a> {
         let filt = filter_all!(f_eq(Attribute::Uuid, udi));
         self.internal_modify(&filt, &modl)
     }
-
+    pub fn set_domain_ldap_allow_unix_pw_bind(
```
Firstyear, line 1562: I don't think you need a dedicated set function here, none of the other domain update bits have it.

Firstyear (changes requested):

server/lib/src/constants/entries.rs
yaleman: Someone needs to rebase on `HEAD`?

server/lib/src/server/mod.rs
NavinShrinivas: Makes sense, on rebasing, get_ava_single accpets Attributes

server/lib/src/idm/ldap.rs
```diff
@@ -642,16 +645,35 @@ mod tests {
 
         let pce = UnixPasswordChangeEvent::new_internal(UUID_ADMIN, TEST_PASSWORD);
 
-        assert!(idms_prox_write.set_unix_account_password(&pce).is_ok());
+        let allow_unix_pw_flag = ModifyEvent::new_internal_invalid(
+            filter!(f_eq(Attribute::Uuid, PartialValue::Uuid(UUID_DOMAIN_INFO))),
+            ModifyList::new_purge_and_set(Attribute::DomainLdapAllowUnixPwBind, Value::Bool(false)),
+        );
+        assert!(idms_prox_write.qs_write.modify(&allow_unix_pw_flag).is_ok());
```
NavinShrinivas, line 652: setting the flag to false here doesnt seem to block the bind in auth ldap...i.e the test on line 666 id failign

server/lib/src/idm/server.rs
```diff
@@ -1361,6 +1361,10 @@ impl<'a> IdmServerAuthTransaction<'a> {
             admin_error!("Failed to start auth ldap -> {:?}", e);
             e
         })?;
+        if self.qs_read.d_info.d_ldap_allow_unix_pw_bind == false {
```
NavinShrinivas, line 1364: this is where the condition should be read as false, but currently is not

server/lib/src/idm/ldap.rs
```diff
@@ -642,16 +645,35 @@ mod tests {
...
             ldaps.do_bind(idms, "", "test").await.unwrap_err() == OperationError::NotAuthenticated
         );
 
         // Now test the admin and various DN's
+
+
+        // Writing test for optional use of POSIX password for bind
+        // NOT ALLOWING UNIX_PW FOR BIND :
+        let admin_t = ldaps.do_bind(idms, "admin", TEST_PASSWORD).await.unwrap(); // Cannot unwrap futher as on disallowed unix_pw_bind it will return Ok(None)
+        assert!(admin_t.is_none() == true);
```
NavinShrinivas, line 666: this test is failing

server/lib/src/server/mod.rs
```diff
@@ -1500,8 +1513,33 @@ impl<'a> QueryServerWriteTransaction<'a> {
     pub(crate) fn reload_domain_info(&mut self) -> Result<(), OperationError> {
         let domain_name = self.get_db_domain_name()?;
         let display_name = self.get_db_domain_display_name()?;
+        let domain_ldap_allow_unix_pw_bind = match self.get_domain_ldap_allow_unix_pw_bind() {
```
Firstyear, line 1516: Here you're trying represent "the possibility of abscence" but you're encoding that into the Result type. Result shows "the possibility of error".

If you want to still do it this way, you want to change the type of get_domain_ldap_allow_unix_pw_bind to `Result<Option<bool>, OperationError>`. This way you show that there is a possibility of error, and if there was no error, that the contained value has a possibility of absecence. Then you con do "unwrap_or(true)" which allows you to set a default value here.

server/lib/src/server/mod.rs
```diff
@@ -1500,8 +1513,33 @@ impl<'a> QueryServerWriteTransaction<'a> {
...
+            Ok(v) => {
+                v
+            },
+            _ => {
+                admin_warn!("Defaulting ldap_allow_unix_pw_bind to true");
+
+                let modl = ModifyList::new_purge_and_set(
```
Firstyear, line 1523: Don't do this internal_modify here - you are *reading* the state of the boolean, not updating it. The reason for this is to consider "on upgrade defaults".

If you *write* to the database, we can't tell if the write was from us or the user, so we have to remain stuck on the default.

But if we allow the bool to be absent from the entry, and we just "unwrap_or(true)", then we can later change the "absent" state from true -> false, and every user gets the default changed. Had you written that to the db we can't do that!

server/lib/src/server/mod.rs
```diff
@@ -1500,8 +1513,33 @@ impl<'a> QueryServerWriteTransaction<'a> {
...
+                    Err(e) => {
+                       return Err(e);
+                    }
+                }
+                true
+            }
+        };
         let domain_uuid = self.be_txn.get_db_d_uuid();
         let mut_d_info = self.d_info.get_mut();
+        if mut_d_info.d_ldap_allow_unix_pw_bind != domain_ldap_allow_unix_pw_bind {
```
Firstyear, line 1540: You don't need this check, just always write it to the structure.

server/lib/src/idm/ldap.rs
```diff
@@ -642,16 +645,35 @@ mod tests {
...
+
+        // ALLOWING UNIX_PW FOR BIND :
+        let mut idms_prox_write = idms.proxy_write(duration_from_epoch_now()).await;
```
Firstyear, line 669: Shouldn't you be setting this to "false" to show what happens if you make it false?

server/lib/src/idm/server.rs
Firstyear: Just set the logic to "what it should be", even if that causes test failures for now.

Firstyear (changes requested):

server/lib/src/idm/ldap.rs
Firstyear: I think your logic here is fliipped btw. when the value is *absent* unix bind should work.
when allow pw bind is true it should work
and when it's false, it should *not* work.

So because you're never setting false, it isn't working I think. I think you need to more clearly seperate out the three cases in the test to match the states that I mentioned (about possibility of abscence, then the two possible bool states).

server/lib/src/server/mod.rs
NavinShrinivas: But if I default the `get` method to return false on absence of an error then i wouldnt have to default anything in this relaod function right?

server/lib/src/server/mod.rs
NavinShrinivas: Ahhh, had my suspicions that the writing defaults to database was causing some messups...this makes sense!

server/lib/src/server/mod.rs
Firstyear: Correct. I was more explaining there are two ways to approach it :)

Firstyear: (Also you want to return true by default as allow unix pw bind is the default for now).

NavinShrinivas:
@Firstyear Should Anon binds happen even if unix binding is not allowed? If I understand right anon binding can be used in kanidm to read the details off of the directory, right?

Firstyear:
> @Firstyear Should Anon binds happen even if unix binding is not allowed? If I understand right anon binding can be used in kanidm to read the details off of the directory, right?

This is an excellent question.

I think the answer is "yes anonymous binds should be allowed" because anonymous binds are a special case in ldap. The goal of this change is to isolate and reduce the use of the unix pw. Since that's the goal, anonymous binds and their existance don't block that.

Anonymous binds can of course be stopped by an alternate control (locking the anonymous account).

NavinShrinivas:
> answer is "yes anonymous binds

Ah great! Test driven developer caught this case for me :). Anyways, this will require me to write a logic checking if the UUID is that of `Anon` user that skip the flag check, I feel this is addition `not clean logic`. Any alternatives?

Firstyear:
We already have to special case the anonymous bind anyway in the simple bind path, so I think it won't require any extra code on your part - just put the check for "is unix bind allowed" in the branch for "when not anonymous".

NavinShrinivas:
Ah got it! Also for tests, I've written the following atm : 

using default flag : 
- `admin` as dn (passes)
- `admin@example.com` as dn (passes) 

using flag as false : 
- anon bind | empty dn and pw (passes) 
- empty dn, non-empty pw (not authenticated, does not pass)
- normal unix bind with `admin` dn (doesn't pass) 

using flag as true : 
- the original set of tests (where all pass) 

I've not tested the default and false flag under every possible DN and pw possibility. Does this suffice?

server/lib/src/constants/entries.rs
```diff
@@ -361,6 +363,7 @@ impl From<Attribute> for &'static str {
     fn from(val: Attribute) -> Self {
         match val {
             Attribute::Account => ATTR_ACCOUNT,
+            Attribute::DomainLdapAllowUnixPwBind => ATTR_DOMAIN_LDAP_ALLOW_UNIX_PW_BIND,
```
Firstyear, line 366: Should be sorted alphabetically in the list

server/lib/src/constants/uuids.rs
```diff
@@ -237,6 +237,8 @@ pub const UUID_SCHEMA_ATTR_AUTH_SESSION_EXPIRY: Uuid =
     uuid!("00000000-0000-0000-0000-ffff00000141");
 pub const UUID_SCHEMA_ATTR_AUTH_PRIVILEGE_EXPIRY: Uuid =
     uuid!("00000000-0000-0000-0000-ffff00000142");
+pub const UUID_SCHEMA_ATTR_DOMAIN_LDAP_ALLOW_UNIX_PW_BIND: Uuid =
```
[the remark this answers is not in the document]
Firstyear: Has this been rebase to master? This is a uuid conflict.

server/lib/src/server/migrations.rs
```diff
@@ -594,6 +595,8 @@ impl<'a> QueryServerWriteTransaction<'a> {
         // First, check the system_info object. This stores some server information
         // and details. It's a pretty const thing. Also check anonymous, important to many
         // concepts.
+
+        // The below only check if the UUID of the classes exists (Should exists from migrations)
```
Firstyear, line 599: What does this comment mean?

server/lib/src/server/mod.rs
```diff
@@ -1500,8 +1511,31 @@ impl<'a> QueryServerWriteTransaction<'a> {
     pub(crate) fn reload_domain_info(&mut self) -> Result<(), OperationError> {
         let domain_name = self.get_db_domain_name()?;
         let display_name = self.get_db_domain_display_name()?;
+        let domain_ldap_allow_unix_pw_bind = match self.get_domain_ldap_allow_unix_pw_bind() {
+            Ok(v) => v,
+            _ => {
+                admin_warn!("Defaulting ldap_allow_unix_pw_bind to true");
+                info!("Defaulting ldap_allow_unix_pw_bind to true");
```
Firstyear, line 1518: I think you only need one log message here. For now make it debug.

server/lib/src/server/mod.rs
```diff
@@ -1500,8 +1511,31 @@ impl<'a> QueryServerWriteTransaction<'a> {
...
+
+                // let modl = ModifyList::new_purge_and_set(
+                //     Attribute::DomainLdapAllowUnixPwBind,
+                //     Value::Bool(true),
```
Firstyear, line 1522: Remove commented code here.

server/lib/src/server/mod.rs
```diff
@@ -1630,7 +1665,7 @@ impl<'a> QueryServerWriteTransaction<'a> {
     #[instrument(level = "info", skip_all)]
     pub fn commit(mut self) -> Result<(), OperationError> {
         self.reload()?;
-
+        info!("here : commiting qs_write");
```
Firstyear, line 1668: delete log message, it's already implied by the instrument line.

server/lib/src/server/mod.rs
```diff
@@ -793,6 +794,15 @@ pub trait QueryServerTransaction<'a> {
             })
     }
 
+    fn get_domain_ldap_allow_unix_pw_bind(&mut self) -> Result<bool, OperationError> {
+        let res = self.internal_search_uuid(UUID_DOMAIN_INFO)?;
+        match res.get_ava_single(Attribute::DomainLdapAllowUnixPwBind) {
```
Firstyear, line 799: Change to
```
self.internal_search_uuid(UUID_DOMAIN_INFO)
    .map(|entry| {
        entry.get_ava_single_bool(Attribute::DomainLdapAllowUnixPwBind).unwrap_or(true)
    })
```

server/lib/src/idm/ldap.rs
```diff
@@ -643,15 +642,51 @@ mod tests {
         let pce = UnixPasswordChangeEvent::new_internal(UUID_ADMIN, TEST_PASSWORD);
 
         assert!(idms_prox_write.set_unix_account_password(&pce).is_ok());
-        assert!(idms_prox_write.commit().is_ok());
+        assert!(idms_prox_write.commit().is_ok()); // Committing all configs
 
-        let anon_t = ldaps.do_bind(idms, "", "").await.unwrap().unwrap();
+        // Test for optional use of POSIX password for bind
+        // Empty UNIX_PW bind, should allow :
```
Firstyear, line 648: Comment here doesn't seem to match the test? Should this mean "allow_unix_pw_bind unset - > binds allowed"?

server/lib/src/idm/ldap.rs
```diff
@@ -643,15 +642,51 @@ mod tests {
...
+        let admin_t = ldaps
+            .do_bind(idms, "admin", TEST_PASSWORD)
+            .await
+            .unwrap()
+            .unwrap();
+        assert!(admin_t.effective_session == LdapSession::UnixBind(UUID_ADMIN));
+        let admin_t = ldaps
```
Firstyear, line 655: Why do this twice?

server/lib/src/idm/ldap.rs
```diff
@@ -643,15 +642,51 @@ mod tests {
...
+            .do_bind(idms, "admin@example.com", TEST_PASSWORD)
+            .await
+            .unwrap()
+            .unwrap();
+        assert!(admin_t.effective_session == LdapSession::UnixBind(UUID_ADMIN));
+
+        // Setting UNIX_PW to false:
```
Firstyear, line 662: As above, improve comment.

Firstyear (changes requested):

server/lib/src/constants/entries.rs
NavinShrinivas: Done

server/lib/src/server/migrations.rs
NavinShrinivas: It was a personal note to keep track of code. Removing it

server/lib/src/server/mod.rs
NavinShrinivas: Sorry, all the info!() was to debug stuff in testing....removing them

server/lib/src/idm/ldap.rs
NavinShrinivas: Yes, teh variables names are wrong, making them right

server/lib/src/server/mod.rs
NavinShrinivas: why tho? can we have many entries? I don't understand why there might be differentiating entries where each define changes to a subset of variables? Does each schema become one entry value?

NavinShrinivas: Also what if internak_search_uuid gives out an error?

Firstyear: That's handled. self.internal_search_uuid() -> returns a Result<Entry, Error>
.map -> for the value of the Ok(entry), get the attribute assuming it's a single-boolean. 
   unwrap_or tream means "if no attribute matechs, or it's the wrong type, return true"

So this handles those cases.

server/lib/src/idm/ldap.rs
NavinShrinivas: they are on different DNs, just wanted to include those in test suites as well....can remove it

server/lib/src/server/mod.rs
```diff
@@ -814,6 +815,16 @@ pub trait QueryServerTransaction<'a> {
             })
     }
 
+    fn get_domain_ldap_allow_unix_pw_bind(&mut self) -> Result<bool, OperationError> {
+        self.internal_search_uuid(UUID_DOMAIN_INFO).map(|entry| {
+            entry
+                .get_ava_single(Attribute::DomainLdapAllowUnixPwBind)
+                .unwrap_or(Value::new_bool(true))
```
Firstyear, line 822: use `get_ava_single_bool()` which already does this for you.

server/lib/src/idm/server.rs
```diff
@@ -1379,6 +1379,10 @@ impl<'a> IdmServerAuthTransaction<'a> {
                 effective_session: LdapSession::UnixBind(UUID_ANONYMOUS),
             }))
         } else {
+            if self.qs_read.d_info.d_ldap_allow_unix_pw_bind == false {
```
Firstyear, line 1382: clippy will complain, this should be `if !self.qs_read.d_info.d_ldap_allow_unix_pw_bind { `

server/lib/src/idm/server.rs
```diff
@@ -1379,6 +1379,10 @@ impl<'a> IdmServerAuthTransaction<'a> {
...
+                security_info!("Bind not allowed through Unix passwords.");
+                return Ok(None);
```
Firstyear, line 1384: Probably should return an invalid credential error instead of Ok(None) here.

server/lib/src/idm/ldap.rs
Firstyear: Why is it failing? What's being returned here? What's happening in the logs?

Firstyear (changes requested):

NavinShrinivas:
Don't know why, ghub is not letting me reply on that last comment. None of the tests are failing....I've just written comments telling that the bind wont happen as the flag is set to false. On non binding replies is Ok(None) accepted? (Considering your second last comment, this is not the case and infact invalid credentials error  is expected)

NavinShrinivas:
> Don't know why, ghub is not letting me reply on that last comment. None of the tests are failing....I've just written comments telling that the bind wont happen as the flag is set to false. On non binding replies is Ok(None) accepted? (Considering your second last comment, this is not the case and infact invalid credentials error is expected)

The reason I use Ok(None) is because all other cases in `auth_ldap()` returns back Ok(None) (say for example, not having posix password configured)

NavinShrinivas:
And furthermore, do_op converts Ok(None) from do_bind into InvalidCredentails and most code invokes binding through do_op. If we were to change what do_bind sends out, should we change do_op as well?

NavinShrinivas:
image
And do_op seems to be expection Ok(None) where it generates the error?

Firstyear:
My mistake, you're right :) Ok(None) is correct here

server/lib/src/idm/ldap.rs
```diff
@@ -643,15 +642,54 @@ mod tests {
...
+        assert!(admin_t.effective_session == LdapSession::UnixBind(UUID_ADMIN));
+        let admin_t = ldaps
+            .do_bind(idms, "admin@example.com", TEST_PASSWORD)
+            .await
+            .unwrap()
+            .unwrap();
+        assert!(admin_t.effective_session == LdapSession::UnixBind(UUID_ADMIN));
 
+        // Setting UNIX_PW_BIND flag to false:
+        // Hence all of the below tests will fail (asserts are still satisfied)
```
Firstyear, line 663: Maybe reword to "authentication will fail" rather than the test failing. The test is succeeding at showing auth failing!

NavinShrinivas:
Hey, do we have to fix these failing builds?

Also, I'd like to add docs regarding this option and its behaviour. Is there any guidelines for writing these docs?

Firstyear:
If the failures are because of something in the PR, then yeah, it needs to be fixed.

As for the docs, leave that to me I think. :)

Firstyear:
I went to check out and edit your branch but it looks completely cooked - I have no idea how you were working on this tbh. Going to fix it in a new pr.

NavinShrinivas:
> I went to check out and edit your branch but it looks completely cooked - I have no idea how you were working on this tbh. Going to fix it in a new pr. 
> 
> EDIT: It's because you are working on your forks master branch @NavinShrinivas  - that's mucking up a whole bunch of things if anyone else tries to clone your fork and then review your work. So It's probably a good future thing to "never do this". 

Yes I felt this pr was really messy too, do you want me to fix this? I'd like to open a proper pr that's cleaner

Firstyear:
> > I went to check out and edit your branch but it looks completely cooked - I have no idea how you were working on this tbh. Going to fix it in a new pr.
> > EDIT: It's because you are working on your forks master branch @NavinShrinivas  - that's mucking up a whole bunch of things if anyone else tries to clone your fork and then review your work. So It's probably a good future thing to "never do this".
> 
> Yes I felt this pr was really messy too, do you want me to fix this? I'd like to open a proper pr that's cleaner

It's going to be super hard to fix. You can't rebase-squash to get a single commit here because you did merge commits. I think you have to do a shit ton of rebasing to rearrance history and make it work. I'm thinking the easiest way will be a new branch, completely clean from master, and copy-paste your changes from the diff viewer here on github.

Trying to resolve this in git is going to be simply too painful I think.

NavinShrinivas:
> diff

What about me cloning you upstream, cherry picking my commits from my fork and pushing that as a new branch onto my fork and I'll make a pr from there

Firstyear:
That would work, but then you have to cherry pick 11 commits, doing a manual conflict resolve at every step.

Firstyear:
This was merged in a separate PR.

PR #2163, opened 2023-10-01
Chasing wooly quadrapeds again

yaleman:
Cleaning up some TODOs and stuff

Checklist

- [x] cargo fmt has been run

### DIFF
--- a/server/lib/src/filter.rs
+++ b/server/lib/src/filter.rs
@@ -658,12 +658,12 @@ impl FilterComp {
                 }
             }
             FilterComp::Or(filters) => {
-                // If all filters are okay, return Ok(Filter::Or())
-                // If any is invalid, return the error.
-                // @ftweedal says an empty or is a valid filter
-                // in mathematical terms.
+                // * If all filters are okay, return Ok(Filter::Or())
+                // * Any filter is invalid, return the error.
+                // * An empty "or" is a valid filter in mathematical terms, but we throw an
+                //   error to warn the user because it's super unlikey they want that
                 if filters.is_empty() {
-                    return Ok(FilterComp::Or(vec![]));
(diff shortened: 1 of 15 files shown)


server/lib/src/filter.rs
```diff
@@ -660,10 +660,10 @@ impl FilterComp {
             FilterComp::Or(filters) => {
                 // If all filters are okay, return Ok(Filter::Or())
                 // If any is invalid, return the error.
-                // TODO: ftweedal says an empty or is a valid filter
+                // @ftweedal says an empty or is a valid filter
                 // in mathematical terms.
                 if filters.is_empty() {
-                    return Err(SchemaError::EmptyFilter);
+                    return Ok(FilterComp::Or(vec![]));
```
Firstyear, line 666: I think these should stay as errors. Is there a reason to make them valid here?

Firstyear (changes requested):

server/lib/src/filter.rs
yaleman: 'cause an empty `Or` is an empty `Or` and thus valid and thus "false"

Firstyear: But if someone requests an empty or, shouldn't we tell them "hey you're query looks a bit weird?" Is there a use case you are trying to resolve here?

yaleman: I dunno, I was mainly just cleaning up TODOs 😄 I can roll it back if you think it's an error.

> TODO: ftweedal says an empty or is a valid filter in mathematical terms.

Firstyear: Also gosh damn that's some *ancient* code.

yaleman: Done!

Firstyear (approved):

Issue #2028, opened 2023-08-25
OpenAPI / Documentation Schema for the REST API


yaleman:
**Is your feature request related to a problem? Please describe.**

The old `RouteMap` implementation went away when we moved to Axum, we need a self-documenting schema def from the routes.

**Describe the solution you'd like**

OpenAPI or some similar accessible spec.

**Describe alternatives you've considered**

"read the source" which is not a good response.

**Additional context**

A few implementations/links to check out:

- https://github.com/tamasfe/aide - folks seem to like it
- https://github.com/juhaku/utoipa - lots of templatey things
- tokio-rs/axum#50

Firstyear:
utoipa seems to be the more popular/maintained, nd it's compile time generated, so I think I prefer that over aide atm.

ClashTheBunny:
I see you got started on this in #2175.  When I tried to use the openapi.json for a terraform module, I get a bunch of schema references that are not defined:

```
"#/components/schemas/ApiTokenPurpose"
"#/components/schemas/AuthAllowed"
"#/components/schemas/AuthIssueSession"
"#/components/schemas/AuthMech"
"#/components/schemas/AuthStep"
"#/components/schemas/Base64UrlSafeData"
"#/components/schemas/BTreeSet"
"#/components/schemas/CredentialDetailType"
"#/components/schemas/CUExtPortal"
"#/components/schemas/CURegState"
"#/components/schemas/Entry"
"#/components/schemas/Filter"
"#/components/schemas/IdentifyUserRequest"
"#/components/schemas/Json"
"#/components/schemas/Modify"
"#/components/schemas/OperationError"
"#/components/schemas/ProtoEntry"
"#/components/schemas/ScimSyncRequest"
"#/components/schemas/time.OffsetDateTime"
"#/components/schemas/UatPurpose"
"#/components/schemas/UatPurposeStatus"
"#/components/schemas/UatStatusState"
"#/components/schemas/Uuid"
```

Is this expected at this stage of development?

<details>
  <summary> Full command output </summary>

```
 tfplugingen-openapi generate \
  --config ./generator_config.yml \
  --output ./provider-code-spec.json \
  ./openapi.json
{"time":"2024-03-09T08:35:50.492532994-06:00","level":"ERROR","msg":"unable to locate reference anywhere in the rolodex","reference":"#/components/schemas/AuthStep"}
[...]

PR #2175, opened 2023-10-04
OpenAPI/swagger docs autogen

yaleman:
Refers to #2028, starting to implement an OpenAPI schema generation for the server. Currently testing with `utoipa` but wow there's a lot of boilerplate and macros and stuff.

Checklist

- [x] cargo fmt has been run


yaleman:
But hey, it happens quickly... 😄 

Screenshot 2023-10-04 at 21 39 37

server/core/src/https/v1.rs
```diff
@@ -38,7 +37,22 @@ pub(crate) struct SessionId {
     pub sessionid: Uuid,
 }
 
-// /v1/raw/create
+#[utoipa::path(
+    post,
+    path = "/v1/raw/create",
```
Firstyear, line 42: I like that you added this with openapi, but lets *not* document the /raw commands ... they are there as breakglass for us to issue emergency help commands if needed.

server/core/src/https/v1_scim.rs
```diff
@@ -151,6 +183,23 @@ pub async fn sync_account_id_put_attr(
     json_rest_event_put_id_attr(state, id, attr, filter, values, kopid).await
 }
 
+#[utoipa::path(
+    get,
+    path = "/scim/v1/Sink",
```
Firstyear, line 188: This is an easter egg, should we document it?

Firstyear (changes requested):
I think my biggest concert here is this looks pretty noisy/manual to implement. Should we limit this to apis we really want people to interact with?

server/core/src/https/v1.rs
yaleman: Done

yaleman:
> I think my biggest concert here is this looks pretty noisy/manual to implement. Should we limit this to apis we really want people to interact with?

The biggest problem I've had trying to implement and work with the Kanidm APIs is the lack of documentation and typing (no judgement, just reality) on input/output, so ... IMO a clearly documented API is needed for **_everything_** to make the DX a lot smoother. 

I agree it's pretty bulky, and I think there's some things that can make it neater with the `axum` integration, but it's much less bad than manually building an OpenAPI schema 🤣

yaleman:
This is as far as I'm going to get at this point, lots of TODOs but WAY better docs for API endpoints etc. and less use of unwrap and expect

Firstyear:
Ready for review?

yaleman:
> Ready for review?

Sure is!

server/core/src/https/errors.rs
```diff
@@ -0,0 +1,69 @@
...
+
+/// The web app's top level error type, this takes an `OperationError` and converts it into a HTTP response.
+#[derive(Debug, ToSchema)]
+pub enum WebError {
+    /// Something went wrong when doing things.
+    OperationError(OperationError),
+    InternalServerError(String),
+}
+
+impl From<OperationError> for WebError {
```
Firstyear, line 18: Is this kind of tied into the error handling stuff we want to do in future?

Firstyear (approved):

server/core/src/https/errors.rs
yaleman: Yep, it's a look at how that kind of thing could work in the future.

PR #2189, opened 2023-10-06
Replication tweaks - try the most recent successful one and error less

yaleman:
* When a peer connection is successful, try it first next time.
* Don't throw errors until we can't connect to *any* peers - it was erroring on each connection.

Somewhere in the socket code something's throwing a *LOT* of debug messages that it'd be good to filter out.

Checklist

- [x] cargo fmt has been run

### DIFF
--- a/server/core/src/repl/mod.rs
+++ b/server/core/src/repl/mod.rs
@@ -8,12 +8,15 @@ use std::net::SocketAddr;
 use std::pin::Pin;
 use std::sync::Arc;
 use std::time::Duration;
-use tokio::net::{TcpListener, TcpStream};
 use tokio::sync::broadcast;
 use tokio::sync::mpsc;
 use tokio::sync::oneshot;
 use tokio::sync::Mutex;
 use tokio::time::{interval, sleep, timeout};
+use tokio::{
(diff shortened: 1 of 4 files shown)


server/core/src/repl/mod.rs
```diff
@@ -450,8 +456,22 @@ async fn repl_task(
...
+
     // Okay, all the parameters are setup. Now we wait on our interval.
     loop {
+        // if it worked last time, then let's use it this time!
+        // I sure hope this works. - @yaleman
+        let mut sorted_socket_addrs = match last_working_address {
+            Some(addr) => vec![addr],
+            None => vec![],
+        };
+        socket_addrs.iter().cloned().for_each(|addr| {
```
Firstyear, line 469: I'd throw in a comment that this is an exponential search/insert process, but that's okay because the number of addrs will be small, so it's cheap over all.

Firstyear (approved):
Minor, but looks great.

server/core/src/repl/mod.rs
yaleman: Done!

PR #2190, opened 2023-10-08
Thread naming and display

yaleman:
I found a thread was `panic`'ing and it wasn't showing which one or why, so this stores a name for the thread on startup and then can show it when we're shutting down.

Checklist

- [x] cargo fmt has been run

### DIFF
--- a/server/core/src/admin.rs
+++ b/server/core/src/admin.rs
@@ -168,7 +168,7 @@ impl AdminActor {
                     }
                 }
             }
-            info!("Stopped AdminActor");
+            info!("Stopped {}", super::TaskName::AdminSocket);
         });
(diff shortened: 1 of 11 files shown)


Firstyear (approved):

Firstyear:
If you aren't in a rush to merge, you could make threadName an enum, and then pust them in a btreemap and .drain() it during .join() instead.

yaleman:
For reference it was ... I think a migration barfing on startup due to a corrupted database, which caused a panic somewhere, which probably should kill the whole thing... but now I've lost the scrollback 😡 

It was solved by rm'ing the DB ^_^

Firstyear (approved):

PR #2191, opened 2023-10-08
20231008 remove expect used

Firstyear:
Removes a ton of expect used. Note clippy is *not* run yet, because I added some unwraps that I plan to get rid of soon.

Checklist


server/core/src/lib.rs
```diff
@@ -173,7 +173,13 @@ macro_rules! dbscan_setup_be {
 
 pub fn dbscan_list_indexes_core(config: &Configuration) {
```
yaleman, line 174: can we add an explicit return type here? 🤔

server/lib/src/filter.rs
```diff
@@ -1287,8 +1287,7 @@ impl FilterResolved {
 
                 // If the f_list_or only has one element, pop it and return.
                 if f_list_new.len() == 1 {
-                    #[allow(clippy::expect_used)]
-                    f_list_new.pop().expect("corrupt?")
+                    f_list_new.remove(0)
```
yaleman, line 1290: This should also be a [clear](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.clear)
```suggestion
                    f_list_new.clear()
```

yaleman (changes requested):
Minor changes for clarity but woo, less expecting things.

server/lib/src/filter.rs
Firstyear: Uhhh? no? clear just deletes everythdng. We want to *pop* the last element and return it.

yaleman: Derp I r dumb. All good 😄

yaleman (approved):

PR #2206, opened 2023-10-10
testing things fixing bugs

yaleman:
Fixes:

* idm_domain_reset_token_key wasn't working because we weren't sending a body and the API couldn't handle it, added a test and fixed it (we weren't testing it)
* idm_domain_set_ldap_basedn - added tests

Checklist

- [x] cargo fmt has been run

### DIFF
--- a/pykanidm/poetry.lock
+++ b/pykanidm/poetry.lock
@@ -208,33 +208,29 @@ files = [
 
 [[package]]
 name = "black"
-version = "23.9.1"
+version = "23.10.0"
 description = "The uncompromising code formatter."
 optional = false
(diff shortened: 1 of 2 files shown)


libs/client/src/lib.rs
```diff
@@ -890,7 +891,8 @@ impl KanidmClient {
         let response = self
             .client
             .delete(self.make_url(dest))
-            .header(CONTENT_TYPE, APPLICATION_JSON);
+            // empty-ish body taht makes the parser happy
```
yaleman, line 894:
```suggestion
            // empty-ish body that makes the parser happy
```

PR #2219, opened 2023-10-14
Splitting the SPAs

yaleman:
I might have also gone down a rabbit-hole and split the admin UI out into its own SPA.

After discussing it with @Firstyear I'll split this into three `yew` apps:

- Login/OAuth2 flows
- Other user-related things (app list, cred-management, profile etc)
- Admin related UI

This'll let us shrink the WASM blobs somewhat and the transition points are where users would expect things to "reload".

Changes:

* split the apps - previously it was one ~2.9MB / 685KB WASM blob, is now: 
  * Login flow WASM (login, OAuth2, reauth) is 998KB/290KB pre/post brotli-ing.
  * User interface is 2.5M/610KB
  * Admin UI is 605KB/144KB
* anything compressible over 16KB also now gets brotli'd during build, saving a bunch of network-time.
* stripped a load of static things out and replaced them with constants or functions that create the things.

Checklist

- [x] cargo fmt has been run


Firstyear:
I have an example code you may want to look at, I'll add you to the repo.

server/web_ui/login_flows/pkg/README.md
```diff
@@ -0,0 +1,144 @@
+# Kanidm - Simple and Secure Identity Management
+
+<p align="center">
```
Firstyear, line 3: Shouldn't this readme be specific to the component?

server/web_ui/shared/src/models.rs
```diff
@@ -1,12 +1,101 @@
...
     let l: Result<(CUSessionToken, CUStatus), _> = TemporaryStorage::get("cred_update_session");
     l.ok()
 }
+
+pub fn push_return_location(l: &str) {
+    TemporaryStorage::set("return_location", l)
+        .expect_throw("failed to set return_location in temporary storage");
+}
+
+pub fn pop_return_location() -> String {
```
Firstyear, line 38: Wasn't this an enum in the past? Should we consider similar here?

Firstyear (changes requested):
Missing cargo.toml for kanidm_web_ui_login

server/web_ui/shared/src/models.rs
yaleman: Yeah this is just me messin' around while splitting it up and trying to make things go. This ... is a messy part that `yew-rs`'s routers might cause some friction with, since it's an SPA-framework and expect some things, but I have ideas.

server/web_ui/login_flows/pkg/README.md
yaleman: It's really just a placeholder to make the build/package process less sad, and I think I can remove it anyhow

yaleman: Yep, gone.

server/web_ui/shared/src/models.rs
yaleman: Turns out to get it working it has to be a string value because they don't share routers/enums, nor do we want them to.

server/web_ui/admin/Cargo.toml
```diff
@@ -0,0 +1,70 @@
+[package]
+name = "kanidmd_web_ui_admin"
+description = "Kanidm Server Web UI - Admin Interface"
+documentation = "https://docs.rs/kanidm/latest/kanidm/"
+
+version = "1.1.0-rc.14-dev"
```
Firstyear, line 6: I guess that wasm still can't inherit workspaces?

server/web_ui/admin/Cargo.toml
```diff
@@ -0,0 +1,70 @@
...
+authors = [
+    "William Brown <william@blackhats.net.au>",
+    "James Hodgkinson <james@terminaloutcomes.com>",
+]
+rust-version = "1.66"
```
Firstyear, line 11: we should just raise this I think.

server/web_ui/admin/build.sh
```diff
@@ -0,0 +1,45 @@
+#!/bin/bash
```
Firstyear, line 1: Is there a reason to split this up between the packages? Could we just have a single "build all the wasms?"
 Or just a way to symlink them all to a common script?

server/web_ui/admin/pkg/LICENSE.md
```diff
@@ -0,0 +1,327 @@
+# Mozilla Public License Version 2.0
+
```
Firstyear, line 2: Does this need to be copied to these folders or can we link to the root license?

server/web_ui/admin/pkg/README.md
```diff
@@ -0,0 +1,144 @@
+# Kanidm - Simple and Secure Identity Management
```
Firstyear, line 1: Should have a "admin specific readme"

server/web_ui/login_flows/pkg/LICENSE.md
```diff
@@ -0,0 +1,327 @@
+# Mozilla Public License Version 2.0
+
+1. Definitions
```
Firstyear, line 3: As above about the license file.

server/web_ui/shared/src/lib.rs
```diff
@@ -0,0 +1,183 @@
...
+            RequestMethod::GET => "GET".to_string(),
+        }
+    }
+}
+
+/// creates the "Kanidm is alpha" banner
+pub fn alpha_warning_banner() -> Html {
+    html!(
+        <div class={CSS_ALERT_WARNING} role="alert">
+        {"🦀 Kanidm is still in early Alpha, this interface is a placeholder! "}
```
Firstyear, line 138: I wonder if we should yeet this at this point, we are pretty mature at this point.

Firstyear (changes requested):

Firstyear:
Not too much, looks really good though <3

server/core/src/https/ui.rs
```diff
@@ -8,58 +8,73 @@ use http::header::CONTENT_TYPE;
 use super::middleware::KOpId;
 use super::ServerState;
 
+/// TODO: deprecate this
```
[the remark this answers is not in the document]
yaleman: Nah it's an old comment I missed. Fixed.

server/web_ui/admin/Cargo.toml
yaleman: Done

server/web_ui/admin/Cargo.toml
yaleman: Sorted, seems they can now

server/web_ui/admin/build.sh
yaleman: Sorted 😄

server/web_ui/admin/pkg/README.md
yaleman: Nah, we don't need one at all, we used to have to have README and LICENSE files for the WASM thing to build at all else it'd get sad.

server/web_ui/shared/src/lib.rs
yaleman: It's only on the admin things, and they're still *very* alpha.

yaleman:
Plz can merge? :uwu:

Firstyear (approved):

PR #2241, opened 2023-10-18
cargo fmt + clippy

scabrero:
Run cargo fmt and fix clippy errors

Checklist

- [X] cargo fmt has been run

### DIFF
--- a/server/lib/src/idm/authsession.rs
+++ b/server/lib/src/idm/authsession.rs
@@ -714,6 +714,15 @@ impl AuthSessionState {
     }
 }
 
+pub(crate) struct AuthSessionData<'a> {
+    pub(crate) account: Account,
+    pub(crate) account_policy: ResolvedAccountPolicy,
+    pub(crate) issue: AuthIssueSession,
+    pub(crate) webauthn: &'a Webauthn,
+    pub(crate) ct: Duration,
+    pub(crate) source: Source,
+}
+
 #[derive(Clone)]
 /// The current state of an authentication session that is in progress.
 pub(crate) struct AuthSession {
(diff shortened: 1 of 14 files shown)


Firstyear:
Hey @scabrero can you check what version of rust/cargo you are using? We normally use the latest, and sometimes cargo fmt has minor differences between versions. We normally cargo fmt the whole code base during every PR, and so it's surprising to see so many changes.

You should be on cargo/rust 1.73 - if you're not, likely you need to update :)

server/lib/src/idm/authsession.rs
```diff
@@ -714,6 +714,15 @@ impl AuthSessionState {
     }
 }
 
+pub(crate) struct AuthSessionData<'a> {
```
yaleman, line 717: This isn't just a clippy/fmt change and should probably be out in `kanidm_proto`

yaleman (changes requested):

Firstyear:
> > Hey @scabrero can you check what version of rust/cargo you are using? We normally use the latest, and sometimes cargo fmt has minor differences between versions. We normally cargo fmt the whole code base during every PR, and so it's surprising to see so many changes.
> > You should be on cargo/rust 1.73 - if you're not, likely you need to update :)
> 
> Ah, that fixed the cargo fmt differences, but brought a new set of clippy errors :) I will update this PR fixing these new ones.

Some of the clippy errors are around unwraps that I intentionally left in as I'm slowly working to bubble up some errors to higher levels. So if you want to take that on, go for it, but if you see "unwrap() shouldn't be used" feel free to leave that for me to clean up in the coming days :)

yaleman:
This needs another rebase but otherwise looks good to go 👍🏻

server/lib/src/idm/authsession.rs
Firstyear: This would be an internal struct too, not proto.

Firstyear (approved):

yaleman (approved):

PR #2245, opened 2023-10-20
20231019 1122 account policy basics

Firstyear:
Relates #1122 - This does a lot of the ground work for account policy. We can compute the policy from the various groups, and it's resultant policy is applied in authentication and pw changes. For now there is still a lot I want to think about on how to proceed, especially about how we set policy by default. 

Checklist


server/lib/src/constants/groups.rs
```diff
@@ -410,7 +410,6 @@ lazy_static! {
     };
 
 
-    /// Builtin IDM Group for allowing migrations of service accounts into persons
```
yaleman, line 413: It'd be really nice if there was some way of automagically making this comment match the description so when one looks at it in the docs/an IDE, you can tell what it does without having to dig deeper 😞

server/lib/src/idm/accountpolicy.rs
```diff
@@ -0,0 +1,153 @@
...
+            privilege_expiry: MAXIMUM_AUTH_PRIVILEGE_EXPIRY,
+            authsession_expiry: MAXIMUM_AUTH_SESSION_EXPIRY,
+            credential_policy: CredentialPolicy::default(),
+        };
+
+        iter.for_each(|acc_pol| {
+            // Take the smaller expiry
+            if acc_pol.privilege_expiry < accumulate.privilege_expiry {
+                accumulate.privilege_expiry = acc_pol.privilege_expiry
+            }
```
yaleman, line 87: `itertools::min/max` is pretty rad for the iterator, would allow a big collapse of the lewp:
```
        let min_priv_expiry = itertools::min(acc_pol.iter().map(|a| a. privilege_expiry));

        // define the policy
        let mut accumulate = ResolvedAccountPolicy {
            privilege_expiry: std::cmp::min(MAXIMUM_AUTH_PRIVILEGE_EXPIRY, min_priv_expiry),
            authsession_expiry: MAXIMUM_AUTH_SESSION_EXPIRY,
            credential_policy: CredentialPolicy::default(),
        };

      
```

Firstyear: Yeah, but we can only go over the iterator *once*. We can't do this because it requires us to iterate over the iter multiple times.

yaleman: Ah, because we're doing a from on the parent iter I guess. Boooooo.

yaleman:
So, looking at this - we're storing the policy outcome on the accounts, rather than computing/caching it at runtime - that **_sounds_** like a scaling/replication/maintenance-of-the-outcomes challenge in bigger environments with lots of changes to groups etc, but I guess it's faster than having to pull all the policies and compute them on check? 🤔

Firstyear:
> So, looking at this - we're storing the policy outcome on the accounts, rather than computing/caching it at runtime - that **_sounds_** like a scaling/replication/maintenance-of-the-outcomes challenge in bigger environments with lots of changes to groups etc, but I guess it's faster than having to pull all the policies and compute them on check? 🤔

No? We store the account policy on the groups, then we compute it everytime we need it. 

This way we don't have cache invalidation etc. We just have to look at the users groups and we get the policy. So it will always reflect the latest changes. 

We only cache the account/policy in transactions for things like authsessions or credupdatesessions where we need the policy to be stable over the duration of the operation.

Firstyear:
In theory we *are* caching though - we just lean on the entry cache to cache the groups for us so that it's fast to access when we need it.

yaleman:
> No? We store the account policy on the groups, then we compute it everytime we need it.
> 

Clearly I was awake at that point in the morning 😊 Looks good, given that description and ... trying again to read it.

book/src/account_policy.md
```diff
@@ -0,0 +1,86 @@
+# Account Policy
+
+Account Policy defines the security requirements that accounts must meet and influences users
+sessions.
+
+Policy is defined on groups so that membership of a group influences the security of it's members.
```
yaleman, line 6:
```suggestion
Policy is defined on groups so that membership of a group influences the security of its members.
```

book/src/account_policy.md
```diff
@@ -0,0 +1,86 @@
...
+This allows you to express that if you can access a system or resource, then the account must also
+meet the policy requirements.
+
+## Default Account Policy
+
+By default Account Policy is applied to `idm_all_accounts`. This provides the defaults that
```
yaleman, line 12:
```suggestion
A default Account Policy is applied to `idm_all_accounts`. This provides the defaults that
```

book/src/account_policy.md
```diff
@@ -0,0 +1,86 @@
...
+To set the maximum authentication session time
+
+```
+kanidm group account-policy auth-expiry <group name> <seconds>
+kanidm group account-policy auth-expiry my_admin_group 86400
+```
+
+## Setting Maximum Privilege Time
+
+The privilege-expiry time defines how long a session retains it's write privileges after a
```
yaleman, line 78:
```suggestion
The privilege-expiry time defines how long a session retains its write privileges after a
```

book/src/account_policy.md
```diff
@@ -0,0 +1,86 @@
...
+
+```
+kanidm group account-policy enable <group name>
+kanidm group account-policy enable my_admin_group
+```
+
+## Setting Maximum Session Time
+
+The auth-session value influences the maximum time in seconds that an authenticated session can
+exist for. After this time, the user must reauthenticate.
```
yaleman, line 62:
```suggestion
exist. After this time, the user must reauthenticate.
```

yaleman (approved):
LGTM 🥳

yaleman:
^ such rebase like wow

Issue #2248, opened 2023-10-23
Lack of debian derivative documentation


TheRealGramdalf:
I'd like to start by saying thank you: kanidm looks like an amazing IDM solution for homelabbers who want good user management and sane, secure defaults. I love the idea, the intent, and the execution of how you've implemented this.

That being said, there are a couple of problems that prevent me, a semi-advanced linux user from deploying this in the way I'd like. Part of it could be user error/lack of knowledge, missing documentation, or any combination of the two. I have some "constructive criticism" if you will, and a couple of questions:

##### Kanidm Book
- As it stands right now, the Kanidm book looks like it's at a great starting point - but it needs some cleanup and clarification before it will be user friendly and accurate.
    - For example, client tools: It wasn't entirely clear to me how the client configuration works, what options exist, and how they work. With my previous knowledge and a little bit of looking, it makes sense (more or less) to me, but changes or clarifications would be helpful.
    - To this effect, I offer to help where I can - I feel like I'm in a good place knowledge wise to write documentation that is technically  accurate (as in, an accurate representation of the code), and I would be willing to take on the role of documentation maintainer. If this offer intrigues you, let me know and I can elaborate.

##### Integration
- I realize that this is primarily due to the project still being in beta - not all features/api endpoints/etc are finalized, and as such writing and maintaining documentation isn't necessarily the best use of dev time. With that in mind, however, the most recent release states that it is a beta, and release candidates should be starting soon™. With _that_ in mind, I have some questions regarding integration with debian and derivatives:
    - My environment (as much as is relative) is as follows:
```
Dell Poweredge R810
├─ OS: Proxmox VE @ v8.x
|     └─ Portainer BE 2.9.x @ Docker
├─ CPU: Intel Xeon E7540 @ 2.00GHz
├─ RAM: (32x4GB) DDR3 @ 1333 MT/s
└─ HBA: LSI SAS 9201-16e @ FW 20.00.07.0
      └─ HP SAS expander 468405-001
          └─ ZFS pools
              └─ 2x 450GB SSD @ Mirror (docker storage)
...
```
    - As it currently stands, I am running docker directly on the proxmox host, using Portainer to manage my containers. 
        - While docker is currently running as root, I fully intend to move to a rootless install once I finish my system prototype
        - I would like to use Kanidm to manage users (persons) for my various services (jellyfin, nextcloud, etc.), as well as unix users (service accounts) on the proxmox host (for managing file permissions as needed for docker rootless and bind mounts)
[...]

PR #2252, opened 2023-10-25
.deb package build and docs fixes

### LINKED ISSUE #2157: Missing permission on upload_to_releases for Debian packages
I took a look at the Debian flow, and I think I know [why it fails](https://github.com/kanidm/kanidm/actions/runs/6321269095/job/17165314339). The package builds go fine, but the upload to releases isn't.

I think it's because it's missing a permission here: https://github.com/kanidm/kanidm/blob/034ddd624a5792dc74d27112eb994aee78282180/.github/workflows/debian_package_kanidm.yml#L54-L73

```yaml
permissions:
      contents: write
```

It needs the `contents: write` permission in order to be allowed to upload to GitHub Releases. If I understand it correctly, that's what that action is doing. I think by default that's a permission the token should have, but I suspect you might've changed the repository default to read-only. That would also align with why there's an explicit `packages: write` in https://github.com/kanidm/kanidm/blob/034ddd624a5792dc74d27112eb994aee78282180/.github/workflows/docker_build_kanidm.yml#L39-L48

yaleman:
Closes #2157

Refers to #2248 - breaks up the PAM/nsswitch docs a little

Checklist

- [x] cargo fmt has been run

### DIFF
--- a/book/src/SUMMARY.md
+++ b/book/src/SUMMARY.md
@@ -33,6 +33,9 @@
 
 - [Service Integrations](integrations/readme.md)
   - [PAM and nsswitch](integrations/pam_and_nsswitch.md)
+    - [Fedora](integrations/pam_and_nsswitch/fedora.md)
+    - [SUSE / OpenSUSE](integrations/pam_and_nsswitch/suse.md)
+    - [Troubleshooting](integrations/pam_and_nsswitch/troubleshooting.md)
   - [SSH Key Distribution](integrations/ssh_key_dist.md)
   - [Oauth2](integrations/oauth2.md)
   - [LDAP](integrations/ldap.md)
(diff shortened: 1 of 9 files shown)


book/src/SUMMARY.md
```diff
@@ -33,6 +33,9 @@
 
 - [Service Integrations](integrations/readme.md)
   - [PAM and nsswitch](integrations/pam_and_nsswitch.md)
+    - [Fedora](integrations/pam_and_nsswitch/fedora.md)
```
Firstyear, line 36: Minor, sort fedora/centos after suse.

Firstyear (approved):

book/src/SUMMARY.md
yaleman: I figured you'd ask for this 🤣

book/src/SUMMARY.md
```diff
@@ -33,6 +33,9 @@
...
+    - [SUSE / OpenSUSE](integrations/pam_and_nsswitch/suse.md)
```
yaleman, line 37:
```suggestion
    - [SUSE / OpenSUSE](integrations/pam_and_nsswitch/suse.md)
    - [Fedora](integrations/pam_and_nsswitch/fedora.md)
```

PR #2260, opened 2023-10-27
Ldap allow unix pw bind

### LINKED ISSUE #2027: Prevent posix password from operating with LDAP without a config option

NavinShrinivas:
Fixes #2027 

Checklist

- [x] cargo fmt has been run
- [ ] book chapter included (if relevant) [TO-DO]
- [ ] design document included (if relevant) [NOT NEEDED]


yaleman (approved):
What happens on an existing system when we're

PR #2279, opened 2023-11-01
Minor improvements to incoming replication

### LINKED ISSUE #2182: Check incoming RUV that the current servers id has not advanced.
This needs to be checked so that we can detect potential duplicate server id conflicts.

### LINKED ISSUE #2243: Check incoming replica time
When a consumer receives data from a supplier, we should check that the incoming maximum CID is not greater than 5 minutes ahead of our clock as this may indicate ongoing clock drift within the topology.

The question becomes "if we detect clock skew, how do we react?"

* Ignore it - because of how we order changes, clock skew even to large amounts doesn't matter for correctness, it just means that one nodes changes will "persist" over others.
* Disable the supplier that is ahead - this would stop replication from that supplier since it has a skew but do we actually need to do this? 

A consideration is that because we use a lamport clock if a server jumps ahead then back, then the maximum CID will remain at the maximum until the true system clock catches up. This could easily occur with a machine clock being in local vs utc and ntp needs to sync.

Firstyear:
Fixes #2243 and Fixes #2182 

Checklist

### DIFF
--- a/server/lib/src/repl/ruv.rs
+++ b/server/lib/src/repl/ruv.rs
@@ -480,6 +480,45 @@ impl<'a> ReplicationUpdateVectorWriteTransaction<'a> {
         self.ranged.clear();
     }
 
+    pub(crate) fn incremental_preflight_validate_ruv(
+        &self,
+        ctx_ranges: &BTreeMap<Uuid, ReplCidRange>,
+        txn_cid: &Cid,
+    ) -> Result<(), OperationError> {
+        // Check that the incoming ranges, for our servers id, do not exceed
+        // our servers max state. This can occur if we restore from a backup
+        // where the replication state is older than what our partners have,
+        // meaning that the context may have diverged in a way we can't then
+        // resolve.
+
+        if let Some(our_cid_range_max) = self
+            .ranged
+            .get(&txn_cid.s_uuid)
+            .and_then(|range| range.last().copied())
+        {
+            if let Some(incoming_cid_range) = ctx_ranges.get(&txn_cid.s_uuid) {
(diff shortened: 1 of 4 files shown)


server/lib/src/repl/ruv.rs
```diff
@@ -480,6 +480,45 @@ impl<'a> ReplicationUpdateVectorWriteTransaction<'a> {
...
+        // resolve.
+
+        if let Some(our_cid_range_max) = self
+            .ranged
+            .get(&txn_cid.s_uuid)
+            .and_then(|range| range.last().copied())
+        {
+            if let Some(incoming_cid_range) = ctx_ranges.get(&txn_cid.s_uuid) {
+                if incoming_cid_range.ts_max > our_cid_range_max {
+                    error!("The incoming data contains changes from our servers uuid, and those changes are *newer* than our data. This can occur if this consumer was restored from backup. Replication is unable to proceed as datacorruption may occur. You must refresh this consumer immediately.");
```
yaleman, line 501: Updated wording, this'd be a good E-REPL-12345 text 😄 

```suggestion
                    error!("The incoming data contains changes matching this server's UUID, and those changes are newer than the local version. This can occur if the server was restored from a backup which was taken before sending out changes. Replication is unable to proceed as data corruption may occur. You must refresh this consumer immediately to continue.");
```

yaleman (changes requested):

server/lib/src/repl/ruv.rs
Firstyear: Yeah, I know. I've tried to make the replication error messages really good.

yaleman (approved):

PR #2284, opened 2023-11-02
20231101 add id cert to unixint

Firstyear:
This is the other half of integrating hsm-crypto into unix-integration. This setups the hsm types and interfaces so that a provider doesn't need to worry about key storage or persistence. 

Checklist

- [ ] cargo fmt has been run


unix_integration/src/resolver.rs
```diff
@@ -95,21 +100,71 @@ where
...
+                error!(?err, "Client was unable to configure hsm keys");
+            })?;
+
+        drop(hsm_lock);
+
+        dbtxn.commit().map_err(|_| ())?;
 
         if pam_allow_groups.is_empty() {
-            eprintln!("Will not be able to authorise user logins, pam_allow_groups config is not configured.");
+            warn!("Will not be able to authorise user logins, pam_allow_groups config is not configured.");
```
yaleman, line 158: This really feels like it should just make the program bail 🤔

Cargo.toml
```diff
@@ -75,13 +75,16 @@ sshkey-attest = { git = "https://github.com/kanidm/webauthn-rs.git", rev = "2218
 # webauthn-rs-proto = { path = "../webauthn-rs/webauthn-rs-proto" }
 # sshkey-attest = { path = "../webauthn-rs/sshkey-attest" }
 
+kanidm-hsm-crypto = { path = "../hsm-crypto" }
```
yaleman, line 78: 🤔

yaleman (commented):
Lookin' sensible so far 🥳

Cargo.toml
Firstyear: This is because I still need to finish the hsm-crypto side and publish a version.

unix_integration/src/resolver.rs
Firstyear: Yeah, I was thinking that too. But I'm not sure either. Is there a reason you'd install a kanidm resolver and *not* need pam_allow_groups to have entries? My only example here is I have a samba server in a container and it needs to *resolve* users, but not authenticate them, so this is empty in it's config iirc.

Firstyear:
yep, @dmulder was happy with it to via email. I need to finish adding the other hsm identity cert parts next, then I think a clean up and good to go.

Firstyear:
Waiting on https://github.com/kanidm/hsm-crypto/pull/18

unix_integration/src/resolver.rs
yaleman: Ah, I guess that'd work. Feels kinda icky to be using the IDM for one thing and not another 🤣

Firstyear: It's still IDM though, it's resolving identities. So kinda a bit in scope :)

unix_integration/src/idprovider/interface.rs
```diff
@@ -86,8 +91,52 @@ pub enum AuthCacheAction {
     PasswordHashUpdate { cred: String },
 }
 
+pub struct KeyStore<'a> {
```
dmulder, line 94: Can we wrap this in a RwLock, or do something similar? I'm getting errors because I can't pass the keystore across threads. Himmelblau is a bit unique in that we can have multiple backend providers, so I have a main provider that communicates with the child providers. Passing the keystore to the child providers throws the error.

dmulder (changes requested):

unix_integration/src/idprovider/interface.rs
```diff
@@ -86,8 +91,52 @@ pub enum AuthCacheAction {
...
+            .delete_tagged_hsm_key(tag)
+            .map_err(|_err| IdpError::KeyStore)
+    }
+}
+
 #[async_trait]
 pub trait IdProvider {
+    async fn configure_hsm_keys(
+        &self,
+        _keystore: &mut KeyStore,
```
dmulder, line 133: Passing the keystore reference here is what causes the error.

unix_integration/src/idprovider/interface.rs
Firstyear: What's the error? It might actually need to be Pin to make it work.

unix_integration/src/idprovider/interface.rs
Firstyear: It's also possible you can wrap this in your own lock on your side, but you still may need to dance around lifetimes here.

PR #2292, opened 2023-11-05
Moving daemon tracing to OpenTelemetry

yaleman:
Notes

* This changes the "core" logging pipeline to the opentelemetry-tracing one, but *doesn't* enable trace sending or the exporter by default.
* Tracing-forest still prints to `stdout`
* 

Checklist

- [x] cargo fmt has been run


server/core/src/actors/v1_write.rs
```diff
@@ -1568,7 +1568,14 @@ impl QueryServerWriteV1 {
         let res = idms_prox_write
             .qs_write
             .purge_tombstones()
-            .and_then(|_| idms_prox_write.commit());
+            .and_then(|changed| {
+                // we don't need to commit the txn if there were no changes
```
Firstyear, line 1572: Actually we do, because we need this on tombstones *at least* to keep the RUV "touched". I'm working on this a bit more soon.

server/lib/src/server/migrations.rs
```diff
@@ -756,7 +759,8 @@ impl<'a> QueryServerWriteTransaction<'a> {
         res?;
 
         // Delete entries that no longer need to exist.
-        let delete_entries = [UUID_IDM_ACP_OAUTH2_READ_PRIV_V1];
+        // TODO: Shouldn't this be a migration?
```
Firstyear, line 762: Probably, but too late now. :)

Firstyear (changes requested):
One change with tombstone handling, but otherwise all good.

server/core/src/actors/v1_write.rs
yaleman: Oh, I thought it was only commiting the TXN not doing... lots of other things. I'll back this out.

Firstyear (approved):

Issue #2295, opened 2023-11-06
OpenAPI validation issues


Water261:
Current error log when attempting to generate client using [openapi-generator](https://github.com/OpenAPITools/openapi-generator)
[error.log](https://github.com/kanidm/kanidm/files/13262358/error.log)

yaleman:
If you get the chance could you test against the above PR?

PR #2296, opened 2023-11-06
Update OpenAPI schema gen to actually... be kinda sorta valid.

yaleman:
Refers #2295 - I thought `utoipa` was validating the schema, turns out I was wrong...

Checklist

- [x] cargo fmt has been run

### DIFF
--- a/proto/src/v1.rs
+++ b/proto/src/v1.rs
@@ -235,6 +235,7 @@ pub enum OperationError {
     NoMatchingAttributes,
     CorruptedEntry(u64),
     CorruptedIndex(String),
+    // TODO: this should just be a vec of the ConsistencyErrors, surely?
     ConsistencyError(Vec<Result<(), ConsistencyError>>),
     SchemaViolation(SchemaError),
     Plugin(PluginError),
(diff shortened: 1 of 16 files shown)


Water261:
It almost works, just complains about a few things now. Two structs `ScimSyncRequest` and `IdentifyUserRequest` are missing from the file and anything that's wrapped in `Json<T>`  also breaks. Should just be `T` instead.

proto/src/v1.rs
```diff
@@ -235,6 +235,7 @@ pub enum OperationError {
     NoMatchingAttributes,
     CorruptedEntry(u64),
     CorruptedIndex(String),
+    // TODO: this should just be a vec of the ConsistencyErrors, surely?
```
Firstyear, line 238: Yeah it should. I had some past ideas here but forgot them. Likely consistency error needs to wrap operr and no the other way for ? To work too.

Firstyear (approved):

proto/src/v1.rs
yaleman: We should be able to something like a `filter_map` so it's just a `Vec` without the `Result` bit... will deal with it later.

Issue #290, opened 2020-07-17
Use environment variables for configuration


phozzy:
Currently in order to run a server you have to do that using the following command:
`/sbin/kanidmd server -c /data/server.toml`
or something similar.
That means that you have to provide a server configuration using a configuration file. That is nice for static environments, but for more dynamic environments such as k8s or similar ones it is nice to provide configuration using environment variables. It is also has a sense if you use systemd service, since systemd also allows you to inject environment variable in a systemd service.
So we can use one to one mapping of server configuration file parameters to environment variables (almost):
```
export BINDADDRESS="127.0.0.1:8443"
export LDAPBINDADDRESS="127.0.0.1:3636"
export DB_PATH="/tmp/kanidm.db"
export LOG_LEVEL="perffull"
```
And here comes a difference for environment variables for keys and certificates. For example in k8s environment you can keep this sensitive information using k8s-secrets. And use it after in one of the two ways:
* using config-maps, in this scenario those secrets appear inside a container a standard files, so nothing particularly changes here, and you create environment variables containing file paths:
```
export TLS_CA_PATH="/cert/insecure/ca.pem"
export TLS_CERT_PATH="/cert/insecure/cert.pem"
export TLS_KEY_PATH="/cert/insecure/key.pem"
```
the downside of this method is that in case you change the content of your k8s-secret, your deployment may be not aware (this may be not actual now, but it was when I worked with it) of those changes and will not restart your containers in order to reread the content of your k8s-secrets. And here comes the second way.
* using environment variables:
```
export TLS_CA="-----BEGIN CERTIFICATE-----\nblablabal....\nASEASDFA...aEFAE\n...\n...\n-----END CERTIFICATE-----"
export TLS_CERT="-----BEGIN CERTIFICATE-----\nblablabal....\nASEASDFA...aEFAE\n...\n...\n-----END CERTIFICATE-----"
export TLS_KEY_PATH="-----BEGIN PRIVATE KEY-----\n...\n...\n...\n-----END PRIVATE KEY-----"
```
That basically it, of course those variables naming and number can vary, but the idea should be clear.
Is a dependancy for #282

phozzy:
[Here](https://kubernetes.io/docs/concepts/configuration/secret/) is a general information about k8s secrets.

ryankurte:
[...]

PR #2329, opened 2023-11-23
Adding env var configs for the server

yaleman:
Refers #290 - adds env var config for *most* of the server config options available in the config file.

Checklist

- [x] cargo fmt has been run

### DIFF
--- a/server/core/src/config.rs
+++ b/server/core/src/config.rs
@@ -213,26 +213,16 @@ impl ServerConfig {
                     self.ldapbindaddress = Some(value.to_string());
                 }
                 "ROLE" => {
-                    self.role = match ServerRole::from_str(&value) {
-                        Ok(val) => val,
-                        Err(err) => {
-                            return Err(format!(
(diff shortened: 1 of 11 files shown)


server/core/src/config.rs
```diff
@@ -47,10 +44,20 @@ pub struct OnlineBackup {
...
     #[serde(default = "default_online_backup_versions")]
-    /// How many past backup versions to keep
+    /// How many past backup versions to keep, defaults to 7
     pub versions: usize,
 }
 
+impl Default for OnlineBackup {
+    fn default() -> Self {
+        OnlineBackup {
+            path: "/dev/null".to_string(),
```
Firstyear, line 54: Seems like a chaotic default 🤔

I wonder if we could make a backups dir next to the DB or something?L

Firstyear (approved):

server/core/src/config.rs
yaleman: Done. Urgh, messy.

Firstyear (approved):

PR #2382, opened 2023-12-19
Add improved domain migration framework and default MFA

### LINKED ISSUE #1435: Enforce MFA on persons by default
Currently we have two account classes (service accounts and persons). For service accounts they authenticate with a high entropy server side generated password, or an api-token. These tend to be very strong authentication mechanisms so aren't a concern today.

However, persons can have either:

* Password
* Password + TOTP
* Webauthn (Passkey)

I am considering if we want to shift our baseline for persons such that Password-Only is not allowed on their accounts. This would represent a significant improvement to user security.

Initially we would still allow password only auth, but we would issue a warning on authentication that this will be disallowed in the future. Users can then update to have password + totp and/or password. 

At a later time we can then fully and outright remove password as a singular auth type, and enforce that a password must always be paired with totp.

Thoughts? Is this a good improvement?

Firstyear:
Fixes #1435 - This upgrades our domain migrations to perform "once off" operations. This way when MFA is enforced it can be undone by admins without the changes being removed.

We can use this in future for other migrations too. 

Checklist


server/lib/src/server/migrations.rs
```diff
@@ -121,16 +122,42 @@ impl QueryServer {
...
+
+        // This is what tells us if the domain entry existed before or not.
+        let db_domain_version = match write_txn.internal_search_uuid(UUID_DOMAIN_INFO) {
             Ok(e) => Ok(e.get_ava_single_uint32(Attribute::Version).unwrap_or(0)),
             Err(OperationError::NoMatchingEntries) => Ok(0),
             Err(r) => Err(r),
         }?;
-        admin_debug!(?domain_info_version);
+
+        warn!(?db_domain_version);
```
yaleman, line 135: This should be a debug and say what we're doing
```suggestion
        info!(?db_domain_version, "Before setting internal domain info");
```

server/lib/src/server/migrations.rs
```diff
@@ -121,16 +122,42 @@ impl QueryServer {
...
+        write_txn.set_phase(ServerPhase::DomainInfoReady);
+
+        // This is the start of domain info related migrations which we will need in future
+        // to handle replication. Due to the access control rework, and the addition of "managed by"
+        // syntax, we need to ensure both node "fence" replication from each other. We do this
+        // by changing domain infos to be incompatible during this phase.
+
+        // The reloads will have populated this structure now.
+        let domain_info_version = write_txn.get_domain_version();
+        warn!(?domain_info_version);
```
yaleman, line 160: This should be a debug and say what we're doing
```suggestion
        info!(?db_domain_version, "After setting internal domain info");
```

yaleman (changes requested):

Firstyear:
Added your comments, thanks!

yaleman (approved):

PR #2388, opened 2023-12-21
daemon: Fix inverted logic on cache dir check

dmulder:
This logic is backwards in the daemon. We should print the "may not be RW" warning when the directory is read-only, not when the directory is *NOT* read-only.

Checklist

- [x] cargo fmt has been run


yaleman (approved):

yaleman:
Good catch!

Issue #2353, opened 2023-12-02
FR: kanidm system oauth2 set-origin


yaleman:
**Is your feature request related to a problem? Please describe.**

Be able to set the Origin for an OAuth2 RS

**Describe the solution you'd like**

`kanidm system oauth2 set-origin <client_id> <origin_url>`

**Describe alternatives you've considered**

yeah nah

**Additional context**

Suggested by @mweinelt

Water261:
Going to tackle this issue

PR #2418, opened 2024-01-06
Change OAuth2 RS Origin from the CLI

Water261:
Fixes #
#2353 
Checklist

- [x] cargo fmt has been run

### DIFF
--- a/tools/cli/src/cli/oauth2.rs
+++ b/tools/cli/src/cli/oauth2.rs
@@ -315,9 +315,9 @@ impl Oauth2Opt {
                         None,
                         Some(origin),
                         None,
-                        true,
-                        true,
-                        true,
+                        false,
+                        false,
+                        false,
                     )
                     .await
                 {
--- a/tools/cli/src/cli/oauth2.rs
+++ b/tools/cli/src/cli/oauth2.rs
@@ -28,6 +28,7 @@ impl Oauth2Opt {
             Oauth2Opt::CreateBasic { copt, .. } | Oauth2Opt::CreatePublic { copt, .. } => {
                 copt.debug
             }
+            Oauth2Opt::SetOrigin { nopt, .. } => nopt.copt.debug,
         }
     }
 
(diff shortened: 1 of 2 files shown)


tools/cli/src/cli/oauth2.rs
```diff
@@ -305,6 +306,25 @@ impl Oauth2Opt {
...
+            Oauth2Opt::SetOrigin { nopt, origin } => {
+                let client = nopt.copt.to_client(OpType::Write).await;
+                match client
+                    .idm_oauth2_rs_update(
+                        &nopt.name,
+                        None,
+                        None,
+                        Some(origin),
+                        None,
+                        true,
```
yaleman, line 318: The bools at the end shouldn't be true, they reset all the secrets associated with the RS.

yaleman (changes requested):

tools/cli/src/cli/oauth2.rs
Water261: 👍 Was just a guess, didn't actually know what to do

yaleman (approved):

tools/cli/src/cli/oauth2.rs
yaleman: No worries, that's why we have a review process 😄

yaleman:
Thanks!

andreas-bulling:
Hm, I was just looking for this feature but while it looked as if this was merged, the command is not available in the CLI. At least not in version 1.10.2

andreas-bulling:
Also not in the latest 1.11.0-dev...

Firstyear:
@andreas-bulling `kanidm system oauth2 add-redirect-url`. In the future please open a new issue instead of posting in old issues.

andreas-bulling:
I assumed that you prefer not to have a large number of issues, particularly new issues on an "old" topic. Plus, commenting here instead of opening a new one directly puts the comment/question in context and all people involved at the time are still involved.

But that's fine - sorry for that. I'll open new issues in the future then.

PR #2423, opened 2024-01-10
Upgrade replication to use anchors

### LINKED ISSUE #2174: Consumers that do not recieve any changes in an extended period will cause replication desync.
If a replica is online and performing incremental replication, if that replica does *not* generate a change within the changestate trim interval, then it can lead to the node desyncing. This is because it's ruv min / max converge to a single CID, which then is older than the changestate trim interval. On the next change event, then the ruv would "jump" ahead and cause the servers ruv window to exceed the other replicas ruv window for this server.

Some possible solutions:

* Keepalive (touch) on a schedule, perhaps with the RUV trim events to cause frequent but small changes. 
* Have the RUV add blank CID's during a trim (currently if there are no changes in a write txn then no CID is recorded to the RUV).

Relates #68

Firstyear:
Fixes #2174 - Due to how the RUV worked, if a server did not accept a write within changelog-max-age, it's ts_min and max would converge to a value lower than the changelog period. When that server then accepted a new write, it would "jump" the ts_min and max ahead. This could lead to a condition where a consumer could then not accept from the supplier as there was no overlap in the changelog windows.

To resolve this, we use the tombstone purge mechanism to create "anchors", which are increments to the ruv containing no changes. This way all servers RUV's are continually advancing within the change window.

During a replication these anchors are then sent to partners which use them to create empty RUV datapoints, giving a longer trail of CID's that can be trimmed allowing ts_min to stay within a reasonable window.

Checklist

- [  ] cargo clippy has been run


server/lib/src/be/mod.rs
```diff
@@ -1257,11 +1262,16 @@ impl<'a> BackendWriteTransaction<'a> {
...
+        // Mark a new maximum for the RUV by inserting an empty change. This
+        // is important to keep the changestate always advancing.
+        self.get_ruv().insert_change(cid, IDLBitRange::default())?;
+
         // We plan to clear the RUV up to this cid. So we need to build an IDL
         // of all the entries we need to examine.
-        let idl = self.get_ruv().trim_up_to(cid).map_err(|e| {
-            admin_error!(?e, "failed to trim RUV to {:?}", cid);
+        let idl = self.get_ruv().trim_up_to(trim_cid).map_err(|e| {
+            admin_error!(?e, "failed to trim RUV to {:?}", trim_cid);
```
yaleman, line 1274:
```suggestion
            admin_error!(?e, "During tombstone cleanup, failed to trim RUV to {:?}", trim_cid);
```

server/lib/src/repl/tests.rs
```diff
@@ -1881,6 +1878,143 @@ async fn test_repl_increment_consumer_ruv_trim_past_valid(
     drop(server_b_txn);
 }
 
+// Test two synchronised nodes where changes no are occuring - this situation would previously
```
yaleman, line 1881: Is [no how you make repl](https://youtu.be/WADdeYZYmrE?t=4)?

```suggestion
// Test two synchronised nodes where changes are not occuring - this situation would previously
```

yaleman (approved):
Can we please add the Anchored RUV description somewhere other than in a PR, maybe in the docs, or the docstring of `ReplAnchoredCidRange`? 😄

Issue #41, opened 2019-04-17
Add support for API-application passwords


Firstyear:
It's common that applications (like email, imap etc), need passwords and can't supply 2fa. Allow generation of passwords for these services, and allow them to have constrained rights related to their operations and interactions.

Firstyear:
@mweinelt You want this issue for your mail setup.

TheRealGramdalf:
I'm not sure how this would really work out, but it would be nice to use this with [languagetool](https://github.com/languagetool-org/languagetool) to support user authentication at some level. You'd likely need some sort of proxy (I'm not sure if traefik works for that or not) that interfaces with Kanidm, since that is probably out of scope for Kanidm itself.

yaleman:
@TheRealGramdalf please have those kinds of discussions in the repos that they relate to (ie, languagetool's).

TheRealGramdalf:
> @TheRealGramdalf please have those kinds of discussions in the repos that they relate to (ie, languagetool's).

Sorry, I meant it would be nice to have Kanidm integrate such that api tokens are issued and managed by Kanidm, but that I might need some sort of third-party proxy to put between languagetool and Kanidm. Given that, I'm not sure what the best way to implement it would be - what protocol would Kanidm use to create/vet API tokens? Is this something Oauth2 could cover, or is there another protocol I'm unaware of?

Firstyear:
This issue is for discussion of this specific feature, not for discussion of third party integration details.

ieugen:
Hi, 

I believe I already found a use-case for this.
I have kanidm providing Auth (OpenID Connect) for Nextcloud which provides a webdav endpoint that I use in KOReader on an ebook reader. 
I use passkeys and pass+otp for kanidm. 
KOreader does not support any of the above and would probably only work with app passwords or if I switch to LDAP - which I would like to avoid, because I kind of like passkeys now :) . 

https://github.com/koreader/koreader/ .

I am eager to test application passwords ASAP.

pando85:
Hello @ieugen,

As far as I know the recommended way for accessing Nextcloud through third-party apps is using Nextcloud application passwords: https://docs.nextcloud.com/server/latest/user_manual/en/files/access_webdav.html#third-party-webdav-clients

At least I use it like that and it works perfectly.

ieugen:
> Hello @ieugen,
> 
[...]

Issue #22, opened 2019-02-02
Event auditing


Firstyear:
In some cases we want to record audit events on objects for transparency. For example, on failed authentication we want to write to the object about the failed authentication. We want to audit account impersonations, password changes, etc. 

For some events, these should imply a write to the object itself that allows the object to self-read it's audit information about what has happened and who is using the account.

yaleman:
The more I think about the transport/pipeline for these is that OpenTelemetry's model is one of the best options - Kanidm publishes the events into a pipeline and then if folks want to dump them to disk/stdout then that's doable, or directly publish to one of the many SIEM/logging providers, or a database etc.

Example here: https://github.com/open-telemetry/opentelemetry-rust/blob/18f6c01ee35f2e7909fe46a815ecd6f958ecef71/examples/logs-basic/src/main.rs

Firstyear:
Yeah, I think for now internally to Kanidm each server will have it's own queue (quwu) then I think opentelemetry can be a backend to that. I wouldn't want opentelemetry in the internal server paths as I don't know much about it's internals/performance, so better to control the queueing internally then we can have that feed opentelemetry in a seperate thread/worker.

yaleman:
Yep, as I understand it, logs are just hooked into the default log provider and it sits off the side doing its thing.

Firstyear:
> Yep, as I understand it, logs are just hooked into the default log provider and it sits off the side doing its thing.

Yep, and later we can have that task funnel events into something else like opentelemetry etc.

dvv:
Hi!

I wonder what's the status of the progress?
Can it be used to catch details of changes made to person/group/oauth2 via API or client tool?
If it's not at all intended for the above how it's possible to provide kinda eventsource of changes made to kanidm objects?

TIA

yaleman:
At the moment there's no design or implementation.

PR #2427, opened 2024-01-12
docs: Add application passwords design document

scabrero:
Draft PR to discuss the design.

Related to #41 

Checklist

- [ ] cargo fmt has been run


book/src/developers/designs/application_passwords.md
```diff
@@ -0,0 +1,188 @@
...
+The user can login to the webui or command line and list what linked
+applications exist on their account that require application passwords.
+
+The user can then request a new "application password" for the mail server for
+their laptop. They can copy-paste the generated password to their mail client
+which uses this password on their behalf. Similar he can request a new
+application password to access the web server basic auth.
+
+As a user can request several application passwords for each application, the
+username will have a label appended like "username@laptop" or "username@phone".
```
yaleman, line 43: The username@app format's a problem given we already use SPN's which are `user@domain`, maybe `username+app` ?

book/src/developers/designs/application_passwords.md
```diff
@@ -0,0 +1,188 @@
...
+
+The mail services and web services are configured to point to Kanidm's LDAP
+gateway with a customized search base DN.
+
+The user can login to the webui or command line and list what linked
+applications exist on their account that require application passwords.
+
+The user can then request a new "application password" for the mail server for
+their laptop. They can copy-paste the generated password to their mail client
+which uses this password on their behalf. Similar he can request a new
```
yaleman, line 39:
```suggestion
which uses this password on their behalf. Similar they can request a new
```

book/src/developers/designs/application_passwords.md
```diff
@@ -0,0 +1,188 @@
+# Rationale
+
+Kanidm exists to provide an authentication source for external applications.
+These applications need to have standardised ways to integrate with Kanidm to
+allow that application to interact and trust Kanidm's authentication results.
+
+For web based applications we offer Oauth2/OIDC. For Linux machines we offer a
+Kanidm specific HTTPS channel for identifying users. We also offer an LDAPS
+gateway for applications that don't support other protocols but we want users to
+use their "UNIX password".
```
Firstyear, line 10: The point of this is also to remove the ability for unix password to do anything outside of unix integration, that way we can remove this from ldap. :)

book/src/developers/designs/application_passwords.md
Firstyear: There actually isn't a need for this at all. 

Because the imap server is configured to ldap `app=imap,dc=idm,dc=example,dc=com`, and http to `app=http,dc=idm,dc=example,dc=com` then we can use username/spn as usual. The distinction/routing of the application password to use *isn't part of the username*. It's the applications ldap basedn that routes to the specific application that we are interacting with at this point in time.


As a more concrete example:

```
ldapwhoami -D name=william,app=imap,dc=idm,dc=example,dc=com -w "imap-application-password-here"

ldapwhoami -D name=william,app=http,dc=idm,dc=example,dc=com -w "http-application-password-here"
```

That's why both imap/http would have the search basedns move to app=imap,dc=idm,dc=example,dc=com or app=http,dc=idm,dc=example,dc=com

book/src/developers/designs/application_passwords.md
```diff
@@ -0,0 +1,188 @@
...
+Since application passwords are related to applications, on delete of an
+application all entries that have a bound application password should be removed
+from user accounts.
+
+### Access controls
+
+The IDM administrators group can manage the applications and user's application
+passwords.
+
+A user can manage his own application passwords.
```
Firstyear, line 170: their own

book/src/developers/designs/application_passwords.md
```diff
@@ -0,0 +1,188 @@
...
+  application: Uuid,
+  label: String,
+  password: Password,
+}
+
+struct ValueSetApplicationPassword {
+  map: Map<Uuid, ApplicationPassword>
+}
+```
+
```
Firstyear, line 140: I think this may be inverted.

You need ValueSetApplicationPassword to be a map of:

```

type ApplicationPasswords = BTreeMap<Uuid, (String, Password)>;

type ApplicationUuid = Uuid;

struct ValueSetApplicationPassword {
  map: Map<ApplicationUuid, ApplicationPasswords>
}
```

This way when the authentication request comes in for app=X, we can get the applications UUID in the first step. This allows us to look up the account and then get application passwords by the UUID of the app. 

In the design you propose, we would need to iterate over all possible application passwords to find the one that matches rather than being limited to getting just the ones related to the application ID.

Firstyear (changes requested):

book/src/developers/designs/application_passwords.md
scabrero: It is not the app what is appended but the label. As a user can have multiple passwords for the same application it helps to identify which one a client application is using. For example, a user configures his mail client on his phone and his laptop and several months later he wants to delete the phone's one, but he was not careful enough to properly label the passwords at creation time. If the label is appended to the username, he could open the mail application settings on the phone and check the username to identify the app password to remove on the server.

This is just an idea I took from Roundcube because it is useful to myself, but if we agree it is not convenient no problem on dropping it.

book/src/developers/designs/application_passwords.md
scabrero: Ok, updated. It will impact queries by Uuid `PartialValue::ApplicationPassword(Uuid)` but those should be less frequent than authentication requests.

Firstyear: I doubt those are going to be the common thing though. We should be optimising for the common case here.

book/src/developers/designs/application_passwords.md
Firstyear: When it would be confusing is our eventual trusts setup because then you'd have username@domain@label. So I think it gets pretty chaotic there. 

I'm looking at other systems like fastmail, and they have the labels as something for the user in the webui to know what each application password was used for, rather than it being a part of the username. I think username@label being tied to an app pw would create some confusion to what users have to enter.

So lets drop the @label here I think.

scabrero:
Updated addressing all comments and extended to include the new CLI commands.

book/src/developers/designs/application_passwords.md
```diff
@@ -0,0 +1,215 @@
...
+  map: Map<ApplicationUuid, ApplicationPasswords>
+}
+```
+
+Each value in the set is queried by its UUID. This defines the value as
+`Value::ApplicationPassword(Uuid, ApplicationPassword)`. The
+`ApplicationPassword` type implements `PartialEq` so two application passwords
+are equal if their label is equal and they refer to the same application.
+
+(NOTE: We could consider extending the `CredentialType` to add
```
Firstyear, line 147: I think we don't want to do this, credential type is already pretty overloaded from past design flaws. I think the proposed design here is clearer and going to be nicer :)

book/src/developers/designs/application_passwords.md
```diff
@@ -0,0 +1,215 @@
...
+
+### Reference integrity
+
+Since application passwords are related to applications, on delete of an
+application all entries that have a bound application password should be removed
+from user accounts.
+
+### Access controls
+
+The IDM administrators group can manage the applications and user's application
```
Firstyear, line 168: I think this should be an "application administrator" group. It should be able to define new applications etc. 

The application should also allow "managed by" so that it can have delegated administration.

Finally, "application user password admins" should only be able to assert if a pw is present, or delete them - they should NOT be able to create them on users. Only a user should be able to self-create their own application password.

book/src/developers/designs/application_passwords.md
```diff
@@ -0,0 +1,215 @@
...
+The rootdse needs to be extended to include the applications as additional
+naming contexts.
+
+If the session is related to an application we should only accept application
+passwords in the bind. The user needs to be a member of the application. Binds
+to an application password must be limited by account validity and expiration.
+
+We need to add a cache of available applications for lookup.
+
+(NOTE: Caching and reload needs more explanation)
```
Firstyear, line 189: That's fine, we can get to that later.

book/src/developers/designs/application_passwords.md
```diff
@@ -0,0 +1,215 @@
...
+
+### kanidm CLI
+
+The `kanidm` command line tool will be extended with the following commands to
+manage the applications:
+
+* kanidm application create <NAME>
```
Firstyear, line 196: applications also need service-account api-tokens with them. This is because the application needs to be able to search ldap with elevated read permissions potentially. 

So I think before we define what the CLI will look like, we need to look at what we may need to configure.

book/src/developers/designs/application_passwords.md
```diff
@@ -0,0 +1,215 @@
...
+* kanidm application delete <NAME>
+* kanidm application list
+* kanidm application get <NAME>
+* kanidm application add-members <NAME> <MEMBER> ... <MEMBER>
```
Firstyear, line 200: I think to be consistent, shouldn't we have this be like oauth2 and we reference which groups and their members are allowed to access an application, rather than the application being a group itself?

book/src/developers/designs/application_passwords.md
yaleman: "Admin needs to define an app" rather than "the user defines a label, and uses it" feels like a usability problem, TBH.

Firstyear: Well the admin has to create the application first to say "this imap server will use application passwords" then the user who is a memberof the allowed access group may create an application password to use said service. So I don't see what the issue is?

yaleman: clearly I don't remember reading this in full at all, ima try again. ignore me for now 😮‍💨

Firstyear: No problem <3

book/src/developers/designs/application_passwords.md
```diff
@@ -0,0 +1,215 @@
...
+implemented as an authentication for all applications, we can use this to
+provide *application specific* password based authentication and remove the
+ability to bind with the UNIX password.
+
+# User experience
+
+The administrator configures two applications on their Kanidm instance. One is
+"mail" for a generic SMTP+IMAP service. The other is HTTP basic auth to a legacy
+web server. Once applications are added, the administrator adds the allowed
+users to the applications as if they were regular groups.
```
Firstyear, line 30: I thought about this more and I think that for consistency, we should act the same as oauth and access controls. Rather than the application entry acting as a group, we should have the application entry nominate a single group as the "receiver" of the application, and then members of that group are allowed to access the application.  Does this seem reasonable?

scabrero: Yes, I prefer your proposal, but then users will loose the ability to inherit `memberof="applicaiton X"` (NOTE in "Application Entires" subsection)

book/src/developers/designs/application_passwords.md
scabrero: Replied below.

book/src/developers/designs/application_passwords.md
scabrero: Ok

book/src/developers/designs/application_passwords.md
```diff
@@ -161,55 +154,61 @@ needed.
 
 Since application passwords are related to applications, on delete of an
 application all entries that have a bound application password should be removed
-from user accounts.
+from user accounts. The associated group will be automatically deleted too.
```
Firstyear, line 157: What associated group? It shouldn't be deleted when you delete the application. Imagine someone set the application user group to 'idm_all_persons', and then deleted the application. This would be a potential exploit to delete and damage data.

book/src/developers/designs/application_passwords.md
```diff
@@ -84,19 +83,18 @@ the UNIX one.
...
-(NOTE: It could be good to make this a group such that members of the group
-inherit "memberof=application X". This would make presentation of application
-password links in the UI easier since we can then do a search on (memberof &&
-class = application). Alternately we don't need this since we can in-memory
-cache applications and just compare their member sets to the user sessions
-memberof)
+(NOTE: Should we also disallow the creation of application passwords for an
+application if the user is not a member of the associated group? This will
+complicate reference integrity, as removing a member will require to remove the
+application password.)
```
Firstyear, line 93: Referential integrity is to "clean up" references as a cascade when a delete occurs. It's NOT there to impose constraints on what is or isn't allowed to exist.

We can implement that logic in the idm layer so that on the api call to create a new application password can check if the user is a memberof the needed access groups. That way we can enforce this constraint at application password create time. 
The question *then* is how we handle the application password if the user is removed from the group that can access the application. IMO it should NOT trigger the delete. You can imagine a scenario where someone says "I lost my laptop, please disable my access temporarily". You can move them out the group and back in when they find the device for example. There are plenty of reasons to move the user out then into the application group if needed that shouldn't delete the application password. 

(It's also an argument that we should have booleans on each app password that says if that password is enabled, so that the user can selectively enable/disable application passwords).

book/src/developers/designs/application_passwords.md
```diff
@@ -161,55 +154,61 @@ needed.
...
 
-The IDM administrators group can manage the applications and user's application
-passwords.
+The "Application administrators" group will manage the applications, and
+applications will allow "managed by" so that they can have delegated
+administration.
 
-Users can manage their own application passwords.
+The "Application user passwords administrators" group will be able to list the
+users's application passwords and delete them but only the users will be able
```
Firstyear, line 166: If we add the enable/disable flag, they should be able to enable or disable them too.

book/src/developers/designs/application_passwords.md
```diff
@@ -161,55 +154,61 @@ needed.
...
+passwords.
 
 ### LDAP
 
-The bind DN regular expression needs to adjusted to detect and determine the
-bind dn if it is related to an application or not. The application bind dn
-regular expression will capture the user name, the label and the application
-name.
+As application passwords will replace binding with UNIX passwords, the
+set-ldap-allow-unix-password-bind domain setting will be removed.
```
Firstyear, line 173: No? That flag exists for a reason, it's there as a transition mechanism so that as users start to roll this out they can selectively disable unix binds from being allowed. If we just remove that feature and flag we will break existing user set ups. We need to always consider upgrades and migration paths.

book/src/developers/designs/application_passwords.md
yaleman: The "disable this password" thing feels like a too-many-knobs situation (and a "why not just delete and create a new one). 

- there's prior art in that individual AWS credentials can be disabled and re-enabled
- it'd be useful to test disabling a credential from a "scream testing" perspective, to see if it'll break anything in complex legacy workflows (which is why they're using LDAP in the first place).

On the delete thing, that should be a system-level cleanup task TBH, disable the credential when they're removed from the group and have an updated/last used flag on the cred, with a method to remove "floating" creds after x time.

Firstyear: yep i agree. It's also easier to add later than remove something we don't need later.

book/src/developers/designs/application_passwords.md
scabrero: Ok, fixed.

book/src/developers/designs/application_passwords.md
scabrero: You are right. Fixed.

book/src/developers/designs/application_passwords.md
scabrero: Ok. Please check the doc update to verify I have captured your ideas correctly.

book/src/developers/designs/application_passwords.md
```diff
@@ -0,0 +1,229 @@
...
+generated for them. These are optional since an anonymous bind to kanidm and
+searching under the basedn or application base dn will continue to work.
+
+Application should have a URL reference to help admins identify where the
+application may be located or accessed.
+
+(NOTE: Future, it could be good to allow customisable instructions for users on
+where to go to use their app password?)
+
+(NOTE: To make it easier to manage tokens, it could be good to have a
```
Firstyear, line 109: Seems like a reasonable idea? That or we just make it so that LDAP applications are an extension of service accounts?

book/src/developers/designs/application_passwords.md
```diff
@@ -0,0 +1,229 @@
...
+"token-capable" class type to prevent duplication this with service-accounts,
+and allows adding tokens to future types).
+
+### Accounts
+
+The user may wish to have multiple passwords per application. Each password must
+have a label to identify it, a enabled/disabled flag and a last used timestamp.
```
Firstyear, line 116: I think in an earlier comment yaleman suggested we don't added the enable/disable flag yet as prior art doesn't have these on other systems. It's also easy to add this later - it's harder to take it away once we add it.

book/src/developers/designs/application_passwords.md
```diff
@@ -0,0 +1,229 @@
...
+
+Person accounts will need a new `Attribute::ApplicationPassword` that stores a
+`ValueSetApplicationPassword`. Each value in the set is a new type to manage these
+secrets and their labeling and the references to the applications.
+
+```
+struct ApplicationPAssword {
+    label: String,
+    enabled: bool,
+    last_used: Duration,
```
Firstyear, line 135: Should we be storing last_used here or should this be part of #22  - I think updating this on every bind could get *really* heavy on DB writes very quickly.

book/src/developers/designs/application_passwords.md
```diff
@@ -0,0 +1,229 @@
...
+administration.
+
+The "Application user passwords administrators" group will be able to list the
+users's application passwords and delete them but only the users will be able
+to, additionally to listing and deleting, self-create their own application
+passwords.
+
+### LDAP
+
+The rootdse needs to be extended to include the applications as additional
```
Firstyear, line 189: I don't think so, because this could get very noisy very fast, and clients could then try to walk all the different DIT's

Firstyear (changes requested):

book/src/developers/designs/application_passwords.md
scabrero: Application schema class could supplement service account to add its own attributes, like the associated group.

book/src/developers/designs/application_passwords.md
scabrero: Ok, I understood it wrong. Reverted.

book/src/developers/designs/application_passwords.md
scabrero: OK, removed.

book/src/developers/designs/application_passwords.md
scabrero: Good point. I will keep it simple for now and remove the field.

book/src/developers/designs/application_passwords.md
Firstyear: I think that seems reasonable.

scabrero: :+1:

scabrero:
Apart from application caching I think this is ready to go.

Firstyear (approved):

PR #2428, opened 2024-01-13
2390 1980 allow native applications

### LINKED ISSUE #1980: Allow multiple redirect_uris in oauth2

This issue is almost the same as https://github.com/kanidm/kanidm/issues/1290 , I am using kanidm as the identity provider for a self-hosted [netbird management server](https://netbird.io). The problem I have is that for the PKCE authorization flow, it uses internal urls : `http://localhost:53000`, but in kanidm the url registered is `https://netbird.example.selfhosted:443`, which is different. Therefore, I cannot setup the mobile client for netbird.

It is possible to change part of the redirect url in the flow to `https://netbird.example.selfhosted:53000`, but the port cannot change as it ~identifies the device trying to connect, and thus the redirect_uri check still fails.

Reading the previous issue, I see that adding multiple redirect_uris would come with its issues, but would resolve the problem here. I don't know if only comparing part of the uri in https://github.com/kanidm/kanidm/blob/master/server/lib/src/idm/oauth2.rs#L1246-L1254 and leaving out the port would be a solution.

### LINKED ISSUE #2319: Custom claims for OIDC tokens

I can't use Kanidm with [Minio](https://min.io/) via OIDC because I can't find a way to return a `policy` claim. Minio states the following in [its OIDC prerequisites](https://min.io/docs/minio/linux/operations/external-iam/configure-openid-external-identity-management.html#openid-connect-oidc-compatible-identity-provider):

> Ensure each user identity intended for use with MinIO has the appropriate claim configured such that MinIO can associate a policy to the authenticated user. An OpenID user with no assigned policy has no permission to access any action or resource on the MinIO cluster.

I'd like to configure a `policy` claim on user tokens. In my case, I'd be happy to configure this token on a per-user basis, as I have a small number of users.

I may choose to integrate Minio using its LDAP support, but I'd definitely prefer to use OIDC.

**Additional context**

I'm guessing this might involve a "custom claim map" as mentioned at https://github.com/kanidm/kanidm/issues/612#issuecomment-1596424088

### LINKED ISSUE #2390: RFC8252 - OAuth2.0 for Native apps

Revisiting the "OAuth2 for localhost" discussion - [RFC 8252 Section 7.3](https://www.ietf.org/rfc/rfc8252.html#section-7.3) explicitly calls this out as a thing, and [Section 8.1](https://www.ietf.org/rfc/rfc8252.html#section-8.1) notes that PKCE is a reliable mitigation for "you started this flow, then something else tries to steal it".

- allow registration/management of additional origins for OAuth2 Resource Servers
- warn/reject `localhost` origins, maybe? per [8.3](https://www.ietf.org/rfc/rfc8252.html#section-8.3)
- enforce that if an origin has a host of `localhost` or `127.*`, then PKCE *must* be enabled
  - basic RS? adding one of those URIs? reject
  - creating an RS and specifying a matching origin? nope
- [always ask for confirmation per 8.6](https://www.ietf.org/rfc/rfc8252.html#section-8.6) for `localhost/127.*` redirects

none

**Additional context**

This is used by [Hashicorp Vault for OIDC CLI logins](https://developer.hashicorp.com/vault/tutorials/auth-methods/oidc-auth) - `http://localhost:8250/oidc/callback` is the URL, but can be configured.

Firstyear:
Fixes #2390 Fixes #1980 Fixes #2319 

This adds custom claims and support for native applications (via localhost redirection).

Checklist


book/src/integrations/oauth2.md
```diff
@@ -200,32 +200,48 @@ kanidm system oauth2 reset-secrets
...
-imagepath=../images
-title=WARNING
-text=Changing these settings MAY have serious consequences on the security of your resource server. You should avoid changing these if at all possible!
-}}
+To create or update a claim map on a client:
 
-<!-- deno-fmt-ignore-end -->
+```
+kanidm system oauth2 update-claim-map <name> <claim_name> <kanidm_group_name> [values]...
+kanidm system oauth2 update-claim-map nextcloud account_role nextcloud_admins admin
```
yaleman, line 217: Can this be extended to show multiple claims just so it's clear, I assume they're space-delimited on the CLI?

```suggestion
kanidm system oauth2 update-claim-map nextcloud account_role nextcloud_admins admin secondclaim
```

book/src/integrations/oauth2.md
```diff
@@ -235,24 +251,53 @@ applications that act as the OAuth2 client and its corresponding webserver that
 server. In this case the SPA is unable to act as a confidential client since the basic secret would
 need to be embedded in every client.
 
+Another common example is native applications that use a redirect to localhost. These can't have a
+client secret embedded, so much act as public clients.
```
yaleman, line 255:
```suggestion
client secret embedded, so must act as public clients.
```

tools/cli/src/opt/kanidm.rs
```diff
@@ -900,12 +962,28 @@ pub enum Oauth2Opt {
...
+        name: String,
+    },
+
     /// Use the 'name' attribute instead of 'spn' for the preferred_username
     PreferShortUsername(Named),
     #[clap(name = "prefer-spn-username")]
     /// Use the 'spn' attribute instead of 'name' for the preferred_username
     PreferSPNUsername(Named),
-    /// Set the origin of a client
+    /// Set the origin of an oauth2 client
```
yaleman, line 986:
```suggestion
    /// Set the origin of an OAuth2 client
```

Firstyear: All the cli helps have lower case here.

yaleman: 😢

yaleman (approved):

PR #2456, opened 2024-01-25
20240125 2217 client credentials grant

### LINKED ISSUE #2217: OAuth 2 Client Credentials Grant

OAuth 2.0 has support for client credentials, which can provide for machine-to-machine services to gain an access token. Please consider adding this support.

For example, Prometheus natively supports this which would allow one to scrape services behind protected endpoints: https://prometheus.io/docs/alerting/latest/configuration/#oauth2 

I think [this is the relevant RFC](https://datatracker.ietf.org/doc/html/rfc6749#section-4.4).

I'd like to be able to configure kanidm for client credentials granting.

Not using kanidm, basic authentication (where possible), not protecting machine-to-machine traffic with authentication.

Firstyear:
Fixes #2217 - This adds support for client credential grants.

Checklist


Firstyear:
So the current question I'm "stuck" on is scopes/groups.

Currently, we rely on groups to create scope mappings. However, once we allow the oauth2rs to have an access token, it needs to be able to access scopes. It also needs to have an id that can be introspected.

We also have to make the oauth2rsname become name else a user could name collide with the oauth2rs and a naive consumer may incorrectly believe the user's access token is the token of the oauth2rs, granting incorrect or elevated privileges.

This walks us pretty close to the "oauth2rs is an account". Where this has implications is that we then need it to have the class "Account" associated to it, and it will show up in things like idm_all_accounts.

* make special exemptions for the oauth2rs in this situation?

An example of this is in SPN handling - we need this since prefer shortname is a preference.  However, SPN's currently only generate on account and group. I've currently added a special exemption here for oauth2rs, but it feels "messy". But it also doesn't cross any further boundaries so far. 

* Alter accounts to be "something that can auth in some way" and then move things like the passkey schema support into person / service account directly?

This would make oauth2rs natively an account, and avoid it having the ability to also act as a service account. It simply becomes an account that only can auth to it's associated relying parties.

* Make oauth2rs a full blown service account

This would allow some edge cases to work, but I don't really like this. It does however solve the problems by declaring the oauth2rs as a proper account with auth access etc. 



I guess this comes down to:

* Should oauth2rs be classified as an "account" since it may now authenticate to a relying party? 
* If it is an account, how should it behave?

Firstyear:
After a walk, I realised that I'm potentially confusing two things.

We have the idea of an identity - a unique entity, something that can be ... well, identified. We already have this with our internal Ident structures.

And then we have an account which is an identity that can authenticate. (is account the right word at that point?). 

So I think here what I'm confusing is those two. I propose we add an Identity Class that applies to all service-accounts and persons, and in this case oauth2rs. However, we leave account as "identities that can authenticate". This identity idea then works nicely to trusts in future where an identity may need to be trusted even though it indirectly authenticates such as in oauth2rs.

yaleman:
I think "account" is still appropriate terminology for the OAuth2RS in this mode - it's a combination of an identifier and credential which allows a client to authenticate.

At the moment we have:

- Service Account `class=[object,account,service]`
- Person `class=[object,account,person]`

Why not add:

- OAuth2RS `class=[object,account,OAuth2ResourceServer]` (which the entryclass already exists)

Firstyear:
Because he class "account" also allows authentication via passkeys and other bits. It's one of those "legacy" hold overs. It also would then allow oauth2rs to authenticate in it's own right. So maybe this is a prompt for me to clean that all up ...

yaleman:
Yeah, sounds like we need to limit auth per class in that case, since ... I guess service accounts could auth by passkeys too?

Firstyear:
Well it's starting to come up a little bit on the credentials client request and the ldap application design that we probably need to tidy up the idea of "account" a bit here, so I think now is the time. my thinking is "account" says "here is a unique identity that can authenticate in some way" but doesn't define the authentication methods. Then we move things like passkey/password to the actualy edge types like person/service accounts themself to precisely say "a service account can use generated passwords" and "persons can use passkeys".

Because i suspect this might come up with domain join too, that then you have "accounts" for machines, and they will need a completely different auth process. 

Anyway, so I'll do a bit of that tidy here.

yaleman:
Yep, that sounds sensible to me... "account" is an entity that can auth, etc etc.

Firstyear:
Wouldn't really be a kanidm release eve without me landing some insane huge change the day before ...

yaleman (changes requested):
Minor nitpicks but otherwise wooooo

yaleman (approved):

PR #2457, opened 2024-01-25
Application passwords implementation

scabrero:
This dirty implementation was done as an exercise to learn rust and kanidm internals.

It does not completely follow the design document #2427, but still it will be helpful to get some feedback and check if I am on the right track.

Related to #41

Checklist

- [ ] cargo fmt has been run


server/lib/src/credential/apppwd.rs
```diff
@@ -0,0 +1,37 @@
...
+    pub(crate) application: Uuid,
+    pub(crate) label: String,
+    pub(crate) password: Password,
+}
+
+impl ApplicationPassword {
+    pub fn new(
+        application: Uuid,
+        label: &str,
+        cleartext: &str,
```
Firstyear, line 17: crypto policy should be an argument to this function. Cryptopolicy minimum is NOT what you want because when we start up  we do a calculation to find a baseline that's required on this CPU (https://github.com/kanidm/kanidm/blob/master/server/lib/src/idm/server.rs#L151) . We store this in IDM Server Write transactions for you to use, https://github.com/kanidm/kanidm/blob/master/server/lib/src/idm/server.rs#L132 . This way you can pass it in to password hash calls such as here https://github.com/kanidm/kanidm/blob/master/server/lib/src/idm/server.rs#L1707

server/lib/src/idm/account.rs
```diff
@@ -683,6 +704,38 @@ impl Account {
...
     }
+
+    pub fn verify_application_password(
+        &self,
+        app_uuid: &Uuid,
+        app_label: &str,
+        cleartext: &str,
+        ct: Duration,
+    ) -> Result<Option<UserAuthToken>, OperationError> {
+        for ap in self.apps_pwds.values() {
```
Firstyear, line 715: Why iterate here? You can dho self.apps_pwds.get(app_uuid) to get all the passwords related to that application uuid, and then you limit how many iterations you need to do.

server/lib/src/idm/account.rs
```diff
@@ -683,6 +704,38 @@ impl Account {
...
+            if *app_uuid == ap.application && app_label == ap.label {
+                if ap.password.verify(cleartext).map_err(|e| {
```
Firstyear, line 717: Might be a bit clearer to separate this up to:

```
let verification = ap.password.verify( ... ) ?;
if verification {
    // success path
}
```

server/lib/src/idm/account.rs
```diff
@@ -683,6 +704,38 @@ impl Account {
...
+                    error!(crypto_err = ?e);
+                    e.into()
+                })? {
+                    let session_id = uuid::Uuid::new_v4();
+                    let uat: Option<UserAuthToken> = self.to_userauthtoken(
+                        session_id,
+                        SessionScope::ReadWrite,
```
Firstyear, line 724: Needs to be ReadOnly

server/lib/src/idm/account.rs
```diff
@@ -683,6 +704,38 @@ impl Account {
...
+                        ct,
+                        DEFAULT_AUTH_SESSION_EXPIRY,
```
Firstyear, line 726: We might need a separate session expiry here depending on the application in question? @yaleman Is session time on ldap passwords something we should have configurable in future?

server/lib/src/idm/ldap.rs
```diff
@@ -63,6 +62,12 @@ pub struct LdapServer {
 enum LdapBindTarget {
     Account(Uuid),
     ApiToken,
+    Application(Uuid, Uuid),
```
Firstyear, line 65: You can do enum structs like:

```
LdapBindTarget {
    ApiToken,
   Application {
       app_id: Uuid,
        user_id: Uuid,
    }
}
```

That way it's a bit better to label/mark fields. The reason we hae enums without this is just historical, that this feature didn't always exist in rust (or I didn't know about it). The woes of project age.

server/lib/src/idm/ldap.rs
```diff
@@ -146,17 +153,40 @@ impl LdapServer {
...
+            let app_entries = idms_prox_read.qs_read.internal_search(filter!(f_eq(
+                Attribute::Class,
+                EntryClass::Application.into()
+            )))?;
+            for app_entry in &app_entries {
+                let app_nc = app_entry
+                    .get_ava_single_iname(Attribute::Name)
+                    .map(|s| s.to_string())
+                    .ok_or(OperationError::InvalidEntryState)?;
+                rootdse.attributes.push(LdapPartialAttribute {
```
Firstyear, line 170: I don't think we should be pushing every application to a name context in ldap, I think it could get really noisy, and they aren't "really" seperate name contexts. They're "db views" in a way.

server/lib/src/idm/ldap.rs
```diff
@@ -525,6 +532,107 @@ impl LdapServer {
...
+                request_error!(err = ?e, ?rdn, "Error resolving rdn to target");
+                e
+            })?;
+
+            return Ok(LdapBindTarget::Account(uuid));
+        }
+
+        let basedn = &self.default_naming_context.basedn;
+        let app_binddnre = Regex::new(
+            format!("^(([^=,]+)=)(?P<usr>[^=,]+),(([^=,]+)=)(?P<app>[^=,]+)(,{basedn})$").as_str(),
```
Firstyear, line 584: Should this <app> capture be an optional match ?

server/lib/src/idm/ldap.rs
```diff
@@ -525,6 +532,107 @@ impl LdapServer {
...
+                return Err(OperationError::NoMatchingEntries);
+            }
+
+            let mut idms_prox_read = idms.proxy_read().await;
+            let app_entries = idms_prox_read.qs_read.internal_search(filter!(f_eq(
+                Attribute::Class,
+                EntryClass::Application.into()
+            )))?;
+
+            for app_entry in &app_entries {
```
Firstyear, line 607: We shouldn't need this loop, this is another place where we would use the ldap application cache so we can look up applications by their name from a hashmap.

server/lib/src/idm/ldap.rs
```diff
@@ -525,6 +532,107 @@ impl LdapServer {
...
+                .ok_or(OperationError::NoMatchingEntries)?;
+            let usr_rdn = caps
+                .name("usr")
+                .map(|v| v.as_str().to_string())
+                .ok_or(OperationError::NoMatchingEntries)?;
+            if app_rdn.is_empty() || usr_rdn.is_empty() {
+                return Err(OperationError::NoMatchingEntries);
+            }
+
+            let mut idms_prox_read = idms.proxy_read().await;
```
Firstyear, line 601: This introduces a race condition. You should only have a single read txn over the whole operation. By having a read here that is dropped when the fn ends, then another when you call idms.auth(), there is a possibility that the application you are attempting to bind with that existed in the first transaction *no longer exists* when you get to the second. 

I think here at point you shouldn't be doing any db look ups at all - you should be only extracting the application name if one exists in the bind dn and that's it. It should be up to the call to application_auth_ldap to validate that name exists at that stage and then to do the bind.

Another way to think of this is that this part of the code is "preparing" all your needed inputs to build the auth event - the auth event is what makes all the choices and validations about if the application exists (or doesn't) and then validates the password. The auth event must be a fully and self contained call, without a need to rely on another part of the code to pre-vet anything.

server/lib/src/server/migrations.rs
```diff
@@ -806,6 +806,7 @@ impl<'a> QueryServerWriteTransaction<'a> {
             SCHEMA_ATTR_DENIED_NAME.clone().into(),
             SCHEMA_ATTR_CREDENTIAL_TYPE_MINIMUM.clone().into(),
             SCHEMA_ATTR_WEBAUTHN_ATTESTATION_CA_LIST.clone().into(),
+            SCHEMA_ATTR_APPLICATION_PASSWORD.clone().into(),
```
Firstyear, line 809: This migration framework is no longer called on upgrades. We need to put this elsewhere, but that can wait for later into the PR since for now in "tests" and on new instances this will work.

tools/cli/src/opt/kanidm.rs
```diff
@@ -672,6 +698,34 @@ pub enum ServiceAccountOpt {
...
+    List(CommonOpt),
+    /// Create a new application
+    #[clap(name = "create")]
+    Create(Named),
+    /// Delete an application
+    #[clap(name = "delete")]
+    Delete(Named),
+    /// Add new members
+    #[clap(name = "add-members")]
+    AddMembers(GroupNamedMembers),
```
Firstyear, line 714: Given we plan to just have the application associate a group which has the members, I think this cli interface may be a bit out of date.

Firstyear (changes requested):
I think this PR is a solid start, but there are lots of little places to improve and some subtleties.

I think for you to get more confidence in this, first you should focus on the server-only and ignore the https api and the cli for a bit. 

Your next step should be to make tests (per my email) in https://github.com/kanidm/kanidm/blob/master/server/lib/src/idm/ldap.rs#L618 ) with cases such as:

* Create a new application entry/object
* Checking that the application acts as a service-account and can be bound to
* That ldap searches under the application DN show the expected user/group data
* Generating a new user application password (since it should be server side generated even if the user creates the label)
* User with wrong application password
* User with correct application password
* User with correct application password (A) but binding to wrong application (B)
* User with correct application password (A) to a deleted application entry
* User with correct application password who *isn't* a member of the approved access group
* User who's account has expired / not yet valid with correct password

You can then test these on their own with:

```
cd server/lib
cargo test
```

that way you can focus just on the server operating but without it having to build the https or client layers. Once you are confident in the server, then you can "expand" to include them in the scope. Where as if you just "change the world" it makes refactors of a feature harder as each change requires you to touch the api/cli and more. 

I think the tests will also give you more grounding on what your doing and if it will work in the way you intend. It should also let you exercise the ldap application cache too when you start to implement that. 

Hope that helps - I know it's a lot of feedback, but it's a big PR.

server/lib/src/idm/ldap.rs
Firstyear: As another point, I think kanidm varies to other systems that we prefer to create and hold open a long-running read rather than open and closing a lot of smaller reads. The reason is that the copy-on-write structures we use are actually more effecient when you hold them for a longer time, and we want to assert from the start of any operation that "everything is isolated/atomic". If you ever open/close/open a txn in one request this breaks our ACID guarantees, and it actually introduces performance issues too.

tools/cli/src/opt/kanidm.rs
scabrero: Yes. There will be a new one matching the design doc.

server/lib/src/idm/ldap.rs
scabrero: Ok, I have modified the code to delay application search. But doesn't `name_to_uuid()` call in `do_bind()` has the same problem?

tools/cli/src/opt/kanidm.rs
Firstyear: As mentioned, you should probably focus on writing use case oriented tests next

server/lib/src/idm/ldap.rs
scabrero: Yeah, I figured it out later. Thanks

Issue #2448, opened 2024-01-23
Suggestion: Small improvements to make user onboarding more clear to laymen


illode:
**Is your feature request related to a problem? Please describe.**

The Credential Update / Reset page is a bit unclear to people are familiar with Kanidm / authentication terms (TOTP, Passkeys, etc). As it stands, if I were to create accounts for my friends / family, I would need to preemptively explain a bunch of jargon, whereas ideally I would be able to simply create their account and send the credential setup link and have it be explained there.

**Describe the solution you'd like**
Some of these are totally nitpicky, but since I was already making this issue, I figured I may as well include them. The first three are the main ones.
* Since most people probably have no idea what the terms mean, it might be good to either embed (simplified) explanations, or include links to a site that explains the terms.
* `Password / TOTP` should probably be written `Password + Time-based One Time Password (TOTP) generator` or something along those lines, since people will have no idea what that means.
* It should probably list some examples of OTP apps, e.g. Google Authenticator, FreeOTP, and maybe whatever the Microsoft one is called. Microsoft Authenticator?
* ~~It says `Legacy password paired with other authentication factors.`, but there's only one "other" authentication factor (TOTP). It should probably say `Legacy password paired with TOTP`, or nothing at all since that's implied by the header. Unless you're planning to add more forms of MFA in the future.~~ I see https://github.com/kanidm/kanidm/issues/2417 now, which coincidentally is what I want more than passkeys anyways. Ignore this one.
* `Delete this Insecure Password` should probably just say `Delete insecure password` or `Delete password` , since you can only have one password, the `this` is redundant and implies that there can be multiple passwords somehow.
* At the top of the page, it has the display name + `<account name>@<kanidm domain name>`, e.g. `John Carmack` and `john@kanidm.example.com`. While that makes sense given the context of having multiple separate kanidm instances under different subdomains, it somewhat confusingly looks like an email. Moving the `<account name>` next to the display name might make more sense.
* It might be nice to let the user know how long the authentication for that page lasts. Not really sure about this one, it might encourage them to leave the page open if they how long it lasts.

After writing this, I realized it's kind of confusing in text form, so here is a very rough illustration in screenshot form:
![image](https://github.com/kanidm/kanidm/assets/114271894/3a074411-3922-437f-9828-87a5236f0776)

illode:
Also just noticed that the bad password errors don't have spaces:
[...]

PR #2475, opened 2024-02-03
Credential update tweaks

illode:
Addresses #2448.

I tried to simplify wording that I thought was confusing, unclear, or redundant. The credential update screen is a bit wordier now, but I think it's still an improvement.

I'm not sure if `CUCredState::PolicyDeny` and `CUCredState::AccessDeny` should include the text about passkeys, nor what the difference between them is, but I added it anyways.

Screenshots with the main changes:
image
image

Checklist

- [x] cargo fmt has been run

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -33,6 +33,7 @@
 - philipcristiano
 - Jianchen Zhao (bolu61)
 - Allan Zhang (allan2)
+- illode
 
 ## Acknowledgements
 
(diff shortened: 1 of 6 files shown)


Firstyear (approved):

Firstyear:
> I'm not sure if `CUCredState::PolicyDeny` and `CUCredState::AccessDeny` allow should include the text about passkeys, nor what the difference between them is, but I added it anyways.

The difference is one is policy deny says "You have the permission to change this value, but account policy won't let you". Access deny means you don't have permission to change the value at all, regardless of account policy. There is a bit of subtely around it because it's "rare"that you would hit the access deny case, since that would require the in-build access policy to be disabled and a custom one added that denyed self-write to the password for example. 

Otherwise this looks absolutely fantastic! Please also add yourself to contributors.md :)

Firstyear (approved):

PR #2481, opened 2024-02-06
Add tools for remigration and domain level raising

Firstyear:
@jinnatar - This adds the tools needed to re-run migrations, and force domain levels to be raised. 

Checklist

### DIFF
--- a/server/core/src/actors/v1_write.rs
+++ b/server/core/src/actors/v1_write.rs
@@ -1,5 +1,6 @@
 use std::{iter, sync::Arc};
 
+use kanidm_proto::internal::DomainInfo as ProtoDomainInfo;
 use kanidm_proto::internal::ImageValue;
 use kanidm_proto::internal::Oauth2ClaimMapJoin as ProtoOauth2ClaimMapJoin;
 use kanidm_proto::v1::{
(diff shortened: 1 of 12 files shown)


server/core/src/actors/v1_write.rs
```diff
@@ -1820,4 +1821,57 @@ impl QueryServerWriteV1 {
...
+        skip_all,
+        fields(uuid = ?eventid)
+    )]
+    pub(crate) async fn handle_domain_show(
+        &self,
+        eventid: Uuid,
+    ) -> Result<ProtoDomainInfo, OperationError> {
+        trace!("Begin domain show event");
+        let ct = duration_from_epoch_now();
+        let mut idms_prox_write = self.idms.proxy_write(ct).await;
```
yaleman, line 1836: Is this a write transaction because we just haven't made it a reader? Feels weird to be pulling info and having to start a wtxn

Firstyear: It's because in the admin command handler we only have the write handle :)

yaleman: That'll do it 😄

Issue #2487, opened 2024-02-08
Object graph view


ToxicMushroom:
**Is your feature request related to a problem? Please describe.**
Currently the only way to know what the inheritance is between groups and persons is to look at text output.

**Describe the solution you'd like**
By presenting the kanidm objects in a graph admins can easily see what is a member of what and spot things that seem out of place or verify by comparing the kanidm graph to a graph they're using as reference.

This could be implemented using mermaidjs as a tab in the webUI for admins to view.
Or this could be a cli option to print the graph source files.

**Additional context**
![image](https://github.com/kanidm/kanidm/assets/32853531/0276d288-6fb0-4597-80ea-897793c61c58)

I've had troubles with mermaid in the past where it'd generated broken graphs, and it often generates worse graphs than graphviz. Mermaidjs probably has easier browser integration though.

It is also probably a good idea to show different objects as different shapes/colors/or prefix with a symbol ?

**CLI Ideas**
Since graph can output multiple objects or should perhaps integrate many objects into a single graph, it would not make sense to put it under person or group, putting it toplevel also feels a little bit like bloating the interface but maybe it can be moved into a more fitting category in the future
`kanidm graph <mermaid|graphviz> [filters]`

**Web UI ideas**
Some options I'd think are useful:
 - A filter dropdown to choose which objects to show (groups, builtin_groups, accounts, builtin_accounts, persons, maybe more/less ?)
 - A button to copy or view the graph source
 - Maybe just have multiple graph renderers (mermaidjs and graphviz are the 2 I know) but this might be a little overkill.
 
 Feedback is welcome, sorry if this is too much of an information dump (I'm not good at this :p).

ToxicMushroom:
First testing output of the graph output with mermaid via cli: 
![image](https://github.com/kanidm/kanidm/assets/32853531/39ddb4ab-081a-4b46-8ff4-8ebd744f2817)

Firstyear:
Oh dang, that looks really nice! Great work!

PR #2518, opened 2024-02-11
Feature object graph

ToxicMushroom:
Draft for #2487  

Checklist

- [x] cargo fmt has been run
- [Not relevant] book chapter included (if relevant) 
- [Not relevant] design document included (if relevant)


yaleman:
The web UI bits really should be under the admin interface, since they're "management" kinda things.

ToxicMushroom:
> The web UI bits really should be under the admin interface, since they're "management" kinda things.

Oh thanks I didn't know the admin ui was a thing lol

ToxicMushroom:
@yaleman could you link me that mermaid js blob you found ?

yaleman:
I spent some time looking for it, and now I'm wondering how I found it - I guess they broke it up because ... it's chonky

ToxicMushroom:
@yaleman I've in the meantime been able to build mermaid with only the flowchart diagram renderer. This only reduced the size by one third.
I think we need a mechanism to load predefined js files somehow at runtime. 
Thus the client should know what <files, hashes> it can request.

I'm unsure what the best way to implement this is. Currently all js deps just get filled in at `ui_handler_generic` and thus loaded accross the webui where they aren't needed.

Mermaid when built by itself into 1 blob with all dependencies also still has inline-src which our csp doesn't allow, idk if we can or want to relax this for mermaid specifically ?

ToxicMushroom:
> Since it's an SPA we need to have them line up though, because the headers from the server influence the security the browser applies.

We should be able to keep all the headers and split it somehow into 2 
 - headers who'se js also gets injected in the html to load them on page load
 - headers who'se js's we somehow communicate to the user without them loading it instantly, allowing us to load it later.

> We can't relax it for one thing and doing it in general smells a bit... why do these libraries do this stuff 😡

Then I'll stick with graphviz only unless I can convince the mermaid devs to care

ToxicMushroom:
I still need to deduplicate some code I copied over and finish the cli.
Idk how to deal with duplicated logic/types between the cli and wasm so I hope that that's fine ?

server/web_ui/shared/static/shared.js
```diff
@@ -1,6 +1,52 @@
...
-    var elem = document.getElementById(m);
-    var modal = bootstrap.Modal.getInstance(elem);
+    const elem = document.getElementById(m);
+    const modal = bootstrap.Modal.getInstance(elem);
     modal.hide();
-};
\ No newline at end of file
+}
+
+function changeTag(node, newTag) {
```
yaleman, line 8: Please use rust function names/variables since that's what we're doing elsewhere for now 😄  (`change_tag` vs `changeTag`, `new_tag` vs `newTag`)... and now I think of it, that function's _replacing_ the tag so it would be better calling it `replace_tag`.

server/web_ui/shared/static/shared.js
```diff
@@ -1,6 +1,52 @@
...
+
+function start_graphviz(graph_src) {
+    Viz.instance().then(function(viz) {
+        const graphContainer = document.getElementById("graph-container");
+        if (graphContainer)
+            graphContainer.replaceChildren(viz.renderSVGElement(graph_src))
+    });
+}
+
+export function open_blank(content) {
```
yaleman, line 47: Can you please add a docstring saying it'll open a blank window and fill it with `content` - just so it's clear in an IDE without having to read the code 😄 
```suggestion
/** Opens a popup window `target=_blank` filled with `content`
* @param string content
*/
export function open_blank(content) {
```

yaleman (changes requested):
Looks good so far, just some minor nitpicks. 😄 

Where's the shared types? I didn't notice any in my read over the code, but I'm good like that.

server/web_ui/shared/static/shared.js
ToxicMushroom: This doesn't propagate through the `extern "C" {` bindings for me, maybe a rustrover issue ?
I can add this to the rust function as well if you want

ToxicMushroom:
> Looks good so far, just some minor nitpicks. 😄
> 
> Where's the shared types? I didn't notice any in my read over the code, but I'm good like that.

ObjectType is an enum which is duped

yaleman:
> > Looks good so far, just some minor nitpicks. 😄
> > Where's the shared types? I didn't notice any in my read over the code, but I'm good like that.
> 
> ObjectType is an enum which is duped

That can go into `kanidm_proto::v1`

server/web_ui/shared/static/shared.js
yaleman: Oh, it won't propagate, but documenting what functions do is sensible everywhere 😄

ToxicMushroom:
> > > Looks good so far, just some minor nitpicks. 😄
> > > Where's the shared types? I didn't notice any in my read over the code, but I'm good like that.
> > 
> > 
> > ObjectType is an enum which is duped
> 
> That can go into `kanidm_proto::v1`

mhm I ran into the issue that the derived traits on both are a bit different.
One is a [ValueEnum](https://docs.rs/clap/latest/clap/trait.ValueEnum.html) the other is a [Sequence](https://docs.rs/enum-iterator/latest/enum_iterator/trait.Sequence.html).

Putting them in proto v1 means proto needs 2 extra dependencies, idk if that's okay ? 

There's also [40 lines](https://github.com/ToxicMushroom/kanidm/blob/39cf53380594bc0f1921a8ed3371af117c60d8fe/tools/cli/src/cli/graph.rs#L45-L86) of duped logic which could go into proto v1 (although it isn't protocol related imo :/ )

yaleman:
> > That can go into `kanidm_proto::v1`
> 
> mhm I ran into the issue that the derived traits on both are a bit different. One is a [ValueEnum](https://docs.rs/clap/latest/clap/trait.ValueEnum.html) the other is a [Sequence](https://docs.rs/enum-iterator/latest/enum_iterator/trait.Sequence.html).
> 
> Putting them in proto v1 means proto needs 2 extra dependencies, idk if that's okay ?

`enum_iterator` would be fine, I wouldn't want add `clap` to `kanidm_proto` TBH.

> There's also [40 lines](https://github.com/ToxicMushroom/kanidm/blob/39cf53380594bc0f1921a8ed3371af117c60d8fe/tools/cli/src/cli/graph.rs#L45-L86) of duped logic which could go into proto v1 (although it isn't protocol related imo :/ )

I'd leave it as-is for now

ToxicMushroom:
> > > That can go into `kanidm_proto::v1`
> > 
> > 
> > mhm I ran into the issue that the derived traits on both are a bit different. One is a [ValueEnum](https://docs.rs/clap/latest/clap/trait.ValueEnum.html) the other is a [Sequence](https://docs.rs/enum-iterator/latest/enum_iterator/trait.Sequence.html).
> > Putting them in proto v1 means proto needs 2 extra dependencies, idk if that's okay ?
> 
> `enum_iterator` would be fine, I wouldn't want add `clap` to `kanidm_proto` TBH.
> 
> > There's also [40 lines](https://github.com/ToxicMushroom/kanidm/blob/39cf53380594bc0f1921a8ed3371af117c60d8fe/tools/cli/src/cli/graph.rs#L45-L86) of duped logic which could go into proto v1 (although it isn't protocol related imo :/ )
> 
> I'd leave it as-is for now

alright, the compiler didn't want me using the objecttype from proto anyway :')

server/web_ui/admin/src/components/admin_objectgraph.rs
```diff
@@ -0,0 +1,294 @@
...
+            if entries.is_empty() {
+                <div>
+                  <h5>{ "No graph objects for the applied filters." }</h5>
+                </div>
+            } else {
+                <div class="column">
+                    <div class="hstack gap-3">
+                    {
+                        all::<ObjectType>().map(|ot| {
+                            let str = format!("{:?}", ot);
```
yaleman, line 236:
```suggestion
                            let str = format!("{:?}", ot);
```

Can we please do an `impl Display` for `ObjectType` and have:

- Group -> "Group" 
- BuiltinGroup -> "Built In Group" 
- etc

(it feels like "Group" should be "Local Group" or something to distinguish from Builtin Group, but I'll ponder that and it's fine for now)

server/web_ui/admin/src/components/admin_objectgraph.rs
```diff
@@ -0,0 +1,294 @@
...
+        sb.push_str("}");
+        init_graphviz(&sb.as_str());
+
+        let on_checkbox_click = {
+            let scope = ctx.link().clone();
+            move |_: Event| {
+                let coll = document().get_elements_by_class_name("obj-graph-filter-cb");
+                let mut filters = vec![];
+
+                for i in 0..*&coll.length() {
```
yaleman, line 200: This feels weird, especially with the unchecked_into later... is there a reason it's not just `for option in coll.iter()` or similar?

yaleman (changes requested):
This is looking really good so far, I really like the UI and being able to click through to objects.

I'd recommend running `make webui` from the project root and finding all the unused imports etc, because that's blocking build at the moment - and `viz.js` (and its brotli-pre-compressed friend) isn't in the `server/web_ui/pkg/external` dir in git.

server/core/src/https/javascript.rs
```diff
@@ -34,24 +34,33 @@ pub fn generate_integrity_hash(filename: String) -> Result<String, String> {
...
             Some(val) => {
                 format!(" type=\"{}\"", val.as_str())
             }
             _ => String::from(""),
         };
-        format!(
-            r#"<script async src="/pkg/{}" integrity="sha384-{}"{}></script>"#,
-            self.filepath, &self.hash, &filetype,
-        )
+        if self.dynamic {
```
Firstyear, line 54: I'm a web-boomer, what's the different here with these two types of load?

yaleman: This doesn't actually load the script, the meta tag is used as a placeholder so it can be replaced later with the script tag using `init_graphviz` from `server/web_ui/shared/static/shared.js`. It's a bit of a hack but it works.

server/web_ui/admin/src/components/admin_objectgraph.rs
ToxicMushroom: You can't iterate on HtmlCollection afaik

ToxicMushroom: I went with a different approach now

Firstyear:
> Idk what to do with the conflicting files lol

I think those are the prebuilt items, so a rebase + wasm build then git add <conflicts>; git rebase --continue will do it. You may want to squash your pr to one patch first to make it easier though.

server/web_ui/admin/src/components/admin_objectgraph.rs
```diff
@@ -0,0 +1,339 @@
...
+    ) -> Html {
+        let typed_entries = entries
+            .iter()
+            .filter_map(|entry| {
+                let classes = entry.attrs.get("class")?;
+                let uuid = entry.attrs.get("uuid")?.first()?;
+
+                // Logic to decide the type of each entry
+                let obj_type = if classes.contains(&"group".to_string()) {
+                    if uuid.starts_with("00000000-0000-0000-0000-") {
```
Firstyear, line 154: you could just do "if uuid <= UUID_ANONYMOUS" since that's where the system allocated UUID range ends.

Firstyear: I guess it's not the native uuid type though 🤔

yaleman:
It's pretty weird/annoying that seemingly tiny changes which don't add content have added ~700KB to the user UI wasm O_o

yaleman (changes requested):
`server/core/src/https/tests.rs` has failed due to missing "dynamic" field in the JavaScriptFile initialisation (look for `test_javscriptfile`)

server/web_ui/admin/src/components/admin_objectgraph.rs
ToxicMushroom: Pretty sure it comes out of the json obj as a string yes. 
Would it be preferred to convert it to a UUID ?

ToxicMushroom:
> It's pretty weird/annoying that seemingly tiny changes which don't add content have added ~700KB to the user UI wasm O_o

When I run `make webui` I get a 52kB difference in size on `kanidmd_web_ui_admin_bg.wasm` between:
before my changes: `6fb4fac5d0e27131a06e6ae0451064f0ec1aa1bb`
current latest change: `0deef472dc874ce157fa6df75a049324a86493db`

Tested by commenting out code:
-  enum-iterator adds 16kb
-  main `view_ready` function with most logic ect adds 18kb

`make webui` thus produced different/bigger files on the commit before my changes on my pc at least. Am I building it wrong ? Or was it not ran/committed for a while

yaleman:
I was going by the file sizes reported in the PR:

Screenshot 2024-02-23 at 08 41 08

A lot of it's likely just "WASM is known to be not great at identifying what to compile out" weirdnesses.

server/web_ui/admin/src/components/admin_objectgraph.rs
Firstyear: I think I'm cool to go either way now I think more about it.

server/web_ui/admin/src/components/admin_objectgraph.rs
```diff
@@ -0,0 +1,339 @@
...
+        entries: &[Entry],
+    ) -> Html {
+        let typed_entries = entries
+            .iter()
+            .filter_map(|entry| {
+                let classes = entry.attrs.get("class")?;
+                let uuid = entry.attrs.get("uuid")?.first()?;
+
+                // Logic to decide the type of each entry
+                let obj_type = if classes.contains(&"group".to_string()) {
```
Firstyear, line 153: @yaleman I guess we could also just add `class: Builtin` automatically to anything that is builtin? Makes it a bit easier to distinguish.

server/web_ui/admin/src/components/admin_objectgraph.rs
```diff
@@ -0,0 +1,339 @@
...
+                    if uuid.starts_with("00000000-0000-0000-0000-") {
+                        ObjectType::BuiltinGroup
+                    } else {
+                        ObjectType::Group
+                    }
+                } else if classes.contains(&"account".to_string()) {
+                    if classes.contains(&"person".to_string()) {
+                        ObjectType::Person
+                    } else {
+                        ObjectType::ServiceAccount
```
Firstyear, line 163: explicitily check for service-account (oauth2 clients are now accounts too :P )

server/web_ui/admin/src/components/admin_objectgraph.rs
```diff
@@ -0,0 +1,339 @@
...
+            </>
+        }
+    }
+
+    fn view_error(&self, _ctx: &Context<Self>, msg: &str, kopid: Option<&str>) -> Html {
+        error_page(msg, kopid)
+    }
+
+    async fn fetch_objects() -> Result<Msg, FetchError> {
+        let urls = vec!["/v1/group", "/v1/service_account", "/v1/person"];
```
Firstyear, line 306: We could do this in oneshot to /v1/raw with a search request to get all of these at once. I think that also can request the entries are limited to certain attributes, which will reduce transfer size.

ToxicMushroom: Sounds good

Firstyear: You should be able to use kanidm_proto SearchRequest + Filter to make something like:

```
Or(vec![
    Eq("class", "person"),
    Eq("class", "service_account"),
    Eq("class", "group")
])
```

Then post that to `/v1/raw/search`

ToxicMushroom: Limiting the response to certain attributes looks supported internally but not via search request,
see:
 - [handle_search](https://github.com/ToxicMushroom/kanidm/blob/b1e7cb13a54c8865154123bf3c7f90d51bf03cec/server/core/src/actors/v1_read.rs#L96)
 - [SearchEvent::from_message](https://github.com/ToxicMushroom/kanidm/blob/d5ed335b529d6eaff52c90a995342a32a7cb3dde/server/lib/src/event.rs#L100)

ToxicMushroom: But the oneshot request already works locally and simplified the request code a bit :)

ToxicMushroom:
My latest build ballooned the pkg build again, idk why or how :')
I'm just running `make webui`

yaleman:
> My latest build ballooned the pkg build again, idk why or how :') I'm just running `make webui`

I'd bet it's this line:

```
use serde::Serialize;
```

Try looking for the uses of `serde_wasm_bindgen::Serializer::json_compatible` and replicating that, `serde` is a huge package.

ToxicMushroom:
> > My latest build ballooned the pkg build again, idk why or how :') I'm just running `make webui`
> 
> I'd bet it's this line:
> 
> ```
> use serde::Serialize;
> ```
> 
> Try looking for the uses of `serde_wasm_bindgen::Serializer::json_compatible` and replicating that, `serde` is a huge package.

image

Changing this to use serde_wasm_bindgen only has no size impact

ToxicMushroom:
I rebased it, idk if this is done proper so I saved the old branch if I need to redo it @Firstyear

server/web_ui/admin/src/components/admin_objectgraph.rs
yaleman: Yeah, that would make sense

Firstyear: @ToxicMushroom Can you open an issue that we need to add class: Builtin as a marker?

Firstyear (approved):
I am so impressed by this. You've done a great job to create this, well done! Ensure you are in contributors.md as well please :)

yaleman (approved):

Firstyear (approved):

yaleman (approved):

PR #2523, opened 2024-02-12
Himmelblau requires the machine key for unix_user_get

dmulder:
I need access to the machine key here in order to
send a new request to Azure for user details.

Fixes #

Checklist

- [ ] cargo fmt has been run


Firstyear:
Yeah I figured this change would be coming. I think I had a branch with the same somewhere, but just never did the PR.

dmulder:
> Yeah I figured this change would be coming. I think I had a branch with the same somewhere, but just never did the PR.

I thought about adding it for unix_group_get also, but I don't actually have a use case for that (I can't auth a group in Azure).

Firstyear:
If we need it in the future, we can add it then. :)

PR #2546, opened 2024-02-16
When the HTTPS server fails, handle that gracefully, and check client_ca on startup 😄 

yaleman:
Previously if the client_ca path was set, but it didn't work the HTTPS server would stop silently. This handles it before startup, and also handles internal failures at runtime of the HTTPS server by sending a shutdown signal.

Checklist

- [x] cargo fmt has been run

### DIFF
--- a/server/core/src/lib.rs
+++ b/server/core/src/lib.rs
@@ -748,7 +748,7 @@ impl Display for TaskName {
 
 pub struct CoreHandle {
     clean_shutdown: bool,
-    tx: broadcast::Sender<CoreAction>,
+    pub tx: broadcast::Sender<CoreAction>,
     /// This stores a name for the handle, and the handle itself so we can tell which failed/succeeded at the end.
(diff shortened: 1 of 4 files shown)


server/core/src/lib.rs
```diff
@@ -748,7 +748,7 @@ impl Display for TaskName {
 
 pub struct CoreHandle {
     clean_shutdown: bool,
-    tx: broadcast::Sender<CoreAction>,
+    pub tx: broadcast::Sender<CoreAction>,
```
Firstyear, line 751: Why does this need to be pub?

yaleman: so we can access it in the tokio::select loop

Firstyear: pub(crate) then maybe?

yaleman: Nope, won't work - CoreHandle is over in server/core, while it's being used in server/daemon - different crates 😱

Firstyear (approved):

PR #2558, opened 2024-02-20
prctl compile-time fixes, also chasing lints

yaleman:
Fixes:

- the issues with not handling the prctl `Result`
- my innate need to sate clippy
- should help fix issues with missing wasm-related tools on dev machines

Checklist

- [x] cargo fmt has been run


server/lib/src/be/mod.rs
```diff
@@ -1923,8 +1923,8 @@ impl<'a> BackendWriteTransaction<'a> {
         })?;
         Ok(nsid)
     }
-
     pub fn get_db_s_uuid(&mut self) -> Result<Uuid, OperationError> {
+        #[allow(clippy::blocks_in_conditions)]
```
Firstyear, line 1927: We should make the value / block assign to a value instead and then match.

server/lib/src/entry.rs
```diff
@@ -358,6 +358,8 @@ impl Entry<EntryInit, EntryNew> {
     pub(crate) fn unsafe_from_entry_str(es: &str) -> Self {
         // Just use log directly here, it's testing
         // str -> proto entry
+
```
Firstyear, line 361: I swear this function is one of my deepest regrets.

Firstyear (changes requested):
You may be able to set:
https://doc.rust-lang.org/clippy/lint_configuration.html#allow-unwrap-in-tests 
in lib.rs to make this easier.

server/lib/src/entry.rs
yaleman: we'll excise it one day!

unix_integration/src/unix_passwd.rs
```diff
@@ -99,14 +99,18 @@ wheel:x:481:admin,testuser
 
     #[test]
     fn test_parse_passwd() {
-        for record in parse_etc_passwd(EXAMPLE_PASSWD.as_bytes()).unwrap() {
+        #[allow(clippy::expect_used)]
```
Firstyear, line 102: Shouldn't this use the allow-expect-in-tests thing?

Firstyear (approved):

clippy.toml
```diff
@@ -13,3 +13,7 @@ too-many-arguments-threshold = 8
 type-complexity-threshold = 300
 
 disallowed-types = ["std::collections::HashMap", "std::collections::HashSet"]
+
+allow-expect-in-tests = true
+allow-unwrap-in-tests = true
+allow-dbg-in-tests = true
```
Firstyear, line 19: I have to remember this macro exists. It's good.

Firstyear (approved):

PR #2559, opened 2024-02-20
20240216 308 resource limits

### LINKED ISSUE #308: Add resource limit types to account
Following #307 we need a way to configure resource limits on accounts. These should be a property of the account, and if we want to allow groups to provide limits, we should consider templating or similar. It probably needs some thought how this should look.

Firstyear:
Fixes #308 - add resource limits to account policy

Checklist


tools/cli/src/opt/kanidm.rs
```diff
@@ -170,25 +170,25 @@ pub enum GroupAccountPolicyOpt {
...
+    /// Set the maximum time for privilege session expiry in seconds.
     #[clap(name = "privilege-expiry")]
     PrivilegedSessionExpiry {
         name: String,
         expiry: u32,
         #[clap(flatten)]
         copt: CommonOpt,
     },
-    /// The the webauthn attestation ca list that should be enforced
+    /// The webauthn attestation ca list that should be enforced
```
yaleman, line 189:
```suggestion
    /// The WebAuthn attestation CA list that should be enforced
```

tools/cli/src/opt/kanidm.rs
```diff
@@ -197,6 +197,25 @@ pub enum GroupAccountPolicyOpt {
...
+    /// search operation.
+    #[clap(name = "limit-search-max-results")]
+    LimitSearchMaxResults {
+        name: String,
+        maximum: u32,
+        #[clap(flatten)]
+        copt: CommonOpt,
+    },
+    /// Sets the maximum number of entries that are examined during
+    /// a partially indexed search. This does not affect fully
```
yaleman, line 210: We need to document the difference between "fully" and "partially" indexed somewhere, but that can get done in another task.

yaleman (dismissed):

tools/cli/src/opt/kanidm.rs
Firstyear: I actually hesitated about adding this one, but it's needed in conjunction with search max results. I also thought about having it automatically set to a value, or hiding it, but it all gets weird. 

Generally we've hidden a lot about the way the search internals work, because realistically, the defaults have been "so good" no one has thought to ask about them! 

Anyway, fully and partial indexed is highlighted as an error in the logs if it occurs and you hit the filter test max limit.

yaleman (approved):

PR #2573, opened 2024-02-22
20240221 2489 cleanup api v1

### LINKED ISSUE #2489: Clean up v1 API
Some API elements in the v1 interface are internals that may change version to version. These should be moved to the internal.rs file, since this crate represents what we do/don't change between versions.

Firstyear:
Fixes #2489 

This cleans up most of the api between internal an v1. I was over-cautious in moving things to internal, so that we have the choice to move things back as required. I'm already seeing that we may need to move back UserAuthToken and ApiToken as we need to keep these stable over releases (and likely they need some way to handle versions .... ) 

Checklist


yaleman (approved):
Such chonk, many moves. Looks good for now 😄

PR #2591, opened 2024-02-27
20230224 2437 orca remodel

### LINKED ISSUE #1131: Develop a method for setting up a dev/test environment full of expected config

It'd be handy to have a way of yeeting in a load of configuration elements to have a test/dev environment full of things.

To be clear, this is for *development* purposes, not for demo/real usage.

- [x] groups
- [x] persons
 - [ ] with SSH keys
 - [ ] various auth methods
- [ ] service accounts
- [x] oauth2 rp's
- [ ] RADIUS config/users
- [ ] 🦀🔪 

An automated/documented process for creating them so we have a way to have a ready-to-go environment.

Donating my old socks to fast food megacorps to help supply their feedstock.

**Additional context**

### LINKED ISSUE #2437: Orca setup script is broken
### I expected the following: 
Orca to setup the dev env and run as usual. 

### Kanidm version details
kanidm 1.1.0-rc.15-dev cf87993a1c
2024-01-18T19:30:50.278831Z DEBUG kanidm: Using 20 worker threads
Linux sebastiano-XPS-15-9520 6.2.0-39-generic #40-Ubuntu SMP PREEMPT_DYNAMIC Tue Nov 14 14:18:00 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux

### Any other comments
Basically it always fails the authentication bit as the admin password variable is always blank since the command supposed to set it: `ADMIN_PW=$(grep -E "^admin password" "${KANI_TEMP}/kanifile" | awk '{print $NF}')` doesn't behave as it used to do. Basically in 0e504517159ab0d6eb325ce38e0cc2320b7394c7 this happened in setup_dev_environment.sh: 
```diff
- echo "admin password:     ${ADMIN_PASS}"
```
and since `kanifile` comes from: `../../scripts/setup_dev_environment.sh | tee "${KANI_TEMP}/kanifile"`) that's what broke it. My first instinct was adding the admin password thing back but gate it behind a flag in to make it run only when required by orca. On the other hand I don't really know why it was removed in the first place, so I figured @yaleman has a better understanding on what is going on there and I might ask for some feedback before doing anything :)

Firstyear:
Fixes #2437 Fixes #1131 - This is a remodel of orca, which jettisons the previous behaviour in favour of an actor based framework. We can then use different "models" to create tests, and the preflight mode allows us to generate databases full of content.

Checklist

- [ ] cargo fmt has been run


tools/orca/src/populate.rs
```diff
@@ -0,0 +1,134 @@
...
+
+    preflight_flags.push(Flag::DisableAllPersonsMFAPolicy);
+
+    // PHASE 1 - generate a pool of persons that are not-yet created for future import.
+    // todo! may need a random username vec for later stuff
+
+    // PHASE 2 - generate persons
+    //         - assign them credentials of various types.
+    let mut persons = Vec::with_capacity(profile.person_count() as usize);
+    let mut person_names = BTreeSet::new();
```
Seba-T, line 63: Just some minor nitpicking as it took me some seconds to figure out what was going on :p 
```suggestion
    let mut person_usernames = BTreeSet::new();
```

Seba-T:
Besides that it looks good to me. Also I was a bit confused by the naming of `PopulateData` and `Preflight`; the first one because I thought it would actually populate the db, but that happens only inside preflight. The second one feels like its only purpose is to 'preflight' orca,  but if I understood correctly the idea was to create a general tool to fill the db with expected content, maybe a self-referencing name could help, i.e. "populate-db". It would also be closer to the wording used in the CLI help page 
`populate      Create a new state file that is populated with a complete dataset, ready to be loaded into a kanidm instance`
 `preflight     Populate the data for the test into the Kanidm instance`

tools/orca/src/run.rs
```diff
@@ -0,0 +1,162 @@
...
+    Stop,
+}
+
+async fn execute_inner(
+    warmup: Duration,
+    test_time: Option<Duration>,
+    mut control_rx: broadcast::Receiver<Signal>,
+) -> Result<(Instant, Instant), Error> {
+    // Delay for warmup time.
+    // TODO: Read warmup time from profile.
```
Seba-T, line 56: Can't we pass it as a parameter and read it inside the calling function `execute` since we already have it there?

Firstyear: Well yeah, I just hadn't added it yet.

Firstyear:
I certainly don't think this is perfect by any means I but I think it captures what I wanted in a rewrite.

We have the ability to swap out data collectors to gather stats in different ways.

We can run tests "forever" until a user sends a ctrl-c, and the stats will still work.

We can have different "actors" that have swappable models of behaviours they should exhibit.

We have real, multicultural name data for our users now! 

And there are deterministic seed generation features so that we can rebuild state files. 

I'm sure there is so much that can be improved. I've been really unsure about which way to go with certain parts in this process. I really don't think it's perfect by any means.

So I'd love to hear from both @Seba-T and @yaleman about what you think here.

Firstyear:
@ToxicMushroom If you would like some practice at code review, I'd love to hear your thoughts too.

ToxicMushroom:
> @ToxicMushroom If you would like some practice at code review, I'd love to hear your thoughts too.

I'll see this weekend

yaleman:
Seems sensible so far, will need to play with it a bit down the line 😄

Seba-T:
Hold on did "profile.json" disappear by any chance? Or am I tripping?

yaleman:
Screenshot 2024-03-05 at 10 29 59

it's still listed in the docs and ... hasn't been deleted, plz share mushrooms (not @ToxicMushroom)

tools/orca/src/stats.rs
```diff
@@ -0,0 +1,107 @@
...
+    Start(Instant),
+    End(Instant),
+    StopNow,
+}
+
+pub trait DataCollector {
+    fn run(
+        &mut self,
+        stats_queue: Arc<SegQueue<EventRecord>>,
+        ctrl: Arc<ArrayQueue<TestPhase>>,
```
Seba-T, line 21: What would the downsides be if we used a mpsc channel and just awaited on new messages instead of polling? I think this way there would be no need to do a `spawn_blocking` when calling the run function

Firstyear: The tokio mpsc is quite a bit slower because it uses a semaphore to lock and submit to the queue rather than lock free submission.

Seba-T:
> plz share mushrooms (not @ToxicMushroom)

Ofc, beware that they hit hard! :D 
Now that I think about it I suspect William had included the file when he opened the draft pr but he then removed it when he rebased

tools/orca/src/stats.rs
Seba-T: Oh cool I didn't know about that. I always thought that if you could make it asynchronous then you probably should. Like the fancier the better :p

Seba-T:
And also I don't fully understand how you can set the transition delay 🤔

Firstyear:
> And also I don't fully understand how you can set the transition delay 🤔

model_basic.rs line 30.

The transition delay is just something internal to any model, it's up to the model to call it and perform it's own delay internally during a transition.

Seba-T:
> > And also I don't fully understand how you can set the transition delay 🤔
> 
> model_basic.rs line 30.
> 
> The transition delay is just something internal to any model, it's up to the model to call it and perform it's own delay internally during a transition.

Ohh I see now. Would it make sense to include `next_transition` inside the `ActorModel` trait? Or is the `Transition` struct just something that is made available to the models if they want to use it but they can also freely implement custom transition logic that doesn't involve it?

tools/orca/README.md
```diff
@@ -0,0 +1,58 @@
+# Orca - A Kanidm Load Testing Tool
+
+Make a profile
+
+```
+orca setup-wizard --idm-admin-password ... \
+  --admin-password ... \
+  --control-uri 'https://localhost:8443' \
+  --profile ./profile.json
```
ToxicMushroom, line 9: .json example but it writes toml

tools/orca/README.md
```diff
@@ -0,0 +1,58 @@
...
+Test the connection
+
+```
+orca conntest --profile ./profile.json
+```
+
+Generate a State File
+
+```
+orca generate --profile ./profile.json --state ./state.json
```
ToxicMushroom, line 21: state does use json, maybe keep it consistent ?

Firstyear:
Initially I did that, but it got pretty messy because then you have the caller calling next_transition but also needs to give the result from next_transition back to transition so it can proceed, else the actor doesn't know what it decided to do when next_ was called. So I opted for this as a simpler model.

The real only risk here is actors with long delays will prevent/delay stop of the tool, so we likely need actors that do long delays have a small sleep/yield cycle. But that's a future thing.

ToxicMushroom:
I'm also curious why the loadtester doesn't output anything ?
e.g. time, throughput or is that not what it's for

tools/orca/src/generate.rs
```diff
@@ -1,164 +1,134 @@
...
-const N_GROUPS: usize = 1500;
-const N_MEMBERSHIPS: usize = 10;
-const N_NEST: usize = 4;
+    let surnames = std::include_str!("../names-dataset/dataset/Surnames.txt");
 
-pub(crate) fn doit(output: &Path) {
-    info!(
-        "Performing data generation into {}",
-        output.to_str().unwrap(),
+    let surnames = surnames.split('\n').collect::<Vec<_>>();
```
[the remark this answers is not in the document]
yaleman: Lines is good for when the dataset's not constrained, but takes more effort to turn into a `Vec`

tools/orca/README.md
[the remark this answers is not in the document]
yaleman: I'd expect our to be a test/dev thing that folks shouldn't really be `cargo install`ing, TBH

Seba-T:
> Initially I did that, but it got pretty messy because then you have the caller calling next_transition but also needs to give the result from next_transition back to transition so it can proceed, else the actor doesn't know what it decided to do when next_ was called. So I opted for this as a simpler model.
> 

Mmh makes sense. And what about keeping the call to `self.next_transition()` inside the `transition` function while also adding `next_transition` to the trait? Or do you envision models that could actually never call `next_transition` so no need to put it in the trait? 
> The real only risk here is actors with long delays will prevent/delay stop of the tool, so we likely need actors that do long delays have a small sleep/yield cycle. But that's a future thing.

Something like the loop inside `DataColletor` that checks for messages every 100ms?

tools/orca/src/generate.rs
```diff
@@ -1,164 +1,134 @@
...
-        accounts,
-        precreate,
-        connections,
-    };
+    // PHASE 3 - generate groups for integration access, assign persons.
+
+    // PHASE 4 - generate groups for user modification rights
+
+    // PHASE 5 - generate excess groups with nesting. Randomly assign persons.
+
```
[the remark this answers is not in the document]
Firstyear: Probably, but I think they'll naturally fill in once @Seba-T starts to work on them.

tools/orca/Cargo.toml
```diff
@@ -18,23 +18,21 @@ test = true
 doctest = false
 
 [dependencies]
+async-trait = { workspace = true }
 clap = { workspace = true }
 crossbeam = { workspace = true }
 csv = { workspace = true }
-dialoguer = { workspace = true }
+# dialoguer = { workspace = true }
```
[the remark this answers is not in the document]
Firstyear: Because I'm stupid :) 

I'll remove it.

tools/orca/README.md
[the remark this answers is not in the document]
Firstyear: Yeah, this will not be cargo installed, more likely we'll just run it from dev.

tools/orca/README.md
Firstyear: Opps.

tools/orca/README.md
Firstyear: profile should be human editable and readable, state shouldn't.

Firstyear:
Okay, anything else? It's not intended to be perfect, but a skeleton that myself and @Seba-T can keep building on, so I'd like to merge and keep going :)

yaleman (changes requested):

yaleman:
- fixed an issue with TOML handling u64 numbers (it can't, because it thinks they're i64's)
- clippy-pats.

yaleman (approved):

Seba-T:
> Okay, anything else? It's not intended to be perfect, but a skeleton that myself and @Seba-T can keep building on, so I'd like to merge and keep going :) 

Yeah 100% agreed. Perhaps there were also some spelling stuff here and there but I figured we were better off if I fixed them myself once I start working on it rather than endlessly going back and forth.

PR #2602, opened 2024-03-01
20240301 systemd uid

### LINKED ISSUE #393: subid ranges sourced from the network store
I tried to run an container as a user with podman.
I get this error.

```
ERRO[0000] cannot find UID/GID for user rene: No subuid ranges found for user "rene" in /etc/subuid - check rootless mode in man pages.
WARN[0000] using rootless single mapping into the namespace. This might break some images. Check /etc/subuid and /etc/subgid for adding sub*ids
Resolved "alpine" as an alias (/etc/containers/registries.conf.d/000-shortnames.conf)
Trying to pull docker.io/library/alpine:latest...
Getting image source signatures
Copying blob ba3557a56b15 done
Copying config 28f6e27057 done
Writing manifest to image destination
Storing signatures
  Error processing tar file(exit status 1): potentially insufficient UIDs or GIDs available in user namespace (requested 0:42 for /etc/shadow): Check /etc/subuid and /etc/subgid: lchown /etc/shadow: invalid argument
Error: Error committing the finished image: error adding layer with blob "sha256:ba3557a56b150f9b813f9d02274d62914fd8fce120dd374d9ee17b87cf1d277d": Error processing tar file(exit status 1): potentially insufficient UIDs or GIDs available in user namespace (requested 0:42 for /etc/shadow): Check /etc/subuid and /etc/subgid: lchown /etc/shadow: invalid argument
```

Looking on the internet I found that this is a known issue with non local users.

See [Active issue about this problem, subid ranges sourced from the network store](https://github.com/shadow-maint/shadow/issues/154)

### LINKED ISSUE #2601: gidnumber can fall into systemd container ranges
When creating a user in Kanidm and enabling a posix account for them, the associated user's GID number can fall into the systemd container UID range 524288-1879048191 (https://systemd.io/UIDS-GIDS/).

The immediately noticeable effect of this is if the user logs on through `pam_kanidm`, they cannot view their own user journal because systemd considers their UID to be a systemd container user and will not create dedicated journal files for them (https://github.com/systemd/systemd/blob/main/src/journal/journald-server.c#L447 https://github.com/systemd/systemd/blob/main/src/basic/uid-classification.h#L18-L20).

```
$ id
uid=1688334410(concord) gid=1688334410(concord) groups=1688334410(concord),2029292874(posix_users),3982434320(sysadmins) context=unconfined_u:unconfined_r:unconfined_t:s0

$ journalctl --user
Hint: You are currently not seeing messages from the system.
      Users in groups 'adm', 'systemd-journal' can see all messages.
      Pass -q to turn off this notice.
No journal files were opened due to insufficient permissions.
```

Additionally, the gidnumber can fall into the range 2147483648-4294967294 which systemd considers to be dangerous to use:
>Note that the range 2147483648…4294967294 (i.e. 2^31…2^32-2) should be handled with care. Various programs (including kernel file systems — see devpts — or even kernel syscalls – see setfsuid()) have trouble with UIDs outside of the signed 32-bit range, i.e any UIDs equal to or above 2147483648. It is thus strongly recommended to stay away from this range in order to avoid complications. This range should be considered reserved for future, special purposes.

Finally, it looks like it may be technically possible for the gidnumber to land on 2^32-1, which appears to be reserved:
>4294967295, aka “32-bit (uid_t) -1” → This UID is not a valid user ID, as setresuid(), chown() and friends treat -1 as a special request to not change the UID of the process/file. This UID is hence not available for assignment to users in the user database.

Firstyear:
Fixes #2601 Fixes #393 - gid numbers can be part of the systemd nspawn range.

Previously we allocated gid numbers based on the fact that uid_t is a u32, so we allowed 65536 through u32::max. However, there are two major issues with this that I didn't realise. The first is that anything greater than i32::max (2147483648) can confuse the linux kernel. 

The second is that systemd allocates 524288 through 1879048191 to itself for nspawn.

This leaves with with only a few usable ranges.

1000 through 60000
60578 through 61183
65520 through 65533
65536 through 524287
1879048192 through 2147483647

The last range being the largest is the natural and obvious area we should allocate from. This happens to nicely fall in the pattern of 0x7000_0000 through 0x7fff_ffff which allows us to take the last 24 bits of the uuid then applying a bit mask we can ensure that we end up in this range. 

There are now two major issues.

We have now changed our validation code to enforce a tighter range, but we may have already allocated users into these ranges. 

External systems like FreeIPA allocated uid/gid numbers with reckless abandon directly into these ranges. 

As a result we need to make two concessions.

We *secretly* still allow manual allocation of id's from 65536 through to 1879048191 which is the nspawn container range. This happens to be the range that freeipa allocates into. We will never generate an ID in this range, but we will allow it to ease imports since the users of these ranges already have shown they 'don't care' about that range. This also affects SCIM imports for longer term migrations. 

Second is id's that fall outside the valid ranges. In the extremely unlikely event this has occurred, a startup migration has been added to regenerate these id values for affected entries to prevent upgrade issues. 

An accidental effect of this is freeing up the range 524288 to 1879048191 for other subuid uses.

Checklist


Firstyear:
I want to also note that I have tested this with a backup from my database where I have users in some of these ranges. Migrations occured correctly and values in the valid ranges were untouched.

yaleman (changes requested):
We should really make this clear one and show folks how to check, 'cause a change in UID/GID can blow up file permissions if they're unlucky. If the patch below gets applied they can search using:

```shell
cargo run --bin kanidm -- raw search '{"and":[{"pres":"gidnumber"},{"eq":["class","object"]}]}' --output json \
| jq '.[] | select(.gidnumber<=1879048192 or .gidnumber>=2147483647)'
```

```diff
diff --git a/tools/cli/src/cli/raw.rs b/tools/cli/src/cli/raw.rs
index c53f211da..30227e1a6 100644
--- a/tools/cli/src/cli/raw.rs
+++ b/tools/cli/src/cli/raw.rs
@@ -9,7 +9,7 @@ use kanidm_proto::internal::{Filter, Modify, ModifyList};
 use kanidm_proto::v1::Entry;
 use serde::de::DeserializeOwned;
 
-use crate::RawOpt;
+use crate::{OutputMode, RawOpt};
 
 fn read_file<T: DeserializeOwned, P: AsRef<Path>>(path: P) -> Result<T, Box<dyn Error>> {
     let f = File::open(path)?;
@@ -36,13 +36,24 @@ impl RawOpt {
                 let filter: Filter = match serde_json::from_str(sopt.filter.as_str()) {
                     Ok(f) => f,
                     Err(e) => {
-                        error!("Error -> {:?}", e);
+                        error!("Error parsing filter -> {:?}", e);
                         return;
                     }
                 };
 
                 match client.search(filter).await {
-                    Ok(rset) => rset.iter().for_each(|e| println!("{}", e)),
+                    Ok(rset) => match sopt.commonopts.output_mode {
+                        #[allow(clippy::expect_used)]
+                        OutputMode::Json => {
+                            println!(
+                                "{}",
+                                serde_json::to_string(&rset).expect("Failed to serialize entry!")
+                            )
+                        }
+                        OutputMode::Text => {
+                            rset.iter().for_each(|e| println!("{}", e));
+                        }
+                    },
                     Err(e) => error!("Error -> {:?}", e),
                 }
             }

```

server/lib/src/server/migrations.rs
```diff
@@ -863,6 +863,80 @@ impl<'a> QueryServerWriteTransaction<'a> {
...
+                    f_andnot(f_lt(
+                        Attribute::GidNumber,
+                        PartialValue::Uint32(crate::plugins::gidnumber::GID_UNUSED_C_MIN)
+                    ))
+                ]),
+            ]))
+        ]));
+
+        let modlist = modlist!([m_purge(Attribute::GidNumber)]);
+
```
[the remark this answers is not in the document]
Firstyear: Yep, I can add that.

Firstyear:
Temporarily blocked on https://github.com/kanidm/kanidm/discussions/2607

server/lib/src/plugins/gidnumber.rs
```diff
@@ -33,31 +76,44 @@ fn apply_gidnumber<T: Clone>(e: &mut Entry<EntryInvalid, T>) -> Result<(), Opera
...
         let gid_v = Value::new_uint32(gid);
         admin_info!("Generated {} for {:?}", gid, u_ref);
         e.set_ava(Attribute::GidNumber, once(gid_v));
         Ok(())
     } else if let Some(gid) = e.get_ava_single_uint32(Attribute::GidNumber) {
         // If they provided us with a gid number, ensure it's in a safe range.
-        if gid <= GID_SAFETY_NUMBER_MIN {
-            admin_error!(
-                "Requested GID {} is lower or equal to a safe value {}",
+        if (gid >= GID_REGULAR_USER_MIN && gid <= GID_REGULAR_USER_MAX)
```
Firstyear, line 91: @dmulder

Firstyear:
@yaleman ready to go :)

server/daemon/src/main.rs
```diff
@@ -170,6 +175,52 @@ async fn submit_admin_req(path: &str, req: AdminTaskRequest, output_mode: Consol
...
+                            debug!("from_level             : {}", item.from_level);
+                            debug!("to_level               : {}", item.to_level);
+                            info!("status                 : PASS");
+                        }
+                        ProtoDomainUpgradeCheckStatus::Fail6To7Gidnumber => {
+                            info!("upgrade_item           : gidnumber range validity");
+                            debug!("from_level             : {}", item.from_level);
+                            debug!("to_level               : {}", item.to_level);
+                            info!("status                 : FAIL");
+                            info!("description            : The automatic allocation gidnumbers for posix accounts was found to allocate numbers into systemd-reserved ranges. These can no longer be used.");
```
yaleman, line 213:
```suggestion
                            info!("description            : The automatically allocated gidnumbers for POSIX accounts was found to allocate numbers into systemd-reserved ranges. These can no longer be used.");
```

server/lib/src/constants/mod.rs
```diff
@@ -45,13 +45,30 @@ pub const SYSTEM_INDEX_VERSION: i64 = 31;
  */
 pub type DomainVersion = u32;
 
+/// Domain level 0 - this indicates that this instance
+/// is a new install and has never had a domain level
+/// previously.
 pub const DOMAIN_LEVEL_0: DomainVersion = 0;
```
yaleman, line 51: these are starting to feel like... numbers, or an enum is required 😄

yaleman (changes requested):

server/lib/src/constants/mod.rs
Firstyear: The reason I want them as const, is so when we *delete* them from the constants, we find all the places we were relying on the value in the code base.

tools/cli/src/opt/kanidm.rs
```diff
@@ -444,6 +451,13 @@ pub enum PersonPosix {
     Set(AccountPosixOpt),
     #[clap(name = "set-password")]
     SetPassword(AccountNamedOpt),
+    /// Reset the gidnumber of this person to the generated default
```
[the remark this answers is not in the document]
Firstyear: Stable when you call reset.

server/lib/src/plugins/gidnumber.rs
```diff
@@ -12,14 +12,60 @@ use crate::utils::uuid_to_gid_u32;
...
+pub const GID_REGULAR_USER_MAX: u32 = 60000;
+
+/// Systemd homed claims 60001 through 60577
+
+pub const GID_UNUSED_A_MIN: u32 = 60578;
+pub const GID_UNUSED_A_MAX: u32 = 61183;
+
+/// Systemd dyn service users 61184 through 65519
+
+pub const GID_UNUSED_B_MIN: u32 = 65520;
```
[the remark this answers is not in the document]
Firstyear: It's technically a valid range though. We don't *advertise* that we allow it, but we do. We only indicate that users should allocate in the block `UNUSED_C`. 

The problem is systems like ldap or ipa that may have used this range.

server/lib/src/constants/mod.rs
yaleman: Yeah, enum would make sense in that case, but meh, yaks. 😄

yaleman (approved):

PR #2603, opened 2024-03-01
Adding a builtin class for all built-in things

### LINKED ISSUE #2599: Add a class BuiltIn or similar to denote builtin entities.
Currently the only way to check if an entity (account, oauth2, group, ..) are built in (present on initial launch).
Is by looking at their uuid, since they are statically assigned to be close to 0.

Having code depend on this property of uuids might not be the best in terms of coupling and clearness.
(We might forget that it was checked this way and break things in the future).
It is also unclear in code when you're checking that a uuid is very small/leading with 0's.

We could add an extra attribute to the builtin entities like `class: builtin` (lowercase to stay in style with all the other classes)

This would only add this overhead on builtin entities. 

A list of builtin uuids but that just seems way less efficient.

**Additional context**
https://github.com/kanidm/kanidm/pull/2518#discussion_r1500219646

yaleman:
Fixes #2599

Checklist

- [x] cargo fmt has been run

### DIFF
--- a/server/lib/src/server/migrations.rs
+++ b/server/lib/src/server/migrations.rs
@@ -868,7 +868,6 @@ impl<'a> QueryServerWriteTransaction<'a> {
             SCHEMA_ATTR_LIMIT_SEARCH_MAX_FILTER_TEST_DL6.clone().into(),
             SCHEMA_CLASS_ACCOUNT_POLICY_DL6.clone().into(),
             SCHEMA_CLASS_SERVICE_ACCOUNT_DL6.clone().into(),
-            SCHEMA_CLASS_BUILTIN.clone().into(),
         ];
 
         idm_schema_classes
@@ -936,6 +935,10 @@ impl<'a> QueryServerWriteTransaction<'a> {
     pub fn initialise_schema_idm(&mut self) -> Result<(), OperationError> {
         admin_debug!("initialise_schema_idm -> start ...");
 
+        // ⚠️  DOMAIN LEVEL 1 SCHEMA ATTRIBUTES ⚠️
+        // Future schema attributes need to be added via migrations.
+        //
+        // DO NOT MODIFY THIS DEFINITION
         let idm_schema_attrs = [
             SCHEMA_ATTR_SYNC_CREDENTIAL_PORTAL.clone().into(),
             SCHEMA_ATTR_SYNC_YIELD_AUTHORITY.clone().into(),
@@ -950,7 +953,10 @@ impl<'a> QueryServerWriteTransaction<'a> {
         }
         debug_assert!(r.is_ok());
 
-        // List of IDM schemas to init.
+        // ⚠️  DOMAIN LEVEL 1 SCHEMA ATTRIBUTES ⚠️
+        // Future schema classes need to be added via migrations.
+        //
+        // DO NOT MODIFY THIS DEFINITION
         let idm_schema: Vec<EntryInitNew> = vec![
             SCHEMA_ATTR_MAIL.clone().into(),
             SCHEMA_ATTR_ACCOUNT_EXPIRE.clone().into(),
@@ -1022,8 +1028,10 @@ impl<'a> QueryServerWriteTransaction<'a> {
 
         debug_assert!(r.is_ok());
 
-        // !!! DOMAIN LEVEL 1 SCHEMA CLASSES
-        // !!! Future schema classes need to be added via migrations.
+        // ⚠️  DOMAIN LEVEL 1 SCHEMA CLASSES ⚠️
+        // Future schema classes need to be added via migrations.
+        //
+        // DO NOT MODIFY THIS DEFINITION
         let idm_schema_classes_dl1: Vec<EntryInitNew> = vec![
             SCHEMA_CLASS_ACCOUNT.clone().into(),
             SCHEMA_CLASS_ACCOUNT_POLICY.clone().into(),
@@ -1107,6 +1115,10 @@ impl<'a> QueryServerWriteTransaction<'a> {
         debug_assert!(res.is_ok());
         res?;
 
+        // ⚠️  DOMAIN LEVEL 1 ENTRIES ⚠️
+        // Future entries need to be added via migrations.
+        //
+        // DO NOT MODIFY THIS DEFINITION
         let idm_entries: Vec<BuiltinAcp> = vec![
             // Built in access controls.
             IDM_ACP_RECYCLE_BIN_SEARCH_V1.clone(),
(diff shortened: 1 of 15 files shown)


Firstyear (changes requested):

server/lib/src/server/migrations.rs
```diff
@@ -1013,6 +1013,7 @@ impl<'a> QueryServerWriteTransaction<'a> {
         let idm_schema_classes: Vec<EntryInitNew> = vec![
             SCHEMA_CLASS_ACCOUNT.clone().into(),
             SCHEMA_CLASS_ACCOUNT_POLICY.clone().into(),
+            SCHEMA_CLASS_BUILTIN.clone().into(),
```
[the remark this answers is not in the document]
yaleman: Ah, will do 😄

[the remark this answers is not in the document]
Firstyear: I really should put a big warning there saying "NEVER CHANGE THIS ITS THE WRONG SPOT"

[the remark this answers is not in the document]
yaleman: So... they can be there, just have to be in the migration as well? Or rename  `idm_schema_classes: Vec<EntryInitNew>` to `idm_schema_classes_dl1`  with a note about future things being migrations?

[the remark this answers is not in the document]
Firstyear: Whene you have it (idm_schema_migrate )? is only run on "first install" of an instance. So an upgrade or re-run won't apply it.

But on a first run *all* the domain migrations are applied too. I'd say only put it into the dl 5 to 6 migration just because that gives testing options with the new test framework we merged (if we choose to use it here).

Firstyear (approved):

Firstyear:
> Ugh, because it's attached to all the builtin objects I have to go through and add the class to them all as a migration too 🙄

My actual thought about this was instead of adding it in this way, we just do a migration to touch them all, but also have something in base.rs that says if uuid <= UUID_ANONYMOUS then add class builtin.

yaleman:
> > Ugh, because it's attached to all the builtin objects I have to go through and add the class to them all as a migration too 🙄
> 
> My actual thought about this was instead of adding it in this way, we just do a migration to touch them all, but also have something in base.rs that says if uuid <= UUID_ANONYMOUS then add class builtin.

Happy to revert and change it, but the touch-them migration's there now

server/lib/src/server/migrations.rs
```diff
@@ -894,6 +896,35 @@ impl<'a> QueryServerWriteTransaction<'a> {
                 err
             })?;
 
+        let mut modlist: Vec<_> = Vec::new();
```
Firstyear, line 899: Can't you just do:

```
let filter = f_lt(Attribute::Uuid, PartialValue::Uuid( uuid::uuid!("00000000-0000-0000-0001-000000000000") ));
let modlist = m_pres(Attribute::Class, EntryClass::Object)

server.internal_modify(&filter, &modlist)
```

I think it may need a change in server/lib/src/valueset/uuid.rs to impl fn lessthan though .... (see server/lib/src/valueset/uint32.rs for inspiration )

yaleman: I guess I can, didn't see the need to add more internal functionality to do it... 😄 Added lessthan... I'll have at go at the filter at some point.

yaleman: Welp. That broke all the replication tests because it adds it to *everything*, including schema objects:

```
de763aa4-b78b-49d0-bff0-7e7f3b00ee98 TRACE    ┝━ 📍 [trace]: Entry::validate -> target | self.attrs: {"attributename": ValueSetIutf8 { set: {"name"} }, "class": ValueSetIutf8 { set: {"attributetype", "builtin", "object", "system"} }, "description": ValueSetUtf8 { set: {"The shortform name of an object"} }, "index": ValueSetIndex { set: [Equality, Presence] }, "last_modified_cid": ValueSetCid { set: [00000000000001709380154264643000-5919e721-9c78-48ed-ab74-f11d0efc5d19] }, "multivalue": ValueSetBool { set: [false] }, "phantom": ValueSetBool { set: [false] }, "replicated": ValueSetBool { set: [true] }, "sync_allowed": ValueSetBool { set: [true] }, "syntax": ValueSetSyntax { set: [Utf8StringIname] }, "unique": ValueSetBool { set: [true] }, "uuid": ValueSetUuid { set: [00000000-0000-0000-0000-ffff00000002] }}
de763aa4-b78b-49d0-bff0-7e7f3b00ee98 DEBUG    ┝━ 🐛 [debug]: invalid class: "builtin" | event_tag_id: 0
de763aa4-b78b-49d0-bff0-7e7f3b00ee98 ERROR    ┝━ 🚨 [error]: Schema Violation in refresh validate InvalidClass(["builtin"]) | event_tag_id: 1
de763aa4-b78b-49d0-bff0-7e7f3b00ee98 ERROR    ┕━ 🚨 [error]: Failed to refresh schema entries
thread 'repl::tests::qs_test_repl_refresh_basic' panicked at server/lib/src/repl/tests.rs:120:5:
assertion failed: repl_initialise(&mut server_b_txn,
            &mut server_a_txn).and_then(|_| server_a_txn.commit()).is_ok()
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

yaleman: Fixed it, added some more filters. The "lessthan" feels icky, but I guess as long as there's only one UUID for an object it should be fine :P

server/lib/src/server/migrations.rs
```diff
@@ -895,6 +897,23 @@ impl<'a> QueryServerWriteTransaction<'a> {
                 err
             })?;
 
+        // all the built-in objects get a builtin class
+        let filter = f_and(vec![
+            f_lt(
+                Attribute::Uuid,
+                PartialValue::Uuid(uuid::uuid!("00000000-0000-0000-0001-000000000000")),
```
Firstyear, line 904: uuid as constant for "START_OF_DYNAMIC_RANGE" or something like that?

server/lib/src/server/migrations.rs
```diff
@@ -895,6 +897,23 @@ impl<'a> QueryServerWriteTransaction<'a> {
...
+            ),
+            f_or(vec![
```
Firstyear, line 906: Why the or? Just tag everything that ships as built in, including schema because then we can tell what's ours and what's not.

Firstyear (changes requested):
https://github.com/kanidm/kanidm/blob/master/server/lib/src/plugins/base.rs#L78

At this point everything has a UUID. So you can do an 

```
if uuid < START_OF_DYN_RANGE {
    entry.add_ava(Class, Builtin)
}
```

This way we don't have to think about it ever again.

server/lib/src/server/migrations.rs
yaleman: Because it was breaking all the replication tests before I added this, complaining that "builtin" wasn't valid.

Firstyear: Can you mak ethe change on the PR then I'll solve the replication issue?

yaleman: so just drop the and/or combo on filter on lt(uuid)?

Firstyear: Yep

server/lib/src/server/migrations.rs
yaleman: `DYNAMIC_RANGE_MINIMUM_UUID` done!

Firstyear (approved):

PR #2608, opened 2024-03-02
Better WebAuthn and other error responses

yaleman:
Fixes that thing where if you don't have UV you should get a *slightly* nicer error.

Checklist

- [x] cargo fmt has been run


proto/src/internal/error.rs
```diff
@@ -145,3 +146,129 @@ impl PartialEq for OperationError {
...
+
+impl OperationError {
+    /// This is bad but I don't feel that bad.
+    ///
+    /// It takes something like `CU0001WebauthnAttestationNotTrusted` and turns it
+    /// into `CU0001 - Webauthn Attestation Not Trusted` if it can, otherwise you just get the normal
+    /// debug format with spaces
+    ///
+    /// Probably shouldn't use this with any of the complex types because it'll get weird quick!
+    pub fn variant_as_nice_string(&self) -> String {
```
Firstyear, line 158: Are we using this anywhere?

yaleman: yeah, aded it [over here](https://github.com/yaleman/kanidm/blob/031acc0bbc43e67774fbce31e6c80a5e1deac8cd/server/core/src/https/errors.rs#L73) as a first run to get a nice response to the user

Firstyear: Wouldn't it be better to just have a to_str that matches rather than a regex? Seems a bit overkill ...

yaleman:
@Firstyear I cleaned up the regex silliness 😄 

... those cache test fails are weird, I can't tell how I touched anything in a way that'd break *just* them?

server/core/src/https/errors.rs
```diff
@@ -40,30 +40,32 @@ impl IntoResponse for WebError {
                 (StatusCode::INTERNAL_SERVER_ERROR, inner).into_response()
             }
             WebError::OperationError(inner) => {
-                let (response_code, headers) = match &inner {
+                match &inner {
```
Firstyear, line 43: The cache test error starts here. It's because you have (_, None, _) in the response, and that confuses the headers (I think).

But also, it looks like you're trying to stuff the error code as a string rather than just using the json version we have. Is there a reason to avoid the json here? I think you're trying to make it so the *server* sends the pretty error message here rather than an error code, when really the webui should be getting the error code and then renders the message from that.

Firstyear: (Reverting this one segment btw fixes all the tests. )

yaleman: > But also, it looks like you're trying to stuff the error code as a string rather than just using the json version we have. Is there a reason to avoid the json here? I think you're trying to make it so the _server_ sends the pretty error message here rather than an error code, when really the webui should be getting the error code and then renders the message from that.

Previously it was running `serde_json::to_string` on the `OperationError`, and if that didn't work it'd just string-format it anyway, so we were already just dumping a bare string at the user and showing them that. I'm trying to make it *slightly* less bad for any client using the system without a rewrite of the whole error-handling on both the client/webui/CLI (until we work out how to do *that* rewrite) 😄

yaleman: Reverted it because ... downstreams are JSON-decoding the response, ew.

Firstyear (approved):

PR #2640, opened 2024-03-09
Begin the basis of the key provider model

### LINKED ISSUE #2513: Implement Signing/Encryption Key Domains
Implement key-domains, where a key-domain is responsible for signing task over a subset of elements within Kanidm. This might be an Oauth2rs, service acounts, user auth tokens, or more. 

Each key domain should have a way to rotate keys, invalidate former keys, and the ability to increase their cryptographic standards in use. 

Investigate if this can apply to identity verification.

Firstyear:
Fixes #2513 

Checklist

- [ ] cargo fmt has been run


Firstyear:
Just need to release compact-jwt to support this :)

proto/src/internal/error.rs
```diff
@@ -136,6 +140,52 @@ pub enum OperationError {
...
+    KP0010KeyObjectSignerToVerifier,
+    KP0011KeyObjectMissingClass,
+    KP0012KeyObjectMissingProvider,
+    KP0012KeyProviderNotLoaded,
+    KP0013KeyObjectJwsEs256DerInvalid,
+    KP0014KeyObjectSignerToVerifier,
+    KP0015KeyObjectJwsEs256DerInvalid,
+    KP0016KeyObjectJwsEs256DerInvalid,
+    KP0017KeyProviderNoSuchKey,
+    KP0018KeyProviderNoSuchKey,
```
yaleman, line 162: Why's there KP0017 and KP0018 both with the same tail-names?

yaleman: (Also... a few others...?)

I guess this is different parts in the code?

proto/src/oauth2.rs
```diff
@@ -124,7 +125,50 @@ impl From<GrantTypeReq> for AccessTokenRequest {
...
+    #[serde_as(as = "StringWithSeparator::<SpaceSeparator, String>")]
+    pub scope: BTreeSet<String>,
+
+    pub nonce: Option<String>,
+
+    pub session_id: Uuid,
+    pub parent_session_id: Option<Uuid>,
+}
+
+/// The respons for an access token
```
yaleman, line 171:
```suggestion
/// The response for an access token
```

proto/src/oauth2.rs
```diff
@@ -124,7 +125,50 @@ impl From<GrantTypeReq> for AccessTokenRequest {
...
+    pub sub: Uuid,
+    /// client_id of the oauth2 rp
+    pub aud: String,
+    /// Expiry in utc epoch seconds
+    pub exp: i64,
+    /// Not valid before.
+    pub nbf: i64,
+    /// Issued at time.
+    pub iat: i64,
+    /// -- not used.
```
yaleman, line 146:
```suggestion
    /// Not used, but part of the spec
```

proto/src/oauth2.rs
```diff
@@ -124,7 +125,50 @@ impl From<GrantTypeReq> for AccessTokenRequest {
...
+where
+    V: Clone,
+{
+    /// The issuer of this token
+    pub iss: String,
+    /// Unique id of the subject
+    pub sub: Uuid,
+    /// client_id of the oauth2 rp
+    pub aud: String,
+    /// Expiry in utc epoch seconds
```
yaleman, line 140:
```suggestion
    /// Expiry in UTC epoch seconds
```

server/core/src/actors/v1_read.rs
```diff
@@ -1096,11 +1099,16 @@ impl QueryServerReadV1 {
         session_token: CUSessionToken,
         eventid: Uuid,
     ) -> Result<CUStatus, OperationError> {
+        let session_token = JweCompact::from_str(&session_token.token)
+            .map(|token_enc| CredentialUpdateSessionToken { token_enc })
+            .map_err(|err| {
+                error!(?err, "malformed token");
```
yaleman, line 1105: This could be a *little* more descriptive 😄

server/core/src/actors/v1_read.rs
```diff
@@ -1125,11 +1133,15 @@ impl QueryServerReadV1 {
         scr: CURequest,
         eventid: Uuid,
     ) -> Result<CUStatus, OperationError> {
+        let session_token = JweCompact::from_str(&session_token.token)
+            .map(|token_enc| CredentialUpdateSessionToken { token_enc })
+            .map_err(|err| {
+                error!(?err, "malformed token");
```
yaleman, line 1139: This could be a _little_ more descriptive 😄

server/core/src/https/v1.rs
```diff
@@ -2908,8 +2908,37 @@ pub async fn debug_ipinfo(
...
+        (status=200, body=Jwk, content_type="application/json"),
+        ApiResponseWithout200,
+    ),
+    security(("token_jwt" = [])),
+    tag = "v1/jwk",
+    operation_id = "public_jwk_key_id_get"
+)]
+pub async fn public_jwk_key_id_get(
+    State(state): State<ServerState>,
+    Path(key_id): Path<String>,
```
yaleman, line 2924: Are key_id's always UUIDs or are they identifiers from the keys themselves? I tried to dig in but couldn't really follow it that deep... 😄

server/lib/Cargo.toml
```diff
@@ -28,6 +28,7 @@ harness = false
 [dependencies]
 base64 = { workspace = true }
 base64urlsafedata = { workspace = true }
+bitflags.workspace = true
```
yaleman, line 31:
```suggestion
bitflags = { workspace = true }
```

yaleman (commented):
Incremental review comments so GitHub doesn't lose 'em

proto/src/internal/error.rs
Firstyear: Yep, each error has a unique "raise site" so that if the error occurs you can identify precisely where it occured.

server/core/src/https/v1.rs
Firstyear: KeyIDs are strings that are provided from the keysource itself. For a TPM it will be the object "name" which is generated by the tpm, and in this case from compact-jwt it's a truncated 256 of the public key. It's a "unique id" of the key.

server/lib/src/idm/oauth2.rs
```diff
@@ -2094,98 +2161,104 @@ impl<'a> IdmServerProxyReadTransaction<'a> {
...
+        })
+
+        /*
+        // https://openid.net/specs/openid-connect-basic-1_0.html#UserInfoErrorResponse
+        Oauth2TokenType::ClientAccess { .. } => {
+            warn!("OpenID userinfo introspection of client access tokens is not currently supported.");
+            Err(Oauth2Error::InvalidToken)
         }
+        Oauth2TokenType::Refresh { .. } => Err(Oauth2Error::InvalidToken),
+        */
```
yaleman, line 2261: This can go now I guess?

yaleman (commented):
Niggles flagged but otherwise, woooooo.

So as far as I can tell, user/client facing changes in this commit:

- enables multiple instance config on the client
- existing OAuth2 sessions will need to reauth

Anything else? It'd be good to explicitly call this out in the commit logs if nothing else.

Firstyear:
> Niggles flagged but otherwise, woooooo.
> 
> So as far as I can tell, user/client facing changes in this commit:
> 
>     * enables multiple instance config on the client
> 
>     * existing OAuth2 sessions will need to reauth
> 
> 
> Anything else? It'd be good to explicitly call this out in the commit logs if nothing else.

Yeah, I was planning to reword the commit to include this. :) 

I think the other change is the process to view key ids and reset them.

Firstyear:
Updated with your feedback, I'll reword the commit message when we squash :)

yaleman (approved):
Sweet, all looks good here ❤️

PR #2649, opened 2024-03-11
Add MFA capabilites

dmulder:
Adds the ability to send a generic MFA request. Himmelblau requires that we keep track of a bunch of urls, cookies, tokens, etc in order to continue the MFA auth after talking to PAM, so that is the reason for the `data` blob (which is a json string). Perhaps there is a way we can standardize this into something reasonable that Kanidm can also use for MFA.
This specifically is what that json (converted from a struct) looks like:
```rust
#[derive(Deserialize, Serialize)]
pub struct MFAAuthContinue {
    pub mfa_method: String,
    pub msg: String,
    max_poll_attempts: Option<u32>,
    polling_interval: Option<u32>,
    session_id: String,
    flow_token: String,
    ctx: String,
    canary: String,
    url_end_auth: String,
    url_begin_auth: String,
    url_post: String,
}
```

MS supplies literally hundreds of random bits, and I'm only storing the bare minimum. It's possible the required data could change. MS has not written up any specification or standard for this.

Checklist

- [ ] cargo fmt has been run

### DIFF
--- a/unix_integration/pam_kanidm/src/pam/mod.rs
+++ b/unix_integration/pam_kanidm/src/pam/mod.rs
@@ -53,6 +53,9 @@ use tracing_subscriber::filter::LevelFilter;
 use tracing_subscriber::fmt;
 use tracing_subscriber::prelude::*;
 
+use std::thread;
+use std::time::Duration;
+
 pub fn get_cfg() -> Result<KanidmUnixdConfig, PamResultCode> {
     KanidmUnixdConfig::new()
         .read_options_from_optional_config(DEFAULT_CONFIG_PATH)
@@ -105,6 +108,38 @@ impl TryFrom<&Vec<&CStr>> for Options {
 pub struct PamKanidm;
 
 pam_hooks!(PamKanidm);
+
+macro_rules! match_sm_auth_client_response {
+    ($expr:expr, $opts:ident, $($pat:pat => $result:expr),*) => {
+        match $expr {
+            Ok(r) => match r {
+                $($pat => $result),*
+                ClientResponse::PamAuthenticateStepResponse(PamAuthResponse::Success) => {
+                    return PamResultCode::PAM_SUCCESS;
+                }
+                ClientResponse::PamAuthenticateStepResponse(PamAuthResponse::Denied) => {
+                    return PamResultCode::PAM_AUTH_ERR;
+                }
+                ClientResponse::PamAuthenticateStepResponse(PamAuthResponse::Unknown) => {
+                    if $opts.ignore_unknown_user {
+                        return PamResultCode::PAM_IGNORE;
+                    } else {
+                        return PamResultCode::PAM_USER_UNKNOWN;
+                    }
+                }
+                _ => {
+                    // unexpected response.
+                    error!(err = ?r, "PAM_IGNORE, unexpected resolver response");
+                    return PamResultCode::PAM_IGNORE;
+                }
+            },
+            Err(err) => {
+                error!(?err, "PAM_IGNORE");
+                return PamResultCode::PAM_IGNORE;
+            }
+        }
+    }
+}
 
 impl PamHooks for PamKanidm {
     fn acct_mgmt(pamh: &PamHandle, args: Vec<&CStr>, _flags: PamFlag) -> PamResultCode {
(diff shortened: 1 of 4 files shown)


unix_integration/src/unix_proto.rs
```diff
@@ -30,28 +30,34 @@ pub struct DeviceAuthorizationResponse {
     pub message: Option<String>,
 }
 
+#[derive(Serialize, Deserialize, Debug)]
+pub enum MFAType {
+    Code, // PAM must prompt for an authentication code
```
Firstyear, line 35: `///` is a doc comment if you put it on the line above the enum variant.

unix_integration/src/unix_proto.rs
```diff
@@ -30,28 +30,34 @@ pub struct DeviceAuthorizationResponse {
...
+    Poll, // PAM can immediately return. The provider will poll for an external response
```
Firstyear, line 36: Should poll be part of the clients request? Because else how does the pam module go "idle" waiting for the resolver to respond?

unix_integration/src/unix_proto.rs
```diff
@@ -30,28 +30,34 @@ pub struct DeviceAuthorizationResponse {
...
     Denied,
     Password,
-    DeviceAuthorizationGrant { data: DeviceAuthorizationResponse },
-    /*
+    DeviceAuthorizationGrant {
+        data: DeviceAuthorizationResponse,
+    },
     MFACode {
+        msg: String,
+        mfa_type: MFAType,
```
Firstyear, line 50: Is there a reason for this to be an inner enum rather than just different variants of PamAuthResponse?

unix_integration/src/resolver.rs
```diff
@@ -1111,6 +1115,10 @@ where
                 // Can continue!
                 auth_session
             }
+            (auth_session, PamAuthResponse::MFACode { .. }) => {
+                // Can continue!
```
Firstyear, line 1119: How can this continue?

unix_integration/src/idprovider/interface.rs
```diff
@@ -63,11 +63,21 @@ pub struct UserToken {
...
-    DeviceAuthorizationGrant { data: DeviceAuthorizationResponse },
+    DeviceAuthorizationGrant {
+        data: DeviceAuthorizationResponse,
+    },
+    MFACode {
+        msg: String,
+        mfa_type: MFAType,
+        // Additional data required by the provider to complete the
+        // authentication, but not required by PAM
+        data: String,
```
Firstyear, line 79: Maybe it should be Vec<u8> then to allow other encodings or structures? Why not just have more request variants for each type of possible additional data so we get typing?

unix_integration/pam_kanidm/src/pam/mod.rs
```diff
@@ -309,6 +309,53 @@ impl PamHooks for PamKanidm {
...
+                                Ok(_) => {}
+                                Err(err) => {
+                                    if opts.debug {
+                                        println!("Message prompt failed");
+                                    }
+                                    return err;
+                                }
+                            }
+
+                            timeout = 300; // Default 5 minutes timeout
```
Firstyear, line 351: Seems like a pretty long timeout? Is the resolver polling and responding?

Firstyear (commented):
I'm a bit confused about the logic? Shouldn't the flow be per the diagram? Where the pam module is told to present the challenge then it loops polling the resolver (at a rapid interval, say once a second), and then the resolver itself can asynchronously complete the polling to azure in the bg?

From your code it seems like the polling is the opposite direction? 

Screenshot 2024-03-12 at 11 48 36

unix_integration/src/idprovider/interface.rs
dmulder: I like that idea.

unix_integration/src/unix_proto.rs
dmulder: The poll is happening in the id provider. I only named it 'Poll' meaning what the id provider is doing. I guess that's confusing. The Poll type is operating the same as the DeviceAuthorizationGrant, which is, print a prompt, then immediately call the daemon and wait (the reason for the long timeout) for it to acknowledge the auth completed.

dmulder:
> I'm a bit confused about the logic? Shouldn't the flow be per the diagram? Where the pam module is told to present the challenge then it loops polling the resolver (at a rapid interval, say once a second), and then the resolver itself can asynchronously complete the polling to azure in the bg?
> 
> From your code it seems like the polling is the opposite direction?

It's mirroring the exact same behavior as the DeviceAuthorizationGrant code already in there. MS provides some MFA options where we're expected to print a message saying, for example "We're calling you on X phone number and you need to acknowledge this authentication is coming from you". Then we just poll in the id provider and wait for MS to respond. There's also one where the user opens an app on their phone, and they have to press a button confirming the authentication attempt.
But I also have to handle inputs sometimes, such as an Authenticator code app, or a text message. That's what the Code prompt is for.

dmulder:
Looking over your diagram, I'd say that process makes sense, but that's not currently what we're doing for the DAG auth. Instead, pam just sends a single request to the provider for a response with a long timeout.

dmulder:
I've pushed a bunch of changes which redesigns the MFA requests after your suggestions.

unix_integration/src/unix_proto.rs
dmulder: I think I've resolved this one by modeling the diagram you provided.

Firstyear:
@dmulder Do you think the updated approach seems to work better in your view?

unix_integration/src/resolver.rs
Firstyear: Oh sorry, I read it wrong, I think I opened github wrong and thought this was the offline path.  Sorry!

unix_integration/pam_kanidm/src/pam/mod.rs
```diff
@@ -309,6 +312,61 @@ impl PamHooks for PamKanidm {
...
+                        timeout = cfg.unix_sock_timeout;
+                        req = ClientRequest::PamAuthenticateStep(PamAuthRequest::MFACode {
+                            cred,
+                            data,
+                        });
+                        continue;
+                    }
+                    ClientResponse::PamAuthenticateStepResponse(PamAuthResponse::MFAPoll {
+                        msg,
+                        max_poll_attempts,
```
Firstyear, line 343: Shouldn't the fact we are in a poll state be in the auth session? And it it tracks this? 

Said better, the resolver is controlling the whole flow. The pam module is just an information pipe. So the resolver when it presents the challenge to pam along with the poll interval. then pam module implicitly moves to "start polling". Then pam module just sends a "poll" message with no options. 

It's up to the resolver to track the timeout in the authsession, and indicate that the pam module continues to wait, errors, or has succeeded etc. 

To me I'd assume the PamAuthRequest::MFAPoll shouldn't have any optinos at all, and when we start the poll we tell the client in the clientResponse only the polling interval. 

It's up to the resolvers auth_session state to track the number of polls etc.

Does that seem like it could be a better approach?

Firstyear: I guess what I'm saying is the pam module should have less logic, it's just polling the resolver until the resolver decides and tells it what to do.

Firstyear (changes requested):

unix_integration/pam_kanidm/src/pam/mod.rs
dmulder: Ok, I see what you mean. Looking into it.

dmulder: Well, I've tried today, but I'm having trouble wrapping my head around it. I suppose the auth provider would start the polling, but where would the auth_session init happen (a polling init)? This seems like an awful lot of complexity for something so simple.

Firstyear: Maybe think of it like this. The pam module is just asking "are we there yet" like a kid on a road trip. 

In the background the auth_session has spawn an extra task that does the communication to the provider in it's own internal loop and on completion can update a status that's behind an async mutex in the auth session.

And yeah, it seems complex, but we are dealing with multiple moving parts (pam, resolver and azure ad) as well as a human. The human can ctrl c things, network errors, or anything else. So we have to model it as an event driven system and have states that are transitioned through. 

And the reason we add this now in this way, is that we can re-use the same process and parts later in other things because what we are really modeling is a generic-concurrent auth checker.

WRT to the init, it's the same as any interaction with pam. The session starts https://github.com/kanidm/kanidm/blob/master/unix_integration/src/resolver.rs#L867 and then we have next_req. So the auth_session can immediately make this `PamAuthResponse::BeginPollWait { msg: String, interval: u8 }`. In the background auth_session has already started the task to communicate and poll itself to azure etc. 

Then pam sets "PamAuthRequest::Poll". The auth session + resolver checks their async mutex about the task state and sees it's still "Waiting" and would return PamAuthResponse::PollWait.

If an error happens then the response is PamAuthResponse::Denied. 

If pam requests PamAuthRequest::Poll and the auth session + resolver is ready and the status is "denied, set Denied in the response. If it's "Success" then you can make the next choice to either return PamAuthResponse::Success or you can then start something like PamAuthResponse::PinEnroll to get the user to set their PIN on first use or something.

dmulder: > Maybe think of it like this. The pam module is just asking "are we there yet" like a kid on a road trip.

I get the broader picture. I'm having trouble executing your vision of it.

> WRT to the init, it's the same as any interaction with pam. The session starts https://github.com/kanidm/kanidm/blob/master/unix_integration/src/resolver.rs#L867 and then we have next_req. So the auth_session can immediately make this `PamAuthResponse::BeginPollWait { msg: String, interval: u8 }`. In the background auth_session has already started the task to communicate and poll itself to azure etc.

In himmelblau, I don't even know at this point that I'll be doing a poll (when in `pam_account_authenticate_init()`). That doesn't happen until after a response from `unix_user_online_auth_step()`. This is because I first send just a username+password req to Azure, and if it responds with an error saying MFA is required, then we start that dance.

> Then pam sets "PamAuthRequest::Poll". The auth session + resolver checks their async mutex about the task state and sees it's still "Waiting" and would return PamAuthResponse::PollWait.

I also don't see the purpose of leaving a task polling here. This is a single call every X # of seconds (as specified by Azure). IMHO, we should just wait in PAM, sending a request to the daemon every so often (the interval specified by Azure), until the daemon responds with an auth success.
Why spin a daemon to poll Azure, when all it's going to do is sit there waiting for us to send a request from PAM for the auth status?

dmulder: MFA_challenge drawio

This is what it's currently doing. The resolver should only send a poll status request to Azure if PAM asks for it, IMHO. There's no reason for the resolver to be spinning and polling simultaneously as PAM is spinning and sending status requests to the resolver.

Currently my 'wait' response is just sending a new `PamAuthResponse::MFAPoll` to PAM. This isn't necessary, but was just a sort of simplification (that's the reason for the extra data, is I just combined it into a single request). Perhaps we could have a sub loop within the initial `PamAuthResponse::MFAPoll` that doesn't require passing the same data over and over?

dmulder: > Currently my 'wait' response is just sending a new `PamAuthResponse::MFAPoll` to PAM. This isn't necessary, but was just a sort of simplification (that's the reason for the extra data, is I just combined it into a single request). Perhaps we could have a sub loop within the initial `PamAuthResponse::MFAPoll` that doesn't require passing the same data over and over?

I implemented this as a sub-loop in the latest changes here.

Firstyear: And of course, then you don't need "max attempts" or counting or anything. Pam just keeps polling until the resolver says "no" . Then it's completely up to the resolver and it's timeouts and stuff with it's session to azure ad which influences the whole thing. The logic now firmly becomes tied only to the resolver and azure communication and it's timeouts, rather than multiple possible timeouts that can occur.

Firstyear: The other problem is with adjusting the timeout like https://github.com/kanidm/kanidm/blob/master/unix_integration/pam_kanidm/src/pam/mod.rs#L306 then this taints the timeout for the rest of the pam module session where it's now possible to leave the user waiting for a really long time if the resolver hasn't returned.

Firstyear: for comparison, what do you think are the benefits to the way you wanted to do it? Now I'm getting in my own head about if my ideas are right or not.

Firstyear: I'm getting in my head now about if I was right in what I said at all. Maybe there is an easier way, I'm not sure. I feel like I've sent you on a wild chase now. 

Possibly the right answer is in front of us: 

https://github.com/kanidm/kanidm/blob/master/unix_integration/pam_kanidm/src/pam/mod.rs#L244

pam here has a timeout so that if the resolver takes too long for an expected action, it fails. In my head I was thinking this timeout detects disconnects of the resolver, but that's handled by the err case in the loop. 

When the resolver says "wait" we tell pam an upper limit to wait for. Then the resolver can do it's work and no polling needed, and when the resolver is done it just sends back a response. 

I feel extremely stupid not seeing the simple and obvious answer here. I think we don't want polls at all. 


Screenshot 2024-03-15 at 12 21 06

Pam calls the resolver, which sends back "present the mfa challenge to the user". When it's displayed to the user, pam messages us to say "yep, they've seen it". That now sets pam blocking on the socket to wait for a response message. In the "present challenge" we send a "long timeout" adjustment.

Then the resolver can sit there and poll on it's own schedule as needed. Internally the resolver should have a "proper" timeout incase things take too long. 

This way the resolver then only needs to send back one message wen the communication is complete, and then pam can advance as well. 

I think this is way simpler, doesn't need any polling, and will be better for you to implement, and covers our needs. 

What do you think? I feel really really really bad giving you such a run around on this, I'm really sorry.

dmulder: Well, that's how the DAG works right now in there. I assumed you meant you wanted that redesigned to do polling as well (I was going to do this next).

dmulder: So this whole conversation wasn't a waste. I actually am going to argue in favor of polling.

The way the current patches are written, PAM is waiting in between sending requests to the resolver. The resolver does nothing unless prompted to check the status.

If we do a long wait instead, PAM is waiting on the response, and the resolver is spinning and calling Azure periodically.

Either way will work, but think about what happens when a user terminates the authentication attempt. If PAM is just periodically requesting an update, then the whole session immediately terminates (because the resolver wasn't doing anything anyway, unless PAM asked it to). If pam is waiting instead on a timeout, then the resolver **doesn't stop asking** Azure for an update until the max tries have completed.

So, consider the implications if say 1000 users did the same thing... started to authenticate, but bailed out all at once. Now we've got our resolver bogged down with many requests, none of which will terminate until we reach **max_poll_attempts * polling_interval**, and none of which PAM cares to see completed anyway. This doesn't scale well.

Firstyear: This is a very valid and legitimate concern, but there is a way to handle this, and it may be cleaner than we expect.

If the pam session terminates then https://github.com/kanidm/kanidm/blob/master/unix_integration/src/daemon.rs#L206 drops and yields None, going to https://github.com/kanidm/kanidm/blob/master/unix_integration/src/daemon.rs#L410

We can use this behaviour to then signal that the session dropped to the underlying resolver. All we have to do is add https://docs.rs/tokio/latest/tokio/sync/broadcast/index.html 

We setup the broadcast at the start, and then for requests that need it, we send the subscriber side to the cachelayer such as `cachelayer.pam_account_authenticate_step(auth_session, pam_next_req, liveness_subscriber)`. 

Then within the authenticate step, if we ever wait on something (such as our loop for polling azure) we do it in a select such as:

```
while max_poll > 0 {
    max_poll--;
select! {
    Ok(something) = azure.poll() => {}
   _ = timeout => { continue }
    _ = liveness_subscriber => { break }
}
```

The trick with the liveness_subscriber is any message on it *or* if the broadcast sender drops, it causes an event on the subscriber allowing immediate cancellation. 

Then around line 410 in daemon we just have the broadcast.send(()) just to cause the subscriber to activate and cancel.

Does that seem like a good option? It means we keep all the max poll logic and timeouts in the resolver (himmleblau) while also having a way to signal to spawned tasks that they have been canceled.

dmulder: That sounds ok. I've been pulled away to Samba code for a bit, but I'll get back to this soon.

Firstyear: Turns out it won't work anyway. After sleeping on it, you are correct that we need polling. The reason is that the outer daemon loop won't be on the recv().await, it will be awaiting the inner resolver task.

So we do need polling. 

I suggest we go back to what I suggested before, where we instruct pam to just poll in a loop until a result is ready. All the timeouts and logic for max attempts/polls should be in the resolver (not pam). and then we can use the same broadcast channel idea to solve notifying the resolver that the pam client has disconnected. How does that sound?

dmulder: I have no idea how to implement what you're suggesting. This has gotten far too complex for something so straight forward and simple. sigh

dmulder: I'm going to push back and say I still think polling on the resolver is the wrong approach. I think the polling should happen in PAM. This makes the whole thing very straight forward and simple. This would mean PAM asks every so often for an update, meanwhile the resolver does nothing until getting a explicit request from the resolver.
Why do you want the polling to happen in the resolver? I still don't think you've given a good reason.

Firstyear: > I have no idea how to implement what you're suggesting. This has gotten far too complex for something so straight forward and simple. sigh

We have an event driven system, it's never simple. 

I'll update the PR then.

PR #2653, opened 2024-03-14
first markov attempt

Seba-T:
Fixes nothing, it actually creates more problems xD. Rn the people are unable to modify their names, I imagine that's because they would need a privileged session for that. Also for some weird reason they try to view their profiles before authenticating, which in theory shouldn't be happening if I implemented Markov the right way (I have a strong feeling I didn't).
Checklist

- [x] cargo fmt has been run


tools/orca/Cargo.toml
```diff
@@ -27,10 +27,14 @@ hashbrown = { workspace = true }
 kanidm_client = { workspace = true }
 kanidm_proto = { workspace = true }
 mathru = { workspace = true }
+ndarray = "0.15.6"
 rand = { workspace = true }
 rand_chacha = { workspace = true }
 serde = { workspace = true, features = ["derive"] }
 serde_json = { workspace = true }
+statrs = "0.16.0"
+strum = "0.26.2"
```
Firstyear, line 36: What are these being added for?

tools/orca/src/models/model_basic.rs
```diff
@@ -43,6 +43,7 @@ impl ActorModel for ActorBasic {
         let (result, event) = match transition.action {
             TransitionAction::Login => model::login(client, person).await,
             TransitionAction::Logout => model::logout(client, person).await,
+            _ => Err(Error::InvalidState),
```
Firstyear, line 46: File should be named basic.rs since we are in the dir "models".

tools/orca/src/models/model_markov.rs
```diff
@@ -0,0 +1,143 @@
...
+use rand::{
+    distributions::{self, Distribution},
+    rngs::StdRng,
+    Rng, SeedableRng,
+};
+use statrs::distribution::{Multinomial, Normal};
+use strum::EnumCount;
+use strum_macros::EnumCount;
+
+// The idea is to have a multinomial distribution for each state that represents
```
Firstyear, line 21: I think you went "too hard too fast" on the stats side.

I think you need to back up a bit and make some simpler models. Stuff that just does a linear set of actions first in a loop. Think "login, read attr, set attr, logout, repeat". I think we need those in place first because that gives the foundation of "how to create the entries in the db". 

From there we need ways to direct those models based on attributes of the person. If in group X do Y, else do Z. again, in a loop.

And then once we have that, then we can start to add in randomisation into the actions of the persons. Remember also some of these linear-loop models are good for other tests. Consider a model of login and loop on infinite searches. That's a search benchmark now! 

So I think backing out of this a bit and really stepping back to create simpler models first and then work up will be more valuable.

Firstyear (changes requested):

tools/orca/Cargo.toml
Seba-T: statrs for the Multinomial distribution and strum to count enum variants

tools/orca/src/models/model_basic.rs
Seba-T: wopsies I'll fix it

tools/orca/src/models/model_markov.rs
Seba-T: > I think you went "too hard too fast" on the stats side.
> 
> I think you need to back up a bit and make some simpler models. Stuff that just does a linear set of actions first in a loop. Think "login, read attr, set attr, logout, repeat". I think we need those in place first because that gives the foundation of "how to create the entries in the db".
> 

I got this working, now onto the group part!

tools/orca/Cargo.toml
Firstyear: I think we should remove these for now.

tools/orca/src/models/markov.rs
```diff
@@ -0,0 +1,143 @@
+use std::{slice::Iter, time::Duration};
```
Firstyear, line 1: Lets remove/disable this module for now.

tools/orca/src/models/read_write_attr.rs
```diff
@@ -0,0 +1,104 @@
...
+        // Once we get to here, we want the transition to go ahead.
+        let (result, event) = match transition.action {
+            TransitionAction::Login => model::login(client, person).await,
+            TransitionAction::ReadAttribute => model::person_get(client, person).await,
+            TransitionAction::WriteAttribute => model::person_set(client, person).await,
+            TransitionAction::Logout => model::logout(client, person).await,
+            _ => Err(Error::InvalidState),
+        }?;
+
+        // Given the result, make a choice about what text.
```
Firstyear, line 53: About what text?

tools/orca/src/state.rs
```diff
@@ -1,11 +1,12 @@
 use crate::error::Error;
 use crate::model::ActorModel;
+use crate::models::markov::DISTR_MATRIX_SIZE;
 use crate::profile::Profile;
 use serde::{Deserialize, Serialize};
 use std::collections::BTreeSet;
 use std::path::Path;
 
-/// A serialisable state representing the content of a kanidm database and potential
+/// A serializable state representing the content of a kanidm database and potential
```
Firstyear, line 9: English AU spelling, it was correct before.

Firstyear (commented):
I think we remove the markov thing (or disable it for now) then we proceed with this.

tools/orca/src/state.rs
Seba-T: Wops my bad

tools/orca/src/models/read_write_attr.rs
Seba-T: Just some absent minded cut & paste from the Basic model :p

tools/orca/src/models/markov.rs
Seba-T: sure :)

tools/orca/src/models/model_markov.rs
Seba-T: > From there we need ways to direct those models based on attributes of the person. If in group X do Y, else do Z. again, in a loop.
> 
I realized this could be done in two different ways. That is: do we want to be able to dynamically change a person's behavior by changing their groups, which means querying kani each time, or is it enough to just use whatever group we assigned at the start and we don't care about possible changes?

tools/orca/src/state.rs
```diff
@@ -112,3 +117,14 @@ pub struct Person {
     pub credential: Credential,
     pub model: Model,
 }
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct Group {
```
Firstyear, line 122: Okay, so. 

I think you should stop and think about this some more because you're on the right track, but this current approach will hit limits pretty quick.

Something we'll need is to have persons with *multiple* roles, and those roles gain permissions from groups. In your current approach every person only gets one actor role but why is that the case?

We create groups, then assign a role to the group, then assign persons to those groups.

This way persons can have M of N roles, and then within the person it can see "I'm in these roles, so I can make these choices or actions". 

So shouldn't this model

Firstyear (changes requested):

tools/orca/src/state.rs
Seba-T: Yeah in hindsight it was a quite naive attempt :p. What I'd like to better understand before diving deeper is the mechanism that will determine how persons behave. The best idea I've had so far relies on kani to determine who is entitled to do what,  while on the orca side persons don't know (at least at the start) what they can do. Here's my plan: at first we basically let everyone attempt to do everything, and when they get a 403 they don't try that specific action again. After a little while we "reset" their memory of allowed actions and we repeat the process.  We proceed in this fashion until the end of the test. 
The rationale behind this approach is that during the test itself we could be interested in changing some roles, and I couldn't come up with a better way to dynamically fetch what one can do or cannot do. 
What do you think? Should I start coding something that more or less works like this?

Firstyear: Rather than adding in the randomness, why not just make some persons that behave in a static manner first?

Seba-T: Sure!

tools/orca/src/error.rs
```diff
@@ -5,7 +5,8 @@ pub enum Error {
     KanidmClient,
     ProfileBuilder,
     Tokio,
-    Interupt,
+    Interrupt,
     Crossbeam,
     InvalidState,
+    InvalidInput(String),
```
Firstyear, line 11: I want to avoid encoding strings into error messages, because then you just end up witnh InvalidInput and 13 different possible string errors. So use a generic version (InvalidInput) or make a new enum variant with the specific error encoded. Generally with orca, because the user is there at the terminal, a generic-variant is fine since the error!() line is what encodes the error info.

tools/orca/src/generate.rs
```diff
@@ -55,7 +58,20 @@ pub async fn populate(_client: &KanidmOrcaClient, profile: Profile) -> Result<St
     // PHASE 1 - generate a pool of persons that are not-yet created for future import.
     // todo! may need a random username vec for later stuff
 
-    // PHASE 2 - generate persons
+    // PHASE 2 - generate groups for integration access, assign roles to groups
+    // actually these are just groups to decide what each person is supposed to do with their life
+    let groups = vec![
+        Group::new(
+            "readers".to_string(),
```
Firstyear, line 65: Why not make this the Enum for the ActorRole? Because we'll need a role like "Ignore" so that groups can be created just for the purposes of membercship etc.

tools/orca/src/model.rs
```diff
@@ -1,15 +1,37 @@
...
 pub enum TransitionAction {
-    Login,
-    Logout,
+    Login = 0,
+    Logout = 1,
+    ReadAttribute = 2,
+    WriteAttribute = 3,
+}
+
+impl TryFrom<i32> for TransitionAction {
```
Firstyear, line 21: Why does this need i32?

tools/orca/src/model.rs
```diff
@@ -84,6 +124,96 @@ pub async fn login(
...
+            ))
+        }
+    }
+}
+
+pub async fn person_set(
+    client: &KanidmClient,
+    person: &Person,
+) -> Result<(TransitionResult, EventRecord), Error> {
+    // Should we measure the time of each call rather than the time with multiple calls?
```
Firstyear, line 165: Each call timed individually so they can become their own event records.

Firstyear (changes requested):

tools/orca/src/error.rs
Seba-T: Yeah actually I just realized I ended up not needing that at all ☠️☠️. So never mind I guess. 
But I'll keep it in mind for the future :)

tools/orca/src/generate.rs
Seba-T: Yeah actually I have no clue what I was smoking when I wrote that. Switched to the enum

tools/orca/src/model.rs
Seba-T: Well ... it doesn't. It was the result of some VERY poor design choices that I've since abandoned :p

tools/orca/src/model.rs
Seba-T: I think it makes sense to keep it as a single transition and return something like `(TransitionResult, Vec<EventRecord>)` (Actually I came up with another enum to avoid having vecs everywhere). Alternatively we could use two separate transitions but I don't really see the point here as these two calls are basically always done together. In general having the ability to perform multiple calls timed individually could help limit "useless" transitions while still having precise measurements.

Firstyear: I guess it depends what we want to measure. I still think any singluar transition should be it's own result and event record. If we have to make multiple calls, we either coalesce them in one transition and start/pause/start/stop the timer. Or we have the State Login -> State Logging In -> State Complete. And then we get three distinct events.

Seba-T: Alright so I think we will need a "reauth" state, if we want to have a separate transition for that

Firstyear: Do you mean "reauth" as in for priv reauthentication? Or as in to logout and in? Because to logout and in, you'd move back to State::Unauthenticated

Seba-T: Yeah it’s about priv reauth, that’s why I wanted to bind it together with the modify operation

Firstyear: Oky, need re-auth should be it's own state, and then internally the actor can track it's priv state and if it needs the re-auth or not. It shouldn't be bounh with modify. 

Have you worked with state machines before.?

Seba-T: Alright sounds reasonable. 

> Have you worked with state machines before.?

Yeah but I didn't think this through that much have I

PR #2659, opened 2024-03-16
fix(docs): capitalization fixes

alexvonme:
There were two files in documentation where the first letter of Kanidm should have been capitalized.


alexvonme:
I agree with @JFortun on #2644, capitalization rules seem unclear.

May I suggest always capitalizing Kanidm unless exclusively referring to a specific package? Otherwise, it seems likely to cause needless confusion.

yaleman (approved):

yaleman:
Yes that's what we would expect but sometimes humans are forgetful. 🫤

PR #2665, opened 2024-03-18
Offer configuration of images for Oauth2 resources 

### LINKED ISSUE #2056: Set custom image/icon for applications

It would be nice to set a custom image/icon for each app (e.g. oauth2 integrations) that shows up on the web interface to replace the default crab mascot.

This may or may not be relevant to #1996 or #1955.

Perhaps something like:

```shell
kanidm system oauth2 set-logo appname https://example.com/image.svg
```

I could print the icon on a post-it note and stick it over the top of the default image on my screen. 😁

ToxicMushroom:
Fixes #2056 

Checklist

- [x] cargo fmt has been run


tools/cli/src/cli/oauth2.rs
```diff
@@ -259,6 +262,47 @@ impl Oauth2Opt {
                     Err(e) => handle_client_error(e, nopt.copt.output_mode),
                 }
             }
+            Oauth2Opt::SetImage { nopt, path, image_type } => {
+                let client = nopt.copt.to_client(OpType::Write).await;
+                let img_res: Result<ImageValue, &str> = (move || {
+                    let path = Path::new(path);
+                    let file_name = path
+                        .file_name().ok_or("Please pass a file")?
+                        .to_str().ok_or("Path contains non utf-8")?
```
Firstyear, line 271: nit, it would be "invalid utf-8" here.

yaleman:
> Should I prepend the configured origin set in `server.toml` together with `/ui/images/oauth2/{rs_name}` to create the full icon url that `AppLink::Oauth2` wants ?

Context would help for this question - the Kanidm client library has make_url which can do this for you, and on the web UI you should be using relative URLs

ToxicMushroom:
sry, it was about object here that gets returned in `pub async fn applinks_get`
 at `/v1/self/_applinks`
the actual line that should specify the icon url is here
https://github.com/ToxicMushroom/kanidm/blob/097db70c3d4964194fd29f8771b3ee4ed55912bf/server/lib/src/idm/applinks.rs#L59

ToxicMushroom:
But your answer implies that this should return a relative path which a URL is not I think ? Or I'd have to parse the icon url in the web_ui and throw away the host and protocol to make it a rel path.

yaleman:
> But your answer (because this icon field will be used by the web_ui to download the icon) implies that this should return a relative path which a URL is not I think ? Or I'd have to parse the icon url in the web_ui and throw away the host and protocol to make it a rel path.

Given this is an as-yet completely unused feature, we can define what it is right now in this PR 😄 Realistically since the URL is based on the name (`/ui/images/oauth2/{rs_name}`) and the name is in the applink struct already, that icon field could just be a `bool` called `has_custom_icon` which means the webUI accesses `/ui/images/oauth2/{rs_name}` to get the image.

ToxicMushroom:
I'm still gonna add an admin web interface for it but the current code can already be reviewed if you want to review it in smaller parts :)

server/web_ui/shared/src/lib.rs
```diff
@@ -113,6 +98,53 @@ pub async fn do_request<JV: AsRef<JsValue>>(
     Ok((kopid, status, body, headers))
 }
 
+/// Build and send a request to the backend, with some standard headers and pull back
```
ToxicMushroom, line 101: I missed this

server/lib/src/server/access/mod.rs
```diff
@@ -626,11 +622,7 @@ pub trait AccessControlsTransaction<'a> {
                         false
                     } else if !requested_rem.is_subset(&rem) {
                         security_error!("requested_rem is not a subset of allowed");
-                        security_error!(
-                            "requested_rem: {:?} !⊆ allowed: {:?}",
-                            requested_rem,
-                            rem
-                        );
+                        security_error!("requested_rem: {:?} !⊆ allowed: {:?}", requested_rem, rem);
```
Firstyear, line 625: What version of rust are you on? Looks like cargo fmt fights here.

server/web_ui/user/src/views/apps.rs
```diff
@@ -121,14 +123,30 @@ impl AppsApp {
...
+                            has_image,
                         } => {
                             let redirect_url = redirect_url.to_string();
+                            let node_ref = NodeRef::default();
+                            if *has_image {
+                                let node_ref = node_ref.clone();
+                                let name = name.clone();
+                                spawn_local(async move {
+                                    let (_, status, blob, _) =
+                                        get_blob(format!("/ui/images/oauth2/{name}").as_str()).await.unwrap();
```
Firstyear, line 135: Theres a server uri handler for this?

server/lib/src/server/access/mod.rs
ToxicMushroom: rustc 1.77.0-nightly (5518eaa94 2024-01-29)

server/web_ui/user/src/views/apps.rs
ToxicMushroom: yep https://github.com/ToxicMushroom/kanidm/blob/8175253bae820bd9584ba808a9bc8033c993b2f5/server/core/src/https/oauth2.rs#L96

ToxicMushroom:
I discussed with yaleman on the matrix channel about the blob stuff I'm using atm: 
Since we have an existing carveout for the `data:` scheme because of bootstrap, we thought it'd be best to just use that instead with base64 image data instead of adding `blob: ` to the CSP. 
So I'll refactor that part to use base64 and the server to return base64

ToxicMushroom:
This pr has been stripped down to just the cli part and internal changes, UI will be done in upcoming htmx version as reworking the oauth2 image endpoint to return base64 would conflict with the htmx UI.

ToxicMushroom:
Side note, I tried running `cargo test` multiple times, tried clearing ./target but it keeps eating all my ram, causing my pc to lock up.

tools/cli/src/opt/kanidm.rs
```diff
@@ -1048,6 +1048,16 @@ pub enum Oauth2Opt {
         #[clap(name = "landing-url")]
         url: Url,
     },
+    /// The image presented on the Kanidm Apps Listing page for a oauth2 resource server.
+    #[clap(name="set-image")]
+    SetImage {
+        #[clap(flatten)]
+        nopt: Named,
+        #[clap(name = "file-path")]
+        path: String,
```
Firstyear, line 1057: Should be a pathbuf (we really shouldn't have used strings here in the past ...)

tools/cli/src/opt/kanidm.rs
```diff
@@ -1048,6 +1048,16 @@ pub enum Oauth2Opt {
         #[clap(name = "landing-url")]
         url: Url,
     },
+    /// The image presented on the Kanidm Apps Listing page for a oauth2 resource server.
+    #[clap(name="set-image")]
+    SetImage {
```
Firstyear, line 1053: Needs a corresponding "remove-image" that just clears the image on the entry.

server/lib/src/idm/applinks.rs
```diff
@@ -51,11 +51,13 @@ impl<'a> IdmServerProxyReadTransaction<'a> {
...
                     .map(str::to_string)?;
 
+                let has_image = entry.get_ava_single_image(Attribute::Image).is_some();
+
                 Some(AppLink::Oauth2 {
                     name,
                     display_name,
                     redirect_url,
-                    icon: None,
+                    has_image,
```
Firstyear, line 60: Is there a reason we couldn't just inline the image data here so that it's ready to render?

ToxicMushroom: Applinks would be slower because you'd have to wait to receive all image data. (e.g. `/v1/self/_applinks` but also in-lining it into a template)

Firstyear (approved):

PR #2672, opened 2024-03-21
Dmulder/mfa capabilities

Firstyear:
Supercedes #2649 

This associates MFA data within the CredHandler to prevent it being sent between the pam module and the resolver, when the pam module doesn't need that data at all. It's resolver internal state. 

This also adds shutdown_rx to the interface in the init process and steps. This means that you can now have a process like:

```

unix_user_online_auth_init( ..., shutdown_rx ) {

    let (state_tx, state_rx) = tokio::sync::oneshot::channel();

    let cred_handler = MFA { state_rx };

    let task_state = syncstate.clone();

    tokio::task::spawn(move || {
        let decision = loop i < max_azure_poll_attepts {
            select! {
                 _ = shutdown_rx => {
                     break Error
                 }
                 result = azure_client => {
                     match result {
                         break Accept/Deny
                     }
                 }
            }
        };

        state_tx.send(decision);
    });

    ...

    return (AuthRequest, cred_handler)
}

unix_user_online_auth_step ( cred_handler) {
    match cred_handler {
        MFA { state_rx } => {

            // https://docs.rs/tokio/latest/tokio/sync/oneshot/struct.Receiver.html#method.try_recv
            match state_rx {
               Ok(Accept/Deny/Error) => { }
               Err(Empty) => { // Ask to poll again }
               Err(Closed) => { // Error }
            }
        }
        ...
    }
    ....
}

```

Note that this would require another change to AuthCredHandler::MFA to a oneshot receiver, but it means you can now have a fully isolated task with it's own state machine, logic, timeouts, and behaviour, and the pam client/resolver stack is just polling to see if that task is completed and has a decision. 

Checklist


Firstyear:
Theres also probably more opporutunity to clean up device code, as I didn't realise what you were using data for before, when really we can put that in the cred handler to reduce the pam side from having to handle it.

dmulder:
> Theres also probably more opporutunity to clean up device code, as I didn't realise what you were using data for before, when really we can put that in the cred handler to reduce the pam side from having to handle it.

Yes, I had thought of that also. That should be something I could fix after reviewing your code here.

dmulder:
This mostly works, but I'm encountering an issue:
```
error[E0596]: cannot borrow `*shutdown_rx` as mutable, as it is behind a `&` reference
   --> src/common/src/idprovider/himmelblau.rs:838:24
    |
838 |                     if shutdown_rx.try_recv().ok().is_some() {
    |                        ^^^^^^^^^^^ cannot borrow as mutable

For more information about this error, try `rustc --explain E0596`.
```
I'm doing a non-blocking `try_recv` on the `shutdown_rx` to see if a shutdown message has arrived. Apparently `shutdown_rx` needs to be mutable to do a `try_recv`?

dmulder:
Try this patch:
https://github.com/kanidm/kanidm/commit/e773074d9816837846d9883bad4f3364640e9840

Firstyear:
> This mostly works, but I'm encountering an issue:
> 
> ```
> error[E0596]: cannot borrow `*shutdown_rx` as mutable, as it is behind a `&` reference
>    --> src/common/src/idprovider/himmelblau.rs:838:24
>     |
> 838 |                     if shutdown_rx.try_recv().ok().is_some() {
>     |                        ^^^^^^^^^^^ cannot borrow as mutable
> 
> For more information about this error, try `rustc --explain E0596`.
> ```
> 
> I'm doing a non-blocking `try_recv` on the `shutdown_rx` to see if a shutdown message has arrived. Apparently `shutdown_rx` needs to be mutable to do a `try_recv`?

`let mut new_shutdown_rx = shutdown_rx.resubscribe()` will get you a new handle to the subscriber that you can then use.

Firstyear:
https://docs.rs/tokio/latest/tokio/sync/broadcast/struct.Receiver.html#method.resubscribe

dmulder:
> `let mut new_shutdown_rx = shutdown_rx.resubscribe()` will get you a new handle to the subscriber that you can then use.

Would this be problem:

> This [Receiver](https://docs.rs/tokio/latest/tokio/sync/broadcast/struct.Receiver.html) handle will receive a clone of all values sent after it has resubscribed. This will not include elements that are in the queue of the current receiver.

So, does that mean if I get the shutdown signal prior to the `resubscribe` that the signal gets lost? This would be a really short window, regardless.

dmulder:
> So, does that mean if I get the shutdown signal prior to the `resubscribe` that the signal gets lost? This would be a really short window, regardless.

Otherwise, all appears to work as expected. So, looks good to me depending on your opinion about my last comment.

dmulder (approved):
LGTM

Firstyear:
> So, does that mean if I get the shutdown signal prior to the `resubscribe` that the signal gets lost? This would be a really short window, regardless.

This is an excellent observation and question: If you look at the stack trace you'll see:

```
handle_client
pam_account_authenticate_init
unix_user_online_auth_init
```

The shutdown signal can only be generated in handle_client - since we are already in handle client and it is awaiting on the response from pam_account_authenticate_init, then it's not possible for a shutdown to be generated until unix_user_onlin_auth_init returns.

Even still, if you wish to harden this you can do:

```
let mut new_shutdown_rx = shutdown_rx.resubscribe();
if !matches(shutdown_rx.try_recv(), Err(TryRecvError::Empty)) {
    // return here since a message was in the shutdown_rx that may not be in the new_shutdown_rx
}
// After this point the new_shutdown_rx and shutdown_rx both will receive the messages. 
```

> Otherwise, all appears to work as expected. So, looks good to me depending on your opinion about my last comment.

I'm happy to merge it, did you want to do another PR to adjust AuthCredHandler to contain your needed internal state or the oneshot channel for the tasks? Or do you want me to do those now?

dmulder:
> I'm happy to merge it, did you want to do another PR to adjust AuthCredHandler to contain your needed internal state or the oneshot channel for the tasks? Or do you want me to do those now?

Only thing I'd ask for is a `max_poll_attempts` and `polling_interval` in the AuthCredHandler::MFA (both u32). I just stuffed those as strings into AuthCredHandler::MFA { data } for testing this. And I still need that strings data (which is a bunch of urls, etc).

dmulder:
Thanks for doing this!

dmulder:
> Maybe just stuff those two params into AuthCredHandler::MFA, then I think it's good to merge.

I'll just submit a separate MR afterward. I think let's just merge this as is.

dmulder:
I'll also submit another MR to add the shutdown request, etc to the DAG request.

dmulder (approved):
LGTM!

Firstyear:
> I'll also submit another MR to add the shutdown request, etc to the DAG request.

Yeah, I think with this you'll be able to move all your state inside the resolver rather than messaging it via pam. Should make things a bit easier in the end.

yaleman (approved):

PR #2676, opened 2024-03-22
ldap-sync: allow to use attrs more than once

### LINKED ISSUE #2675: ldap-sync can not use attribute twice
### I did this
Having a config for `kanidm-ldap-sync` with the following:

```
person_attr_user_name = "cn"
person_attr_display_name = "cn"
```

and it fails with:
```
Missing required attribute cn (person_attr_display_name)
```

### I expected the following
Both attrs should have the same content coming from the entities `cn`.

### Kanidm version details

- Output of `kanidm(d) version`: `kanidm 1.1.0-rc.15-dev`
- If not a container, how'd you install it: nixos
- Operating System / Version (On Unix please post the output of `uname -a`): nixos, `Linux auth 6.1.79 #1-NixOS SMP PREEMPT_DYNAMIC Fri Feb 23 08:12:53 UTC 2024 x86_64 GNU/Linux`

fooker:
Fixes #2675

Checklist

- [x] cargo fmt has been run


Cargo.toml
```diff
@@ -55,7 +55,7 @@ repository = "https://github.com/kanidm/kanidm/"
 
 # ldap3_client = { path = "../ldap3/client" }
 # ldap3_proto = { path = "../ldap3/proto" }
-# ldap3_client = { git = "https://github.com/kanidm/ldap3.git" }
+ldap3_client = { git = "https://github.com/fooker/ldap3.git", branch = "pr/get_ava" }
```
yaleman, line 58: This probably needs to be removed?

yaleman (changes requested):

Cargo.toml
fooker: I think so. But this needs to have a release of ldap3 first, right?

yaleman: Ah, didn't see you'd PR'd over there because it wasn't mentioned. We'll need to block this PR until ldap3 gets a release, in that case.

yaleman: If you can update the dep to 0.4.4 you should be good to go.

fooker: Done.

yaleman (approved):

PR #2688, opened 2024-03-29
Windows Hello Authentication requirements

dmulder:
This adds the ability in the client to set, then authenticate using a Pin code. This will be used in Himmelblau as the auth value when creating a TPM key. That TPM key is then assigned in Azure as a Windows Hello Key. Afterward, users will be able to authenticate with the Pin, which will load the Windows Hello Key from the TPM, which can be used to sign an authentication request as an Azure Entra ID user.

Checklist

- [x] cargo fmt has been run

### DIFF
--- a/unix_integration/src/resolver.rs
+++ b/unix_integration/src/resolver.rs
@@ -906,9 +906,11 @@ where
                 )
                 .await
         } else {
+            let mut dbtxn = self.db.write().await;
+
             // Can the auth proceed offline?
             self.client
-                .unix_user_offline_auth_init(account_id, token.as_ref())
+                .unix_user_offline_auth_init(account_id, token.as_ref(), &mut dbtxn)
                 .await
         };
 
@@ -1016,10 +1018,10 @@ where
             */
             (
                 &mut AuthSession::InProgress {
-                    account_id: _,
+                    ref account_id,
                     id: _,
                     token: Some(ref token),
-                    online_at_init: _,
+                    online_at_init,
                     ref mut cred_handler,
                     // Only need in online auth.
                     shutdown_rx: _,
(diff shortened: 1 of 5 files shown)


unix_integration/src/resolver.rs
```diff
@@ -1054,6 +1057,28 @@ where
                         // AuthCredHandler::MFA is invalid for offline auth
                         return Err(());
                     }
+                    (&mut AuthCredHandler::SetupPin, _) => {
+                        // AuthCredHandler::SetupPin is invalid for offline auth
+                        return Err(());
+                    }
+                    (&mut AuthCredHandler::Pin, PamAuthRequest::Pin { cred }) => {
+                        // The Pin acts as a single device password, and can be cached for offline
```
Firstyear, line 1065: Do we actually want to cache it? This uses our built-in password cache mechanism, rather than calling the backend with the TPM which should still be present and working even offline. Perhaps this is a time where we actually want to uncomment src/idprovider/interface.rs line 165. I think in the case we have an offline auth, we should be using the TPM here for validation of the PIN, rather than our baked in cache mechanism (as convenient as it is). 

Similar, it probably means when you return from unix_user_online_auth_step, if we are using a PIN then we should set AuthCacheAction::None rather than bubbling up the PIN to be cached in the resolver here.

Firstyear (commented):
Beside the possible change to how we handle pins offline, this looks excellent!

unix_integration/src/resolver.rs
dmulder: Hrm, I didn't consider this.

If we are offline though, providing the PIN would just unlock the TPM key. Ah, I see, you mean uncomment `unix_user_offline_auth_step()`? Then I could just unlock the tpm key and count that as a successful auth?

dmulder: I think my latest changes implement what you've suggested (and I like this better than caching the pin).

Firstyear: Yep, that's exactly what I meant! I think I prefer it more than caching the PIN too, because it means we keep everything nice and contained to the TPM and it's controls :)

Firstyear (approved):

PR #2699, opened 2024-04-09
20240409 rework orca markov

Firstyear:
Checklist


tools/orca/src/generate.rs
```diff
@@ -156,6 +126,45 @@ pub async fn populate(_client: &KanidmOrcaClient, profile: Profile) -> Result<St
...
+
+        for p_idx in person_index {
+            let person = persons.get_mut(p_idx).unwrap();
+
+            // Add the person to the group.
+            group.members.insert(person.username.clone());
+
+            // Add the reverse links, this allows the person in the test
+            // to know their roles
+            person.member_of.insert(group.name.clone());
```
Seba-T, line 163: Since `member_of` and `roles` are so tightly correlated (we also have the `requires_membership_to` method on `ActorRole`), wouldn't it make sense to couple them together in a single struct/enum to avoid possible missmatches? ie we only keep the `roles` field and if we want to know the groups we're in we just use `requires_memebership_to`.

Seba-T:
And also F in the chat for the Markov model 🥲 (/s)

tools/orca/src/generate.rs
Firstyear: I actually don't even think we need the "memberof" on the person tbh. I think we should remove it since ActorRole is wht defines what can be done.

Firstyear:
> And also F in the chat for the Markov model 🥲 (/s)

It can come back in future *once* we add more foundations.

tools/orca/src/generate.rs
Seba-T: Yeah that’s what I was also thinking!

Seba-T:
> > And also F in the chat for the Markov model 🥲 (/s)
> 
> It can come back in future _once_ we add more foundations.

Yeah admittedly it was a bit iffy the way I did it :/

Seba-T (approved):

Seba-T:
@yaleman I suppose gh is waiting for your review before triggering the automerge? (although I don't really know how all of this works :p)

yaleman (approved):

yaleman:
Yeeeet

yaleman:
Needs a rebase though.

PR #2713, opened 2024-04-19
Update design for KRC

Firstyear:
Checklist

- [ ] cargo fmt has been run

### DIFF
--- a/book/src/developers/designs/replication_coordinator.md
+++ b/book/src/developers/designs/replication_coordinator.md
@@ -127,7 +127,7 @@ server to be "authoritative".
 ### KRC Configuration
 
 The KRC is implied as a server node by the *lack* of a `krc_client` configuration. This informs
-the node that it must not contact other nodes for it's replication topology, and it prepares the node
+the node that it must not contact other nodes for its replication topology, and it prepares the node
 for serving that replication metadata. This is analgous to a single node operation configuration.
 
 ```
@@ -151,21 +151,21 @@ krc_url = https://kanidmd_a
 krc_ca_dir = /path/to/ca_dir
 ```
 
-The domain will automatically add a `Default Site`. The KRC implies it's own membership to "Default Site"
-and it will internally add itself to the`Default Site`
+The domain will automatically add a `Default Site`. The KRC implies its own membership to "Default Site"
+and it will internally add itself to the `Default Site`.
 
 The KRC can then issue Tokens that define which Site a new replica should join. Initially we will
 only allow `Default Site` (and will disallow creation of other sites).
 
-The new replica will load it's krc token from the environment variable `KANIDMD_KRC_TOKEN_PATH`. This
+The new replica will load its KRC token from the environment variable `KANIDMD_KRC_TOKEN_PATH`. This
 value will contain a file path where the JWT is stored. This is compatible with systemd credentials and
 docker secrets. By default the value if unset will be defined by a profile default
 (`/etc/kanidm/krc.token` or `/data/krc.token`).
 
 A new replica can then contact the `krc_url` validating the presented TLS chain with the roots
 from `krc_ca_dir` to assert the legitimacy of the KRC. Only once these are asserted, then the
-KRC token can be sent to the instance as a `Bearer` token. The new replica will also provide it's
-mTLS certificate and it's server uuid.
+KRC token can be sent to the instance as a `Bearer` token. The new replica will also provide its
+mTLS certificate and its server UUID.
 
 
 Once validated, the KRC will create or update the server's replica entry. The replica entry
@@ -190,7 +190,7 @@ If a server's entry in the database is marked to be `Revoked` then it will remai
 be inelligible for replication participation. This is to allow for forced removal of a potentially compromised node.
 
 
-The KRC will periodically examine it's RUV. For any server entry whose UUID is not contained in the RUV, and whose
+The KRC will periodically examine its RUV. For any server entry whose UUID is not contained in the RUV, and whose
 "time first seen + trime window" is less than now, then the server entry will be REMOVED for inactivity
 since it has now been trimmed from the RUV.
 
@@ -228,7 +228,7 @@ Imagine the following example. Here, Node A is acting as the KRC.
 This would allow Node A to be aware of B, C, D and then create a full mesh.
 
 We wish to decommision Node A and promote Node B to become the new KRC. Imagine at this point
-we cut over Node D to point it's KRC at Node B.
+we cut over Node D to point its KRC at Node B.
 
 ```
 ┌─────────────────┐                ┌─────────────────┐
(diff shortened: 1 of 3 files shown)


Firstyear:
Updated design, going to leave it to cook over the weekend, then I'll impl it next week.

book/src/developers/designs/replication_coordinator.md
```diff
@@ -133,160 +106,217 @@ kanidmd replication get-certificate
 For each node, a replication configuration is created in json. For A pulling from B.
```
yaleman, line 106: ^ `kanidmd replication get-certificate` should be formatted to show as a shell command

book/src/developers/designs/replication_coordinator.md
```diff
@@ -133,160 +106,217 @@ kanidmd replication get-certificate
...
+bindaddress = "[::]:8444"
+
+# krc_url = https://private.name.of.krc.node
+krc_url = https://kanidmd_a
+# must contain ca that signs kanidmd_a's tls_chain.
+krc_ca_dir = /path/to/ca_dir
 ```
 
-The KRC is also configured with replication groups.
+The domain will automatically add a `Default Site`. The KRC implies it's own membership to "Default Site"
```
yaleman, line 154:
```suggestion
The domain will automatically add a `Default Site`. The KRC implies its own membership to "Default Site"
```

book/src/developers/designs/replication_coordinator.md
```diff
@@ -133,160 +106,217 @@ kanidmd replication get-certificate
...
+and it will internally add itself to the`Default Site`
```
yaleman, line 155:
```suggestion
and it will internally add itself to the `Default Site`.
```

book/src/developers/designs/replication_coordinator.md
```diff
@@ -133,160 +106,217 @@ kanidmd replication get-certificate
...
+
+The KRC can then issue Tokens that define which Site a new replica should join. Initially we will
+only allow `Default Site` (and will disallow creation of other sites).
+
+The new replica will load it's krc token from the environment variable `KANIDMD_KRC_TOKEN_PATH`. This
```
yaleman, line 160:
```suggestion
The new replica will load its KRC token from the environment variable `KANIDMD_KRC_TOKEN_PATH`. This
```

book/src/developers/designs/replication_coordinator.md
```diff
@@ -133,160 +106,217 @@ kanidmd replication get-certificate
...
+value will contain a file path where the JWT is stored. This is compatible with systemd credentials and
+docker secrets. By default the value if unset will be defined by a profile default
+(`/etc/kanidm/krc.token` or `/data/krc.token`).
+
+A new replica can then contact the `krc_url` validating the presented TLS chain with the roots
+from `krc_ca_dir` to assert the legitimacy of the KRC. Only once these are asserted, then the
+KRC token can be sent to the instance as a `Bearer` token. The new replica will also provide it's
```
yaleman, line 167:
```suggestion
KRC token can be sent to the instance as a `Bearer` token. The new replica will also provide its
```

book/src/developers/designs/replication_coordinator.md
```diff
@@ -133,160 +106,217 @@ kanidmd replication get-certificate
...
+mTLS certificate and it's server uuid.
```
yaleman, line 168:
```suggestion
mTLS certificate and its server UUID.
```

book/src/developers/designs/replication_coordinator.md
```diff
@@ -133,160 +106,217 @@ kanidmd replication get-certificate
...
+in the database will contain the active mTLS cert of the replica and a reference to the replication
+site that the token referenced.
+
+This will additionally add the "time first seen" to the server entry.
+
+From this, for each server in the replication site associated to the token, the KRC will provide
+a replication config map to the new replica providing all URL's and mTLS certs.
+
+
+Anytime the replica checks in, if the KRC replication map has changed a new one will be provided,
```
yaleman, line 181: Why not just return the "current" map every time with an `ETag` header (usually used to identify changes), and the client can decide what to do with it... then if it reboots or loses its copy, it can parse and update its internal representation of the state.

This means it's simpler on the KRC side (not tracking who has what) and the client *always* looks at the replication config and goes "hey is this the same as mine".

Generation IDs make sense just from a change tracking perspective on the human/audit side, but there's no loss in returning a relatively small set of data periodically to KRC clients.

book/src/developers/designs/replication_coordinator.md
```diff
@@ -133,160 +106,217 @@ kanidmd replication get-certificate
...
+To determine no changes we use a "generation". This is where any change to a replication site or
+server entries will increment the generation counter. This allows us to detect when a client requires
+a new configuration or not.
+
+
+If a server's entry in the database is marked to be `Revoked` then it will remain in the database, but
+be inelligible for replication participation. This is to allow for forced removal of a potentially compromised node.
+
+
+The KRC will periodically examine it's RUV. For any server entry whose UUID is not contained in the RUV, and whose
```
yaleman, line 193:
```suggestion
The KRC will periodically examine its RUV. For any server entry whose UUID is not contained in the RUV, and whose
```

book/src/developers/designs/replication_coordinator.md
```diff
@@ -133,160 +106,217 @@ kanidmd replication get-certificate
...
+│                 │                │                 │
+│                 │                │                 │
+└─────────────────┘                └─────────────────┘
 ```
 
-This would yield the following arrangement.
+This would allow Node A to be aware of B, C, D and then create a full mesh.
+
+We wish to decommision Node A and promote Node B to become the new KRC. Imagine at this point
+we cut over Node D to point it's KRC at Node B.
```
yaleman, line 231:
```suggestion
we cut over Node D to point its KRC at Node B.
```

book/src/developers/designs/replication_coordinator.md
```diff
@@ -133,160 +106,217 @@ kanidmd replication get-certificate
...
+│                 │                │                 │
+│                 │                │                 │
+│     Node C      │                │     Node D      │
+│                 │                │                 │
+│                 │                │                 │
+└─────────────────┘                └─────────────────┘
 ```
 
-!!! TBD - How to remove / decomission nodes?
+Since we still have the Server Entry records in the Default Site on bothe Node A and Node B, then
```
yaleman, line 257:
```suggestion
Since we still have the Server Entry records in the Default Site on both Node A and Node B, then
```

yaleman (changes requested):

book/src/developers/designs/replication_coordinator.md
Firstyear: If the client has lost it's replication map it'll need to be refreshed and reset anyway in this case because it's then ambiguous if there is other dataloss.

yaleman: Yeah, I'm just thinking it'll be simpler on the server side to make/cache a config and return that every time, less code more betterer

Firstyear: Yeah, but network bandwidth also isn't infinite in replication and bigger topologies means we're sending big config maps all the time. The other part is that at some point the client or the server need to choose "when do I need to change and reload my config". I can see it going either way, it's not set in stone really. I just think it's easier for the server to calc that change since it processes all the changes to replication topology, and can work out "okay here is the info you need" rather than the client.

yaleman: 'k

yaleman (approved):

PR #2714, opened 2024-04-19
Add session limit

### LINKED ISSUE #2693: Excessive session creation makes replicated secondary instances' memory blow up
### I did this

I have a replicated (active-passive, two nodes) kani setup. My custom app used a service account (username + password) to regularly (every 10 minutes per env of which I currently have 3) contact the kani api to ask about things. I used the init session path instead of api tokens. I also properly logout after every contact, however, I don't reuse sessions, but create a new one every time.

### I expected the following

I expected kani to cleanup revoked sessions and it does, but there's a problematic side effect that (the following is my understanding) for revoked sessions there's at least a tombstone (so the information that the session is gone) replicated to other instances and these are not being "garbage collected" (in the same interval or ever) for reasons.

### Kanidm version details

Official rc.16 container running on latest docker on Debian 11.

### Any other comments

image

The blue instance is the passive replica and you see, as described before, the primary has a regular cleanup and stays alive while the secondary hits the system limit at some point and gets killed by the oom killer. This causes a restart and the same thing repeats over and over.

I implemented the use of api tokens now and will observe if this solves the issue on my end.

### Workaround

Don't use sessions for such things. Make use of api tokens (my choice) or at least recycle your sessions.

Firstyear:
Fixes #2693 - Add a limit to sessions that a human would not hit, but prevents rapid or unbounded expansion of the value. 

Checklist

### DIFF
--- a/server/lib/src/valueset/session.rs
+++ b/server/lib/src/valueset/session.rs
@@ -22,6 +22,8 @@ pub struct ValueSetSession {
     map: BTreeMap<Uuid, Session>,
 }
 
+const SESSION_MAXIMUM: usize = 48;
+
 impl ValueSetSession {
     pub fn new(u: Uuid, m: Session) -> Box<Self> {
         let mut map = BTreeMap::new();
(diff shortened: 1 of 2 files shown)


server/lib/src/valueset/session.rs
```diff
@@ -425,7 +427,38 @@ impl ValueSetT for ValueSetSession {
...
+            // At this point we will force a number of sessions to be removed. This
+            // is replication safe since other replicas will also be performing
+            // the same operation on merge, since we trim by session issuance order.
+
+            // This is a "slow path". This is becase we optimise session storage
+            // based on fast session lookup, so now we need to actually create an
+            // index based on time. We need to also clone here since we need to mutate
+            // self.map which would violate mut/imut.
+
+            warn!(
```
yaleman, line 443: This warning should include the entry's ID, so it can be found while troubleshooting.

server/lib/src/valueset/session.rs
```diff
@@ -22,6 +22,8 @@ pub struct ValueSetSession {
     map: BTreeMap<Uuid, Session>,
 }
 
+const SESSION_MAXIMUM: usize = 48;
```
yaleman, line 25: This should really be in `kanidmd_lib::constants` with its friends 😄

yaleman (changes requested):

server/lib/src/valueset/session.rs
Firstyear: It does, about 10 lines below :)

yaleman: Does it though? It shows the session ID that's being trimmed, not the account

Firstyear: At that point in the code, we can't see the account, because we're in the valueset. :( let me check if the caller can log what it's doing ....

yaleman (approved):

PR #2773, opened 2024-05-16
Allow name write privileges to be withheld

### LINKED ISSUE #2772: Option to disable self managing displayname, name, and legal name
### Discussed in https://github.com/kanidm/kanidm/discussions/2770

<div type='discussions-op-text'>

<sup>Originally posted by **scaredmushroom** May 15, 2024</sup>
We have the need that usernames have to be in a specific format and also not be changed w/o approval.

Reasons include that we want to circumvented that students choose unprofessional names ("trolling") and be recognizable to other students and tutors during exercises (especially while working on lab servers – UUIDs are rather impractical here for groupwork).

Hence I wanted to ask if it is somehow possible to disable self managing displaynames, names, and legal names, so that they can be changed on request? If not I also would be thankful for any ideas on how to deal with this problem.</div>

Firstyear:
Fixes #2772 - Allow self write name privileges to be withheld. This is achieved by moving the legal name and display name write to the self name write acp, and adding a group that enables or removes this access. This allows admins to remove idm_all_persons from this group to disable the ability for users to self modify their name related values.

I don't really like this feature but I see it's importance for schools and some businesses. I don't want it to be the default, but I also recognise that it's needed in some environments. 

Checklist


book/src/access_control/intro.md
```diff
@@ -69,6 +70,7 @@ tasks within Kanidm as required.
 | `idm_people_on_boarding`     | create (but not modify) persons. Intended for use with service accounts |
 | `idm_people_pii_read`        | allow read to personally identifying information                        |
 | `idm_people_self_write_mail` | allow self-modification of the mail attribute                           |
+| `idm_people_self_name_write` | allow self-modification of the name related attributes                  |
```
yaleman, line 73: We've got different naming conventions here... I can't live with that inconsistency 😄 

Should be `idm_people_self_write_name` given the current naming ... then again we've got `pii_read` ... so I kinda think renaming `write_mail` is better than renaming `name_write` 🤣

yaleman (changes requested):

book/src/access_control/intro.md
Firstyear: We can rename them for free basicly, it doesn't break anything. I'll change them to be consistent. I'd say we should make the pattern "idm_<class>_<target>_<action>". So that would make it "idm_people_read_pii", "idm_people_self_name_write", "idm_people_self_mail_write". What do you think? Alternately, swap target and action and we just update "idm_people_self_mail_write"

yaleman: Yep, `idm_<target>_<verb>` where `<target>` narrows from largest scope to smallest, so `self_mail` or `all_persons_name` etc. Probably should document that, too.

yaleman: `idm_people_pii_read` though 😄

yaleman (approved):

yaleman (approved):

PR #2779, opened 2024-05-18
2756 - resolve invalid loading of dyngroups at startup

### LINKED ISSUE #2756: New person accounts not being added to dynamic groups
### I did this

`kanidm person create someuser someuser`

### I expected the following

New user would be automatically added to the dynamic groups `idm_all_accounts` and `idm_all_persons` as was the case with previous versions.

### Kanidm version details

- Output of `kanidm(d) version`: 1.2.0
- Are you running it in a container? If so, which image/tag?: N/A
- If not a container, how'd you install it: NixOS module
- Operating System / Version (On Unix please post the output of `uname -a`): Linux 6.6.29 1-NixOS SMP PREEMPT_DYNAMIC Sat Apr 27 15:11:44 UTC 2024 x86_64 GNU/Linux

Firstyear:
Fixes #2756 - dyngroups were not loaded correctly after a restart unless migrations were triggered. This was not found on dev builds as they have a hook to force a re-migration each startup.

This will need some additional handlers to support automating a fix now we just yeeted the sysinfo versions. 

Checklist


server/lib/src/server/migrations.rs
```diff
@@ -127,12 +142,26 @@ impl QueryServer {
...
             if option_env!("KANIDM_PRE_RELEASE").is_some() && !cfg!(test) {
                 write_txn.domain_remigrate(DOMAIN_PREVIOUS_TGT_LEVEL)?;
+                write_txn.reload()?;
             }
         }
 
-        // Reload if anything in migrations requires it - this triggers the domain migrations
-        // which in turn can trigger schema reloads etc.
-        write_txn.reload()?;
+        if domain_target_level >= DOMAIN_LEVEL_7 && domain_patch_level < DOMAIN_TGT_PATCH_LEVEL {
```
yaleman, line 149: If this is a one-shot thing, surely we'd match on "equals DL7" not ">= DL7" because it'll trigger every major version bump.. or are we going to be bumping the minors going forward as well? Seems.. risky.

Firstyear: It's so that in tests DL5 -> DL6 doesn't trigger because it doesn't have this yet.

Think of this flag as like the former "system info version". It's just a monotonic counter that we bump if we have a one-shot fixup we need to apply. It's a rare and exceptional case we have to apply it, but when we do, we need a way to have a fixup run *without* bumping the domain level like we do on a major upgrade (1.2 to 1.3 for example).

So for example, in the example here with dyngroups, lets say we have version 1.3.0 released to users. We found this issue and we need to trigger the fixup. We can't bump the domain level again, that's for 1.3.0 to 1.4.0. But we need a way to do a oneshot task. That's the gap this fills, so that within the 1.3.0 series we can run this once-off task, but also that it *fences* replication between 1.3.0 nodes that aren't on the same fixup level.

yaleman: If they come in at DL8 already then it'll bypass this because they're already in the 7-8 migration, that's the bit I wasn't thinking of... derp.

yaleman (approved):

PR #2780, opened 2024-05-19
feat: add support for ldap compare request

tobikris:
This adds support for the LDAP CompareRequest. The implementation is ~~using~~ based on the existing `Search` operation.

I tried to follow this description of the request: https://ldap.com/the-ldap-compare-operation/.
~~As the `Search` operation is not exposing the reason for errors, I could not return `invalidDNSyntax`.~~
I decided on using the same authentication decision as for the `Search`.

Edit:
After the review, I changed the implementation to be standalone with copy-pasted code from `do_search` instead of calling the method.

Checklist

- [x] cargo fmt has been run

### DIFF
--- a/server/lib/src/idm/ldap.rs
+++ b/server/lib/src/idm/ldap.rs
@@ -477,6 +477,123 @@ impl LdapServer {
         Ok(result)
     }
 
+    #[instrument(level = "debug", skip_all)]
+    async fn do_compare(
+        &self,
+        idms: &IdmServer,
+        cr: &CompareRequest,
+        uat: &LdapBoundToken,
+        source: Source,
+    ) -> Result<Vec<LdapMsg>, OperationError> {
+        admin_info!("Attempt LDAP CompareRequest for {}", uat.spn);
+
+        let (opt_attr, opt_value) = match self.dnre.captures(cr.entry.as_str()) {
+            Some(caps) => (
+                caps.name("attr").map(|v| v.as_str().to_string()),
+                caps.name("val").map(|v| v.as_str().to_string()),
+            ),
+            None => {
+                request_error!("LDAP Search failure - invalid basedn");
+                return Err(OperationError::InvalidRequestState);
+            }
+        };
+
+        let ext_filter = match (opt_attr, opt_value) {
+            (Some(a), Some(v)) => LdapFilter::Equality(a, v),
+            _ => {
+                request_error!("LDAP Search failure - invalid rdn");
+                return Err(OperationError::InvalidRequestState);
+            }
+        };
+
+        let ct = duration_from_epoch_now();
+        let mut idm_read = idms.proxy_read().await;
+        // Now start the txn - we need it for resolving filter components.
+
+        // join the filter, with ext_filter
+        let lfilter = LdapFilter::And(vec![
+            ext_filter.clone(),
+            LdapFilter::Equality(cr.atype.clone(), cr.val.clone()),
+            LdapFilter::Not(Box::new(LdapFilter::Or(vec![
+                LdapFilter::Equality(Attribute::Class.to_string(), "classtype".to_string()),
+                LdapFilter::Equality(Attribute::Class.to_string(), "attributetype".to_string()),
+                LdapFilter::Equality(
+                    Attribute::Class.to_string(),
+                    "access_control_profile".to_string(),
+                ),
+            ]))),
+        ]);
+
+        admin_info!(filter = ?lfilter, "LDAP Compare Filter");
+
+        // Build the event, with the permissions from effective_session
+        let ident = idm_read
+            .validate_ldap_session(&uat.effective_session, source, ct)
+            .map_err(|e| {
+                admin_error!("Invalid identity: {:?}", e);
+                e
+            })?;
+
+        let f = Filter::from_ldap_ro(&ident, &lfilter, &mut idm_read.qs_read)?;
+        let filter_orig = f
+            .validate(idm_read.qs_read.get_schema())
+            .map_err(OperationError::SchemaViolation)?;
+        let filter = filter_orig.clone().into_ignore_hidden();
+
+        let ee = ExistsEvent {
+            ident: ident.clone(),
+            filter,
+            filter_orig,
+        };
+
+        let res = idm_read.qs_read.exists(&ee).map_err(|e| {
+            admin_error!("call to exists failure {:?}", e);
+            e
+        })?;
+
+        if res {
+            admin_info!("LDAP Compare -> True");
+            return Ok(vec![cr.gen_compare_true()]);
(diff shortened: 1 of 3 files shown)


server/lib/src/idm/ldap.rs
```diff
@@ -477,6 +478,94 @@ impl LdapServer {
...
 
+    #[instrument(level = "debug", skip_all)]
+    async fn do_compare(
+        &self,
+        _idms: &IdmServer,
+        cr: &CompareRequest,
+        uat: &LdapBoundToken,
+        _source: Source,
+    ) -> Result<Vec<LdapMsg>, OperationError> {
+        admin_info!("Attempt LDAP CompareRequest for {}", uat.spn);
```
yaleman, line 489: this should probably be a trace/debug level message

tobikris: I can change it of course - but I just replicated the search method: https://github.com/kanidm/kanidm/pull/2780/files/a3e46270e7cba16242619c10505983be01b9eee0#diff-3a38c3a25731778180d0864ffe0f2987822f93891340f37a61d05d4207afc1e7R148

What do you prefer for me to do?

yaleman: m'bad, I went looking for a similar message and didn't find it (was on my phone 😇) leave it as-is, thanks.

server/lib/src/idm/ldap.rs
```diff
@@ -536,9 +625,39 @@ impl LdapServer {
...
+                Some(u) => self
+                    .do_compare(idms, &cr, &u, source)
+                    .await
+                    .map(LdapResponseState::MultiPartResponse)
+                    .or_else(|e| {
+                        let (rc, msg) = operationerr_to_ldapresultcode(e);
+                        Ok(LdapResponseState::Respond(cr.gen_error(rc, msg)))
+                    }),
+                None => {
+                    // Compare can occur without a bind, so bind first.
```
Firstyear, line 638: Might be worth adding a comment here that this is per section 4 of rfc 4513 ( https://www.rfc-editor.org/rfc/rfc4513#section-4 )

server/lib/src/idm/ldap.rs
```diff
@@ -477,6 +478,94 @@ impl LdapServer {
...
+                    code,
+                    matcheddn: "".to_string(),
+                    message: "".to_string(),
+                    referral: vec![],
+                }),
+                ctrl: vec![],
+            }
+        }
+
+        let res = match self.do_search(_idms, &sr, uat, Source::Internal).await {
```
Firstyear, line 515: We should be passing through the Sourch from line 487 here.

server/lib/src/idm/ldap.rs
```diff
@@ -477,6 +478,94 @@ impl LdapServer {
...
+                    if let LdapOp::SearchResultDone(LdapResult {
+                        code: LdapResultCode::Success,
+                        ..
+                    }) = &srd.op
+                    {
+                        return Ok(vec![compare_result_msg(
+                            cr,
+                            match results.len() {
+                                0 => LdapResultCode::NoSuchObject, // entry really does not exist
+                                _ => LdapResultCode::CompareFalse, // entry exists, but condition is not met
```
Firstyear, line 558: Well it could also be that the caller lacks permissions to perform the search.

server/lib/src/idm/ldap.rs
```diff
@@ -477,6 +478,94 @@ impl LdapServer {
...
+            {
+                if !results.is_empty() {
+                    return Ok(vec![compare_result_msg(cr, LdapResultCode::CompareTrue)]);
+                }
+
+                // search again to determine if entry exists at all
+                let mut sr = sr;
+                sr.filter = LdapFilter::Present(Attribute::Spn.to_string());
+
+                let res = self.do_search(_idms, &sr, uat, Source::Internal).await?;
```
Firstyear, line 546: The problem with calling this search twice is that they are done in two separate transactions, so it's possible the entry changes from the first search to the second. As a result, we can't just make do_compare a wrapper over do_search *unless* there is a way to do it in a single search. Most likely we need to "duplicate" do_search and strip out anything that isn't needed.

Firstyear (changes requested):
Thanks for this. There are quite a few comments but I hope we can help out get this to a state we can merge it.

server/lib/src/idm/ldap.rs
tobikris: Good catch. Do you think it makes sense to duplicate this or shall I just remove the not strictly necessary second search and just return `CompareTrue` vs `CompareFalse` based on the first one?

Firstyear: The problem then is you can't distinguish a compareFalse from no such entry. I think it needs to actually be a dedicated operation like do_search, we can't just layer it on top. If that's too much, I'm happy to edit the PR and fix that :)

tobikris: Alright, I will try to implement it properly. Thanks for your help 😊

Firstyear: You should be able to "copy paste" the insides of do_search, and then strip out all the parts excepting base-dn searches. We could probably even refactor a few parts to be common. 

Remember also the internal query server has both search (which returns entries) and exists which gives a true or false if the condition is met. I think you want the latter :)

server/lib/src/idm/ldap.rs
tobikris: added a comment to both operations.

server/lib/src/idm/ldap.rs
tobikris: I am not sure how this should be handled - can you lead me into the right direction, please?

server/lib/src/idm/ldap.rs
tobikris: I have changed the implementation accordingly. I removed quite a lot of code after the copying and I think there is not much deduplication to be achieved - but I can try. Please first let me know if you think the current implementation is correct and I will continue from there. Thanks!

Firstyear (approved):
I think you got this spot on, great work :)

server/lib/src/idm/ldap.rs
Firstyear: Can you also add yourself to contributors.md in the root of the project if you wish? :)

tobikris: Sure, added myself.

Firstyear:
This may need a rebase to latest master before we can merge but otherwise it looks great, thank you!

Firstyear (approved):

tobikris:
> This may need a rebase to latest master before we can merge but otherwise it looks great, thank you!

I rebased, but this reset the auto-merge status.

PR #2791, opened 2024-05-23
WIP: serialization and domain info setting wonkiness

yaleman:
So I started by troubleshooting this:

```shell
$ kanidm system domain set-ldap-allow-unix-password-bind -D idm_admin false 
2024-05-22T08:54:53.594833Z ERROR kanidm_cli: HTTP Error: 422 Unprocessable Entity "f6352010-51a4-4814-b9ce-ab076a455223"
```

That led to an issue with serialization of boolean values passed through a put request. Fun!

Then I found we had `json_rest_event_put_id_attr` and `json_rest_event_put_attr`, so I cleaned that up.

Now I found that testkit was resetting the admin password twice on startup so that's now `admin` and `idm_admin`...

Finally got back to troubleshooting the first issue and... modifying the domain attribute `ldap_allow_unix_pw_bind` doesn't work as either `admin` or `idm_admin`
- when idm_admin tries to modify the domain , they're blocked from searching for it - `404, NoMatchingEntries`
- when admin tries, it throws `403, SystemProtectedObject`

Checklist

- [x] cargo fmt has been run

### DIFF
--- a/server/core/src/https/v1.rs
+++ b/server/core/src/https/v1.rs
@@ -344,6 +344,15 @@ pub async fn json_rest_event_post_id_attr(
         .map_err(WebError::from)
 }
 
+// Okay, so a put normally needs
+///  * filter of what we are working on (id + class)
+///  * a `Map<String, Vec<String>>` that we turn into a modlist.
+///
+/// OR
+///  * filter of what we are working on (id + class)
+///  * a `Vec<String>` that we are changing
+///  * the attr name  (as a param to this in path)
+///
 pub async fn json_rest_event_put_attr(
     state: ServerState,
     id: String,
@@ -376,27 +385,6 @@ pub async fn json_rest_event_post_attr(
         .await
         .map(Json::from)
         .map_err(WebError::from)
-}
-
-// Okay, so a put normally needs
-///  * filter of what we are working on (id + class)
-///  * a `Map<String, Vec<String>>` that we turn into a modlist.
-///
-/// OR
-///  * filter of what we are working on (id + class)
-///  * a `Vec<String>` that we are changing
-///  * the attr name  (as a param to this in path)
-///
-pub async fn json_rest_event_put_id_attr(
-    state: ServerState,
-    id: String,
-    attr: String,
-    filter: Filter<FilterInvalid>,
-    values: Vec<String>,
-    kopid: KOpId,
-    client_auth_info: ClientAuthInfo,
-) -> Result<Json<()>, WebError> {
-    json_rest_event_put_attr(state, id, attr, filter, values, kopid, client_auth_info).await
 }
 
 pub async fn json_rest_event_delete_id_attr(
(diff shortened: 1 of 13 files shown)


yaleman:
@Firstyear I can't track down who is supposed to have access edit this domain attribute or why it's being a jerk 😡

server/core/src/https/v1.rs
```diff
@@ -2423,9 +2411,24 @@ pub async fn domain_attr_put(
...
+    let values: Vec<String> = values
+        .into_iter()
+        .map(|v| {
+            #[allow(clippy::expect_used)]
+            if v.is_boolean() {
+                v.as_bool()
+                    .map(|b| b.to_string())
+                    .expect("Failed to un-wrap thing we already identified as a boolean!")
+            } else {
+                v.as_str().unwrap().to_string()
```
Firstyear, line 2427: I'm not sure I like this, what's the intent here? It should work with strings because we convert in a lower stage from str to bool.

Firstyear (changes requested):

Firstyear:
https://github.com/kanidm/kanidm/blob/master/server/lib/src/constants/acp.rs#L884

Should be who has the access. This ties back to https://github.com/kanidm/kanidm/blob/master/server/lib/src/constants/groups.rs#L122 which admin is a default member of.

server/core/src/https/v1.rs
yaleman: Yeah I'm not a huge fan either. At some point we have to turn it into a string, but if we don't do it here, then it throws a deser issue on the receiever because it's expecting `Vec<String>` not `Vec<string or bool or other rando things>`

None of this was being caught because all the current tests only send string-type-values for attributes.

Firstyear: JSON is "unambiguous" so this means if we have a bool value, that the client didn't str encode it. https://docs.rs/serde_json/latest/serde_json/value/enum.Value.html#method.is_boolean

So that means the issue is actually:

https://github.com/kanidm/kanidm/blob/master/libs/client/src/lib.rs#L1944

Which should have enable.to_string() instead of vec![enable].

yaleman: yeah, I was trying to take _**actual JSON"**_ 'cause that's nicer from the client side. oh wells.

yaleman:
> https://github.com/kanidm/kanidm/blob/master/server/lib/src/constants/acp.rs#L884
> 
> Should be who has the access. This ties back to https://github.com/kanidm/kanidm/blob/master/server/lib/src/constants/groups.rs#L122 which admin is a default member of.

I found the real issue - there's a list of "protected" attributes which didn't include the LDAP one. Cleaned that up now - is here [server/lib/src/plugins/protected.rs#L28](https://github.com/yaleman/kanidm/blob/ffb4dd3a6c30e3a9d4d1e240862d3283df5cbd6b/server/lib/src/plugins/protected.rs#L28)

ffb4dd3 also cleans plugins/protected a bit so we only have to define the some entryclass things once.

Firstyear (approved):

PR #2795, opened 2024-05-24
Resolve incorrect handling of tokens in logout flow

### LINKED ISSUE #2766: Can't sign out properly
### Discussed in https://github.com/kanidm/kanidm/discussions/2760

<div type='discussions-op-text'>

<sup>Originally posted by **MrSpock** May 11, 2024</sup>
Hi,
I'm evaluating kanidm and I stuck with such basic stuff like logout.
When I login to system and click Sign Out I'm getting following log error and web message:
`kanidmd  | 76071ae1-1f1b-40b6-a9f8-d1d2778420b0 INFO     request [ 4.50ms | 9.87% / 100.00% ] method: GET | uri: /v1/logout | version: HTTP/1.1
kanidmd  | 76071ae1-1f1b-40b6-a9f8-d1d2778420b0 INFO     ┝━ handle_logout [ 4.06ms | 22.44% / 90.13% ]
kanidmd  | 76071ae1-1f1b-40b6-a9f8-d1d2778420b0 INFO     │  ┝━ validate_client_auth_info_to_ident [ 3.05ms | 67.69% ]
kanidmd  | 76071ae1-1f1b-40b6-a9f8-d1d2778420b0 INFO     │  │  ┕━ ｉ [info]: A valid limited session value exists for this token | event_tag_id: 10
kanidmd  | 76071ae1-1f1b-40b6-a9f8-d1d2778420b0 INFO     │  ┝━ ｉ [info]: modify initiator | event_tag_id: 10 | name: User( spock@*, fc4a6f61-a843-4dbf-bf27-cfc91b0ec0f6 ) (d8ebf915-0b57-4439-b3d9-ea757cf7585e, read write)
kanidmd  | 76071ae1-1f1b-40b6-a9f8-d1d2778420b0 INFO     │  ┝━ ｉ [info]: search | event_tag_id: 10 | initiator: User( spock@idm*, fc4a6f61-a843-4dbf-bf27-cfc91b0ec0f6 ) (d8ebf915-0b57-4439-b3d9-ea757cf7585e, read write)
kanidmd  | 76071ae1-1f1b-40b6-a9f8-d1d2778420b0 INFO     │  ┝━ ｉ [info]: denied ❌ - no entries were released | event_tag_id: 11
kanidmd  | 76071ae1-1f1b-40b6-a9f8-d1d2778420b0 ERROR    │  ┝━ 🚨 [error]: modify: no candidates match filter, failure Filter(Valid) (not ( (class eq Iutf8("tombstone") or class eq Iutf8("recycled")) ) and (uuid eq Uuid(fc4a6f61-a843-4dbf-bf27-cfc91b0ec0f6) and user_auth_token_session eq Refer(d8ebf915-0b57-4439-b3d9-ea757cf7585e)))
kanidmd  |  | event_tag_id: 4
kanidmd  | 76071ae1-1f1b-40b6-a9f8-d1d2778420b0 ERROR    │  ┕━ 🚨 [error]: Failed to destroy user auth token NoMatchingEntries | event_tag_id: 1
kanidmd  | 76071ae1-1f1b-40b6-a9f8-d1d2778420b0 WARN     ┕━ 🚧 [warn]:  | latency: 4.545756ms | status_code: 404 | kopid: "76071ae1-1f1b-40b6-a9f8-d1d2778420b0" | msg: "client error"`
kanidm-signout

What Am I doing wrong ? :)</div>

I wonder if in this case since the auth token is within the grace window, and since it's being used to logout, we should just write a session stub with the revoked state which will cause the async write back to be dropped? It would resolve this error.

Firstyear:
Fixes #2766 - Fixes invalid handling of tokens in the logout flow. This was an oversight on my part when adding multi-instance handling. 

Checklist

### DIFF
--- a/tools/cli/src/cli/session.rs
+++ b/tools/cli/src/cli/session.rs
@@ -641,17 +641,20 @@ impl LogoutOpt {
                 Ok(j) => j,
                 Err(err) => {
                     error!(?err, "Unable to parse token");
+                    info!("The token can be removed locally with `--local-only`");
                     std::process::exit(1);
                 }
             };
 
             let Some(key_id) = jwsc.kid() else {
                 error!("Invalid token, missing KeyID");
+                info!("The token can be removed locally with `--local-only`");
                 std::process::exit(1);
             };
 
             let Some(pub_jwk) = token_instance.keys().get(key_id) else {
                 error!("Invalid instance, no signing keys are available");
+                info!("The token can be removed locally with `--local-only`");
                 std::process::exit(1);
             };
 
@@ -659,6 +662,7 @@ impl LogoutOpt {
                 Ok(verifier) => verifier,
                 Err(err) => {
                     error!(?err, "Unable to configure jws verifier");
+                    info!("The token can be removed locally with `--local-only`");
                     std::process::exit(1);
                 }
             };
@@ -666,12 +670,14 @@ impl LogoutOpt {
             let uat = match jws_verifier.verify(&jwsc).and_then(|jws| {
                 jws.from_json::<UserAuthToken>().map_err(|serde_err| {
                     error!(?serde_err);
+                    info!("The token can be removed locally with `--local-only`");
                     JwtError::InvalidJwt
                 })
             }) {
                 Ok(uat) => uat,
                 Err(e) => {
                     error!(?e, "Unable to verify token signature, may be corrupt");
+                    info!("The token can be removed locally with `--local-only`");
                     std::process::exit(1);
                 }
             };
--- a/tools/cli/src/cli/session.rs
+++ b/tools/cli/src/cli/session.rs
@@ -589,7 +589,17 @@ impl LogoutOpt {
     }
 
     pub async fn exec(&self) {
+        let mut tokens = read_tokens(&self.copt.get_token_cache_path()).unwrap_or_else(|_| {
+            error!("Error retrieving authentication token store");
+            std::process::exit(1);
+        });
+
         let instance_name = &self.copt.instance;
+        let n_lookup = instance_name.clone().unwrap_or_default();
+        let Some(token_instance) = tokens.instances.get_mut(&n_lookup) else {
+            println!("No sessions for instance {}", n_lookup);
+            return;
+        };
 
         let spn: String = if self.local_only {
             // For now we just remove this from the token store.
@@ -624,15 +634,9 @@ impl LogoutOpt {
                     std::process::exit(1);
                 }
             };
-            // Parse it for the username.
 
-            if let Err(e) = client.logout().await {
-                error!("Failed to logout - {:?}", e);
-                std::process::exit(1);
-            }
-
-            // Server acked the logout, lets proceed with the local cleanup now.
-
+            // Parse it for the SPN. Annoying but it's what we have to do
+            // because we don't know what token was used in the lower to client calls.
             let jwsc = match JwsCompact::from_str(&token) {
                 Ok(j) => j,
                 Err(err) => {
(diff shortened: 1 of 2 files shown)


Firstyear:
Will need to be cherry-picked into 1.2, hopefully won't cause too much issue with a 1.2.2 ....

tools/cli/src/cli/session.rs
```diff
@@ -641,19 +645,24 @@ impl LogoutOpt {
...
-                    Ok(verifier) => verifier,
-                    Err(err) => {
-                        error!(?err, "Unable to configure jws verifier");
-                        std::process::exit(1);
-                    }
-                }
-            } else {
-                error!("Unable to access token public key");
+            let Some(key_id) = jwsc.kid() else {
+                error!("Invalid token, missing KeyID");
```
yaleman, line 649: Since this is still parsing the local token, if it's invalid why not just log a debug message and remove it from the local store? (same with the rest of this)

Firstyear: Yeah, I ummed and ahhed a lot about this. The problem is that we need the *spn* to actually ID the token out the store. The spn is collected down in to_client(), and by this part of the process we *must* have done a successful to_client() which means:

* The token *already* has to have passed signature checks for to_client to have worked: https://github.com/kanidm/kanidm/blob/master/tools/cli/src/cli/common.rs#L213
* If logout() succeeded that means the token was *also* validated.

So by this point, unless something bonkers happens, then the validation will always pass.

I did consider using ReleaseNoVerify here, but I think the above meant it was "safe" by that point. 

Really, this is a bit of a tough one layering wise, because we have to handle token state in the client *and* in the cli itself due to the token store. Tech Debt, yay. 

Anywy, I think an addition that could be wise here is removing https://github.com/kanidm/kanidm/blob/master/tools/cli/src/opt/kanidm.rs#L785 hidden from the local_only flag, allowing the token to be removed *always* bypassing all these other checks.

yaleman: > So by this point, unless something bonkers happens, then the validation will always pass.

so why error out and leave a broken token in the local cache at that point? just break out of the parsing, remove it from the file and carry on

Firstyear: Because we haven't been able to bubble up the spn that we need to remove the token.

yaleman: This seems like a logical loop. 

1. We got the token from the local token store.
2. We tried to parse it
3. That failed 
4. Now we can't remove it because we can't parse it?

Firstyear: Which is exactly why I un-hide the local only removal flow. That allows breaking out of the loop.

yaleman: But it's a broken token, why not just clean the token store and tell the user what happened, instead of `exit(1)`ing with an error message and leaving them with the broken thing still in the cache? 

Fair enough don't try and send it to the server, but it shouldn't be in the token store at the end of the interaction if it's broken - once you've gotten past line 636 (`client.get_token()`) there's *something* found that can be dealt with, so if validation doesn't work then just clean it up.

Firstyear: That's a good point. I think we need to do that in the try_to_client() call though because that affects every possible workflow / interaction.

yaleman: That'd work, too 😄

Firstyear: Reading through this, the logic to add that here is going to be a nightmare actually. I think having unhidden the ability to force remove the token locally is going to have to suffice for now. Also note that kanidm session cleanup catches them too.

Firstyear: Actually there is a way, spoke too soon.

Firstyear: Actually, there isn't.

yaleman (approved):

PR #2805, opened 2024-05-29
Sometimes an MFA Code prompt requires polling

dmulder:
The MS Authenticator app requires both a code
entered from the app, and input on the app for a
entropy code (sent in the same PAM message). This
requires the idprovider to poll and wait for the
user to enter the entropy code on the app.

Fixes #

Checklist

- [ ] cargo fmt has been run


dmulder:
In the case that no polling is actually necessary (usually the case), then only 1 poll request is sent and the idprovider immediately responds.

dmulder:
Going to try this a different way.

Firstyear:
Yeah, I was going to say couldn't we do similar here to what we have for the async task spawn for MFA already to drive this? :)

PR #2820, opened 2024-06-06
Stats collection improvements and a bunch of other stuff

Seba-T:
Maybe this should have been more than one pr but here we are anyways:

- new password change role for actors
- added more granular data collection; now we differentiate between read and write operations, and we can control the label for each operation, which means that some can also get no label if just want to ignore them for whatever reason
- added csv file generation and included information about the test environment such as the number of nodes, the number of users.
- various small improvements here and there (see commits for the full list) 

- [x] cargo fmt has been run


tools/orca/src/stats.rs
```diff
@@ -22,12 +25,44 @@ pub trait DataCollector {
     ) -> Result<(), Error>;
 }
 
-pub struct BasicStatistics {}
+enum OpKind {
+    WriteOp,
+    ReadOp,
+    Other, //TODO! does this make sense?
```
Firstyear, line 31: Could be OpKind "Auth" since that's what covers those? 
Similor for opkind error.

Firstyear (changes requested):
Minor changes but looks good

tools/orca/Cargo.toml
```diff
@@ -19,9 +19,10 @@ doctest = false
 
 [dependencies]
 async-trait = { workspace = true }
+chrono.workspace = true
```
yaleman, line 22: `<crankyoldman>`
please use the same TOML format as the other lines
`</crankyoldman>` 😄 Thanks!

tools/orca/src/stats.rs
Seba-T: Yeah actually makes sense, I used `Other` because I thought it could cover some other future use cases, but maybe it would be too generic

tools/orca/src/model.rs
```diff
@@ -118,6 +121,30 @@ pub async fn person_set_self_mail(
     Ok(parsed_result)
 }
 
+pub async fn person_set_self_password(
+    client: &KanidmClient,
+    person: &Person,
+    pw: &str,
+) -> Result<(TransitionResult, EventRecord), Error> {
+    // Should we measure the time of each call rather than the time with multiple calls?
```
yaleman, line 129: We should be using tracing [instrumentation](https://docs.rs/tracing/0.1.38/tracing/attr.instrument.html) then you can get all the various call timings.

tools/orca/Cargo.toml
Seba-T: 😂, sure!

tools/orca/src/model.rs
Seba-T: Actually I'm wondering why we didn't use it in the first place @Firstyear? 🤔

Firstyear: Because I didn't know you could use it like this to pull the timings out. 

Lets change it to tracing instrumentation in a follow up PR I think.

Seba-T: Makes lots of sense. Will do that once this gets merged ig

Firstyear (approved):

PR #2821, opened 2024-06-06
Add development taint flag to prevent mismatch of server versions

### LINKED ISSUE #2817: Unable to authenticate using replicated kanidm server
### I did this
Set up replication between two kanidm servers. I went to the login page after verifying that the logs said replication successful, and tried to log in. 
### I expected the following
To log in
### This happened instead
An error has occurred 😔

kp0031keyobjectnotfound Operation ID: df5414d9-9247-4ba5-a2c8-9ef7aee90e43
``` on the login webpage

From the docker logs:
8ba0eda6-1e71-48c0-bf3e-b166e1c227a5 INFO     ┝━ auth [ 4.49ms | 90.22% ]
8ba0eda6-1e71-48c0-bf3e-b166e1c227a5 INFO     │  ┝━ ｉ [info]: Begin auth event | event_tag_id: 10 | sessionid: None | req: AuthRequest { step: Init2 { username: "idm_admin", issue: Token, privileged: false } }
8ba0eda6-1e71-48c0-bf3e-b166e1c227a5 INFO     │  ┝━ ｉ [info]: Initiating Authentication Session | event_tag_id: 10 | username: idm_admin | issue: Token | privileged: false | uuid: 00000000-0000-0000-0000-000000000018
8ba0eda6-1e71-48c0-bf3e-b166e1c227a5 INFO     │  ┕━ ｉ [info]: Sending auth result | event_tag_id: 10 | res: Err(KP0031KeyObjectNotFound)
8ba0eda6-1e71-48c0-bf3e-b166e1c227a5 INFO     ┝━ auth_session_state_management [ 1.33µs | 0.03% ] inter: Err(KP0031KeyObjectNotFound)
8ba0eda6-1e71-48c0-bf3e-b166e1c227a5 ERROR    ┝━ 🚨 [error]:  | latency: 4.971765ms | status_code: 500 | kopid: "8ba0eda6-1e71-48c0-bf3e-b166e1c227a5" | msg: "error handling request"
8ba0eda6-1e71-48c0-bf3e-b166e1c227a5 ERROR    ┕━ 🚨 [error]: response failed | classification: Status code: 500 Internal Server Error | latency: 4 ms



### Kanidm version details

- Server 1: kanidmd 1.2.0-dev e8d7010b4b
- Server 2: kanidmd 1.2.2
- Server 1: No
- Server 2: 46534f4ae93   kanidm/server:latest
Server 1: Ubuntu Server 22.04
Server 2: Ubuntu Server 24.04
### Any other comments

### PR DESCRIPTION
**Pull request by Firstyear**
Fixes #2817 - It was previously possible that users could mix-and-match dev and stable releases, which could break replication and databases. This is because the dev version is a pre-release version of a domain level, so the stable and dev versions both believe they have the same migration level, but the dev version is likely extremely behind and has missing elements.

In addition, trying to move from a dev version to a stable version had it's own share of traps - you needed to manually run a domain remigration as you moved to ensure you had all the needed migrations applied, else you risked inconsistency.

This fixes both issues by adding a taint flag. If you have a dev version it "taints" your domain object. This way when you upgrade between dev versions, if the taint is present, you always have to remigrate (as currently expected).

If you move from dev to stable, since the DB still has it's taint flag, then the final domain remigration is automatically carried out, and then the taint flag is cleared.

If you move from stable to dev, the taint flag is re-added forcing migrations to be forced at startup.

In replication if a node attempts to refresh another server, both must have matching taint flags, or absence of. 

If two replicas attempt to incrementally replicate, if the taint flag is present then it must be present on all nodes or none of them.

Checklist

### CODE REVIEW

File: server/lib/src/constants/schema.rs [hunk 1]
```diff
@@ -698,6 +698,14 @@ pub static ref SCHEMA_ATTR_PATCH_LEVEL_DL7: SchemaAttribute = SchemaAttribute {
     ..Default::default()
 };
 
+pub static ref SCHEMA_ATTR_DOMAIN_DEVELOPMENT_TAINT_DL7: SchemaAttribute = SchemaAttribute {
+    uuid: UUID_SCHEMA_ATTR_DOMAIN_DEVELOPMENT_TAINT,
+    name: Attribute::DomainDevelopmentTaint.into(),
+    description: "".to_string(),
```
yaleman, line 704:
```suggestion
    description: "A flag to show that the domain has been run on a development build, and will need additional work to upgrade/migrate.".to_string(),
```

yaleman (dismissed):

yaleman (approved):

PR #2844, opened 2024-06-18
20240613 coalesce delayed writes

Firstyear:
Thanks @Seba-T. No other notes.

Checklist

### DIFF
--- a/server/core/src/actors/internal.rs
+++ b/server/core/src/actors/internal.rs
@@ -100,22 +100,55 @@ impl QueryServerWriteV1 {
         }
     }
 
-    pub(crate) async fn handle_delayedaction(&self, da: DelayedAction) {
+    pub(crate) async fn handle_delayedaction(&self, da_batch: &mut Vec<DelayedAction>) {
         let eventid = Uuid::new_v4();
         let span = span!(Level::INFO, "process_delayed_action", uuid = ?eventid);
+
+        let mut retry = false;
 
         async {
             let ct = duration_from_epoch_now();
             let mut idms_prox_write = self.idms.proxy_write(ct).await;
-            if let Err(res) = idms_prox_write
-                .process_delayedaction(da, ct)
-                .and_then(|_| idms_prox_write.commit())
-            {
-                info!(?res, "delayed action error");
+
+            for da in da_batch.iter() {
+                retry = idms_prox_write.process_delayedaction(da, ct).is_err();
+                if retry {
+                    // exit the loop
+                    break;
+                }
+            }
+
+            if let Err(res) = idms_prox_write.commit() {
+                retry = true;
+                warn!(?res, "delayed action commit error");
             }
         }
         .instrument(span)
-        .await
+        .await;
+
+        if retry {
+            // An error occured, retry each operation one at a time.
+            for da in da_batch.iter() {
+                let eventid = Uuid::new_v4();
+                let span = span!(Level::INFO, "process_delayed_action_retried", uuid = ?eventid);
+
+                async {
+                    let ct = duration_from_epoch_now();
+                    let mut idms_prox_write = self.idms.proxy_write(ct).await;
+                    if let Err(res) = idms_prox_write
+                        .process_delayedaction(da, ct)
+                        .and_then(|_| idms_prox_write.commit())
+                    {
+                        warn!(?res, "delayed action error");
+                    }
+                }
+                .instrument(span)
+                .await
+            }
+        }
+
+        // We're done, clear out the buffer.
+        da_batch.clear();
     }
 
     #[instrument(
--- a/server/core/src/actors/internal.rs
+++ b/server/core/src/actors/internal.rs
@@ -114,13 +114,14 @@ impl QueryServerWriteV1 {
                 retry = idms_prox_write.process_delayedaction(da, ct).is_err();
                 if retry {
                     // exit the loop
+                    warn!("delayed action failed, will be retried individually.");
                     break;
                 }
             }
 
             if let Err(res) = idms_prox_write.commit() {
                 retry = true;
-                warn!(?res, "delayed action commit error");
+                error!(?res, "delayed action batch commit error");
             }
         }
         .instrument(span)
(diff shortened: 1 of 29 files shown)


server/core/src/actors/internal.rs
```diff
@@ -100,22 +100,55 @@ impl QueryServerWriteV1 {
...
+                retry = idms_prox_write.process_delayedaction(da, ct).is_err();
+                if retry {
+                    // exit the loop
+                    break;
+                }
+            }
+
+            if let Err(res) = idms_prox_write.commit() {
+                retry = true;
+                warn!(?res, "delayed action commit error");
```
yaleman, line 123: isn't this an "OH NO LET'S BAIL" situation at this point?

Firstyear: The problem here is that if we are writing a stripe of delayed actions and one of the set fails, we don't want to negatively impact the other elements of that set. So we fall back to one-at-a-time processing to allow everything else to proceed correctly even if a tad slower.

server/core/src/actors/internal.rs
```diff
@@ -100,22 +100,55 @@ impl QueryServerWriteV1 {
         }
     }
 
-    pub(crate) async fn handle_delayedaction(&self, da: DelayedAction) {
+    pub(crate) async fn handle_delayedaction(&self, da_batch: &mut Vec<DelayedAction>) {
```
yaleman, line 103: Since we don't use the batch anywhere else we don't really need the mut and can take ownership of it, by the looks of it?

server/lib/src/be/idl_sqlite.rs
```diff
@@ -1771,7 +1771,6 @@ impl IdlSqlite {
                         "PRAGMA page_size={fs_page_size};
                          PRAGMA cache_size={cache_pages};
                          PRAGMA journal_mode=WAL;
-                         PRAGMA synchronous=NORMAL;
```
yaleman, line 1774: What's the default/setting in this case? the [sqlite docs](https://www.sqlite.org/pragma.html#pragma_synchronous) don't seem to mention it? 🫨

Firstyear: It's "synchronous=FULL" but you can relax it to NORMAL if you have WAL. I think I added it earlier, but I reliased now it's not a good idea to set in that layer as it doesn't make much performance change, and an admin can always just do it direct into the underlynig sqlite file if they want to change it.

yaleman: Ah, OK. they do mention that "NORMAL" is normally the best case in WAL mode, but more sync more better.

server/core/src/actors/internal.rs
yaleman: 'k, just thinking if the commit fails after an error, then something horrible's going on.

Firstyear: Yeah, but even if we panic we can't do much about it I don't think. Well a panic would poison the write lock and prevent any new writes I guess ....

Firstyear: I am wondering if we should make fn commit panic instead ....

server/core/src/actors/internal.rs
Firstyear: We re-use the vec as a buffer here.

Seba-T:
> Thanks @Seba-T. No other notes.

*pretends this wasn't an absolute sadistic experience on your end* "You're welcome 🥰"

server/core/src/actors/internal.rs
yaleman: well I don't think a panic would be right but an `Err` at least :D

server/lib/src/server/mod.rs
```diff
@@ -1311,6 +1309,11 @@ impl QueryServer {
 
         let key_providers = Arc::new(KeyProviders::default());
 
+        // These needs to be pool_size minus one to always leave a DB ticket
+        // for a writer. But it also needs to be at least one :)
+        debug_assert!(pool_size > 0);
+        let read_ticket_pool = if pool_size - 1 > 0 { pool_size } else { 1 };
```
yaleman, line 1315: If pool_size is `2` you still get `2` in this case, which doesn't match the comment - suggested code makes this a little easier to understand as well 😄 

```suggestion
        let read_ticket_pool = std::cmp::min(pool_size-1,1);
```

Firstyear: Actually wait this will make it 0. It needs to be cmp::max

server/lib/src/plugins/dyngroup.rs
```diff
@@ -208,22 +230,53 @@ impl DynGroup {
...
+                    // The *dyn group* isn't changing, it's that a member OF the dyn group
+                    // is being added. This means the dyngroup isn't part of the set that
+                    // needs update to MO, only the affected members do!
+                    /*
+                    if pre.get_ava_set(Attribute::DynMember)
+                        != d_group.get_ava_set(Attribute::DynMember)
+                    {
+                        affected_uuids.insert(*dg_uuid);
+                    }
+                    */
```
yaleman, line 257: Redundant now? 
```suggestion
```

server/lib/src/plugins/dyngroup.rs
```diff
@@ -344,11 +386,31 @@ impl DynGroup {
...
```
yaleman, line 398: Some copypasta, perhaps?

```suggestion
```

yaleman (approved):
Woo, such fast, many zoom.

server/lib/src/plugins/dyngroup.rs
Firstyear: I left it there as a "just in case", bunt I guess it can just be removed. I still think we need the comment though.

server/lib/src/plugins/dyngroup.rs
Firstyear: As above, we should keep the comment.

yaleman (approved):

PR #2845, opened 2024-06-20
Added configurable thread count for orca and server

Seba-T:
Checklist

- [x] cargo fmt has been run

### DIFF
--- a/server/daemon/src/main.rs
+++ b/server/daemon/src/main.rs
@@ -529,6 +529,11 @@ async fn kanidm_main() -> ExitCode {
         return ExitCode::FAILURE;
     }
 
+    if let Some(threads) = sconfig.thread_count {
+        // we don't want to use more threads than available!
+        config.update_threads_count(std::cmp::min(threads, config.threads));
+    }
+
     config.update_db_arc_size(sconfig.get_db_arc_size());
     config.update_role(sconfig.role);
     config.update_output_mode(opt.commands.commonopt().output_mode.to_owned().into());
--- a/tools/orca/src/main.rs
+++ b/tools/orca/src/main.rs
@@ -220,7 +220,9 @@ fn main() -> ExitCode {
                     return ExitCode::FAILURE;
                 }
             };
-            let runtime = create_tokio_runtime(state.thread_count);
+            // here we need to create one less worker compared to the desired amount since we later call `spawn_blocking`, which consumes
+            // an extra thread all on its own
+            let runtime = create_tokio_runtime(state.thread_count.map(|t| t - 1));
             // We have a broadcast channel setup for controlling the state of
             // various actors and parts.
             //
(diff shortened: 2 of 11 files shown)


tools/orca/src/main.rs
```diff
@@ -205,74 +206,88 @@ async fn main() -> ExitCode {
                     return ExitCode::FAILURE;
                 }
             };
-
+            // here we need to create one less worker compared to the desired amount since we later call `spawn_blocking`, which consumes
+            // an extra thread all on its own
+            let runtime = build_tokio_runtime(state.thread_count.map(|t| t - 1));
```
Firstyear, line 211: No, you don't need to do this. spawn blocking is a separate thread pool. Just set the thread count to the number of threads.

tools/orca/src/main.rs
```diff
@@ -205,74 +206,88 @@ async fn main() -> ExitCode {
...
                     }
                 }
-            }
+            })
         }
-    };
+    }
+}
+
+fn build_tokio_runtime(threads: Option<usize>) -> Runtime {
```
Firstyear, line 284: doc comments to describe how this works with threads whene None means to use the default multi thread scheduler.

server/daemon/src/main.rs
```diff
@@ -529,6 +529,11 @@ async fn kanidm_main() -> ExitCode {
         return ExitCode::FAILURE;
     }
 
+    if let Some(threads) = sconfig.thread_count {
```
Firstyear, line 532: Where are you getting the threads value from? This comparison doesn't make any sense.

Firstyear: Should be 
```
std::cmp::min(threads, std::threads::avaliable_parallelism())
```
https://doc.rust-lang.org/stable/std/thread/fn.available_parallelism.html

Firstyear (changes requested):

server/daemon/src/main.rs
Seba-T: Well that would be the idea since the default for sconfig.thread_count is actually the parallelism, even though if that ever gets changed then I agree it will stop making any sense.
But I also thought who the hell is gonna change it multiple times? So I felt it wasn't *too* bad, but maybe it is

tools/orca/src/main.rs
Seba-T: But my fear was that since that uses a separate thread pool, if we have 20 workers and 1 spawn blocker, won't this use 21 threads?

server/daemon/src/main.rs
Firstyear: But it's not available_parallelism, you're never calling available parallelism. So what is config.threads even set to here? I think you're over-engineering this.

tools/orca/src/main.rs
Firstyear: And why is that a problem? That's what spawnblocking does. If you want to maek it 19 workers, set it to 19.

PR #2846, opened 2024-06-21
added orca docker file, make target and credential reset capabilities

Seba-T:
Checklist

- [x]  This pr contains no AI generated code
- [x] cargo fmt has been run

### DIFF
--- a/tools/orca/Dockerfile
+++ b/tools/orca/Dockerfile
@@ -1,0 +1,69 @@
+# This builds the kanidm CLI tools
+ARG BASE_IMAGE=opensuse/tumbleweed:latest
+# ARG BASE_IMAGE=opensuse/leap:15.5
+
+FROM ${BASE_IMAGE} AS repos
+ADD ../../scripts/zypper_fixing.sh /zypper_fixing.sh
+RUN --mount=type=cache,id=zypp,target=/var/cache/zypp /zypper_fixing.sh
+
+FROM repos AS builder
+ARG KANIDM_FEATURES
+ARG KANIDM_BUILD_PROFILE
+ARG KANIDM_BUILD_OPTIONS=""
+
+# Set the build profile
+ENV KANIDM_BUILD_PROFILE=${KANIDM_BUILD_PROFILE:-container_generic}
+ENV RUSTFLAGS="-Clinker=clang -Clink-arg=-fuse-ld=/usr/bin/ld.mold"
+
+RUN \
+    --mount=type=cache,id=zypp,target=/var/cache/zypp \
+    zypper install -y --no-recommends \
+        sccache \
+        cargo \
+        clang \
+        make \
+        automake \
+        autoconf \
+        libopenssl-3-devel \
+        pam-devel \
+        libudev-devel \
+        sqlite3-devel \
+        rsync \
+        mold
+
+COPY . /usr/src/kanidm
+
+WORKDIR /usr/src/kanidm/
+
+# build the CLI
+RUN \
+    --mount=type=cache,id=cargo,target=/cargo \
+    --mount=type=cache,id=sccache,target=/sccache \
+    export CARGO_HOME=/cargo; \
+    export SCCACHE_DIR=/sccache; \
+    export RUSTC_WRAPPER=/usr/bin/sccache; \
+    export CC="/usr/bin/clang"; \
+    cargo build -p orca ${KANIDM_BUILD_OPTIONS} \
+        --target-dir="/usr/src/kanidm/target/" \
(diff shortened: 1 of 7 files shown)


yaleman:
This'll need another rebase from the changes in #2847

tools/orca/Dockerfile
```diff
@@ -0,0 +1,69 @@
...
+        --release && \
+    cargo install \
+        --git https://github.com/kanidm/webauthn-rs.git \
+        --rev 5f4db4172f8e22aedc68c282d177e98db2b1892f \
+        --force fido-mds-tool \
+        --target-dir="/usr/src/kanidm/target/" && \
+    sccache -s
+
+# == Construct the orca container starting on the kanidm server devel image
+FROM kanidm/server:devel
```
yaleman, line 58: this should be based on "repos", and you need to set the "CMD" line to match (see the server `DockerFile`)

Seba-T: But won't I have to build the entire server from scratch again if I base it on repos? I thought this was saving some time by using the prebuild image, also to avoid having a huge copypasta from the server Dockerfile. I've added the cmd tho :)

yaleman: No, because what you're doing is building the binaries starting at line 9 in a container called `builder` then starting the new/final container here, and copying the binaries across from `builder` at lines 63-64.

yaleman (approved):

yaleman (changes requested):
uh, accidental approval based on no sleep

tools/orca/Dockerfile
Seba-T: Ah yeah you're right this was really dumb :p

Seba-T:
@yaleman Also I think the credential reset inside orca is a really dumb idea too, since it assumes we're running orca in the same container as the server itself, which is basically never gonna happen. So at this point I might just leave the fixed docker file for the orca image?

tools/orca/src/profile.rs
```diff
@@ -45,6 +45,16 @@ impl Profile {
...
     }
 
+    pub fn set_admin_password(&mut self, pw: &str) {
+        self.admin_password.clear();
+        self.admin_password.push_str(pw);
+    }
+
+    pub fn set_idm_admin_password(&mut self, pw: &str) {
+        self.idm_admin_password.clear();
+        self.idm_admin_password.push_str(pw);
```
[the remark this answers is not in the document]
Seba-T: some weird efficiency paranoia :p Idk I felt it was better to update the string content rather than allocate a new one? As if it mattered since it literally does it two times tho

yaleman:
> @yaleman Also I think the credential reset inside orca is a really dumb idea too, since it assumes we're running orca in the same container as the server itself, which is basically never gonna happen. So at this point I might just leave the fixed docker file for the orca image?

You won't always be running it in a container - especially with Orca where we're using it for dev testing etc. I'm not sure doing it with command manipulation's necessarily going to be the best way, since again, it's a dev tool so you won't have the binaries on the command line...

Seba-T:
> > @yaleman Also I think the credential reset inside orca is a really dumb idea too, since it assumes we're running orca in the same container as the server itself, which is basically never gonna happen. So at this point I might just leave the fixed docker file for the orca image?
> 
> You won't always be running it in a container - especially with Orca where we're using it for dev testing etc. I'm not sure doing it with command manipulation's necessarily going to be the best way, since again, it's a dev tool so you won't have the binaries on the command line...

Yeah I completely agree. I say I nuke the credential reset thing and we just keep the make + Dockerfile, which in some instances turns out to be useful (ie what I'm doing now, but I don't think I'll be the last person to ever do this). What do you think?

yaleman:
> > > @yaleman Also I think the credential reset inside orca is a really dumb idea too, since it assumes we're running orca in the same container as the server itself, which is basically never gonna happen. So at this point I might just leave the fixed docker file for the orca image?
> > 
> > 
> > You won't always be running it in a container - especially with Orca where we're using it for dev testing etc. I'm not sure doing it with command manipulation's necessarily going to be the best way, since again, it's a dev tool so you won't have the binaries on the command line...
> 
> Yeah I completely agree. I say I nuke the credential reset thing and we just keep the make + Dockerfile, which in some instances turns out to be useful (ie what I'm doing now, but I don't think I'll be the last person to ever do this). What do you think?

sounds sensible to me 👍🏻

Seba-T:
@yaleman everything should be fixed now!

yaleman (approved):

yaleman (approved):

PR #2854, opened 2024-06-26
20240620 htmx

Firstyear:
Fixes #

Checklist

- [ ] cargo fmt has been run


server/core/templates/base_htmx.html
```diff
@@ -0,0 +1,27 @@
...
+	</head>
+	<body hx-boost="true">
+	(% block body %)(% endblock %)
+	</body>
+
+	<footer class="footer mt-auto py-3 bg-light text-end">
+		<div class="container">
+			<span class="text-muted">Powered by <a href="https://kanidm.com">Kanidm</a></span>
+		</div>
+	</footer>
```
2xsaiko, line 25: This should probably be inside \<body>

Firstyear: Good point, will fix that :)

Firstyear:
There is still a ton to do here but this is the "absolutely minimum". This has working password and totp flows and a minimal landing page for apps. I need to still make error handling better as well as add a guard to check for login state.

server/core/templates/apps.html
```diff
@@ -0,0 +1,10 @@
+(% extends "base_htmx.html" %)
+
+(% block title %)Error(% endblock %)
+
+(% block head %)
+(% endblock %)
+
+(% block body %)
+	(( apps_partial|safe ))
```
yaleman, line 9: we shouldn't be rendering the whole body here as a `apps_partial|safe`, surely - just the tags for the list of them? this feels... risky

server/core/Cargo.toml
```diff
@@ -15,9 +15,14 @@ repository = { workspace = true }
 test = true
 doctest = false
 
+[features]
+ui_htmx = []
```
yaleman, line 19: Probably either need to commit your changes or add this to `server/daemon/Cargo.toml` under its features:

```
ui_htmx = ["kanidmd_core/ui_htmx"]
```

yaleman (commented):
We should probably [overlay the new HTMX things as specific routes](https://learn.microsoft.com/en-us/azure/architecture/patterns/strangler-fig) until the WASM bits are completely replaced, it'd make it easier to test - I can't get to /ui/reset to log in on a test instance 😢

server/core/templates/base_htmx.html
```diff
@@ -0,0 +1,26 @@
...
+		<title>(% block title %)(( title )) - Kanidm(% endblock %)</title>
+
+		<link rel="icon" href="/pkg/img/favicon.png" />
+		<link rel="stylesheet" href="/pkg/external/bootstrap.min.css" integrity="sha384-EVSTQN3/azprG1Anm3QDgpJLIm9Nao0Yz1ztcQTwFspd3yD65VohhpuuCOmLASjC"/>
+		<script src="/pkg/external/bootstrap.bundle.min.js" integrity="sha384-MrcW6ZMFYlzcLA8Nl+NtUVF0sA7MsXsP1UyJoMp4YLEuNSfAP+JcXn/tWtIaxVXM"></script>
+		<script src="/pkg/external/htmx.min.1.9.2.js" integrity="sha384-L6OqL9pRWyyFU3+/bjdSri+iIphTN/bvYyM37tICVyOJkWZLpP2vGn6VUEXgzg6h"></script>
+
+		(% block head %)(% endblock %)
+	</head>
+	<body hx-boost="true">
```
ToxicMushroom, line 17: Makes the footer sit at the bottom again
```suggestion
	<body hx-boost="true" class="flex-column d-flex h-100">
```

`/pkg/style.css` is also not linked as stylesheet anymore/yet (causing the `logo-square.svg` on login to stay quite large on mobile sizes)

server/core/templates/apps.html
ToxicMushroom: To my understanding `apps_partial` should be responsible for safely rendering its app items, and we have to assume it did this correctly otherwise we cannot include html from other dynamic templates

without safe askama escapes all html tags:
image

yaleman: Since the `apps_partial` is just another rendered template it should extend the parent template, rather than being yeeted in like this.

yaleman:
> There is still a ton to do here but this is the "absolutely minimum". This has working password and totp flows and a minimal landing page for apps. I need to still make error handling better as well as add a guard to check for login state.

I'd probably make the error handling in the login flow just part of the parent login template, instead of having to have a totp errors and password errors etc.

yaleman:
> > We should probably [overlay the new HTMX things as specific routes](https://learn.microsoft.com/en-us/azure/architecture/patterns/strangler-fig) until the WASM bits are completely replaced, it'd make it easier to test - I can't get to /ui/reset to log in on a test instance 😢
> 
> Then we need to prepend all the routes and resources with some /htmx-dev prefix ? Would it not be simpler to run 2 servers, 1 with the htmx feature on and one without ?

The axum router uses the "most specific" as a match, so specifying the new view routes will be fine. This is how the current overloads work, see: https://github.com/kanidm/kanidm/blob/eb005e13969bd803d20142697a79430aa10c0d8d/server/core/src/https/mod.rs#L263-L274

Firstyear:
> We should probably [overlay the new HTMX things as specific routes](https://learn.microsoft.com/en-us/azure/architecture/patterns/strangler-fig) until the WASM bits are completely replaced, it'd make it easier to test - I can't get to /ui/reset to log in on a test instance 😢

Yes, but then when we go to actually swap we'll need to up-end every url that we just spent time developing and testing. I'd rather it be a hard feature flag instead.

yaleman:
> > We should probably [overlay the new HTMX things as specific routes](https://learn.microsoft.com/en-us/azure/architecture/patterns/strangler-fig) until the WASM bits are completely replaced, it'd make it easier to test - I can't get to /ui/reset to log in on a test instance 😢
> 
> Yes, but then when we go to actually swap we'll need to up-end every url that we just spent time developing and testing. I'd rather it be a hard feature flag instead.

I mean define the new URLs at the new place as more specific than the current ones - which you're already doing, because we're rendering individual server-side views instead of the current yew-handles-the-wildcards - and leave the current wildcards and "prune" the WASM routers when the feature flag is on.

Firstyear:
> > There is still a ton to do here but this is the "absolutely minimum". This has working password and totp flows and a minimal landing page for apps. I need to still make error handling better as well as add a guard to check for login state.
> 
> I'd probably make the error handling in the login flow just part of the parent login template, instead of having to have a totp errors and password errors etc.

Yes, this was my plan. I just need to actually, you know. Do it :)

Firstyear:
> > > We should probably [overlay the new HTMX things as specific routes](https://learn.microsoft.com/en-us/azure/architecture/patterns/strangler-fig) until the WASM bits are completely replaced, it'd make it easier to test - I can't get to /ui/reset to log in on a test instance 😢
> > 
> > 
> > Yes, but then when we go to actually swap we'll need to up-end every url that we just spent time developing and testing. I'd rather it be a hard feature flag instead.
> 
> I mean define the new URLs at the new place as more specific than the current ones - which you're already doing, because we're rendering individual server-side views instead of the current yew-handles-the-wildcards - and leave the current wildcards and "prune" the WASM routers when the feature flag is on.

Thats exactly what's happening?

Firstyear:
Anyway, I think we should merge this "as is" so that I can finish the login flows and let @ToxicMushroom start on the new app views in parallel.

yaleman:
> > > > We should probably [overlay the new HTMX things as specific routes](https://learn.microsoft.com/en-us/azure/architecture/patterns/strangler-fig) until the WASM bits are completely replaced, it'd make it easier to test - I can't get to /ui/reset to log in on a test instance 😢
> > > 
> > > 
> > > Yes, but then when we go to actually swap we'll need to up-end every url that we just spent time developing and testing. I'd rather it be a hard feature flag instead.
> > 
> > 
> > I mean define the new URLs at the new place as more specific than the current ones - which you're already doing, because we're rendering individual server-side views instead of the current yew-handles-the-wildcards - and leave the current wildcards and "prune" the WASM routers when the feature flag is on.
> 
> Thats exactly what's happening?

You're pruning the axum router so users can't access the WASM, not the yew-internal routers...

yaleman (approved):

yaleman:
kk, leaving it as a feature flag while dev is ongoing

ExplodingDragon:
> We should probably [overlay the new HTMX things as specific routes](https://learn.microsoft.com/en-us/azure/architecture/patterns/strangler-fig) until the WASM bits are completely replaced, it'd make it easier to test - I can't get to /ui/reset to log in on a test instance 😢

 Will kanidm remove wasm completely and use htmx in the future? 😢

Firstyear:
> > We should probably [overlay the new HTMX things as specific routes](https://learn.microsoft.com/en-us/azure/architecture/patterns/strangler-fig) until the WASM bits are completely replaced, it'd make it easier to test - I can't get to /ui/reset to log in on a test instance 😢
> 
> Will kanidm remove wasm completely and use htmx in the future? 😢

Yes. We've found it too heavy and hard to modify as we need to expand to add new features.

ExplodingDragon:
> > > We should probably [overlay the new HTMX things as specific routes](https://learn.microsoft.com/en-us/azure/architecture/patterns/strangler-fig) until the WASM bits are completely replaced, it'd make it easier to test - I can't get to /ui/reset to log in on a test instance 😢
> > 
> > 
> > Will kanidm remove wasm completely and use htmx in the future? 😢
> 
> Yes. We've found it too heavy and hard to modify as we need to expand to add new features.

htmx seems to be very vulnerable to xss attacks, A filter exists when our cli adds anything?

yaleman:
> htmx seems to be very vulnerable to xss attacks, A filter exists when our cli adds anything?

what does the cli have to do with HTMX?

ExplodingDragon:
> > htmx seems to be very vulnerable to xss attacks, A filter exists when our cli adds anything?
> 
> what does the cli have to do with HTMX?

Sorry for the description , if I enter something with `<script>` in nickName will it be rejected or not, I don't quite understand, I'm a rust noob.

yaleman:
> > > htmx seems to be very vulnerable to xss attacks, A filter exists when our cli adds anything?
> > 
> > 
> > what does the cli have to do with HTMX?
> 
> Sorry for the description , if I enter something with `<script>` in nickName will it be rejected or not, I don't quite understand, I'm a rust noob.

There are other protections in place for that in our content-security-policy headers, and `askama`, the templating library.

> (A similar issue does not seem appropriate for discussion here, so let's end it.)

Good plan, for now 😄 Perhaps a github discussion?

yaleman (approved):

2xsaiko:
> htmx seems to be very vulnerable to xss attacks, A filter exists when our cli adds anything?

HTMX is absolutely not any more vulnerable to XSS attacks than the current code. What does that even mean? XSS is an issue of the server not properly sanitizing user input.

Firstyear:
> Sorry for the description , if I enter something with `<script>` in nickName will it be rejected or not, I don't quite understand, I'm a rust noob.

Correct. But regardless of this, all views that we render are generated from askama templates which performs the correct filtering of values, so even if you were to set your users display name to something like <script> it will be sanitised. If you have an example of a working xss we would like to know about it, otherwise it's speculation and we have already ensure we take steps to correctly template and process user input in safe ways.

PR #2868, opened 2024-07-03
[htmx] Apps page

ToxicMushroom:
Not tracked in an issue afaik.
This pr adds the apps page as we had with wasm.

Checklist

- [x] cargo fmt has been run


server/core/src/https/views/apps.rs
```diff
@@ -27,20 +25,53 @@ struct AppsView {
...
 }
 
 pub(crate) async fn view_apps_get(
-    State(_state): State<ServerState>,
-    Extension(_kopid): Extension<KOpId>,
+    State(state): State<ServerState>,
+    Extension(kopid): Extension<KOpId>,
     HxRequest(hx_request): HxRequest,
-    VerifiedClientInformation(_client_auth_info): VerifiedClientInformation,
+    VerifiedClientInformation(client_auth_info): VerifiedClientInformation,
```
Firstyear, line 54: https://docs.rs/axum/latest/axum/response/type.Result.html

We should use this here I think to handle the possible app_links error.

server/core/src/https/views/apps.rs
```diff
@@ -27,20 +25,53 @@ struct AppsView {
...
 ) -> Response {
     // Because this is the route where the login page can land, we need to actually alter
     // our response as a result. If the user comes here directly we need to render the full
     // page, otherwise we need to render the partial.
+    let handler = get_transformer(kopid.eventid);
```
Firstyear, line 59: I don't like this kind of weird factory/fn factory this returns,  Ithink there are simpler ways to achieve this.

server/core/src/https/views/login.rs
```diff
@@ -29,14 +29,14 @@ use crate::https::{
...
+use time::OffsetDateTime;
 use super::{HtmlTemplate, UnrecoverableErrorView};
 
 #[derive(Template)]
 #[template(path = "login.html")]
-struct LoginView<'a> {
-    username: &'a str,
-    remember_me: bool,
+pub struct LoginView<'a> {
+    pub(crate) username: &'a str,
```
Firstyear, line 38: Why did you change the visibility of these?

server/core/src/https/views/login.rs
```diff
@@ -90,6 +90,30 @@ pub async fn view_index_get(
...
 }
 
+// Maybe cursed ?
+pub async fn view_logout_get(
+    State(state): State<ServerState>,
+    VerifiedClientInformation(_client_auth_info): VerifiedClientInformation,
+    Extension(_kopid): Extension<KOpId>,
+    _jar: CookieJar,
+) -> Response {
+    let mut bearer_cookie = Cookie::new(COOKIE_BEARER_TOKEN, "");
```
Firstyear, line 100: It seems weird that just deleting the cookie isn't working, because it certainly works in other parts of the auth flow. Are you sure this is needed?

server/core/templates/apps_partial.html
```diff
@@ -1,3 +1,29 @@
-<main class="flex-shrink-0 form-signin">
-	<h2>Apps Go Here</h2>
+<main class="p-3 x-auto">
+	<div class="(( crate::https::ui::CSS_PAGE_HEADER ))">
+		<h2>Applications list</h2>
+    </div>
+(% if apps.is_empty() %)
```
Firstyear, line 5: Indent these in to match properly.

server/core/templates/apps.html
```diff
@@ -1,10 +1,15 @@
...
 (% endblock %)
 
+(% block nav %)
+	(% include "navbar.html" %)
+(% endblock %)
+
 (% block body %)
 	(( apps_partial|safe ))
-(% endblock %)
+	(% include "signout_modal.html" %)
```
Firstyear, line 14: Maybe we need a template for a base_with_nav that also has the signout and things? That way something like apps can extend that? Since we'll need this navbar + signout in other views too.

server/core/src/https/views/apps.rs
ToxicMushroom: Yeah me neither, but I need to acquire the kopid for displaying errors somehow, switching to `axum::response::Result` does do away with the implementation here.

This is what I have right now but it's kinda the same issue
`.map_err(|old| { HtmxError::from((kopid, old)) })`

server/core/src/https/views/login.rs
ToxicMushroom: This was so the _to_be_replaced_ get_transformer function could construct LoginView by itself.

server/core/src/https/views/login.rs
ToxicMushroom: Yeah, path seems to be required in my testing on firefox, `bearer=; Path=/; Expires=Thu, 04 Jul 2024 15:15:46 GMT` deletes the cookie for me.

The cookiejar remove method: `bearer=; Max-Age=0; Expires=Wed, 05 Jul 2023 15:57:02 GMT`  does not

> because it certainly works in other parts of the auth flow

Probably because bearer_cookie.set_path("/"); is set causes this different behavior compared to `auth-session-id`

server/core/src/https/ui.rs
```diff
@@ -8,6 +8,13 @@ use axum::{Extension, Router};
 use super::middleware::KOpId;
 use super::ServerState;
 
+// when you want to put big text at the top of the page
+pub const CSS_PAGE_HEADER: &str = "d-flex justify-content-between flex-wrap flex-md-nowrap align-items-center pt-0 pb-0 mb-3 border-bottom";
```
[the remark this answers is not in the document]
ToxicMushroom: I don't think we can create 'shortcuts' to classes in plain css ?

ToxicMushroom:
I've duplicated the error type into an HtmxError because I didn't want to break the wasm error stuff, hope that that's okay.
I also changed the error handling on the login page mostly to test it out, can revert if it's a problem :)

server/core/src/https/ui.rs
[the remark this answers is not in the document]
yaleman: Yep, this isn't CSS, it's a list of classes, 'tis fine... this is being migrated over from `server/web_ui/shared/src/constants.rs`. We probably should have a "constants" or "prelude" module for this stuff as it'll be everywhere, in `server/core/src/https/` ?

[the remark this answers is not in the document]
ToxicMushroom: > We probably should have a "constants" or "prelude" module for this stuff as it'll be everywhere, in server/core/src/https/ ?

Yes, sounds good to me :)

server/core/src/https/views/login.rs
Firstyear: That seems like a weird issue and one we should look into because we need to be able to use .remove.

server/core/src/https/views/login.rs
Firstyear: I think we drop this part and if there is a specific issue we follow up with it in isolation (especially because I just did a ton of changes to the login flow)

Firstyear: And because you shouldn't need to construct the LoginView anywhere else. You should redirect to the login page, not render it inplace.

server/core/src/https/views/login.rs
```diff
@@ -61,35 +63,50 @@ pub async fn view_index_get(
...
-                remember_me: false,
-            })
-            .into_response()
+    // manual error handling because we don't want to see a not logged/session expired error
+    // on the login page
+    match auth_result {
+        Ok(_) => Ok(Redirect::to("/ui/apps").into_response()),
+        Err(err) => {
+            match err {
+                OperationError::NotAuthenticated | OperationError::SessionExpired => {
```
Firstyear, line 79: These are impossible states at this point, so instead of making an infinite login loop, render the error as it was before. (Better yet, undo this whole section of your pr).

server/core/src/https/errors.rs
```diff
@@ -1,11 +1,72 @@
...
+
+/// The web app's top level error type, this takes an `OperationError` and converts it into a HTTP response.
+#[derive(Debug, ToSchema)]
+pub enum HtmxError {
+    /// Something went wrong when doing things.
+    OperationError(Uuid, OperationError),
+    // InternalServerError(Uuid, String),
+}
+
+impl From<(&KOpId, OperationError)> for HtmxError {
```
Firstyear, line 25: This from syntax makes calling HtmxError really messy because now it's "HtmxError::from((a, b))" all the time (note the double parens.). 

It should just be "HtmxError::new()"

server/core/src/https/errors.rs
```diff
@@ -1,11 +1,72 @@
...
+impl IntoResponse for HtmxError {
+    fn into_response(self) -> Response {
+        match self {
+            // HtmxError::InternalServerError(_kopid, inner) => {
+            //     (StatusCode::INTERNAL_SERVER_ERROR, inner).into_response()
+            // }
+            HtmxError::OperationError(kopid, inner) => {
+                let body = serde_json::to_string(&inner).unwrap_or(inner.to_string());
+                let response = match &inner {
+                    OperationError::NotAuthenticated | OperationError::SessionExpired => {
```
Firstyear, line 40: These two should just redirect you to the login page instead.

server/core/src/https/views/mod.rs
```diff
@@ -47,7 +50,7 @@ pub fn view_router() -> Router<ServerState> {
         .merge(guarded_router)
 }
 
-struct HtmlTemplate<T>(T);
+pub struct HtmlTemplate<T>(pub T);
```
Firstyear, line 53: Why is this pub?

server/core/src/https/errors.rs
```diff
@@ -1,11 +1,72 @@
...
+
+use kanidm_proto::internal::OperationError;
+
+use crate::https::middleware::KOpId;
+use crate::https::views::errors::ErrorPartialView;
+use crate::https::views::HtmlTemplate;
+
+/// The web app's top level error type, this takes an `OperationError` and converts it into a HTTP response.
+#[derive(Debug, ToSchema)]
+pub enum HtmxError {
```
Firstyear, line 19: I suspect this should be in "views/error" to avoid HtmlTemplate becoming pub.

Firstyear (changes requested):

server/core/src/https/views/login.rs
Firstyear: (I say that mostly because I think the whole thing is going to conflict with the PR I just put in, and I think there are better ways to fix the logout issue potentially, so I'll do that in a follow up PR. )

server/core/templates/base_htmx_with_nav.html
```diff
@@ -0,0 +1,37 @@
+<!DOCTYPE html>
```
Firstyear, line 1: Minor but this can just extend the base_htmx as well. You can nest templates :)

server/core/templates/apps_partial.html
```diff
@@ -1,3 +1,28 @@
...
+	(% else %)
+	<div class="row row-cols-1 row-cols-sm-2 row-cols-md-3 g-3">
+		(% for app in apps %)
+		<div class="col-md-3">
+			<div class="card text-center">
+			(% match app %)
+				(% when AppLink::Oauth2 with { name, display_name, redirect_url, has_image } %)
+				<a href="(( redirect_url ))" class="link-dark stretched-link mt-2">
+					(% if has_image %)
+						<img src="/ui/images/oauth2/(( name ))" class="oauth2-img" alt="((display_name)) icon" id="(( name ))">
```
Firstyear, line 16: Mental note, check this url is cached (and our other ui urls are not)

Firstyear (commented):
I think it's almost done!

server/core/templates/apps_partial.html
ToxicMushroom: Yeah this doesn't seem to get cached:
```
HTTP/1.1 200 OK
content-type: image/jpeg
content-length: 14075
content-security-policy: default-src 'self'; base-uri 'self' https:; form-action 'self' https:;frame-ancestors 'none'; img-src 'self' data:; worker-src 'none'; script-src 'self' 'unsafe-eval' 'sha384-YvpcrYf0tY3lHB60NNkmXc5s9fDVZLESaAA55NDzOxhy9GkcIdslK1eN7N6jIeHz' 'sha384-ujb1lZYygJmzgSwoxRggbCHcjc0rB2XoQrxeTUQyRjrOnlCoYta87iKBWq3EsdM2' 'sha384-j55kjUt6fvgxRPzrED9pe3r179Nod1RYTUzquoKlgEnsjxHkXutHcIws8u8/seDf' 'sha384-9w+Roa3ri0y9Ouy1kgFlBIwzZkKygG1EMgpvhBvdNUG2PLMNT2hxwHL4o2bDtOfF' 'sha384-gLRkh0byHbUW4T8eGWfGzvMsvJu6Umb1JY1CPYRKzC6a2y56aj3TQYtHV4eQOisC';
x-content-type-options: nosniff
permissions-policy: fullscreen=(), geolocation=()
referrer-policy: no-referrer-when-downgrade
x-kanidm-version: 1.3.0-dev
strict-transport-security: max-age=86400
x-kanidm-opid: 6719e98f-be80-4def-9849-1babc416ff99
date: Tue, 09 Jul 2024 20:02:52 GMT
```

server/core/templates/base_htmx_with_nav.html
ToxicMushroom: Resolved by [00df717](https://github.com/kanidm/kanidm/pull/2868/commits/00df7171ef215c16b1a66e5d7c1a514a1f9b3d95)

Firstyear (approved):

PR #2889, opened 2024-07-10
Fixes the logout flow in htmx and improves the login error dialog

Firstyear:
Fixes #

Checklist

-  [ x ] This pr contains no AI generated code


server/core/src/https/views/login.rs
```diff
@@ -617,7 +658,7 @@ async fn view_login_step(
                             let mut username_cookie =
                                 Cookie::new(COOKIE_USERNAME, session_context.username.clone());
                             username_cookie.set_secure(state.secure_cookies);
-                            username_cookie.set_same_site(SameSite::Strict);
+                            username_cookie.set_same_site(SameSite::Lax);
```
yaleman, line 661: I clearly missed this on a previous one, why aren't we using `Strict` here (and above, which is scarier)?

Firstyear: https://developer.mozilla.org/en-US/docs/Web/Security/Practical_implementation_guides/Cookies#samesite

tl;dr is if it's strict then the cookie isn't sent on a redirect to kanidm. The redirector can't see the cookie, but lax allows it to be added on that redirection.

yaleman: Ah, 'k. With the other protections in place that'll work 😄 👍🏻

server/core/src/https/views/apps.rs
ToxicMushroom, line 1: These changes make me wonder if my cargo fmt really works
My toolchain is set to stable with `rustc 1.79.0`

Firstyear: Yeah, something seems off, we have to keep reformating after you ..... :(

Firstyear:
Thanks @ToxicMushroom those were good points. I've fixed them all now :)

ToxicMushroom (approved):

yaleman (approved):

PR #2897, opened 2024-07-13
[htmx] Credential Update page 

ToxicMushroom:
Contains credential reset form, update page with (passkeys, attested passkeys, totp, password)

Checklist

- [x] cargo fmt has been run


server/core/src/https/views/reset.rs
```diff
@@ -50,7 +50,72 @@ struct CredResetPartialView {
...
+    VerifiedClientInformation(_client_auth_info): VerifiedClientInformation,
+    jar: CookieJar,
+    Form(new_passwords): Form<NewPassword>,
+) -> axum::response::Result<Response> {
+    let cookie = jar.get(COOKIE_CU_SESSION_TOKEN);
+    return if let Some(cookie) = cookie {
+        let cu_session_token = cookie.value();
+        let cu_session_token = CUSessionToken { token: cu_session_token.into() };
+
+        let eq = new_passwords.new_password == new_passwords.new_password_check;
```
yaleman, line 92: `eq` is confusing here, and can save having to specify the name in line 99 later

```suggestion
        let pwd_equal = new_passwords.new_password == new_passwords.new_password_check;
```

server/core/src/https/views/reset.rs
```diff
@@ -125,10 +189,29 @@ fn get_cu_template(domain_display_name: String, cu_status: CUStatus) -> HtmlTemp
...
+
+
+// Any filter defined in the module `filters` is accessible in your template.
+mod filters {
+    pub fn blank_if<T: std::fmt::Display>(implicit_arg: T, condition: bool) -> ::askama::Result<String> {
+        return if condition {
+            Ok("".into())
+        } else {
+            Ok(format!("{implicit_arg}"))
+        };
```
yaleman, line 208: Don't repeat implementations when you can just do this:

```suggestion
        blank_iff(implicit_arg, &condition)
```

server/core/static/external/cred_update.js
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    function markPwdCheckInvalid() {
+        new_pwd_check.classList.add("is-invalid");
+        new_pwd_check.classList.remove("is-valid");
+        pwd_submit.disabled = true;
+    }
+
+    new_pwd.addEventListener("input", (_) => {
+        if (new_pwd.value === new_pwd_check.value) {
```
yaleman, line 20: Minor UX thing, it'd be nice to differentiate between "You haven't put anything in the second box" and "the two are different"... I always find it annoying when I'm getting red-flagged as "You're doing it wrong" before I've even started typing in the second field 😞 We still need to note that the second box needs filling, but ... not THEY ARE DIFFERENT 😄

ToxicMushroom: Yeah that's true, I'll change that :)

server/core/templates/cred_update/add_password_modal_partial.html
```diff
@@ -6,37 +6,70 @@ <h5 class="modal-title" id="staticPasswordLabel">Add a New Password</h5>
                 <button type="button" class="btn-close" aria-label="Close"></button>
             </div>
             <div class="modal-body">
-                <form class="row g-3 needs-validation" novalidate
-                      hx-post="/ui/reset/pw_check"
-                      hx-trigger="input from:#new-password delay:200ms changed, input from:#new-password-check delay:200ms changed">
-                    <input hidden=true type="text" autocomplete="username" />
+                <form class="row g-3 needs-validation" id="newPasswordForm" novalidate
```
yaleman, line 9: `needs-validation` is from the [bootstrap example site](https://getbootstrap.com/docs/5.0/forms/validation/#custom-styles), if we aren't referencing it, we don't need it.

```suggestion
                <form class="row g-3" id="newPasswordForm" novalidate
```

server/core/templates/cred_update/add_password_modal_partial.html
```diff
@@ -6,37 +6,70 @@ <h5 class="modal-title" id="staticPasswordLabel">Add a New Password</h5>
...
                             id="new-password"
                             placeholder=""
                             type="password"
                             required
+                            autofocus
                     />
+                    <!-- bootstrap hides the feedback if we remove is-invalid from the input above -->
+                    <div id="password-validation-feedback" class="invalid-feedback d-block">
+                        <ul>
+                            (% if let PwdCheckResult::Failure with { pwd_equal, warnings } = check_res %)
```
yaleman, line 38: Move this block outside the `<div>` above it, don't need the tags if we don't have the `check_res` Failure to handle 😄

Firstyear (commented):
This is a good start, but remember just because we did it before with modals and js like this doesn't mean we need to stick to this pattern *now*. I think you're trying to keep it too "the same" when you have the opportunity to approach things differently. For example, we could avoid modals and do something more like what github does or have inline forms, or we could even use htmx a bit more to avoid the js. Don't be limited by what we already have!

server/core/src/https/views/mod.rs
```diff
@@ -66,7 +68,12 @@ pub fn view_router() -> Router<ServerState> {
     // The webauthn post is unguarded because it's not a htmx event.
 
     // Anything that is a partial only works if triggered from htmx
-    let guarded_router = Router::new().layer(HxRequestGuardLayer::new("/ui"));
+    let guarded_router = Router::new()
+        .route("/reset/new_pwd", post(reset::view_new_pwd))
```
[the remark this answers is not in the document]
ToxicMushroom: It's now /reset/add_password but also no longer a modal, it now feels like a full page

proto/src/internal/credupdate.rs
```diff
@@ -388,7 +388,7 @@ impl fmt::Display for PasswordFeedback {
             }
             PasswordFeedback::TooShort(minlength) => write!(
                 f,
-                "Password too was short, needs to be at least {} characters long.",
+                "Password was too short, needs to be at least {} characters long.",
```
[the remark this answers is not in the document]
ToxicMushroom: Well the form gets cleared after submission so the password is no longer there

ToxicMushroom:
Would be great if someone could test out the attested passkeys ui,
I kinda failed at this: https://kanidm.github.io/kanidm/master/accounts/account_policy.html#setting-webauthn-attestation-ca-lists, fido-mds-tool was complaining that my file wasn't valid or something and I hadn't bothered digging further.

remove totp needs to be reworked I think

Firstyear:
I'll test it for you during the week :)

Firstyear:
Seems to work:

Screenshot 2024-07-23 at 15 51 54
Screenshot 2024-07-23 at 15 52 17
Screenshot 2024-07-23 at 15 52 22

Only things are that there is a trailing slash that shouldn't be there, and the "name this passkey" should be autofocused already, and should accept "enter" key to submit the form. Otherwise works well!

Firstyear (dismissed):

Firstyear:
@yaleman what do you think? 

@ToxicMushroom Just needs a rebase is the main thing

server/core/src/https/views/errors.rs
```diff
@@ -21,27 +21,31 @@ use crate::https::middleware::KOpId;
...
+    }
 }
 
 impl IntoResponse for HtmxError {
     fn into_response(self) -> Response {
         match self {
-            // HtmxError::InternalServerError(_kopid, inner) => {
-            //     (StatusCode::INTERNAL_SERVER_ERROR, inner).into_response()
-            // }
+            HtmxError::InternalServerError(_kopid, inner) => {
```
yaleman, line 40: Are we sure we're returning the `kopid` as part of these responses? Dropping the `kopid` makes it harder to troubleshoot.

server/core/static/external/cred_update.js
```diff
@@ -0,0 +1,106 @@
...
+function stillSwapFailureResponse(event) {
+    if (event.detail.xhr.status === 422) {
+        console.log("Still swapping failure response")
+        event.detail.shouldSwap = true;
+        event.detail.isError = false;
+    }
+}
+
+function startPasskeyEnrollment() {
+    let data_elem = document.getElementById('data');
```
yaleman, line 53: If it's never reassigned, it should be a const
```suggestion
    const data_elem = document.getElementById('data');
```

server/core/static/external/cred_update.js
```diff
@@ -0,0 +1,106 @@
...
+    let credentialRequestOptions = JSON.parse(data_elem.textContent);
```
yaleman, line 54: If it's never reassigned, it should be a const

```suggestion
    let credentialRequestOptions = JSON.parse(data_elem.textContent);
```

server/core/static/external/cred_update.js
yaleman: Also we're assuming that the getElementById and then parse work, which probably isn't great - what happens if it's not there for some reason?

server/core/templates/cred_update/add_passkey_partial.html
```diff
@@ -0,0 +1,27 @@
+<div>
+    <div class="row" id="staticPasskeyCreateRow">
+        <script id="data">(( challenge|safe ))</script>
+
+        <!-- Safari requires a human input to start passkey creation -->
+        <button id="passkeyNamingSafariBtn" class="btn btn-primary">Begin Passkey Enrollment</button>
```
yaleman, line 6:
```suggestion
        <button id="passkeyNamingSafariBtn" class="btn btn-primary">Begin Passkey Enrolment</button>
```

yaleman:
Looks great so far!

server/core/static/external/cred_update.js
ToxicMushroom: JS logs an exception and clicking the button will seemingly do nothing.
The element with this id sits in the same file that calls passkeyNamingSubmitBtn: add_passkey_partial.html

I can show an alert browser popup telling them we have some internal data structure mismatch and that they should report this to us ?

yaleman: I'd check for the fail to get by id and parsing errors and log that to the console, because the exception can cause all javascript to stop, which isn't great.

Then pop a JS confirm saying "There was an error handling a server response, would you like to try and reload the page?" which when they click OK, reloads the tab - which should re-initialise everything.

server/core/src/https/views/reset.rs
```diff
@@ -0,0 +1,736 @@
...
+            let cu_status = state
+                .qe_r_ref
+                .handle_idmcredentialupdate(
+                    cu_session_token,
+                    CURequest::TotpGenerate,
+                    kopid.eventid,
+                )
+                .await
+                // TODO: better handling for invalid mfaregstate state, can be invalid if certain mfa flows were interrupted
+                // TODO: We should maybe automatically cancel the other MFA reg
```
Firstyear, line 428: Open an issue so we don't forget.

server/core/src/https/views/reset.rs
```diff
@@ -0,0 +1,736 @@
...
+                .map_err(|op_err| HtmxError::new(&kopid, op_err))?;
+
+            let partial = if let CURegState::TotpCheck(secret) = cu_status.mfaregstate {
+                let uri = secret.to_uri();
+                let svg = match QrCode::new(uri.as_str()) {
+                    Ok(qr) => qr.render::<svg::Color>().build(),
+                    Err(qr_err) => {
+                        error!("Failed to create TOTP QR code: {qr_err}");
+                        "QR Code broke".to_string()
```
Firstyear, line 437: "QR Code Generation Failed" might be a better string?

Firstyear (changes requested):
Minor changes but looking good :)

Firstyear:
Also needs a rebase to master :)

Firstyear (approved):

Firstyear:
@yaleman Up to you now :)

server/core/src/https/views/reset.rs
ToxicMushroom: #2940

Firstyear (approved):

PR #2907, opened 2024-07-16
build profiles: rename release_suse_generic to release_linux

alindt:
Just to avoid confusion — there's nothing SUSE specific about it (anymore?).


Firstyear (commented):
Maybe "release_linux" then because we have growing interesting in freebsd/solaris so we need to account for them having releases too. Otherwise looks good!

alindt:
Good point, renaming to "release_linux".

Firstyear (approved):

PR #2909, opened 2024-07-16
New orca models

Seba-T:
Brand new orca models folks! 

Checklist

- [x] cargo fmt has been run


tools/orca/src/generate.rs
```diff
@@ -122,7 +123,7 @@ pub async fn populate(_client: &KanidmOrcaClient, profile: Profile) -> Result<St
 
         let roles = BTreeSet::new();
 
-        let model = Model::Basic;
+        let model = Model::Writer;
```
Firstyear, line 126: This needs to be configurable ....

tools/orca/src/generate.rs
```diff
@@ -146,15 +147,23 @@ pub async fn populate(_client: &KanidmOrcaClient, profile: Profile) -> Result<St
     // them a baseline of required accounts with some variation. This
     // way in each test it's guaranteed that *at least* one person
     // to each role always will exist and be operational.
+    let groups_person_count_map: HashMap<&String, u64> = profile
```
Firstyear, line 150: I don't understand what this is doing?

tools/orca/src/main.rs
```diff
@@ -97,6 +97,7 @@ fn main() -> ExitCode {
                 admin_password,
                 idm_admin_password,
                 threads,
+                Vec::new(), // for now we don't support specifying person_count_by_group from the wizard
```
Firstyear, line 100: What is person_count_by_group?

tools/orca/src/models/markov.rs
```diff
@@ -0,0 +1,141 @@
...
+use statrs::distribution::{Multinomial, Normal};
+
+// The idea is to have a multinomial distribution for each state that represents
+// the probabilities of each transitions from that state.
+//* Huge issues with this approach! since we take the probabilities matrix from the user there are SOOO many ways things can go wrong
+//* from a syntactic standpoint we need to ensure all values fall within range (between 0 and 1 if we want the probabilities already normalized)
+//* and we have the right number of them.
+//* We also need to safeguard the user from accidentally creating inconsistent transitions, that is we must force some transition probabilities to always be 0,
+//* i.e. the prob of an Unauthenticated user updating their display_name must be 0. All this validation should be done at the SetupWizard stage??
+pub(crate) struct ActorMarkov {
```
Firstyear, line 22: What did I say about Markov?

tools/orca/src/models/read.rs
```diff
@@ -0,0 +1,93 @@
...
+
+impl ActorReader {
+    fn next_transition(&mut self) -> Transition {
+        match self.state {
+            State::Unauthenticated => Transition {
+                delay: Some(Duration::from_secs(2)),
+                action: TransitionAction::Login,
+            },
+            State::Authenticated => Transition {
+                delay: Some(Duration::from_secs(3)),
```
Firstyear, line 69: Your problem here is that by having this at 3 seconds you create "bursts" where every actor delays and then reads then stops then reads. That's why your numbers are so chaotic.

tools/orca/src/models/write.rs
```diff
@@ -0,0 +1,107 @@
...
+            State::Unauthenticated => Transition {
+                delay: Some(Duration::from_secs(2)),
+                action: TransitionAction::Login,
+            },
+            State::Authenticated => Transition {
+                delay: Some(Duration::from_secs(3)),
+                action: TransitionAction::PrivilegeReauth,
+            },
+            State::AuthenticatedWithReauth => Transition {
+                delay: Some(Duration::from_secs(3)),
```
Firstyear, line 76: And here.

tools/orca/src/stats.rs
```diff
@@ -141,15 +141,15 @@ impl DataCollector for BasicStatistics {
             return Err(Error::InvalidState);
         }
 
-        if writeop_times.is_empty() {
-            error!("For some weird reason no write operations were recorded, exiting...");
-            return Err(Error::InvalidState);
-        }
+        // if writeop_times.is_empty() {
```
Firstyear, line 144: Why is this disabled?

Firstyear (changes requested):

tools/orca/src/generate.rs
Seba-T: Yeah I think that was about time. Should I also add an option to the wizard to choose the model when generating the profile.toml file?

tools/orca/src/main.rs
Seba-T: Basically it's an option that allows the user to decide how many people should be in each group instead of having them randomly assigned. This is useful when for example you want EVERYONE to be a member of `role_people_self_write_mail`. If you don't explicitly set the number of users for a given group then it will resort to the old technique we were using (that is randomly assign members). UPDATE: I've realized this is extremely dumb and works only if we assign everyone to a single group. I'll explain below what the broader issue is.

As of now, there are two factors that determine what each actor does: 

-  The first thing is obviously the `TransitionAction` that gets spit out by the `transition` function. For the `Writer` model, it will always be `WriteAttributePersonMail`
- The second factor is the groups the actor is member of. If we are not members of `role_people_self_write_mail` we'll never be able to write our email, regardless of what `TransitionAction` said. 

Before, this wasn't an issue, as all previous models derived their transitions based on what groups they belonged to, like [here](https://github.com/kanidm/kanidm/blob/master/tools/orca/src/models/basic.rs#L91), so they never tried to do something "forbidden".
Unfortunately, when we have stricter requirements, ie we need everyone to belong to a single group, the old approach won't work. We'll have actors that try to do something (i.e write their emails) and fail, as they don't have the right privileges. On the other hand, I think we could simplify it down by having just an option that says "give me the name of a group and I'll make everyone member of that group". I mean maybe in some VERY advanced *future* benchmarks one could say "alright I have 100 users, I want 80 in the mail writing group, 15 in the self read account, and so forth", but as far as I'm concerned I don't think I'll ever need such option.

tools/orca/src/stats.rs
Seba-T: Because when running a write only / read only test we expect to have 0 read events / write events respectively. Idk if you like my workaround to avoid panicking tho, I'm not sure I do either 🥲

tools/orca/src/generate.rs
Seba-T: This is derived from `person_count_by_group`. The idea is that for each group we check if its name is also a key inside the hashmap. If it is we set the number of members of that group to the value stored in the hashmap under its name. When a group name is not a key, then we randomly choose the number of members it should have, like we did before.

tools/orca/src/models/read.rs
Seba-T: Yeah I agree introducing some randomness here will help. I've done it by passing the seed inside `as_dyn_object`, this way I don't need to do something like `Model::Writer(u64)` and `Model::Reader(u64)`, and as such I can use the `Model` enum as a ClapEnum, which looks really clean to me :D. No weird string conversions, only strong typing!

tools/orca/src/generate.rs
Seba-T: Well I did it anyway because I was too amazed by the way enums and clap work together :p

tools/orca/src/models/markov.rs
Seba-T: :///

Firstyear: That's right, remove it.

tools/orca/profile-sample.toml
```diff
@@ -1,10 +1,19 @@
-control_uri = "https://localhost:8443"
+person_count_by_group = [
+  ["role_people_self_write_mail", 500]
+  ["role_people_self_set_password", 0],
```
Firstyear, line 3: I've been thinking about this and I realised what's "off" here is that it's *super* unclear how it actually works when you look at it.

I think this should be:

```

[group."role_people_self_write_mail"]
member_count = 500

```

Then any *unspecified* group is just set to the default.

This can be done in the serde struct for toml with:

```
group: BTreeMap<String, GroupProperties>,

struct GroupProperties {
    member_count: Option<u32>
}
```

I think once you do this, it'll make the *rest* of the PR make a lot more sense, and be far clearer in how it reads.

Firstyear (changes requested):

tools/orca/profile-sample.toml
Seba-T: Yeah I should have done my own research into the toml crate before doing random stuff. I've fixed it now! (along with some extra *tiny* additions which you can see below :D )

tools/orca/Cargo.toml
```diff
@@ -1,46 +1,46 @@
...
+name = "orca"
 
-version = { workspace = true }
-authors = { workspace = true }
-rust-version = { workspace = true }
-edition = { workspace = true }
-license = { workspace = true }
-homepage = { workspace = true }
-repository = { workspace = true }
+authors = {workspace = true}
```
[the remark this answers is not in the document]
Seba-T: I swear I straight up don't know how to stop vscode from doing it automatically on save 😭😭 (or better, I've never bothered to figure it out, looks like now it's about time :p)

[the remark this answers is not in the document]
Seba-T: Should be fixed now :)

tools/orca/profile-sample.toml
```diff
@@ -1,10 +1,24 @@
...
 group_count = 5
+idm_admin_password = ""
+model = "writer"
 person_count = 500
+seed = -1236045086759770365
+test_time = 60
 thread_count = 20
+warmup_time = 10
+
+[group.role_people_self_mail_write]
```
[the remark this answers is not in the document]
Seba-T: Agreed! Even though I must admit I wasn't able to find ANY guide detailing the trickshot you suggested with the hashmap. It only appeared in some sketchy examples but no one bothered to explain how it works :/

tools/orca/Cargo.toml
```diff
@@ -25,6 +25,7 @@ crossbeam = { workspace = true }
 csv = { workspace = true }
 futures-util = { workspace = true, features = ["sink"] }
 hashbrown = { workspace = true }
+idlset = {workspace = true}
```
Firstyear, line 28: Missed one.

tools/orca/profile-sample.toml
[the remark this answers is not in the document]
Firstyear:
```
#[serde(flatten)]
Group: HashMap<String, GroupStruct>

...
#[derive(Deserialize)]
struct GroupStruct {
    member_count: u32
}
```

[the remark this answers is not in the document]
Seba-T: Yeah once you see it in actions it looks kinda trivial, my point is that all of this is a bit .... lacking in documentation? Like the toml crate just doesn't tell you you can do this. Or I'm blind.

tools/orca/Cargo.toml
Seba-T: 😩😩

tools/orca/profile-sample.toml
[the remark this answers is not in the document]
Firstyear: https://serde.rs/field-attrs.html

Firstyear (approved):

Firstyear:
@Seba-T Well done!

PR #2922, opened 2024-07-21
Some debugging and bits, but mostly a design

Firstyear:
So much for "it's a design".

This starts on multi-resolver support, improved system id resolving, better TPM handling and more.

Currently the only "feature" regression is that group overrides no longer work - but I plan to correct that before we proceed further. The problem is currently that it's getting to be "a lot" and I want to try to break it up over a few PR's.

Going to run this at home as a sandbox test first while I take a quick break to implement some scim bits for OpenSUSE. 

Checklist

### DIFF
--- a/book/src/developers/designs/unixd_multi_resolver_2024.md
+++ b/book/src/developers/designs/unixd_multi_resolver_2024.md
@@ -1,0 +1,418 @@
+## Unixd MultiResolver Support
+
+Up until July 2024 the purpose and motivation of the Kanidm Unixd component (`unix_integration` in
+the source tree) was to allow Unix-like platforms to authenticate and resolve users against a Kanidm
+instance.
+
+However, throughout 2023 and 2024 this project has expanded in scope - from the addition of TPM
+support to protect cached credentials (the first pam module to do so!), to use of the framework by
+himmelblau to enable Azure AD authentication.
+
+We also have new features we want to add including LDAP backends (as an alternative to SSSD), the
+ability to resolve local system users, as well as support for PIV and CTAP2 for desktop login.
+
+This has pushed the current design of the resolver to it's limits, and it's increasingly challenging
+to improve it as a result. This will necesitate a major rework of the project.
+
+### Current Architecture
+
+```
+                                                    ┌ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ┐
+
+                ┌───────┐  ┌───────┐  ┌───────┐     │ ┌───────────────────┐     │
+                │       │  │       │  │       │       │                   │
+                │  NSS  │  │  PAM  │  │  CLI  │     │ │   Tasks Daemon    │     │
+                │       │  │       │  │       │       │                   │
+                └───────┘  └───────┘  └───────┘     │ └───────────────────┘     │
+                    ▲          ▲          ▲                     ▲
+            ─ ─ ─ ─ ┼ ─ ─ ─ ─ ─│─ ─ ─ ─ ─ ┼ ─ ─ ─ ─ ┴ ─ ─ ─ ─ ─ ┼ ─ ─ ─ ─ ─ ─ ─ ┤
+            │       ▼          ▼          ▼                     │
+                ┌─────────────────────────────┐           ┌───────────┐         │
+            │   │                             │           │           │
+                │         ClientCodec         │           │   Tasks   │         │
+            │   │                             │           │           │
+                └─────────────────────────────┘           └───────────┘         │
+┌ ─ ─ ─ ─ ─ ┘                  ▲                                ▲
+                               │                                │               │
+│                              ▼                                │
+  ┌───────────────┐      ┌────────────────────────────────┐     │               │
+│ │               │      │                                │     │
+  │  Kani Client  │◀────▶│      Daemon / Event Loop       │─────┘               │
+│ │               │      │                                │
+  └───────────────┘      └────────────────────────────────┘                     │
+│                                   ▲                 ▲
+                                    │                 │                         │
+│                                   ▼                 ▼
+                          ┌──────────────────┐   ┌────────┐                     │
+│                         │                  │   │        │
+                          │    DB / Cache    │   │  TPM   │                     │
+│                         │                  │   │        │
+                          └──────────────────┘   └────────┘                     │
+│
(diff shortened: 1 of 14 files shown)


book/src/developers/designs/unixd_multi_resolver_2024.md
```diff
@@ -0,0 +1,306 @@
...
+
+To protect caches from offline modification content should be optionally encrypted / signed in the
+future.
+
+### Challenges
+
+- The order of backend resolvers needs to be stable.
+- Accounts/Groups should _not_ move/flip-flop between resolvers.
+- Resolvers need to uniquely identify entries in some manner.
+- The ability to store persistent/permananent accounts in the DB that can _not_ be purged in a cache
```
yaleman, line 131: We should identify the purpose of this, and how we would indicate this to the end user, because I know I'd forget about it and be confused why it wouldn't go away when I did a cache clear.

book/src/developers/designs/unixd_multi_resolver_2024.md
```diff
@@ -0,0 +1,306 @@
...
+#### Multiple Resolvers
+
+We would like to support multiple backends simultaneously and in our source tree. This is a major
+motivator of this rework as the himmelblau project wishes to contribute their client layer into our
+source tree, while maintaining the bulk of their authentication code in a separate libhimmelblau
+project.
+
+We also want to support LDAP and other resolvers too.
+
+The major challenge here is that this shift the cache state from the daemon to the client. This
```
yaleman, line 102: Since we (currently) expect to source-control all the clients, why not cache in the daemon and tag them with a client source?

book/src/developers/designs/unixd_multi_resolver_2024.md
```diff
@@ -0,0 +1,306 @@
...
+│ │    Client     │◀─┘     └────────┘  └──────────────────┘                     │
+  │               │
+└ ┴───────────────┴ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ┴
+```
+
+#### Online/Offline State Machine
+
+The major change that that this diagram may not clearly show is that the online/offline state
+machine moves into each of the named clients (excluding files). This also has some future impacts on
+things like pre-emptive item reloading and other tash scheduling. This will require the backends to
```
yaleman, line 179:
```suggestion
things like pre-emptive item reloading and other task scheduling. This will require the backends to
```

book/src/developers/designs/unixd_multi_resolver_2024.md
```diff
@@ -0,0 +1,306 @@
...
+"upcall" into the daemon, as the TPM transaction needs to be passed from the daemon back down to the
+provider. Alternatelly, the provider needs to be able to register scheduled tasks into the daemon
```
yaleman, line 181:
```suggestion
provider. Alternately, the provider needs to be able to register scheduled tasks into the daemon
```

book/src/developers/designs/unixd_multi_resolver_2024.md
Firstyear: I think the context here is "that way we can store things like passkeys to a local user account". You wouldn't want those purged after a cache clear.

Anything from remote services is always cached, but we need to make sure local-things can't be accidentally stomped berause we can't just re-populate them.

book/src/developers/designs/unixd_multi_resolver_2024.md
Firstyear: That's exactly what I'm proposing :)

Firstyear:
The remaining clippy lints are for the main server and I'll do those in a follow up PR.

unix_integration/common/src/unix_passwd.rs
```diff
@@ -5,7 +5,7 @@ use serde::{
 
 use std::fmt;
 
-#[derive(Serialize, Deserialize, Debug)]
+#[derive(Serialize, Deserialize, Debug, Default)]
```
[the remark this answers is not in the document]
Firstyear: Good points!

unix_integration/resolver/src/bin/kanidm-unix.rs
```diff
@@ -228,7 +229,11 @@ async fn main() -> ExitCode {
             } else {
                 match call_daemon(cfg.sock_path.as_str(), req, cfg.unix_sock_timeout).await {
                     Ok(r) => match r {
-                        ClientResponse::Ok => println!("working!"),
+                        ClientResponse::ProviderStatus(results) => {
+                            for provider in results {
+                                println!("{}: {}", provider.name, provider.online);
```
yaleman, line 234:
```suggestion
                                println!("{}: {}", provider.name, "online" if provider.online else "offline");
```

unix_integration/resolver/src/idprovider/system.rs
```diff
@@ -0,0 +1,126 @@
...
+                if group.name != user.name {
+                    error!(name = %user.name, uid = %user.uid, gid = %user.gid, "user private group does not appear to have the same name as the user, this may be a security risk!");
+                }
+                if !(group.members.is_empty()
+                    || (group.members.len() == 1 && group.members.first() == Some(&user.name)))
+                {
+                    error!(name = %user.name, uid = %user.uid, gid = %user.gid, "user private group must not have members, THIS IS A SECURITY RISK!");
+                }
+            } else {
+                warn!(name = %user.name, uid = %user.uid, gid = %user.gid, "user private group is not present on system, synthesising it");
```
yaleman, line 73: since this is (likely) pretty common, this feels like a debug-level message?

unix_integration/resolver/src/bin/kanidm-unix.rs
Firstyear: I actually didn't know you could do that syntax lol.

unix_integration/resolver/src/idprovider/system.rs
Firstyear: I kind of want there to be a stick that says "your distro is doing dumb shit" but the other problem yeh, is how to force people to fix it. nfi.

yaleman (approved):

PR #2929, opened 2024-07-24
[HTMX] User settings

ToxicMushroom:
Fixes # (will fill this in later in case things get delayed due to internal roadblocks)

Currently aims to contain:
- Profile
- Ssh keys
- Reauth for write access and creating cred reset tokens for adding external devices
- Posix/Unix pwd

Checklist
- [ ] cargo fmt has been run


book/src/developers/designs/profile_display.md
```diff
@@ -0,0 +1,136 @@
...
+  E.g. via a popup that shows the old vs new value asking to confirm.
+
+#### TODO: Personal Identifiable Information attributes (currently we don't have these attributes)
+Certain information should not be displayed in the UI without reauthentication:
+ - addresses
+ - phone numbers
+ - personal emails
+ - birthdate
+
+### SSH public keys
```
Firstyear, line 44: I think ssh keys and unix password will move to the cred update flow.

Firstyear:
Probably a good idea to rebase this to main to continue on it :)

book/src/developers/designs/profile_display.md
ToxicMushroom: Oh okay, I think that makes sense

weijiangan: Should we reuse the cred update page content for credential status section? They seem mostly similar other than the ability to update them. Then again, I'm also not sure if there's a need to hide the navbar for credentials reset page.

Firstyear: The credential reset page needs to be outside of the navbar because it happens in a credential update transaction, which also requires a re-auth. That's why we swap to a separate page. I think if we kept the navbar, we'd need a way to implicitly cancel the cred update on navigate away.

weijiangan: It makes sense that reauth doesn't have navbar because you go to the auth screen, but after that on the cred update page it shouldn't be necessary other than for artificially blocking users from navigating to other pages. Do we handle when the user doesn't complete the credential update by canceling implicitly now? My impression is that we don't (navigating away and going back to the update flow and the flow is still alive) and I know from my experience of poking around that it does expire eventually.

The other part of this is me being a UI guy... we should be able to update creds in the app itself without being in a modal screen. We don't need to couple UI and logic. If you agree I will be happy to look into this direction :)

Firstyear: We don't actually, you're right. We just time-out the credential update session in the background if they navigate away. So we could probably change the UI and just rely on that background garbage collect instead if you think the UI can be improved greatly :)

weijiangan: I will also see if I can cancel when the user navigates away... for science

Firstyear (dismissed):

Firstyear:
I'm going to exercise some maintainer power here and yolo merge this because I want to allow @weijiangan to proceed while @ToxicMushroom is on break. Since this is all gated behind a feature flag I'm happy to do so.

Firstyear (approved):

Issue #211, opened 2020-04-29
SAML/SCIM


Firstyear:
A large number of application as a service providers support SAML and SCIM for SSO and retriving identities. Scim is similar to kanidm's native proto and it's data model, so would be an easy retrofit to improve the system and what we expose. SAML also works well with kani able to provide claims and more as an IDP.

We should provide both.

Firstyear:
This is actually mostly done, it's just adding in all the end points now that we have to finish up.

compgeniuses:
Nice wonderful, just myself back in this link SMH form old discussions.

FreeSynergy:
Is SCIM not fully implemented? I always thought it was …

Firstyear:
I'd say it's 80% done - you can access most things via scim and we're filling in gaps as we find them.

So I guess for the "intent" of this issue, it's done, it exists, it works, and overtime it will continue to improve.

everhardt:
Am I right that this issue is about Kanidm as SCIM server, ie. other parties provisioning identities on Kanidm? 

I'm looking for the reverse. I use Kanidm as IdP with OIDC to allow SSO at service providers. At some of these service providers we pay per seat and when I remove an account in Kanidm, that account is not removed at the service provider. If Kanidm would be a SCIM client, it could also deprovision those accounts at the service providers. If this issue does not cover that use case: would it be okay to open a new issue for it?

yaleman:
Yes, it's inbound-only to Kanidm _currently_.

candlerb:
AFAICS, the handbook only mentions SCIM in a few places, and mostly in passing.

* Under [Supported Features > Data Import](https://kanidm.github.io/kanidm/stable/supported_features.html?highlight=scim#data-import)
* API is listed as [not stable](https://kanidm.github.io/kanidm/stable/support.html?highlight=SCIM#api-stability)
* Under [Entry Management](https://kanidm.github.io/kanidm/stable/entry_management.html?highlight=scim#entry-management), as a way of importing users/groups with (H)JSON files pushed out with e.g. ansible or salt.

However, it's not clear to me how to connect to provision accounts by tracking an upstream system like Entra or Google Workspace, presumably via SCIM REST endpoints. Is it even possible? If so, could some documentation be added on how to sync kanidm to these systems via SCIM?

I suggest a SCIM section under the [Sync](https://kanidm.github.io/kanidm/stable/sync/concepts.html) heading.

Firstyear:
[...]

PR #2933, opened 2024-07-26
Add scim proto to kanidm, refactor to improve serde performance.

Firstyear:
Relates #211 - import the scim proto library and improve it with better serde handling. Refactor the uses of scim to match. This will make scim a lot easier to implement in the *very* near future. 

Checklist

- [  ] cargo clippy has been run

### DIFF
--- a/libs/scim_proto/src/user.rs
+++ b/libs/scim_proto/src/user.rs
@@ -1,0 +1,196 @@
+use crate::ScimEntry;
+use base64urlsafedata::Base64UrlSafeData;
+use std::fmt;
+use url::Url;
+use uuid::Uuid;
+
+use serde::{Deserialize, Serialize};
+
+#[derive(Serialize, Deserialize, Debug, Clone)]
+#[serde(rename_all = "camelCase")]
+struct Name {
+    // The full name including all middle names and titles
+    formatted: Option<String>,
+    family_name: Option<String>,
+    given_name: Option<String>,
+    middle_name: Option<String>,
+    honorific_prefix: Option<String>,
+    honorific_suffix: Option<String>,
+}
+
+/*
+// https://datatracker.ietf.org/doc/html/rfc7231#section-5.3.5
+//
+// https://www.iana.org/assignments/language-subtag-registry/language-subtag-registry
+// Same as locale?
+#[derive(Serialize, Deserialize, Debug, Clone)]
+enum Language {
+    en,
+}
+*/
+
+// https://datatracker.ietf.org/doc/html/rfc5646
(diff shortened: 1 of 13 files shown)


libs/scim_proto/src/user.rs
```diff
@@ -0,0 +1,196 @@
...
+            Locale::en_US => write!(f, "en-US"),
+            Locale::de => write!(f, "de"),
+            Locale::de_DE => write!(f, "de-DE"),
+        }
+    }
+}
+
+#[allow(non_camel_case_types)]
+#[derive(Serialize, Deserialize, Debug, Clone)]
+enum Timezone {
```
yaleman, line 60: We should probably use [chrono-tz](https://docs.rs/chrono-tz/latest/chrono_tz/enum.Tz.html) instead of reimplementing it ourselves

Firstyear: Oh I didn't know that was a thing lol. I might do that in another PR since the current tz behaviour isn't really relied on anywhere.

server/lib/src/idm/scim.rs
```diff
@@ -548,7 +548,15 @@ impl<'a> IdmServerProxyWriteTransaction<'a> {
...
         sse: &'b ScimSyncUpdateEvent,
         changes: &'b ScimSyncRequest,
-    ) -> Result<(Uuid, BTreeSet<String>, BTreeMap<Uuid, &'b ScimEntry>, bool), OperationError> {
+    ) -> Result<
+        (
+            Uuid,
+            BTreeSet<String>,
+            BTreeMap<Uuid, &'b ScimEntryGeneric>,
+            bool,
+        ),
```
yaleman, line 557: That's a truly horrifying result type 😄

libs/scim_proto/src/user.rs
yaleman: Yeah, was like "HEY WHAT HOW IS THIS NOT AN IMPORT"

server/lib/src/idm/scim.rs
Firstyear: I think it could be changed, but it's been that way for a while.

libs/scim_proto/src/user.rs
Firstyear: Because I'd never found something that could do it :) 

Also FML the fact we need chrono AND time to do ... time. WHY.

yaleman (approved):

server/lib/src/idm/scim.rs
yaleman: // TODO: less crimes

PR #2942, opened 2024-07-31
Optimized replication trigger

Seba-T:
Well in my defense I have empirical data showing the average replication throughput is 80% lower than the previous approach with a moderate server load. I still am waiting for the max throughput results 

### DIFF
--- a/proto/src/constants.rs
+++ b/proto/src/constants.rs
@@ -43,13 +43,13 @@ pub const DEFAULT_LDAP_LOCALHOST: &str = "localhost:636";
 pub const DEFAULT_REPLICATION_ADDRESS: &str = "127.0.0.1:8444";
 pub const DEFAULT_REPLICATION_ORIGIN: &str = "repl://localhost:8444";
 
-/// Default replication poll window in seconds.
+/// Default replication poll interval in seconds.
 pub const DEFAULT_REPL_TASK_POLL_INTERVAL: u64 = 15;
 
-/// Minimum replication poll window in seconds:
+/// Minimum replication poll interval in seconds:
 pub const MIN_REPL_TASK_POLL_INTERVAL: u64 = 3;
 
-// Estimate of max write ops that can be processed in the corresponding replication poll window
+// Estimate of max write ops that can be processed in the corresponding replication poll interval
 pub const MAX_WRITE_THROUGHPUT_BY_TASK_POLL_INTERVAL: [(u64, u64); 13] = [
     (2945, 3),
     (4081, 4),
@@ -66,7 +66,7 @@ pub const MAX_WRITE_THROUGHPUT_BY_TASK_POLL_INTERVAL: [(u64, u64); 13] = [
     (15893, 15),
 ];
 
-// This represents the percentage of the current MAX_WRITE_THROUGHPUT (determined by the current replication window)
+// This represents the percentage of the current MAX_WRITE_THROUGHPUT (determined by the current replication interval)
 // above which we will not trigger replication as we are too close to the maximum throughput and we risk to lower the server performance
 pub const REPLICATION_TRIGGER_SAFETY_THRESHOLD: f64 = 0.8;
 
--- a/proto/src/constants.rs
+++ b/proto/src/constants.rs
@@ -46,6 +46,30 @@ pub const DEFAULT_REPLICATION_ORIGIN: &str = "repl://localhost:8444";
 /// Default replication poll window in seconds.
 pub const DEFAULT_REPL_TASK_POLL_INTERVAL: u64 = 15;
 
+/// Minimum replication poll window in seconds:
+pub const MIN_REPL_TASK_POLL_INTERVAL: u64 = 3;
+
+// Estimate of max write ops that can be processed in the corresponding replication poll window
+pub const MAX_WRITE_THROUGHPUT_BY_TASK_POLL_INTERVAL: [(u64, u64); 13] = [
+    (2945, 3),
+    (4081, 4),
+    (5294, 5),
+    (6351, 6),
+    (7408, 7),
+    (8280, 8),
+    (9105, 9),
+    (10476, 10),
+    (11919, 11),
+    (12395, 12),
+    (12770, 13),
+    (14293, 14),
+    (15893, 15),
+];
+
+// This represents the percentage of the current MAX_WRITE_THROUGHPUT (determined by the current replication window)
+// above which we will not trigger replication as we are too close to the maximum throughput and we risk to lower the server performance
+pub const REPLICATION_TRIGGER_SAFETY_THRESHOLD: f64 = 0.8;
+
 /// Default grace window for authentication tokens. This allows a token to be
 /// validated by another replica before the backing database session has been
 /// replicated to the partner. If replication stalls until this point then
--- a/proto/src/constants.rs
+++ b/proto/src/constants.rs
@@ -66,7 +66,7 @@ pub const MAX_WRITE_THROUGHPUT_BY_TASK_POLL_INTERVAL: [(u64, u64); 13] = [
     (15893, 15),
 ];
 
-// This represents the percentage of the current MAX_WRITE_THROUGHPUT (determined by the current replication interval)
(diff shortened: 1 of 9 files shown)


server/core/src/repl/mod.rs
```diff
@@ -363,11 +368,60 @@ async fn repl_run_consumer(
...
 
+#[instrument(level="info", skip(idms), fields(eventid=Uuid::new_v4().to_string()))]
+async fn replication_trigger(
+    idms: &IdmServer,
+    last_replication_instant: &Instant,
+    enable_experimental_replication_trigger: bool,
+    replication_interval: Duration,
+) {
+    tokio::select! {
+        _ = async {
```
Firstyear, line 379: This should probably be a function instead, it's really hard to parse this.

server/lib/src/be/idl_arc_sqlite.rs
```diff
@@ -58,6 +59,7 @@ pub struct IdlArcSqlite {
     op_ts_max: CowCell<Option<Duration>>,
     allids: CowCell<IDLBitRange>,
     maxid: CowCell<u64>,
+    write_ops_since_last_repl: CowCell<u64>,
```
Firstyear, line 62: This shouldn,t be here, it should be in the QueryServer instead.

server/lib/src/plugins/write_ops_counter.rs
```diff
@@ -0,0 +1,207 @@
...
+        Self::reset_write_ops_counter(qs);
+        Ok(())
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::prelude::*;
+
+    #[qs_test(domain_level=DOMAIN_LEVEL_7)]
```
Firstyear, line 103: Don't need the domain_level set here.

server/core/src/repl/mod.rs
```diff
@@ -363,11 +368,60 @@ async fn repl_run_consumer(
...
+            let mut one_sec_int = interval(Duration::from_secs(1));
+            // sleep for the minimum window size before we do anything
+            sleep(Duration::from_secs(MIN_REPL_TASK_POLL_INTERVAL)).await;
+            loop {
+                let write_ops_since_last_repl = read_txn.qs_read.get_write_ops_since_last_repl();
+                let time_since_last_repl = last_replication_instant.elapsed().as_secs();
+
+                let Some(write_throughput_index) = time_since_last_repl.checked_sub(MIN_REPL_TASK_POLL_INTERVAL) else {
+                    error!("Negative replication window index detected, aborting...");
+                    return;
```
Firstyear, line 391: We should handle this error better than returning, because a return here will kill the repl task I think?

Firstyear: Just do saturating sub instead.

server/core/src/repl/mod.rs
```diff
@@ -363,11 +368,60 @@ async fn repl_run_consumer(
...
+                };
+
+                let (current_max_throughput, current_repl_window) = MAX_WRITE_THROUGHPUT_BY_TASK_POLL_INTERVAL[write_throughput_index as usize];
```
Firstyear, line 394: No check on the bounds here which could cause out of bounds crash. Use .get() instead.

proto/src/constants.rs
```diff
@@ -46,6 +46,30 @@ pub const DEFAULT_REPLICATION_ORIGIN: &str = "repl://localhost:8444";
...
+pub const MIN_REPL_TASK_POLL_INTERVAL: u64 = 3;
+
+// Estimate of max write ops that can be processed in the corresponding replication poll window
+pub const MAX_WRITE_THROUGHPUT_BY_TASK_POLL_INTERVAL: [(u64, u64); 13] = [
+    (2945, 3),
+    (4081, 4),
+    (5294, 5),
+    (6351, 6),
+    (7408, 7),
+    (8280, 8),
```
Firstyear, line 59: Given this appears linear, it could be a function instead of a constant.

Firstyear (changes requested):
It's a bit messy in it's current form - I think it can be simpler and cleaner. A lot of your tests appear redundant too?

proto/src/constants.rs
yaleman: How was this calculated, surely it's heavily dependent on system architecture etc?

server/core/src/repl/mod.rs
Seba-T: Wopsies hadn't thought of that :p. I was too enthused by finally understanding the whole tokio::select! business :D

server/lib/src/plugins/write_ops_counter.rs
Seba-T: I must admit I blindly copied & pasted from the other tests, not having a clue how the macro works :D

server/core/src/repl/mod.rs
Seba-T: > We should handle this error better than returning, because a return here will kill the repl task I think?

I mean I haven't tested it but I assumed it would just immediately return the future? I'd expect the repl loop to resume normally. Actaully if returning kills the entire repl task, then also this [return](https://github.com/Seba-T/kanidm/blob/53a866bdc0a8b6d47c5aa7c8a1f7a3239df3b12e/server/core/src/repl/mod.rs#L414) should kill it, but I'm pretty sure it doesn't since I've tested it.

server/core/src/repl/mod.rs
Seba-T: wopsies

proto/src/constants.rs
Seba-T: > How was this calculated, surely it's heavily dependent on system architecture etc?

The numbers come from our benchmarks, but you're right in saying they are ... not consistent across different machines. But I have another secret plan to fix that ;-). For now I think it makes sense to just assume they're correct and polish everything else.

Seba-T: > Given this appears linear, it could be a function instead of a constant.

Sure, but I think we first need the big chonky benchmarks we were discussing yesterday before interpolating the points. Also we wouldn't need to compute weird indexes as we're currently doing, it would be much more elegant!

Seba-T:
> It's a bit messy in it's current form - I think it can be simpler and cleaner. A lot of your tests appear redundant too?

I think they all can be coalesced into one per section, i.e one repl test, one plugin test, I kept them separate because I thought it was easier to understand, but maybe I should merge them?

server/lib/src/be/idl_arc_sqlite.rs
Seba-T: sure!

proto/src/constants.rs
```diff
@@ -43,9 +43,33 @@ pub const DEFAULT_LDAP_LOCALHOST: &str = "localhost:636";
...
+    (11919, 11),
+    (12395, 12),
+    (12770, 13),
+    (14293, 14),
+    (15893, 15),
+];
+
+// This represents the percentage of the current MAX_WRITE_THROUGHPUT (determined by the current replication interval)
+// above which we will not trigger replication as we are too close to the maximum throughput and we risk to lower the server performance
+pub const REPLICATION_TRIGGER_SAFETY_THRESHOLD: f64 = 0.8;
```
yaleman, line 71: So we're not going over 0.8% throughput? I'm guessing you mean the *ratio* of `MAX_WRITE_THROUGHPUT` rather than percentage?

Seba-T: Yess it's definitely not 0.8% :D. It should say ratio instead.

Firstyear:
So far I'm skeptical of the possible gains here because as @yaleman said these numbers are specific to the system you benchmarked on, and your benchmarks so far aren't consistent or widely sampled enough to make a firm assessment of an optimal value. I think we delay this until we have more conclusive evidence to support it's inclusion.

proto/src/constants.rs
yaleman: >  For now I think it makes sense to just assume they're correct and polish everything else.

I'll have to disagree on that one, for now.

Firstyear: As do I.

Firstyear:
Closing as I think this isn't needed.

PR #2943, opened 2024-07-31
Update to 1.4.0-dev

Firstyear:
Checklist

- [ x ] `cargo fmt` has been run
- [ ] `cargo clippy` has been run
- [ x ] `cargo test` has been run and passes

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1213,7 +1213,7 @@ dependencies = [
 
 [[package]]
 name = "daemon"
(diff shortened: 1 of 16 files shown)


Firstyear:
So the big "gimmick" here is since we don't do skip upgrades now, we can start to prune out older domain level migrations and older static entries and acps that we don't need.

yaleman (approved):
Wewt.

PR #2685, opened 2024-03-25
Add Grafana integration to OAuth2 documentation

pdostal:
Fixes #
- Adds Grafana integration to OAuth2 documentation
  Significant part of the snippet is from @matthewpi
  As promised in #2597

Checklist

- [ ] This pr contains no AI generated code
- [ ] cargo fmt has been run
- [ ] cargo clippy has been run
- [ ] cargo test has been run and passes
- [x] book chapter included (if relevant)
- [ ] design document included (if relevant)

PR #2950, opened 2024-08-03
docs: Fix typo for oauth2 grafana claim-map example

polarathene:
The docs are a bit out of date in some places.

This fixes a typo, although I'm not sure if it was intentional to omit a line for the `grafana_users` group either?

---

FWIW it's been a little frustrating to try get a basic setup going (_I'm not integrating with Grafana in this case_). The enforced requirement for TLS (_with no opt-out support?_) doesn't help as I'm attempting to do this on Windows 11 (host with web browser) and a completely offline self-hosted setup in WSL2 with Docker.

I've got certs provisioned but may need to add in custom DNS since the containers can't leverage host networking mode, and localhost/loopback is not equivalent to the host.

---

**Known issues in docs:**
- I recall one of the OIDC/OAuth2 URLs documented was incorrect, I think it was the auth URI missing `/ui` path prefix (before `/oauth`), which I found by querying the OIDC client well-known endpoint to get the correct URI.
- Another was with the redirect uri command, this was renamed to landing uri.
- The tools example with docker mounted config to `/home` dir but the container has no such user in `/etc/passwd` thus you can't run the container with that user for that to be relevant.

Might have been others.

I'm open to contributing doc updates for those provided I can get `kanidm` to work the way I would like to (_`compose.yaml` with `docker-mailserver` + `roundcube`, with minimal commands to pair with an IDP outside of the config file itself_). I'll open a discussion with config for reproduction and advice if I get stuck.

## Checklist

### DIFF
--- a/book/src/integrations/oauth2/examples.md
+++ b/book/src/integrations/oauth2/examples.md
@@ -169,7 +169,7 @@ kanidm group create 'grafana_users'
 Setup the claim-map that will set what role each group will map to in Grafana:
 
 ```bash
-kanidmm oauth2 update-claim-map-join 'grafana' 'grafana_role' array
+kanidm system oauth2 update-claim-map-join 'grafana' 'grafana_role' array
 kanidm system oauth2 update-claim-map 'grafana' 'grafana_role' 'grafana_superadmins' 'GrafanaAdmin'
 kanidm system oauth2 update-claim-map 'grafana' 'grafana_role' 'grafana_admins' 'Admin'
 kanidm system oauth2 update-claim-map 'grafana' 'grafana_role' 'grafana_editors' 'Editor'


yaleman:
The Grafana example was community contributed and  omissions are likely since we aren’t Grafana admins. 

TLS isn’t up for discussion, per the FAQ.

polarathene:
> The Grafana example was community contributed and omissions are likely since we aren’t Grafana admins.

I understand. The fix I've contributed is specific to the Grafana guide. However, the other noted issues with the docs **aren't related to Grafana** as they're covered on other pages in the docs.

That Grafana section does have another issue specific to the kanidm guidance that is completely agnostic to the Grafana context, which I would have thought to be glaringly obvious to someone who is far more familiar with `kanidm`. It was missed during review (_[original PR](https://github.com/kanidm/kanidm/pull/2685#discussion_r1681284456)_) I think? (happens)

The guide uses `update-scope-map` before `grafana_users` group is created, which the CLI I don't think had any error for, but it wasn't until after the group was created and I ran that same `update-scope-map` command again that the client actually had scopes applied (_I had been confused why the logged in user page had "No linked applications" until figuring out it was due to that subtle mistake_).

I realize I tend to come across as a bit blunt / crass, but I'm responding purely focused on concerns themselves. No ill-will or judgement towards any project contributors or maintainers. I maintain projects myself, these issues with docs aren't uncommon to miss updating as things change, or not spotting problems during review like the one I mentioned 👍 

---

> TLS isn’t up for discussion, per the FAQ.

The FAQ isn't convincing tbh.

Verbose justification follows, but **TLDR** is kanidm doesn't need to enforce TLS for secure cookies to work, nor does it necessarily provide much value by enforcing it between an LB/RP.

> https://kanidm.github.io/kanidm/stable/frequently_asked_questions.html#can-kanidm-work-without-tls
>
> _**Can Kanidm work without TLS?**_
> _No, it can not. TLS is required due to our use of secure-cookies. secure-cookies is a flag set in cookies that asks a client to transmit them back to the origin site if and only if the client sees HTTPS is present in the URL._
> 
> _If you do NOT have a HTTPS URL, the cookie with the session-id is not transmitted. The server detects this as an invalid-state request in the authentication design, and immediately breaks the connection, **because it appears insecure**._

> _Simply put, we are trying to use settings like secure_cookies to add constraints to the server so that you must perform and adhere to best practices - such as having TLS present on your communication channels._
>
> _This is also why we do not allow the server to start without a TLS certificate being configured._

> https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#block_access_to_your_cookies
>
> _A cookie with the Secure attribute is only sent to the server with an encrypted request over the HTTPS protocol. It's never sent with unsecured HTTP (except on localhost), which means [man-in-the-middle](https://developer.mozilla.org/en-US/docs/Glossary/MitM) attackers can't access it easily._

For reference [here's a copy/paste secure-cookie example](https://webbureaucrat.gitlab.io/articles/setting-and-reading-session-cookies-in-rust-with-actix-web/) with Actix. With [insomnia](https://insomnia.rest/download) I can then `POST` to `/set_session` with JSON `{ "message": "cookie example" }` and a follow-up `GET` request to `/get_session` over HTTPS will return that message in the response, but won't over HTTP.

No surprises there. But I want to point out that HTTPS for secure cookie feature is only required from the client to server connection. A reverse proxy can handle that in front of your service, and this still works even with the service itself only being accessed via HTTP internally on the server from the reverse proxy.

And yes, perhaps you even know that? The other section states:

> https://kanidm.github.io/kanidm/stable/frequently_asked_questions.html#why-disallow-http-without-tls-between-my-load-balancer-and-kanidm
>
> _**Why disallow HTTP (without TLS) between my load balancer and Kanidm?**_
> _Because Kanidm is one of the keys to a secure network, and insecure connections to them are not best practice. This can allow account hijacking, privilege escalation, credential disclosures, personal information leaks and more. The entire path between a client and the server must be protected at all times._

That may be valid in some scenarios, but when both the reverse proxy and service are on the same node, when are those concerns valid from the internal HTTPS / TLS connection between the two vs without? If the attacker has compromised the system itself, I'm not sure if you're getting in pragmatic protection?

---

Caddy docs also share this opinion:

> https://caddyserver.com/docs/caddyfile/directives/reverse_proxy#examples
>
> _Reverse proxy to an HTTPS upstream, but [⚠️ disable TLS verification](https://caddyserver.com/docs/caddyfile/directives/reverse_proxy#tls_insecure_skip_verify). This is NOT RECOMMENDED, since it disables all security checks that HTTPS offers; **proxying over HTTP in private networks is preferred if possible**, because it **avoids the false sense of security**:_
> 
> ```
> example.com {
> 	reverse_proxy 10.0.0.1:443 {
> 		transport http {
> 			tls_insecure_skip_verify
> 		}
> 	}
> }
> ```

I've seen advice to use self-signed certs (_can't quite recall all the links, but there was [this discussion](https://github.com/kanidm/kanidm/pull/2731#discussion_r1586036831)_) just as a way to make kanidm happy with it's enforced TLS. When the CA isn't public, or the cert is specifically internal for this purpose, you'll also find some skipping verification on the chain of trust like that reverse proxy example above, which is technically fine if the CA isn't a separate entity like LetsEncrypt? 🤷‍♂️ (_it's just questionable at that point what actual security benefit you're getting_)

I'm all for secure by default with TLS expected, but I think there are valid scenarios where it should be possible to opt-out. As clarified earlier, kanidm does not need to enforce TLS for secure cookies to work, so long as the client device reached the server over HTTPS.

Enforcing it just adds extra complexity when it's not providing a security benefit which is a known contributor to weakening security (_take traditional passwords for example with policies that enforce letters, numbers, special chars etc when you can have strong passwords with only lower case a-z_).

I'd **strongly urge you to reconsider the TLS stance**. It's not the first time I've seen security practices pushed when they're not necessary (_NIST and similar orgs advising RSA-3072 for compliance when RSA-2048 is perfectly safe for the majority, misinformed users advising RSA-4096 or more instead of ECC, misconceptions with  passwords, etc_).

yaleman:
> > TLS isn’t up for discussion, per the FAQ.
> The FAQ isn't convincing tbh.

It's not up for discussion, so writing treatise on why it's not convincing only wastes everyone's time. Please respect our decision.

Firstyear:
> Verbose justification follows, but **TLDR** is kanidm doesn't need to enforce TLS for secure cookies to work, 

It is needed https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#block_access_to_your_cookies

> nor does it necessarily provide much value by enforcing it between an LB/RP.
>
> I've seen advice to use self-signed certs (_can't quite recall all the links, but there was [this discussion](https://github.com/kanidm/kanidm/pull/2731#discussion_r1586036831)_) just as a way to make kanidm happy with it's enforced TLS. When the CA isn't public, or the cert is specifically internal for this purpose, you'll also find some skipping verification on the chain of trust like that reverse proxy example above, which is technically fine if the CA isn't a separate entity like LetsEncrypt? 🤷‍♂️ (_it's just questionable at that point what actual security benefit you're getting_)

You can use a self-signed CA, just ensure you do CA pinning and verification. If you don't do verification, you're on your own.

> I'm all for secure by default with TLS expected, but I think there are valid scenarios where it should be possible to opt-out. 

Something you don't know here is the history. We have already seen users shoot themselves in the foot by allowing non-TLS setups which led us to enforce TLS only. We aren't going to go backwards on this. 

More likely we will even be *stricter* in future that load balancers shouldn't even terminate your TLS - they should be a TCP balancer only. This will especially be required for PIV.

> The FAQ isn't convincing tbh.

We aren't here to convince you. Our decision is made.

> > TLS isn’t up for discussion, per the FAQ.

polarathene:
## Clarification - Cookie `secure` flag + HTTPS (required only between client and server, not service)

> > Verbose justification follows, but **TLDR** is kanidm doesn't need to enforce TLS for secure cookies to work,
> 
> It is needed https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#block_access_to_your_cookies

No it's not...? I already explained why above (_when I linked to that exact same resource_):

> image

---

## Clarification - TLS between two processes on the same host is redundant?

> You can use a self-signed CA, just ensure you do CA pinning and verification. If you don't do verification, you're on your own.

Sorry, I think there might be a misunderstanding. The self-signed CA was regarding a workaround for satisfying kanidm in a setup where it makes little sense. Perhaps some additional context would help?

For my local dev setup, I provisioned a cert for two FQDNS:
- `auth.example.localhost` (_for Windows 11 host with the web browser_)
- `auth.example.test` (_for Docker containers communicating directly in their subnet_).
- In production, only `auth.example.com` would be used with a public CA provisioned cert.

While kanidm was configured for `auth.example.localhost` as the `domain`.. my reverse proxy (Caddy) was the one managing any actual TLS connections for both domains the cert was provisioned for. Both resolve to the container running Caddy, whilst Caddy would then forward traffic to kanidm (_running in it's own container_) at hostname + port `kanidm-server:8443`.

Process A (Caddy with TLS) terminates TLS and forwards request to Process B (kanidm server), on the same host / instance you control. What concern is mandating TLS resolving in that context? **There's no egress traffic, it seems redundant?** (_it's effectively over loopback_)

Again, I am not advocating to stray from secure by default. Just that opt-out should be supported (_as it [can be for other features like PKCE](https://kanidm.github.io/kanidm/stable/frequently_asked_questions.html?highlight=pkce#why-is-disabling-pkce-considered-insecure) when required_). You still have TLS ensured between client and server traffic, just not between internal processes.

---

## Clarification - User config footguns

> > I'm all for secure by default with TLS expected, but I think there are valid scenarios where it should be possible to opt-out.
> 
> Something you don't know here is the history. We have already seen users shoot themselves in the foot by allowing non-TLS setups which led us to enforce TLS only. We aren't going to go backwards on this.

Was that with TLS by default? They explicitly configured opt-out? That's on them 🤷‍♂️

There is so much about an environment that you don't have control over that they can mess up. You could have an opt-out via feature toggle in `Cargo.toml` so that it's really explicit. While that has increased friction it'd at least be possible to opt-out when it makes sense to.

You allow opt-out with other features that are security focused, the docs have clear warnings discouraging that, but sometimes it's required. Needing to provide kanidm with certs my reverse proxy manages, or provision self-signed certs just to satisfy the requirement to only redundantly establish a TLS connection again is bizarre to enforce.

> More likely we will even be _stricter_ in future that load balancers shouldn't even terminate your TLS - they should be a TCP balancer only. This will especially be required for PIV.

That removes all the benefits that Caddy and Traefik can offer with Layer 4 features, and requires having PROXY protocol support for Layer 7 traffic forwarded to kanidm if you want the original client IP to be preserved? (_I'm not 100% sure since it's HTTPS, but with other protocols like mail servers this has been the case since the only context they have for client IP is the proxy IP at that point otherwise_)

---

## Summary

> > The FAQ isn't convincing tbh.
> 
> We aren't here to convince you. Our decision is made.

Yeah, you've made that very clear.

My feedback on the matter was to clarify why you insisted TLS was mandatory. It appears to be enforced solely on:
- Cookies with `secure` flag (_but you've misunderstood this requirement_)
- Users deploying insecurely due to not using TLS when they should (_not surprising.. good justification for TLS by default, but not for absolute refusal to support opt-out_)

I do not expect a reply, this discussion does not need to continue if you don't value my input (_which engagement thus far seems to convey_).

---

## Feedback - Disappointment

I would like to point out this is the 2nd attempt to contribute positively to the project. While these contributions are minor, the maintainer engagement has typically been defensive / negative. Perhaps you do not want / encourage third-party contributions, as this experience has been the opposite of welcoming / friendly.

My first contribution was closed as @yaleman chose to bundle it into another PR he made afterwards. No appreciation or credit was given, as my PR was replaced, any attribution from the git commit is also neglected.

This is [about to happen a 2nd time](https://github.com/kanidm/kanidm/pull/2961/files#diff-cf68dcd8a84a0bf00f2512f01586fd0883fe8ad24d65c6fcfbcbe849ef12d162R172) (_that PR does not cover the additional docs issues I identified btw_).

As such, while I know this matters little to you, any future feedback and contributions I had in mind for kanidm will not be pursued as they do not appear valued. Due to these interactions thus far, I cannot speak positively about the project (despite technical merits) nor do I have interest to contribute integration guides to kanidm docs or other projects as a result.

While I feel that this experience of mine shouldn't have been the case given the [Code of Conduct](https://github.com/kanidm/kanidm/blob/master/CODE_OF_CONDUCT.md#our-standards) the project has... from similar experiences at other projects I wouldn't be surprised if this comment is viewed as a violation and the CoC itself will be used to deem removal of this comment whilst disregarding your own commitments to adhere to CoC standards.

I truly hope that doesn't occur, but I'm not sure how to otherwise interpret the way my contributions and feedback are being handled.

Firstyear:
> Again, I am not advocating to stray from secure by default. Just that opt-out should be supported (_as it [can be for other features like PKCE](https://kanidm.github.io/kanidm/stable/frequently_asked_questions.html?highlight=pkce#why-is-disabling-pkce-considered-insecure) when required_). You still have TLS ensured between client and server traffic, just not between internal processes.

We already told you that we used to support the opt-out, and we also told you that it caused no end of support problems for us.

Second, the difference in security risk between "letting people disable PKCE" vs "allow plaintext traffic where auth tokens can be stolen" is very large. We don't make these choices without context and experience. The internal communication security is *just as important* as client to server.

> ## Feedback - Disappointment
> 
> I would like to point out this is the 2nd attempt to contribute positively to the project. While these contributions are minor, the maintainer engagement has typically been defensive / negative. Perhaps you do not want / encourage third-party contributions, as this experience has been the opposite of welcoming / friendly.

You spent a lot of time trying to convince us that we were wrong, about choices we already made. Of course we're on the defensive.

> While I feel that this experience of mine shouldn't have been the case given the [Code of Conduct](https://github.com/kanidm/kanidm/blob/master/CODE_OF_CONDUCT.md#our-standards) the project has... from similar experiences at other projects I wouldn't be surprised if this comment is viewed as a violation and the CoC itself will be used to deem removal of this comment whilst disregarding your own commitments to adhere to CoC standards.

While our responses may not have been as "welcoming" as they could have been, we have experience abuse and hostility from community members in similar situations and with similar communication styles. As a result, the way you interacted with us, without knowing it, was setting of alarms for us that you would also become hostile and abusive. That's on us that we should adjust our radar on those matters and we should do better to give the benefit of the doubt.

However, as you stated yourself "I realize I tend to come across as a bit blunt / crass, but I'm responding purely focused on concerns themselves. ". If you put out bluntness, people are conditioned to respond in kind. Giving us a lecture on your views and how we were wrong about TLS, certainly didn't help us feel like you were trying to contribute in a positive and considered manner.

So while we may not have been as welcoming as we could have, we have not discriminated against you, levelled personal insults, harassed you, published your private information, or other conduct that would not pass in a professional environment (at least not that I am aware of. If we have, then I'd like to know).

And of course, while we didn't accept your PR's we do appreciate that you raised the issues and tried to get them fixed, even if communication was difficult. I'm happy to accept you in the CONTRIBUTORS.md file if you want since you have contributed.

polarathene:
## Misunderstandings / Miscommunication?

> We already told you that we used to support the opt-out, and we also told you that it caused no end of support problems for us.

No. You did not state you supported opt-out. Just that TLS was optional previously. I lack the context if it was TLS by default, or if your actual issue was TLS was opt-in.

---

> The internal communication security is _just as important_ as client to server.

I would appreciate an example where that is relevant. The scenario was `process-A` => (_HTTP request `127.0.0.1:80`_) => `process-B`, no browser involved just two processes headless.

Could you please explain the value of TLS in that scenario? What attack are you preventing where the bad actor could not do far worse?

I don't understand how that is just as important as the value of TLS across hosts via network (be that public or internal LAN). This is two processes on the same host, that's a major difference.

---

## Convincing vs Informing

> You spent a lot of time trying to convince us that we were wrong, about choices we already made. Of course we're on the defensive.

Neither my prior PR or this PR (actual content to merge) are related to the TLS discussion that followed here.

- [My first PR](https://github.com/kanidm/kanidm/pull/2948#discussion_r1702452974) was for `VOLUME` directive in `Dockerfile` being a bad practice, with ample justification provided to demonstrate why.
  - Yes you were wrong with that (_acknowledged by merging a separate PR that copied my contribution.. instead of the one I provided?_)
- This PR was specifically a typo fix to docs, with mention of various other outdated info in the docs I ran into while setting up `kanidm` such as incorrect commands (renamed), invalid URLS (renamed), incorrect order of steps (user contributed guide).
  - I additionally expressed some frustration as feedback alongside those issues, where the enforced TLS was only complicating my troubleshooting (_I did get a successful setup, but only after I had success with `rauthy` and Authelia_).

Beyond that.. yeah you are also wrong about `secure` attribute with cookies and I backed that up with evidence. Convincing you isn't necessary when what is disputed is demonstratable that your statement was misguided. It's ok to be defensive, but silly to dismiss evidence.

Anything else regarding your stance on TLS is your choice for the project. I explained why I disagree, but that decision of yours for `kanidm` is a choice, I'm not going to fight you on that.

I am genuinely open towards constructive criticism (_as your CoC encourages_), so when users engage with me on projects I maintain I am open to being wrong (_I love it when I am, it's a win-win!_), I don't see a reason to get defensive and dismiss a contributor as I was treated here.

---

## Communication and Contributors

> That's on us that we should adjust our radar on those matters and we should do better to give the benefit of the doubt. 

Apologies on my part. I do have that effect on people with my communication style, it's unintentional and I'm not quite sure how I can improve on that beyond investing more time in my responses to be less verbose.

> If you put out bluntness, people are conditioned to respond in kind.

Fair.

I am absolutely fine with you being blunt back, what I did not like was how my initial experience contributing to this project is resulting in taking my contributions over into another PR and closing them without any attribution like other contributors are given.

Just because you were given the impression that I was similar to negative experiences you've had in the past does not justify treating my contributions (however minor) in that manner. If you're going to accept and merge them indirectly (_as has been done already_), the very least that could be done is a "thanks!" or comment about why the contribution was bundled into another PR instead of providing actionable feedback when needed. Just taking a contribution like that and closing the original quietly does the opposite of encouraging more improvements.

I have contributed small to projects in the past, where due to the positive experience I returned to provide bigger contributions. In one particular project I became a valuable maintainer for several years even though I don't actually use the software. While neither of us are strangers to users with bad and ungrateful/demanding attitudes, we should try to show contributors how we'd like to engage (positively/welcoming).

I provided a disclaimer early on about my awareness of how my communication can be perceived unintentionally. I had no such context with your responses, or that you were responding in kind due to assumptions of my character. That just communicated to me that this is what interacting with maintainers of the project would be like if I were to continue contributing, whilst I was having more positive engagement with alternative projects when I had feedback to contribute.

---

> If we have, then I'd like to know

It seems that we were thinking of opposite concerns to adhere to in the CoC. This is what I had in mind:

https://github.com/kanidm/kanidm/blob/2e3b09ec8a4e3fddfd84a4502287cde6d7cf9852/CODE_OF_CONDUCT.md?plain=1#L13-L17

I appreciate that you took the time to provide your response. Likewise I made my own assumptions from my own past negative experiences engaging with maintainers where I identified concerns and expressed an opinion, but that conflicted with the maintainers stance where they became very defensive, dismissive, and resorted to threats about CoC violations - rather than acknowledging the contradicting yet valid feedback (_not unlike with the cookies `secure` attribute example above_).

## Right / Wrong vs Correct

For clarity, my interest here was not in about you being wrong and me being right. It was about correctness. I wanted to understand that TLS decision (_the technical reasons behind it_), so that I could correct my own knowledge on the topic otherwise your own. If that leads to you reconsidering your stance on opt-out support, great, but I'm not concerned either way as I'm not invested into the project, just correctness.

polarathene:
> And of course, while we didn't accept your PR's we do appreciate that you raised the issues and tried to get them fixed, even if communication was difficult. I'm happy to accept you in the CONTRIBUTORS.md file if you want since you have contributed.

Thanks but that won't be necessary. The contributions are minor. Gratitude and your response as a whole is sufficient.

I just hope that the feedback here helps the `kanidm` project with better interactions for future contributors 👍

Firstyear:
> I just hope that the feedback here helps the `kanidm` project with better interactions for future contributors 👍

Thanks, We'll certainly do our best.

PR #2963, opened 2024-08-10
TLS, no seriously.

yaleman:
Checklist


book/src/frequently_asked_questions.md
```diff
@@ -6,14 +6,32 @@
...
+feature you add later.
 
-### Can Kanidm work without TLS?
+> [!NOTE]
+>
+> Please respect our decision. We've had this discussion. A lot.
+>
+> You're at the FAQ because you have questions, and this is the final answer on the matter.
+>
+> If this means you won't use Kanidm, thank you for your consideration.
```
Firstyear, line 19: This is a bit harsh, maybe just reword to:

```
> Please respect the maintainers decision on TLS-by-default. 
```

Alternately a new segment

```
### I think TLS should be opt-out for my use case

We previously did allow this. It created a lot of support issues, confusing configurations and headaches. After a string of problems, we had an extensive discussion about the role of Kanidm and decided that both in the interest of secure defaults *and* single-config for uses, that TLS should be required between all components. 
```

book/src/frequently_asked_questions.md
```diff
@@ -6,14 +6,32 @@
...
-No, it can not. TLS is required due to our use of `secure-cookies`. `secure-cookies` is a flag set
-in cookies that asks a client to transmit them back to the origin site if and only if the client
-sees HTTPS is present in the URL.
+### Why not allow HTTP (without TLS) between my load balancer and Kanidm?
+
+Because Kanidm is one of the keys to a secure network, and insecure connections to them are not best
+practice. This can allow account hijacking, privilege escalation, credential disclosures, personal
+information leaks and more.
+
+We believe that the entire path between a client and the server must be protected at all times.
```
Firstyear, line 27: "This includes the path between load balancers and Kanidm"

book/src/frequently_asked_questions.md
```diff
@@ -24,18 +42,16 @@ detects this as an invalid-state request in the authentication design, and immed
 connection, because it appears insecure. This prevents credential disclosure since the
 authentication session was not able to be established due to the lost session-id cookie.
 
-Simply put, we are trying to use settings like `secure_cookies` to add constraints to the server so
+Simply put, we are trying to use settings like secure cookies to add constraints to the server so
```
Firstyear, line 45: "we are using settings" ...

book/src/frequently_asked_questions.md
```diff
@@ -78,6 +94,12 @@ mitigations, authorisation code interception is not prevented by the presence or
 We would strongly encourage OAuth2 client implementations to implement and support PKCE, as it
 provides defense in depth to known and exploited authorisation code interception attacks.
 
+### Why do you allow disabling PKCE but not TLS?
+
+Because there are still many applications where PKCE is not available and it is not trivial to solve
```
Firstyear, line 99: I think I'd rather focus on "in the case that PKCE is absent on a single oauth2 client, the scope of failure is reduced to that single client. In the case of compromise of an internal network between a load balancer and Kanidm, the attacker can access and steal all tokens we ever issue. "

book/src/frequently_asked_questions.md
```diff
@@ -152,4 +175,4 @@ Don't [ask](https://www.youtube.com/watch?v=0QaAKi0NFkA). They just
 
 ## Why won't you take this FAQ thing seriously?
 
-Look, people just haven't asked many questions yet.
+Look, people just haven't asked many questions yet. Why are you still reading this?
```
Firstyear, line 178: Change the sassy part to "Sorry, there are no easter eggs in this document, but there may be elsewhere 🥚"

CODE_OF_CONDUCT.md
```diff
@@ -23,6 +23,7 @@ Examples of unacceptable behavior by participants include:
 - Public or private harassment
 - Publishing others’ private information, such as a physical or electronic address, without explicit
   permission
+- Refusing to accept decisions or reasonable requests of project maintainers
```
Firstyear, line 26: I think we should make it "arguing with maintainers after being asked to stop".

book/src/frequently_asked_questions.md
```diff
@@ -6,14 +6,32 @@
...
+information leaks and more.
+
+We believe that the entire path between a client and the server must be protected at all times.
+
+### Can Kanidm authentication work without TLS?
+
+No, it can not. TLS is required due to our use of the `Secure` flag our cookies, which requires a
+client to transmit them back to the origin site
+[if and only if the client
+sees HTTPS as the protocol in the URL](https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#security).
```
polarathene, line 34: Consider revising: https://github.com/kanidm/kanidm/issues/2969#issuecomment-2283548851

Perhaps something along the lines of:

```suggestion
### Can Kanidm authentication work without TLS?

TLS is required for some features kanidm uses such as cookies tagged with the `Secure` attribute, which requires a
client to transmit them back to the origin site [if and only if the client sees HTTPS as the protocol in the URL](https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#security). While TLS could be terminated before the request reaches kanidm, we feel it is in the best interests of security to always ensure traffic arrives to kanidm encrypted.
```

CODE_OF_CONDUCT.md
yaleman: I figured this covers both the "we have decided to design it this way" and "plz stop"

book/src/frequently_asked_questions.md
yaleman: Yeah,  didn't want to allow the "but you did it in the past" hole to be opened up.

book/src/frequently_asked_questions.md
polarathene: Was there something you didn't like about this suggestion that it was "resolved" / ignored?

---

I am not sure if you even read the suggestion before dismissing it (_as I assume was the case when my related issue was promptly closed and locked_).

I'm not trying to go against your policy on TLS, just improve this part of the docs. It is inaccurate to say that kanidm must enforce TLS for secure cookies to work.

The suggestion better clarifies that it's only necessary for the client request, but not necessarily how that request reaches kanidm. That is explained at the end, with your actual reasons for why TLS is still required for kanidm.

polarathene (changes requested):
Please actually read the suggestion, this is respecting your decision. I am not trying to debate your stance.

book/src/frequently_asked_questions.md
yaleman: I did read your suggestion and I don't feel the distinction's necessary in this case. Historically, we've found that giving people even the slightest chance that we might roll back the TLS stance leads to discussions like these...

Firstyear: I'm going to shut this down *now*.

We aren't going to keep arguing over minutia like this.

If were were to change it:

"Secure cookies and Webauthn demand secure browser contexts. Secure contexts only exist if the url is https:// or http://localhost exactly. Since no one runs their production IDM on localhost then ergo, ipso facto, therfore, TLS is required."

And we'd be back to TLS is required.

Again, we don't know what you are trying to prove here outside of some hair splitting. Drop it please. We're done with this.

polarathene: There appears to be a lack of reading comprehension involved.
- I am **not trying to convince you** to change your stance on TLS.
- I am trying to **improve communicating why** you enforce TLS.

---

> Since no one runs their production IDM on localhost then ergo, ipso facto, therefore, TLS is required."

I was never trying to imply this. You've **inferred that due to bias/skimming**, missing the point entirely.

> Again, we don't know what you are trying to prove here

Was there anything in particular you wanted me to clarify for you?

- There was no need to try proving anything, I provided proof that can you could adapt to a production environment.
- Secure cookies only require HTTPS from the client to function as intended. If kanidm was not directly connected to from the client, TLS is not required, yet the FAQ associates this feature to why kanidm enforces TLS.

It seems like you jumped to some wildly inaccurate assumptions.
- It's not my fault if you skimmed my report and responded with biased assumptions.
- Instead of pursuing a positive interaction like we previously discussed, you chose to outright dismiss me again.
- You seemed very confident that I was the one misunderstanding. Yet you focused on the localhost aspect which was irrelevant.

While the suggestion could be revised better, it's quite clearly well intentioned? (_how does this change negatively affect the project?_):

```diff
-No, it can not. TLS is required due to our use of the `Secure` flag our cookies
+TLS is required for some features kanidm uses such as cookies tagged with the `Secure` attribute
```

```diff
+While TLS could be terminated before the request reaches kanidm, we feel it is in the best interests of security to always ensure traffic arrives to kanidm encrypted.
```

I thought that was much more clearer on why you forbid opt-out? 🤷‍♂️ 

A little further down in the same FAQ entry you have this statement attributing secure cookies as the reason to enforce TLS (_this is all I really cared about correcting_):

https://github.com/kanidm/kanidm/blob/f1dfbcc25373201b0fd306f0507ad82b9297fb79/book/src/frequently_asked_questions.md?plain=1#L31

Would you place trust in a security product that actively chooses to lie to the reader when justifying security decisions? (_especially a FAQ item_)

Both of your responses seem to be in support of handwaving technicalities on security related statements like this. I think that kind of **misinformation does kanidm a disservice**, but if you rather deny a **valid production scenario** where that statement is **provably false**.. no worries 👍 

---

<details>
<summary>Original message (extra details but collapsed for brevity)</summary>

I cannot reply to [your response to my issue](https://github.com/kanidm/kanidm/issues/2969#issuecomment-2285075674) due to the lock.

If you do not want to discuss this further, please disengage instead of responding with comments that clearly depict you're not paying attention by contradicting my statements each time with your assumptions.

> All your example proves is that yes, browsers and curl treat `http://localhost/` as a secure context, and `https://*` as secure context when you are remote to the server. All other situations are insecure.

- `http://localhost` was not the point I was trying to convey (_it was there for the benefit of others that may land there_), and demonstrating that the exception behaviour was not applicable to `http://example.localhost`.
- `https://example.test` was then provided to show `client ==(HTTPS)==> [caddy ==(HTTP)==> service]` where `[ ... ]` represents two processes in the server.
- The client can send a secure cookie over HTTPS to `caddy`, and the functionality still works when received over HTTP to the service (_on the same host, not directly exposed to a network_).
- The FAQ entry misleading claims that kanidm requires TLS for traffic it receives for this secure cookie functionality to work. My suggestion revises the wording in favor of correctness, while still **respecting and conveying the decision to embrace only TLS**...

Why I receive hostility for trying to contribute an improvement that does not degrade what the FAQ item covers for a PR that is specifically revising this topic coverage, I do not understand.

I thought we had discussed this sort of treatment towards contributors, but you seem to be taking this personally and reacting negatively due to some bias, despite the recent apology and commitment to do better going forward...?

---

> We are not stupid. We have read the documentation. We understand how this works.

As I've tried several times to clarify any miscommunication, all I receive in your responses are deflection. I'm not here to insult your intelligence or expertise.

May I remind you in our prior discussion when referring to adhering to your CoC, you had the negative guidelines in mind, while I had the positive ones in mind. Your thought process is interpreting the same information very differently.

> Sending us long write ups and examples of things we already know isn't going to change our minds.

I tried being terse and specific with the PR suggestion provided above.

I don't want them to be long, but the intent was to clearly demonstrate the concern since you've been rather adamant about my prior statements being invalid.

I realize now that the effort is wasted due your reluctance to pay attention.

---

> We require secure contexts to interact with Kanidm between any client to the server.

The secure context I tried to correct was between client and server.

Kanidm itself did not require it to support cookies with the `Secure` attribute. This was demonstrated, but for whatever reason dismissed... you can literally try it across separate systems if you think that makes a difference. I made the effort to make that simple for you to try.

To claim that any connection after the clients direct connection to the server must have TLS for the secure cookie to be sent to the service is wrong.

As security professionals I do not understand why you are so against amending the FAQ entry per the suggestion provided?

> Multiple parts of Kanidm beside cookies require secure contexts to operate.

This was not disputed. In fact the amendment I suggested further clarified that distinction. It still endorsed/supported kanidm requiring TLS.

---

> We require TLS between load balancers and Kanidm because an attacker on your internal network is a real threat, and must not be ignored.

The scenario was as depicted `client ==(HTTPS)==> [caddy ==(HTTP)==> service]`.

- In what situation is this a real threat to worry about from an attacker where TLS avoids it?
- If the traffic flowed across the internal network to other hosts, absolutely agree. This is between two processes on the same server host. There is no internal network concern.

> We do not care if you want to overlook that threat. Our job as developers is to raise the bar even if you want to deny that the threat exists.

The threat you express is very vague. If you want to be taken seriously with that you would have no issue providing information on how that scenario I provided could actually be compromised. Chances are low TLS will make a notable difference given the requirements for the compromise.

To deny the potential for threats is silly, but that does not mean there is a threat. Just like TLS itself has had it's own fair share of failings, risk assessment and to what lengths you'll go to protect against such is up to you. Surely we're on the same page with that.

I've been through this dance far too many time. I seek clarity on a claim that a security professional insists is insecure, but no matter how much evidence I provide that contradicts it, they have nothing actually tangible to support their claim, often refusing to acknowledge the very real proof. I care about correctness, being proven wrong is a positive, often the other party does not seem to share this view.

I don't mean to be disrespectful, but if you **truly believe in the importance of security**... surely you can understand that the skepticism is valid/fair? It would be very bad security wise to place trust in the security expertise of someone so confident with their statements, yet is averse to backing up their assertions when they appear contradictory? Yet that's what seems like is occurring here.

Please don't make generic claims that you're not willing to detail further, that's deflection (_often with an intent that it is irrefutable as a means to avoid requiring explicit proof_). Too many parrot misinformation as it is.

---

> We have previously suffered a lot of issues from real users of Kanidm when we DID allow opt-out of TLS in the past.

Right, which like many things you responded with, I had already cited in the issue... Did you not notice that because of skimming?

</details>

> We have made our position and arguments very clear. You may not agree with us, but we aren't changing this policy.

🤦‍♂️ That issue, nor the PR suggestion ever attempted to change your policy on TLS...

Firstyear (approved):

PR #2980, opened 2024-08-16
Add CORS for REST API

SinnySupernova:
Fixes #

Checklist

- [X] `cargo fmt` has been run
- [ ] `cargo clippy` has been run


examples/server_container.toml
```diff
@@ -81,6 +81,10 @@ domain = "idm.example.com"
 #   origin = "https://idm.example.com"
 origin = "https://idm.example.com:8443"
 #
+# Allowed CORS origins. This allows deployment of an
```
Firstyear, line 84: I wonder how this will play with webauthn or cookies? Since we need both to be related to origin, shouldn't we just set cors to origin and subdomains of origin instead? Alternately we need to enforce all the allow_origins are subdomains of origin.

server/core/src/https/mod.rs
```diff
@@ -253,6 +255,13 @@ pub async fn create_https_server(
 
     let trust_x_forward_for = config.trust_x_forward_for;
 
+    let cors_allowed_origins: Option<Vec<HeaderValue>> = config
+        .cors_allowed_origins
+        .as_ref()
+        .map(|os| os.into_iter().map(|o| o.parse().ok()).collect())
+        .flatten()
+        .and_then(|hvs: Vec<HeaderValue>| if hvs.is_empty() { None } else { Some(hvs) });
```
yaleman, line 263: Since it starts and ends as an option you should be able to avoid the re-wrapping-with-Some by dropping the flatten, I think?

examples/server_container.toml
yaleman: I wouldn't note it as an alternative webUI for Kanidm, more "allow access to the Kanidm endpoints from additional domains".

Origin's already covered as it's not a cross-origin request, but I'm not sure about sub/parent domains (I'm pretty sure they'd be blocked)

This needs more documentation in the book

yaleman (changes requested):
Changes I can see as required before this can be considered for merge:

- needs to take into account the existing CORS headers that are being set on other endpoints (search for `ACCESS_CONTROL_ALLOW_ORIGIN` and there's heaps of references)
- testing to ensure that the configured headers are appearing (or not aren't) when configuration is set or not
- explanation in the docs

server/core/src/https/mod.rs
SinnySupernova: It's Option<Option<Vec>> initially, the last and_then converts Some(empty_vec) into None, otherwise the Option<Vec> remains Option<Vec>

SinnySupernova:
Still pending: tests (soon™) and more docs
Where would be a good place to put it in the book?

yaleman:
I'd say under https://kanidm.github.io/kanidm/master/security_hardening.html

"By default we block [cross-origin resource sharing](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS) but if you do need it for some reason then configure it by doing ..."

yaleman:
Also, this all really should only apply to the REST API endpoints, not the javascript/static content etc, folks shouldn't be ever cross-loading that.

SinnySupernova:
- added book chapter at the end of security hardening
- CORS layer currently only applies to endpoints in `v1.rs` (unchanged from the first commit on branch)
- added a test

I'm thinking to add panic during server start when cors config option can't be parsed properly

Firstyear:
I'd rather it be a proper error than a panic though.

examples/server.toml
```diff
@@ -81,6 +81,10 @@ domain = "idm.example.com"
 #   origin = "https://idm.example.com"
 origin = "https://idm.example.com:8443"
 #
+# Allowed CORS origins. This allows access to the Kanidm
+# endpoints from additional domains. Default is empty
```
Firstyear, line 85: I think the open question is still how you plan to handle authentication here - cookies are tied to the origin and similar with webauthn. How do you plan to work with that? Am I missing something?

SinnySupernova: You're right. Currently kanidm [expicitly sets domain](https://github.com/kanidm/kanidm/blob/master/server/core/src/https/v1.rs#L2957) from the server state. Cookie auth support needs some work IF we want that, but token auth still works. Webauthn spec allows [related origins](https://w3c.github.io/webauthn/#sctn-related-origins), which could allow webauthn to work too.

SinnySupernova: e.g. `id.example.com` is kanidm, and `app.example.com` is the web app, then they could use a common RP  ID `example.com`.

Firstyear: That's new in the editors draft and not supported on webauthn-rs yet - it's also unlikely to work on a lot of existing credentials as single origin is baked into a lot of passkey providers. 

The more I think about this the more it starts to scare me a bit. I think if you want to make a custom ui, it has to be a subdomain of the IDM - too many things have to change otherwise, and it opens the door too wide to exploitation. I think a custom UI as a subdomain is a reasonable requirement for a deployment to meet.

CEbbinghaus: A custom UI served under a subpath is also possible. The Reverse proxy can map `idm.example.org/AlternateUI` to an entirely independent service which should be fully CORS compatible

Firstyear: Yes, but then we'd either need to host files from a static dir, or we'd need to be a reverse proxy and that opens up a huge can of security issues.

CEbbinghaus: Oh, My point wasn't that we support it but rather that this functionality is available without us having to do anything by just using a third party reverse proxy Infront of kanidm. I run caddy and could easily configure a route to a different app which would completely circumvent the CORS problem (at least for calls made from the UI)

Firstyear:
Closing due to the age of this PR - if you want to re-open it please let us know.

Issue #2983, opened 2024-08-18
REST API: SSH key tags not gettable


yaakov-h:
Using Kanidm 1.3.0:

Looking at the Swagger docs and Nginx request logs, the APIs for SSH public keys seem to be:

`/v1/person/{name}/_ssh_pubkeys` - returns all SSH public key values as a JSON array
`/v1/person/{name}/_attr/ssh_publickey` - returns all keys as a JSON array of strings formatted as `"<tag>: <value>"`
`/v1/person/{name}/_ssh_pubkeys/{tag}` - returns a single SSH public key value if you know it's "tag".

There appears to be no way to programmatically get a SSH public key's tag.

I would have expected one of `/_ssh_pubkeys` or `/_attr/ssh_publickey` to return a JSON object along the lines of:

```json
{
  "tag1": "type base64 user@host",
  "tag2": "type base64 user@host"
}
```

however that only current result I can get, which then requires string parsing, is:

```json
[
  "tag1: type base64 user@host",
  "tag2: type base64 user@host"
]
```

As an aside, it would also be nice to have an API that just returns the `authorized_keys` file's expected format, without JSON.

yaleman:
It'll likely be a HashSet/Btree being serialized which is... fun to unpick

yaleman:
Actually it's ... even harder to unpick, because of how it all works internally. I'm poking around to see what can be done about it.

yaleman:
this will have to wait for #2315

yaakov-h:
Why does separating keys and values depend on SCIM?

yaleman:
Because @Firstyear's going to sort a bunch of internal state-getting stuff on the way to it. There's no simple "oh just pull out keys and values" magic, I tried for a few hours today and couldn't find the way to sort it.

... that and we have to add endpoints for this, because changing the existing ones would break all existing clients.

Firstyear:
@yaakov-h Internally Kanidm attributes are all strongly typed and contain maps. So something like ssh-key is internally represented as:

```
key: label
value: SshPubKey {
    type
    public
    comment
}
```

The problem though is how we *format* these for presentation. 

Currently, the code for rendering these out as a "protocol value" aka something our (very aging and not good) REST API exposes also shares the same path with *LDAP*. 

This means that currently, SSH keys are in an awkward spot where they have to be presented the same way that LDAP expects them to be - which then limits us based on how SSSD behaves.

This means that either we need to:

* Add an LDAP value render here in this case.
[...]

Issue #1998, opened 2023-08-18
OAuth2 Scope map is string-serialised 


yaleman:
The OAuth2 API returns the scope map in a rather unfortunate format, which is probably fine for your internal use because you share the serializing/deserializing logic among tools. From the outside to parse a response like this:

```json
{ "oauth2_rs_scope_map": ["administrators@idm.example.com: {\"admin\", \"displayname\", \"email\", \"preferred_username\", \"profile\"}', 'internal@idm.example.com: {\"oidc\"}"]}
```

What it should look like:
```json
{ "oauth2_rs_scope_map": { "administrators@idm.example.com" : ["admin", "displayname", "email", "preferred_username", "profile"], "internal@idm.example.com": ["oidc"]}}
```

_Originally posted by @tumbl3w33d in https://github.com/kanidm/kanidm/issues/1954#issuecomment-1683751538_

yaleman:
@tumbl3w33d which specific endpoint were you calling at this point?

tumbl3w33d:
`/v1/oauth2`

yaleman:
Ah, figured this one out, it comes from Rust/serde serialising the attribute lists as values, which all have to be the one type at some point (because strict typing) and that ends up being a string.

we serialize the response as (python-typed): 
```
[{
  "attrs" : Dict[str, List[str]]
}]
```

Pulling it out of the JSON response we get (for example):

```json
{
  "oauth2_rs_sup_scope_map": [
    "test_users@localhost: {\"admin\"}"
  ]
}
```

It's hacky but in the interim you can parse it slightly easier and more accurately with this in python:

```shell
python -q
>>> import json
>>> mapping = "test_users@localhost: {\"admin\"}"
>>> user, scopes = mapping.split(":")
>>> scopes = scopes.replace("{", "[").replace("}", "]")
>>> scopes = json.loads(scopes)
>>> print(f"{user=} {scopes=}")
user='test_users@localhost' scopes=['admin']
```

The core issue is that we don't use `list`s we use `set`s, which is why the `{}` end up in there, but JSON doesn't have that syntax. Type-Casting the response is not out of scope, but ... this is partially me making notes as to how it got to this point 😄

Firstyear:
Where abouts in proto is this code based? We should probably make it emit the correctly typed json else we might trip up other things.

Firstyear:
[...]

Issue #1954, opened 2023-08-07
API consistency


tumbl3w33d:
I noticed an inconsistency in the API when it comes to the data type of `gidnumber`. When retrieving a group via `GET` to `/v1/group`, the `gidnumber` is returned as (list of) **string**, however, when you `POST` to the unix config of that group via `/v1/group/:id/_unix`, it asks you for an (optional) **integer**. I would have expected to `POST` an empty list instead of `null` or use `DELETE`.

It's just a minor thing, hardly worth opening the issue for and it reminds me of the not so perfect status codes the API returns (see #1904). Would you prefer to ignore such things, or have an issue per finding or rather collect everything in a long-lasting "API improvement" issue?

Edit: I actually didn't yet figure out how to delete the `gidnumber` from a group. Neither `DELETE`, nor `POST`ing `null` works, even though the latter returns `200`, because well, the integer is optional, it just doesn't do anything. 😅

yaleman:
Any improvement to dev consistency and ergonomics is *always* worth doing - it's just whether we've hit that use case yet 😄 

1. The `GET` verb probably should return a single integer, but the default response of the database is a list; I'd say that could be nicer, and we should change that response (since there should only ever be one `gidnumber` for one `group`.
2. I'm not sure what the expectation is for `POST`ing a null to that `/v1/group/:id/_unix` endpoint - it feels like we should be requiring that value or throwing an error TBH.
3. Removing the `gidnumber` could be a `DELETE` for sure - I guess we just haven't used that yet. 😄 - have you tried the `/v1/group/:id/_attr/:attr` which has a `DELETE` verb?

tumbl3w33d:
>  the default response of the database is a list; I'd say that could be nicer, and we should change that response (since there should only ever be one gidnumber for one group

I thought that's the general pattern. Pretty much everything I've seen is returned as list by the API, no matter how likely it is that more than one is returned. If you would want to change that, I guess a lot of code/interface would change. I'm neutral on that and got used to handle it in my client.

> I'm not sure what the expectation is for POSTing a null to that /v1/group/:id/_unix endpoint - it feels like we should be requiring that value or throwing an error TBH.

Agree. I'm no fan of associating `null` with any meaning. It was just me trying to get rid of the gidnumber (without success).

> Removing the gidnumber could be a DELETE for sure - I guess we just haven't used that yet. smile - have you tried the /v1/group/:id/_attr/:attr which has a DELETE verb?

[...]

Issue #1476, opened 2023-03-21
Batch importing of users


mbyrdap:
**Is your feature request related to a problem? Please describe.**

Being as every bit of infrastructure seems to be moving to IaC style provisioning and deployment, I thought being able to turn up an IDM server with kanidm, and have the users present to get working, would be a good idea.

**Describe the solution you'd like**

A simple command line parameter with a filename as an argument to add/update/set the users with their authentication options. 

**Describe alternatives you've considered**

a messy bash script 

**Additional context**

mbyrdap:
> but like with a database, you don't specify the full contents of it from your code... this what a backup's for, and that functionality's already in the system.

Why would you not deploy your IDM with an accurate data source? 

For HA, portable, etc., what sense is it restoring a backup vs keeping the internal state of users accurate at deployment?

yaleman:
Sorry if it came off wrong, that was a late-night response I could have worded better. 

Scoping the request's important because we already have a lot of tools floating around and things like the SCIM endpoints; extending the `iam-migrations` package would allow for batch creation for people doing migrations.

> For HA, portable, scaling, onprem + cloud presence, etc.

Clustering/sync (in development now) and backups covers all of these options, especially considering the token encryption requirements in the backend.

Again, I'm not against this, I just want to try and make sure we're not trying to oversell things or reinvent the wheel.

Firstyear:
I think improving our batch operation interfaces and user creation mechanisms is the way to go here, but having uses in a "config" file I'm not sure about yet. I can see it both ways that it would be valuable to template and create these, but I can also see that it adds a complexity once we have replication (we'd essentially need to create an equivalent of the MS AD FSMO roles then.).

Firstyear:
@mbyrdap I'm coming back to this ATM and I'm wondering if this is solved via a scim endpoint you can issue updates to. 

Or is this that you want a fileshare that picks up changes and can apply them?

Firstyear:
I think this is actually solved by #211

q2dg:
There's no documentation about how to do it with kanidmd command or similar...

Firstyear:
@q2dg There are multiple approaches:

* Synchronisation: https://kanidm.github.io/kanidm/stable/sync/concepts.html
* The static entry management feature: https://kanidm.github.io/kanidm/stable/entry_management.html 
* Using our SCIM api located at  `/scim/v1/Entry` or  `/scim/v1/Person`
* Or our REST API at `/v1/person`

[...]

PR #2992, opened 2024-08-22
20240820 SCIM value

Firstyear:
Add the ability for valuesets to resolve to SCIM values. 

This is the first part of many related to addition of SCIM to the server. Following this we will need to add (in rough order that I plan to do these):

* extra metadata for creation and modification time of entries (based on replication metadata)
* Ability to mark scim entries with access control states of attributes (for @ToxicMushroom )
* An entry state "hash" so during a post we can tell the entry revision and reject the update if it has changed in between. 
* SCIM filter / query handling
* SCIM Value to Valueset (allowing modifies and creates)
* The external API for users to tools can access SCIM types.
* More strictly enforce some schema class types as "base" types for Scim resource type classification
* Allow resolving uuid references -> spn in scim values

A number of other schema changes are likely needed in this process - the attribute status "multivalue" should be merged into the syntax type. This will allow differentiating a "single utf8 string" vs a "multiple utf8 string".

Relates #211 #280 #330 #1476 #1954 #1998 #2983

Checklist


Firstyear:
> Add the ability for valuesets to resolve to SCIM values.
> 
> This is the first part of many related to addition of SCIM to the server. Following this we will need to add (in rough order that I plan to do these):
> 
>     * extra metadata for creation and modification time of entries (based on replication metadata)
> 
>     * Ability to mark scim entries with access control states of attributes (for @ToxicMushroom )
> 
>     * An entry state "hash" so during a post we can tell the entry revision and reject the update if it has changed in between.
> 
>     * SCIM filter / query handling
> 
>     * SCIM Value to Valueset (allowing modifies and creates)
> 
>     * The external API for users to tools can access SCIM types.
> 
> 
> A number of other schema changes are likely needed in this process - the attribute status "multivalue" should be merged into the syntax type. This will allow differentiating a "single utf8 string" vs a "multiple utf8 string".
> 
> Relates #211 #280 #330 #1476 #1954 #1998 #2983
> 
> Checklist
> 
>     * [ x ] This PR contains no AI generated code
> 
>     * [ ]  book chapter included (if relevant)
> 
>     * [ ]  design document included (if relevant)

Firstyear:
As usual, none of this is exposed via the API yet, that's a future change :)

libs/scim_proto/src/lib.rs
```diff
@@ -38,11 +39,58 @@ pub enum ScimAttr {
...
 }
 
+impl ScimAttr {
+    pub fn parse_as_datetime(&self) -> Option<Self> {
+        let s = match self {
+            ScimAttr::String(s) => s,
+            _ => return None,
+        };
+
+        use time::format_description::well_known::Rfc3339;
```
yaleman, line 56: crimes!

```suggestion
        use time::format_description::well_known::Rfc3339;
```

server/lib/src/utils.rs
```diff
@@ -74,6 +75,20 @@ pub fn readable_password_from_random() -> String {
...
+    let mut buf = String::with_capacity(alloc_len);
+    set.iter().for_each(|s| {
+        buf.push_str(s);
+        buf.push(' ');
+    });
+
+    // Remove the excess trailing space.
+    let _ = buf.pop();
+
+    buf
```
yaleman, line 89: (this is likely wrong due to ownership but you get the idea)

```suggestion
   set.iter().collect::<Vec<String>>().join(",")
```

join doesn't seem to be implemented on BTreeSet

server/lib/src/valueset/address.rs
```diff
@@ -163,6 +163,27 @@ impl ValueSetT for ValueSetAddress {
...
+                .map(|a| {
+                    let mut complex_attr = ScimComplexAttr::default();
+
+                    complex_attr.insert("formatted".to_string(), a.formatted.clone().into());
+                    complex_attr
+                        .insert("stretAddress".to_string(), a.street_address.clone().into());
+                    complex_attr.insert("locality".to_string(), a.locality.clone().into());
+                    complex_attr.insert("region".to_string(), a.region.clone().into());
+                    complex_attr.insert("postalCode".to_string(), a.postal_code.clone().into());
+                    complex_attr.insert("country".to_string(), a.country.clone().into());
```
yaleman, line 179: can we make these key-strings an enum like we do with `Attribute` so we don't have to go back and fix them later? :D

ToxicMushroom: stretaddress seems to also be a typo here

yaleman: 👨🏻‍🍳 perfect example 😄

server/lib/src/utils.rs
Firstyear: Yeah, but this is btreeset to allow fast merging of scopes in oauth2. So I think we leave it as is :(

server/lib/src/valueset/address.rs
Firstyear: Yep, will do .

libs/scim_proto/src/lib.rs
```diff
@@ -193,4 +194,207 @@ mod tests {
         let s = serde_json::to_string_pretty(&u).expect("Failed to serialise RFC7643_USER");
         eprintln!("{}", s);
     }
+
+
+
+    // =========================================================
```
yaleman, line 200: In this case for attributes I'd waaaaaaaaaaaaaay prefer to do the from/into string handlers from `serde` over manually implementing serde; as we've only got one "custom" attribute type, we don't have to try and catch multiples with the custom serde's, and the "fallback" is much simpler. We can also use it for other things like display etc.

Firstyear: Yeah, this was me just experimenting to demonstrate possible approaches. 

The other thing is that there is a difference here with attributes and subattributes. We need to start to consider that too, but it can be handled in a similar way. 

(Also at some point we need to prune old/unused values).

libs/scim_proto/src/lib.rs
```diff
@@ -1,4 +1,5 @@
-#![deny(warnings)]
+// #![deny(warnings)]
```
yaleman, line 1: RIP clippy

Firstyear: temporary, I'll uncomment once I tidy the test.

server/lib/src/idm/scim.rs
```diff
@@ -546,15 +546,7 @@ impl<'a> IdmServerProxyWriteTransaction<'a> {
...
-    ) -> Result<
-        (
-            Uuid,
-            BTreeSet<String>,
-            BTreeMap<Uuid, &'b ScimEntryGeneric>,
-            bool,
-        ),
-        OperationError,
-    > {
+    ) -> Result<(Uuid, BTreeSet<String>, BTreeMap<Uuid, &'b ScimEntry>, bool), OperationError> {
```
yaleman, line 549: can we make the return a struct so it's clearer to folks what is coming back? I know it's in your head now but that's a ... lot of stuff 😄

server/lib/src/valueset/image/mod.rs
```diff
@@ -417,6 +417,27 @@ impl ValueSetT for ValueSetImage {
         Box::new(self.set.iter().map(|image| image.hash_imagevalue()))
     }
 
+    fn to_scim_value(&self) -> Option<ScimValue> {
+        // TODO: This should be a reference to the image URL, not the image itself!
+        // Does this mean we need to pass in the domain / origin so we can render
+        // these URL's correctly?
```
yaleman, line 423: if we return a relative url (ie `/foo/bar/ref` instead of `https://frontend_url:port/foo/bar/ref`) then that would work for clients, be valid, and be way simpler...

Firstyear:
This is pretty much mergable - by splitting scim to a "serialise" vs "deserialise" path, we can have asymmetric types allowing the server to strongly type (and more effeciently represent) outgoing data. For the incoming path, we can then represent optionality and generic values instead and do schema aware parsing. 

This is currently not publicly exposed, so it's safe to merge. It's pretty much the full valueset -> scim value code.

yaleman (dismissed):
more scim more betterrerere

yaleman (approved):

PR #3000, opened 2024-08-24
Constrain scopes on refresh token usage

### LINKED ISSUE #2527: Oauth2 refresh incorrectly constrains scopes on refresh
The Oauth2 rfc allows a refresh request to request any scope that was *granted* not just any scope that was requested. As a result, since we issue supplemental scopes, this can trip up our logic and incorrectly block refreshes. We should check the full set of issued scopes for validity. 

We should also allow the refresh to constrain and remove scopes by this process.

Firstyear:
When a refresh token is used, allow constraint of the scopes.

Extend tests for session handling

Use the correct invalidGrant response when a refresh token is expired.

Fix docs command.

-

Fixes #2527 

Checklist

### DIFF
--- a/book/src/database_maintenance.md
+++ b/book/src/database_maintenance.md
@@ -25,7 +25,7 @@ definitions.
 ```bash
 docker stop <container name>
 docker run --rm -i -t -v kanidmd:/data \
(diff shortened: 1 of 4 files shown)


yaleman (approved):
woot

PR #3004, opened 2024-08-24
fix(daemon): handling IPv6 addresses in healthcheck

### LINKED ISSUE #3002: docker reports unhealthy --> CRITICAL: Connection failed: error sending request for url (https://[::]:8443/status)
### I did this

Docker reports container is unhealthy. Everything seems to work fine.

📜 Using config file: "/data/server.toml" 00000000-0000-0000-0000-000000000000 ERROR 🚨 [error]: CRITICAL: Connection failed: error sending request for url (https://[::]:8443/status) Logging pipeline completed shutdown

### I expected the following

docker reporting the container is healthy.

### Kanidm version details

- Output of `kanidm(d) version`:  1.3.3
- Are you running it in a container? If so, which image/tag?: kanidm/server:latest

### Any other comments

server.toml contains:
bindaddress = "[::]:8443"

yaleman:
# Change summary

- fix(daemon): handling IPv6 addresses in healthcheck

Fixes #3002

Checklist


server/daemon/src/main.rs
```diff
@@ -1044,10 +1044,24 @@ async fn kanidm_main(
 
             let healthcheck_url = match &sopt.check_origin {
                 true => format!("{}/status", config.origin),
-                false => format!("https://{}/status", config.address),
+                false => {
+                    if config.address.starts_with("[::]") {
+                        format!(
+                            "https://{}/status",
+                            config.address.replace("[::]", "localhost")
```
Firstyear, line 1051: Why not just make it `[::1]`? Why replace the `[::1]` to localhost? Seems like we should only need to worry about the `[::]`

yaleman: Testing it now again locally, it works fine either way so I... must have been doing something weird the other day... going back to OP on #3002 to find out what's going on there.

Firstyear: The only reason to fallback to localhost is to allow swapping over to 127.0.0.1

yaleman:
unneeded at this time, can't repro parent issue

Firstyear:
I thinrk we should still merge this, just the `[::]` to `[::1]` version only.

yaleman:
> I thinrk we should still merge this, just the `[::]` to `[::1]` version only.

It works fine as-is inside the container and outside, so I'm not sure there's any valid reason to change it?

2xsaiko:
I'm pretty sure treating :: as localhost is undefined behavior. At least it's that way for 0.0.0.0 which is only defined as "every interface when opening a listen socket" in the standard, but not what happens if you open a client socket. Something like that at least. It doesn't work on Windows for example.

Firstyear (approved):

PR #3013, opened 2024-08-28
[htmx] Make it harder to miss the save button on the cred update page

### LINKED ISSUE #2504: [UX] Make sure users submit their credential changes before leaving

I've noticed a few users struggle with the necessity to submit changes to passkeys twice, i.e.:
* you press "add passkey"
* kani opens an overlay with button "begin passkey enrollment" which you press
* you do the passkey magic depending on your system
* kani asks for a name for the passkey, so you enter one and press submit
* now the passkey appears in the list of your passkeys

At this point some users close the tab and leave. Then they wonder why they cannot login with the passkey they just added. What happened? After the procedure described above you need to submit the entire authentication settings form to actually store the changes. I think that's not optimal from UX perspective.

It's the same behavior when deleting passkeys and should be adjusted as well. There already is an extra confirmation dialog so you don't accidentally delete passkeys.

When I submit the name of the passkey, it should really be stored. Maybe the other submit button of the entire form is pointless? I don't know all the use cases.

You give affected people another onboarding token and tell them to submit harder.

**Additional context**

Can add more info if it's not clear from the description.

Another minor thing (but edge case) is that the "Updating Credentials" view doesn't update the list of passkeys while it's opened and you refresh the tab. That means if you change the set of credentials on another device (e.g., mobile via QR code) you won't see it in this opened tab on your workstation and I wouldn't know what happens if you submit the entire form without your tab being aware of that change. The entire problem would probably disappear if this surrounding form would go away and be replaced with direct submits as suggested above.

ToxicMushroom:
# Change summary

- Fixed floating style save/cancel bar
- Only the save button is green
- Use "add" instead of "submit" on buttons in other places

beforeunload seems a bit drastic and conflicts with reloads, I think this should already help a lot
otherwise 2504 can be reopened :)

Fixes #2504

Checklist

Some screenshots of what it looks like:
image
image


server/core/templates/credentials_update_partial.html
```diff
@@ -91,21 +91,27 @@
...
-                        class="btn btn-success btn-lg"
-                        type="submit"
-                        hx-post="/ui/api/cu_commit"
-                        hx-boost="false"
-                        (% if !warnings.is_empty() %)
-                        disabled
-                        (% endif %)
-                >Submit Changes</button>
-            </span>
+        <div id="cred-update-commit-bar">
```
weijiangan, line 97: bootstrap has an alert component I think it'll be better if you use that. I don't know if we have any ambitions to make kanidm look amazing (looking at you @Firstyear) but it'll be nicer if everything looks part of something. Also the point of using something like Bootstrap is to minimize custom CSS and they become quite unmaintainable pretty quickly

weijiangan:
I have a few non-blocking nitpicks on CSS but I think it's really cool you're able to get it done so quickly and functionality-wise it's complete.

server/core/templates/credentials_update_partial.html
yaleman: I'd definitely request [bootstrap alerts](https://getbootstrap.com/docs/5.3/components/alerts/) for this please. You should be able to get it to float over with a z-index if that's your jam.

yaleman: Actually, [Toasts](https://getbootstrap.com/docs/5.3/components/toasts/) are another option which might be more effective/dynamic.

ToxicMushroom: image
Like this ?

Firstyear (approved):

server/core/templates/credentials_update_partial.html
Firstyear: > bootstrap has an alert component I think it'll be better if you use that. I don't know if we have any ambitions to make kanidm look amazing (looking at you @Firstyear) 

I'm not skilled enough as a web dev to make it look nice - i'll defer to you and others who actually have those skills :) 

My background is all in server side stuff, so I'll trust your decisions :)

PR #3016, opened 2024-08-29
20240828 Support Larger Images, Allow Custom Domain Icons

### LINKED ISSUE #2057: In-system file storage

Firstyear:
# Change summary

- Allow images up to 256kb
- Validate oauth2 image operation on new ui
- Allow custom domain info images to be set
- present custom domain info images on login.

Fixes #2057 #2056 

Checklist


server/core/src/https/mod.rs
```diff
@@ -269,13 +270,20 @@ pub async fn create_https_server(
...
                 Router::new()
                     .route("/", get(|| async { Redirect::to("/ui") }))
                     .nest("/ui", views::view_router())
-                    .layer(middleware::compression::new())
-                    .route("/ui/images/oauth2/:rs_name", get(oauth2::oauth2_image_get))
+                    .merge(static_compressed_router)
             } else {
                 Router::new()
+                    .route("/ui/images/oauth2/:rs_name", get(oauth2::oauth2_image_get))
+                    .layer(middleware::compression::new())
```
yaleman, line 286: The ordering of the layering here is off - the "else" arm should have the compression layer above the images endpoint - because we don't want to be wasting cycles compressing images

Firstyear: Actually, the order doesn't matter. The router applies the layer as a whole to any route on the router.

server/core/src/https/oauth2.rs
```diff
@@ -107,20 +107,20 @@ pub(crate) async fn oauth2_image_get(
...
-        Ok(image) => (
+        Ok(Some(image)) => (
             StatusCode::OK,
             [(CONTENT_TYPE, image.filetype.as_content_type_str())],
             image.contents,
         )
             .into_response(),
+        Ok(None) => {
+            warn!(?rs_name, "No image set for oauth2 client");
+            (StatusCode::NOT_FOUND, "").into_response()
```
yaleman, line 118: This could be a `Redirect` to the default image

Firstyear: Can't redirect in an img tag.

server/core/src/https/ui.rs
```diff
@@ -83,11 +80,7 @@ pub(crate) async fn ui_handler_generic(
 
     let jstags = jsfiles.join("\n");
 
-    let body = format!(
-        include_str!("ui_html.html"),
-        domain_display_name.as_str(),
-        jstags,
-    );
+    let body = format!(include_str!("ui_html.html"), domain_display_name, jstags,);
```
yaleman, line 83:
```suggestion
    let body = format!(include_str!("ui_html.html"), domain_info.display_name(), jstags);
```

server/core/src/https/ui.rs
```diff
@@ -40,35 +41,31 @@ pub(crate) fn spa_router_login_flows() -> Router<ServerState> {
...
-    kopid: KOpId,
     wasmloader: &str,
+    domain_info: DomainInfoRead,
 ) -> Response<String> {
-    let domain_display_name = state
-        .qe_r_ref
-        .get_domain_display_name(kopid.eventid)
-        .await
-        .unwrap_or_default();
+    let domain_display_name = domain_info.display_name();
```
yaleman, line 68:
```suggestion
```

server/core/src/https/ui.rs
yaleman: Something something the borrow checker might be cranky but it should be an easy fix 😄

Firstyear: I'll try it :)

server/core/src/https/mod.rs
yaleman: That's not what the [docs say](https://docs.rs/axum/latest/axum/middleware/index.html#applying-middleware)

> When you add middleware with Router::layer (or similar) all previously added routes will be wrapped in the middleware. Generally speaking, **this results in middleware being executed from bottom to top.**

server/core/src/https/oauth2.rs
yaleman: Yeah you can, works fine 😄

yaleman: Throw a HTTP 303/307 and it'll just load the file it's told to.

server/core/src/https/mod.rs
Firstyear: Argh it's so confusing. The actualy detail is here https://docs.rs/axum/latest/axum/struct.Router.html#method.layer which makes you right, but for a different reason. #2753

server/core/src/https/oauth2.rs
Firstyear: In this case, I think I actually want to leae the 404. The flow is that *if* the server sends back custom-image == true, then we go to get the image, if the image doesnt exist we should 404 - it's correctly showing the image doesn't exist! 

The fallback of the default image should be processed *before* we even get to this point.

yaleman (dismissed):

yaleman (approved):

PR #3019, opened 2024-09-01
[htmx] Admin ui for groups and users management

ToxicMushroom:
# Change summary

Fixes part of #996

Todo:
 - [ ] limit navigation based on access 
 - [x] do navigation.
 - user
   - [x] view
   - [x] list
 - group
   - [ ] view
   - [ ] list

Checklist

### DIFF
--- a/proto/Cargo.toml
+++ b/proto/Cargo.toml
@@ -38,6 +38,7 @@ uuid = { workspace = true, features = ["serde"] }
 webauthn-rs-proto = { workspace = true }
 sshkey-attest = { workspace = true }
 sshkeys = { workspace = true }
+tracing = { workspace = true }
 
 [dev-dependencies]
 enum-iterator = { workspace = true }
--- a/proto/Cargo.toml
+++ b/proto/Cargo.toml
@@ -38,7 +38,6 @@ uuid = { workspace = true, features = ["serde"] }
 webauthn-rs-proto = { workspace = true }
 sshkey-attest = { workspace = true }
 sshkeys = { workspace = true }
-tracing = { workspace = true }
 
 [dev-dependencies]
 enum-iterator = { workspace = true }
--- a/proto/src/scim_v1/server.rs
+++ b/proto/src/scim_v1/server.rs
@@ -4,11 +4,12 @@ use super::ScimSshPublicKey;
 use crate::attribute::Attribute;
 use crate::internal::UiHint;
 use scim_proto::ScimEntryHeader;
-use serde::Serialize;
+use serde::{Deserialize, Serialize};
 use serde_with::{base64, formats, hex::Hex, serde_as, skip_serializing_none};
 use std::collections::{BTreeMap, BTreeSet};
 use time::format_description::well_known::Rfc3339;
 use time::OffsetDateTime;
+use tracing::debug;
 use url::Url;
 use utoipa::ToSchema;
 use uuid::Uuid;
@@ -28,7 +29,7 @@ pub struct ScimEntryKanidm {
     pub attrs: BTreeMap<Attribute, ScimValueKanidm>,
 }
 
-#[derive(Serialize, Debug, Clone, ToSchema)]
+#[derive(Serialize, Deserialize, Debug, Clone, ToSchema)]
 pub enum ScimAttributeEffectiveAccess {
     /// All attributes on the entry have this permission granted
     Grant,
@@ -49,7 +50,7 @@ impl ScimAttributeEffectiveAccess {
     }
 }
 
-#[derive(Serialize, Debug, Clone, ToSchema)]
+#[derive(Serialize, Deserialize, Debug, Clone, ToSchema)]
 #[serde(rename_all = "camelCase")]
 pub struct ScimEffectiveAccess {
     /// The identity that inherits the effective permission
@@ -209,7 +210,7 @@ pub struct ScimOAuth2ClaimMap {
     pub values: BTreeSet<String>,
 }
 
-#[derive(Serialize, Debug, Clone, PartialEq, Eq, ToSchema)]
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq, ToSchema)]
 #[serde(rename_all = "camelCase")]
 pub struct ScimReference {
     pub uuid: Uuid,
@@ -255,6 +256,92 @@ pub enum ScimValueKanidm {
     OAuth2ClaimMap(Vec<ScimOAuth2ClaimMap>),
     KeyInternal(Vec<ScimKeyInternal>),
     UiHints(Vec<UiHint>),
+}
+
+#[serde_as]
+#[derive(Serialize, Deserialize, Debug, Clone, ToSchema)]
+pub struct ScimPerson {
+    pub uuid: Uuid,
+    pub name: String,
+    pub displayname: Option<String>,
+    pub spn: String,
+    pub description: Option<String>,
+    pub mails: Vec<ScimMail>,
+    pub managed_by: Option<ScimReference>,
+    pub groups: Vec<ScimReference>,
+}
+
+impl TryFrom<ScimEntryKanidm> for ScimPerson {
+    type Error = ();
+
+    fn try_from(scim_entry: ScimEntryKanidm) -> Result<Self, Self::Error> {
+        let attr_str = |attr: &Attribute| -> Option<&str> {
+            match scim_entry.attrs.get(attr) {
+                Some(ScimValueKanidm::String(inner_string)) => Some(inner_string.as_str()),
+                Some(sv) => {
+                    debug!("SCIM entry had the {} attribute but it was not a ScimValueKanidm::String type, actual: {:?}", attr, sv);
+                    None
+                }
+                None => None,
+            }
+        };
+
+        let attr_mails = || -> Option<&Vec<ScimMail>> {
+            match scim_entry.attrs.get(&Attribute::Mail) {
+                Some(ScimValueKanidm::Mail(inner_string)) => Some(inner_string),
+                Some(sv) => {
+                    debug!("SCIM entry had the {} attribute but it was not a ScimValueKanidm::Mail type, actual: {:?}", Attribute::Mail, sv);
+                    None
+                }
+                None => None,
+            }
+        };
+
+        let attr_reference = |attr: &Attribute| -> Option<&ScimReference> {
+            match scim_entry.attrs.get(attr) {
+                Some(ScimValueKanidm::EntryReference(refer)) => Some(refer),
+                Some(sv) => {
+                    debug!("SCIM entry had the {} attribute but it was not a ScimValueKanidm::ScimReference type, actual: {:?}", attr, sv);
+                    None
+                }
+                None => None,
+            }
+        };
+
+        let attr_references = |attr: &Attribute| -> Option<&Vec<ScimReference>> {
+            match scim_entry.attrs.get(attr) {
+                Some(ScimValueKanidm::EntryReferences(refs)) => Some(refs),
+                Some(sv) => {
+                    debug!("SCIM entry had the {} attribute but it was not a ScimValueKanidm::EntryReferences type, actual: {:?}", attr, sv);
+                    None
+                }
+                None => None,
+            }
+        };
+
+        let uuid = scim_entry.header.id;
+        let name = attr_str(&Attribute::Name).ok_or(())?.to_string();
+        let displayname = attr_str(&Attribute::DisplayName).map(|s| s.to_string());
+        let spn = attr_str(&Attribute::Spn).ok_or(())?.to_string();
+        let description = attr_str(&Attribute::Description).map(|t| t.to_string());
+        let mails = attr_mails().cloned().unwrap_or_default();
+        let groups = attr_references(&Attribute::DirectMemberOf)
+            .cloned()
+            .unwrap_or(vec![]);
+
+        let managed_by = attr_reference(&Attribute::EntryManagedBy).cloned();
+
+        Ok(ScimPerson {
+            uuid,
+            name,
+            displayname,
+            spn,
+            description,
+            mails,
+            managed_by,
+            groups,
+        })
+    }
 }
 
 impl From<bool> for ScimValueKanidm {
--- a/proto/src/scim_v1/server.rs
+++ b/proto/src/scim_v1/server.rs
@@ -9,7 +9,6 @@ use serde_with::{base64, formats, hex::Hex, serde_as, skip_serializing_none};
 use std::collections::{BTreeMap, BTreeSet};
 use time::format_description::well_known::Rfc3339;
 use time::OffsetDateTime;
-use tracing::debug;
 use url::Url;
 use utoipa::ToSchema;
 use uuid::Uuid;
@@ -263,7 +262,7 @@ pub enum ScimValueKanidm {
 pub struct ScimPerson {
     pub uuid: Uuid,
     pub name: String,
-    pub displayname: Option<String>,
+    pub displayname: String,
     pub spn: String,
     pub description: Option<String>,
     pub mails: Vec<ScimMail>,
@@ -275,61 +274,67 @@ impl TryFrom<ScimEntryKanidm> for ScimPerson {
     type Error = ();
 
     fn try_from(scim_entry: ScimEntryKanidm) -> Result<Self, Self::Error> {
-        let attr_str = |attr: &Attribute| -> Option<&str> {
-            match scim_entry.attrs.get(attr) {
-                Some(ScimValueKanidm::String(inner_string)) => Some(inner_string.as_str()),
-                Some(sv) => {
-                    debug!("SCIM entry had the {} attribute but it was not a ScimValueKanidm::String type, actual: {:?}", attr, sv);
-                    None
-                }
-                None => None,
-            }
-        };
-
-        let attr_mails = || -> Option<&Vec<ScimMail>> {
-            match scim_entry.attrs.get(&Attribute::Mail) {
-                Some(ScimValueKanidm::Mail(inner_string)) => Some(inner_string),
-                Some(sv) => {
-                    debug!("SCIM entry had the {} attribute but it was not a ScimValueKanidm::Mail type, actual: {:?}", Attribute::Mail, sv);
-                    None
-                }
-                None => None,
-            }
-        };
-
-        let attr_reference = |attr: &Attribute| -> Option<&ScimReference> {
-            match scim_entry.attrs.get(attr) {
-                Some(ScimValueKanidm::EntryReference(refer)) => Some(refer),
-                Some(sv) => {
-                    debug!("SCIM entry had the {} attribute but it was not a ScimValueKanidm::ScimReference type, actual: {:?}", attr, sv);
-                    None
-                }
-                None => None,
-            }
-        };
-
-        let attr_references = |attr: &Attribute| -> Option<&Vec<ScimReference>> {
-            match scim_entry.attrs.get(attr) {
-                Some(ScimValueKanidm::EntryReferences(refs)) => Some(refs),
-                Some(sv) => {
-                    debug!("SCIM entry had the {} attribute but it was not a ScimValueKanidm::EntryReferences type, actual: {:?}", attr, sv);
-                    None
-                }
-                None => None,
-            }
-        };
-
         let uuid = scim_entry.header.id;
-        let name = attr_str(&Attribute::Name).ok_or(())?.to_string();
-        let displayname = attr_str(&Attribute::DisplayName).map(|s| s.to_string());
-        let spn = attr_str(&Attribute::Spn).ok_or(())?.to_string();
-        let description = attr_str(&Attribute::Description).map(|t| t.to_string());
-        let mails = attr_mails().cloned().unwrap_or_default();
-        let groups = attr_references(&Attribute::DirectMemberOf)
-            .cloned()
-            .unwrap_or(vec![]);
+        let name = scim_entry
+            .attrs
+            .get(&Attribute::Name)
+            .and_then(|v| match v {
+                ScimValueKanidm::String(s) => Some(s.clone()),
+                _ => None,
+            })
+            .ok_or(())?;
 
-        let managed_by = attr_reference(&Attribute::EntryManagedBy).cloned();
+        let displayname = scim_entry
+            .attrs
+            .get(&Attribute::DisplayName)
+            .and_then(|v| match v {
+                ScimValueKanidm::String(s) => Some(s.clone()),
+                _ => None,
+            })
+            .ok_or(())?;
+
+        let spn = scim_entry
+            .attrs
+            .get(&Attribute::Spn)
+            .and_then(|v| match v {
+                ScimValueKanidm::String(s) => Some(s.clone()),
+                _ => None,
+            })
+            .ok_or(())?;
+
+        let description = scim_entry
+            .attrs
+            .get(&Attribute::Description)
+            .and_then(|v| match v {
+                ScimValueKanidm::String(s) => Some(s.clone()),
+                _ => None,
+            });
+
+        let mails = scim_entry
+            .attrs
+            .get(&Attribute::Mail)
+            .and_then(|v| match v {
+                ScimValueKanidm::Mail(m) => Some(m.clone()),
+                _ => None,
+            })
+            .unwrap_or_default();
+
+        let groups = scim_entry
+            .attrs
+            .get(&Attribute::DirectMemberOf)
+            .and_then(|v| match v {
+                ScimValueKanidm::EntryReferences(v) => Some(v.clone()),
+                _ => None,
+            })
+            .unwrap_or_default();
+
+        let managed_by = scim_entry
+            .attrs
+            .get(&Attribute::EntryManagedBy)
+            .and_then(|v| match v {
+                ScimValueKanidm::EntryReference(v) => Some(v.clone()),
+                _ => None,
+            });
 
         Ok(ScimPerson {
             uuid,
--- a/proto/src/scim_v1/server.rs
+++ b/proto/src/scim_v1/server.rs
@@ -4,7 +4,7 @@ use super::ScimSshPublicKey;
 use crate::attribute::Attribute;
 use crate::internal::UiHint;
 use scim_proto::ScimEntryHeader;
-use serde::{Deserialize, Serialize};
+use serde::Serialize;
 use serde_with::{base64, formats, hex::Hex, serde_as, skip_serializing_none};
 use std::collections::{BTreeMap, BTreeSet};
 use time::format_description::well_known::Rfc3339;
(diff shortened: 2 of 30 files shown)


server/core/src/https/extractors/mod.rs
```diff
@@ -181,6 +181,19 @@ impl FromRequestParts<ServerState> for VerifiedClientInformation {
...
 
+// Mockup for accessInfo, should probably be its own non-conflicting transaction type thing like domaininforead in https://github.com/kanidm/kanidm/pull/3016/files
+pub struct AccessInfo {
+    pub has_any_admin_access: bool,
+}
+
+impl AccessInfo {
+    pub fn new() -> Self {
+        AccessInfo {
+            has_any_admin_access: true,
```
Firstyear, line 192: For now, can we hide this behind a feature flag?

server/core/src/https/extractors/mod.rs
```diff
@@ -181,6 +181,19 @@ impl FromRequestParts<ServerState> for VerifiedClientInformation {
     }
 }
 
+// Mockup for accessInfo, should probably be its own non-conflicting transaction type thing like domaininforead in https://github.com/kanidm/kanidm/pull/3016/files
```
Firstyear, line 184: We'll need to set some info into the cookies about this.

server/core/src/https/mod.rs
```diff
@@ -140,6 +140,7 @@ pub fn get_js_files(role: ServerRole) -> Result<JavaScriptFiles, ()> {
                 ("external/bootstrap.bundle.min.js", None, false, false),
                 ("external/htmx.min.1.9.12.js", None, false, false),
                 ("external/cred_update.js", None, false, false),
+                ("external/admin.js", None, false, false),
```
Firstyear, line 143: probably shouldn't be in external, external means "sourced elsewhere". Should be in the root. 

Also I don't like extra js :P

server/core/src/https/views/admin/accounts.rs
```diff
@@ -0,0 +1,81 @@
...
+        EntryClass::Account.into()
+    )]));
+    let base: Vec<Entry> = state
+        .qe_r_ref
+        .handle_internalsearch(client_auth_info.clone(), filter, None, kopid.eventid)
+        .map_err(|op_err| HtmxError::new(&kopid, op_err))
+        .await?;
+
+    let accounts = base.into_iter().map(|entry: Entry| {
+        let uuid = entry.attrs.get(ATTR_UUID).unwrap_or(&vec![]).first().unwrap_or(&"".to_string()).clone();
```
Firstyear, line 57: This feels really messy/hacky. We should get the entry either with a search method for scim (similar to https://github.com/kanidm/kanidm/blob/master/server/core/src/actors/v1_scim.rs#L206 ) or making a new version of https://github.com/kanidm/kanidm/blob/master/server/core/src/actors/v1_read.rs#L422 which DOES NOT call into_proto_array() so that we have the strongly typed EntrySealed version instead.

Firstyear: Most likely we want the scim version because that will have pagination.

server/core/src/https/views/admin/admin.rs
```diff
@@ -0,0 +1,45 @@
+use crate::https::extractors::{AccessInfo, VerifiedClientInformation};
+use crate::https::middleware::KOpId;
```
Firstyear, line 2: Is this page needed? Seems like it doesn't do much?

server/core/templates/admin/admin_group_mail_partial.html
```diff
@@ -0,0 +1,16 @@
+<li id="groupMail(( mail|hash ))" class="list-group-item d-flex flex-row justify-content-between">
+    <div class="d-flex align-items-center">
+        <input id="mailValue(( mail|hash ))" value="(( mail ))" hidden="hidden">
+        (( mail ))
+    </div>
+    <div class="buttons float-end">
+        (% if can_edit %)
+        <button hx-target="#groupMail(( mail|hash ))"
+                hx-swap="delete"
+                hx-delete="/ui/admin/group/(( group_uuid ))/mail"
```
Firstyear, line 10: This won't work as I don't think you have a "delete" action defined?

server/core/templates/admin/admin_group_mail_partial.html
```diff
@@ -0,0 +1,16 @@
+<li id="groupMail(( mail|hash ))" class="list-group-item d-flex flex-row justify-content-between">
+    <div class="d-flex align-items-center">
+        <input id="mailValue(( mail|hash ))" value="(( mail ))" hidden="hidden">
```
Firstyear, line 3: Interesting idea to use the hash as a stable key to id which element to remove.

Firstyear (changes requested):
Okay, I think for now pause on features. Let's get what you have already to a workable state.

I think the big issue you are hitting is you're using the older/messier proto entry apis. This is why you have this wild unwray_or(vec) chains that really aren't going to be accepted in the PR. 

I think this is a case where you need to stop and step back. What were the problems you hit? What's been the barrier? Sometimes when you go in and write a bunch of stuff you find all these problems, and then you have built that experience to say "oh, right if only I did it a different way" and we throw out what you have and start a fresh. 

I think we don't have to throw out all of this, but a lot of the routes/handlers in the group path will need a rewrite. I can already see that you probably learnt a lot here in the process. 

I think the big thing you needed to be willing to do here is write something in v1_write.rs or v1_scim.rs that handles the strong types in a single transaciton. view_group_member-post is a great example of this, because you end up making multiple transactions in one call which is NOT thread safe at all because of race conditions. Where what you needed was to do a single txn that does the search and modify in a single go, or just send the modify event directly rather than mucking about with the member search. 

Anyway, I think for now, lets put the brakes on this. Let's rework what you have to get it into shape. 

I would suggest your first step is to go to https://github.com/kanidm/kanidm/blob/master/server/core/src/actors/v1_scim.rs and make a scim_entry_search that returns a set of scim entries.

Firstyear:
I will admit, something that is working against you here is our lack of docs on how to interact with the server apis. This part of the code will need some handlers in v1_write.rs or a new module that has strong types that we expect for the webui. The reason I suggest scim here is it already has strong types so you don't need to diy that.

Firstyear:
Lets also rebase to master, I will try to work with you more on this to help out :)

server/core/src/https/mod.rs
ToxicMushroom: I'll try htmx OOB swaps

server/core/src/https/views/admin/admin.rs
ToxicMushroom: It's unused yes

server/core/src/https/views/admin/accounts.rs
```diff
@@ -0,0 +1,94 @@
...
+            push_url,
+            HtmlTemplate(AccountsView {
+                access_info: AccessInfo::new(),
+                partial: accounts_partial,
+            }),
+        ).into_response()
+    })
+}
+
+fn get_scim_attr_string(entry: &ScimEntryKanidm, attr: &Attribute) -> Option<String> {
```
Firstyear, line 83: It would likely be better to have this as a fn on the ScimEntryKanidm, and it should return an &str instead, it's up to the caller to decide if they need to clone or take ownership or not.

Firstyear: And it makes the code a bit neater, like "entry.attr_str(&Attribute::Name)" would be how you call it instead.

server/core/src/https/views/admin/admin.rs
Firstyear: Should we remove it then? Just to help keep this a bit more manageable as a PR

server/core/src/https/views/admin/groups.rs
```diff
@@ -0,0 +1,617 @@
...
+
+pub(crate) async fn view_group_save_post(
+    State(state): State<ServerState>,
+    HxRequest(_is_htmx): HxRequest,
+    Extension(kopid): Extension<KOpId>,
+    VerifiedClientInformation(client_auth_info): VerifiedClientInformation,
+    Path(guuid): Path<Uuid>,
+    Form(data): Form<GroupDetailsFormData>,
+) -> axum::response::Result<Response> {
+    let filter = filter_all!(f_and!([f_eq(Attribute::Class, EntryClass::Group.into())]));
```
Firstyear, line 153: For now, let's replace this with a todo!() or a 500 error or something as we need to rework this a bit, and it'll help to keep it clearer.

server/core/src/https/views/admin/groups.rs
```diff
@@ -0,0 +1,617 @@
...
+
+pub(crate) async fn view_group_new_member_post(
+    State(state): State<ServerState>,
+    HxRequest(_is_htmx): HxRequest,
+    Extension(kopid): Extension<KOpId>,
+    VerifiedClientInformation(client_auth_info): VerifiedClientInformation,
+    Path(guuid): Path<Uuid>,
+    Form(data): Form<GroupAddMemberFormData>,
+) -> axum::response::Result<Response> {
+    let mut ors = vec![f_id(data.member.as_str())];
```
Firstyear, line 212: Same here re 500 / todo!()

server/core/src/https/views/admin/groups.rs
```diff
@@ -0,0 +1,617 @@
...
+    State(state): State<ServerState>,
+    HxRequest(_is_htmx): HxRequest,
+    Extension(kopid): Extension<KOpId>,
+    VerifiedClientInformation(client_auth_info): VerifiedClientInformation,
+    Path(guuid): Path<Uuid>,
+    Form(data): Form<GroupAddMailFormData>,
+) -> axum::response::Result<Response> {
+    let filter = filter_all!(f_and!([f_eq(Attribute::Class, EntryClass::Group.into()),]));
+    let mail = data.mail;
+    state
```
Firstyear, line 299: And here.

server/core/src/https/views/admin/groups.rs
```diff
@@ -0,0 +1,617 @@
...
+
+    MemberInfo {
+        uuid,
+        name,
+        spn,
+        displayname,
+    }
+}
+
+fn entry_into_groupinfo(entry: &Entry) -> GroupInfo {
```
Firstyear, line 558: This needs to be deleted.

server/core/src/https/views/admin/groups.rs
```diff
@@ -0,0 +1,617 @@
...
+    let mut groups: Vec<_> = base
+        .iter()
+        .map(|entry: &Entry| entry_into_groupinfo(entry))
+        .collect();
+    groups.sort_by_key(|gi| gi.uuid.clone());
+    groups.reverse();
+    Ok(groups)
+}
+
+fn entry_into_memberinfo(entry: &Entry) -> MemberInfo {
```
Firstyear, line 520: And this.

server/core/src/https/views/admin/groups.rs
```diff
@@ -0,0 +1,617 @@
...
+        .qe_r_ref
+        .handle_internalsearch(client_auth_info.clone(), filter, None, kopid.eventid)
+        .map_err(|op_err| HtmxError::new(&kopid, op_err))
+        .await?;
+
+    let mut groups: Vec<_> = base
+        .iter()
+        .map(|entry: &Entry| entry_into_groupinfo(entry))
+        .collect();
+    groups.sort_by_key(|gi| gi.uuid.clone());
```
Firstyear, line 515: This is going to be super costly, I think we don't sort here - we'll need an open issue to have sorting done in the server backend where it can be done muche more effectively.

proto/src/scim_v1/server.rs
```diff
@@ -216,6 +216,21 @@ pub enum ScimValueKanidm {
     KeyInternal(Vec<ScimKeyInternal>),
 }
 
+impl ScimEntryKanidm{
+    pub fn attr_str(&self, attr: &Attribute) -> Option<&str> {
+        match self.attrs.get(attr) {
+            Some(ScimValueKanidm::String(inner_string)) => Some(inner_string.as_str()),
+            Some(sv) => {
+                eprintln!("Scim entry did had the {attr} attr but expected ScimValueKanidm::String, actual: {sv:?}");
```
yaleman, line 224: No eprintln's or inline variables please 😄

```suggestion
                debug!("SCIM entry had the {} attribute but it was not a ScimValueKanidm::String type, actual: {:?}", attr, sv);
```

server/core/Cargo.toml
```diff
@@ -17,6 +17,7 @@ doctest = false
 
 [features]
 ui_htmx = []
+ui_admin = []
```
yaleman, line 20: Why's this a feature in addition to the `ui_htmx` feature?

server/core/templates/admin/admin_partial_base.html
```diff
@@ -0,0 +1,54 @@
+<main class="container-xxl pb-5">
+    <div class="d-flex flex-sm-row flex-column">
+        <div class="list-group side-menu">
+            <a href="/ui/admin" hx-target="#main" class="list-group-item list-group-item-action (% block overview_item_extra_classes%)(%endblock%)">
```
yaleman, line 4: Why all the blocks for attributes and not just variables?

server/core/templates/admin/admin_group_create_partial.html
```diff
@@ -0,0 +1,50 @@
...
+
+    <fieldset class="mb-4">
+        <legend>Optional Attributes</legend>
+        <div class="mb-3">
+            <label for="guidInput" class="form-label">Posix guid
+                <a target="_blank" href="https://kanidm.github.io/kanidm/stable/accounts/posix_accounts_and_groups.html#posix-accounts-and-groups">help</a>
+            </label>
+            <input type="number" class="form-control" id="guidInput" name="guid">
+            <div class="invalid-feedback">
+                Please choose a valid posix group uid.
```
yaleman, line 30:
```suggestion
                Please choose a valid POSIX group ID.
```

server/core/templates/admin/admin_group_member_partial.html
```diff
@@ -0,0 +1,14 @@
+<li id="groupMember(( member.uuid ))" class="list-group-item d-flex flex-row justify-content-between">
+    <div class="d-flex align-items-center">
+        <a href="/ui/admin/group/(( member.uuid ))/view" hx-target="#main">(( member.spn ))</a>
+    </div>
+    <div class="buttons float-end">
+        (% if can_edit %)
+        <button hx-target="#groupMember(( member.uuid ))"
+                hx-swap="delete"
+                hx-delete="/ui/admin/group/(( group_uuid ))/member/(( member.uuid ))"
+                hx-confirm="Are you sure you want to remove (( member.name )) from the group ?"
```
yaleman, line 10:
```suggestion
                hx-confirm="Are you sure you want to remove (( member.name )) from the group?"
```

server/core/templates/admin/admin_group_view_partial.html
```diff
@@ -0,0 +1,116 @@
...
+    </a>
+(% endif %)
+
+(% include "admin_group_details_partial.html" %)
+
+<hr>
+<label class="mt-3 fw-bold">Emails</label>
+<form hx-validate="true" hx-ext="bs-validation">
+(% if group.mails.len() == 0 %)
+    <p>none</p>
```
yaleman, line 33:
```suggestion
    <p>There are no email addresses associated with this group.</p>
```

server/core/templates/admin/admin_group_view_partial.html
```diff
@@ -0,0 +1,116 @@
...
+                            hx-post="/ui/admin/api/group/(( group.uuid ))/mail"
+                            hx-include="#groupAddEmailInput"
+                            hx-target="#addEmailField"
+                            hx-swap="beforebegin"
+                            type="button">
+                        Add Email
+                    </button>
+                </div>
+                <div class="invalid-feedback">
+                    Please choose a valid email.
```
yaleman, line 64:
```suggestion
                    Please enter a valid email address.
```

server/core/templates/admin/admin_group_view_partial.html
```diff
@@ -0,0 +1,116 @@
...
+</form>
+
+<label class="mt-3 fw-bold">Account Policy (ACP)</label>
+<div>
+(% if group.acp.enabled %)
+    Nothing here yet 🙈🙈
+(% else %)
+    Currently disabled.
+    (% if can_edit %)
+        <button hx-post="/ui/api/group/(( group.uuid ))/enable_acp" hx-confirm="Currently no known way to disable again, ok to continue." class="btn btn-success d-block">Enable</button>
```
yaleman, line 79: The message here is confusing, perhaps:

```suggestion
        <button hx-post="/ui/api/group/(( group.uuid ))/enable_acp" hx-confirm="Kanidm does not currently support removing these policies, are you sure you want to continue?" class="btn btn-success d-block">Enable</button>
```

server/core/templates/admin/admin_group_view_partial.html
```diff
@@ -0,0 +1,116 @@
...
+            </li>
+        (% endif %)
+    </ol>
+(% endif %)
+</form>
+
+<label class="mt-3 fw-bold">Account Policy (ACP)</label>
+<div>
+(% if group.acp.enabled %)
+    Nothing here yet 🙈🙈
```
yaleman, line 75:
```suggestion
<!-- TODO: implement ACP view for groups -->
    Nothing here yet 🙈🙈
```

server/core/templates/admin/admin_group_view_partial.html
```diff
@@ -0,0 +1,116 @@
...
+                        hx-post="/ui/admin/api/group/(( group.uuid ))/member"
+                        hx-include="#groupAddMemberInput"
+                        hx-target="#addMemberField"
+                        hx-swap="beforebegin"
+                        type="button">
+                    Add Member
+                </button>
+            </div>
+            <div class="invalid-feedback">
+                Please choose a valid email.
```
yaleman, line 108: see previous suggestion about wording

server/core/templates/admin/admin_groups_partial.html
```diff
@@ -0,0 +1,36 @@
...
+    <ul class="list-group">
+    (% for group in groups %)
+        <li class="list-group-item d-flex flex-row justify-content-between">
+            <div class="d-flex align-items-center">
+                <a href="/ui/admin/group/(( group.uuid ))/view" hx-target="#main">(( group.name ))</a> <span class="text-secondary d-none d-lg-inline-block mx-4">(( group.uuid ))</span>
+            </div>
+            <div class="buttons float-end">
+                (% if can_rw %)
+                    <button hx-target="#main" hx-get="/ui/admin/group/(( group.uuid ))/edit" type="button" class="btn btn-outline-secondary">Edit</button>
+                    <button hx-target="#main" hx-post="/ui/admin/group/(( group.uuid ))/delete" hx-confirm="Are you sure you want to delete group (( group.name )) ?" type="button" class="btn btn-danger">Delete</button>
```
yaleman, line 28:
```suggestion
                    <button hx-target="#main" hx-post="/ui/admin/group/(( group.uuid ))/delete" hx-confirm="Are you sure you want to delete group (( group.name ))?" type="button" class="btn btn-danger">Delete</button>
```

server/core/templates/admin/admin_partial_base.html
```diff
@@ -0,0 +1,54 @@
...
+                Saved.
+            </div>
+        </div>
+    </div>
+
+
+    <div class="toast-container position-fixed bottom-0 end-0 p-3">
+        <div id="permissionDeniedToast" class="toast" role="alert" aria-live="assertive" aria-atomic="true">
+            <div class="toast-header">
+<!--              error icon  <img src="..." class="rounded me-2" alt="...">-->
```
yaleman, line 44:
```suggestion
```

yaleman (changes requested):

server/core/Cargo.toml
ToxicMushroom: I thought that was what William requested here https://github.com/kanidm/kanidm/pull/3019#discussion_r1774305484. Perhaps to ship it in release while hiding it from default view because the hiding wasn't implemented yet

Firstyear: I'm okay with this as the feature gate for now.

server/core/templates/admin/admin_partial_base.html
ToxicMushroom: They're used to mark the active menu-item by the concrete page by implementing the block. 
I felt that this was clearer although hacky and more cohesive than passing variables from rust

server/core/templates/admin/admin_group_view_partial.html
ToxicMushroom: Should the `Please choose a valid POSIX group ID.` then also become `Please enter a valid POSIX group ID.` to be consistent

ToxicMushroom:
It's still a bit rough but a lot more reviewable now

ToxicMushroom:
For `limit navigation based on access` I wanted to discuss how to best implement that.

Please add your thoughts :) 

For context: examples of items are the admin button on the navbar and actually can't think of any others at the moment but I'm sure it'll arise in the future. 

---------
- Option 1 I thought of was: 
  - A DomainInfo equivalent extractor but with all the accessinfo needed for navigating the webui.
with other words: navbar- and sidemenu-item visibility-influencing access rights, combined into one AccessInfoAll struct.

➖ This means wasteful db lookups for many pages. (bad performance scaling)
➕ Easy to use developer wise, only passing 1 struct into navigation elements

---------
- Option 2:
  - AccessInfo struct per navigational menu. AccessInfoNavbar, AccessInfoAdmin, AccessInfoProfile...

➖ Lots of structs that are heavily tied to UI get into the internal files.. (very bad imo)
➕ No wasted db lookups, still quite easy to use.

---------
- Option 3:
  - Only AccessInfo for the big Navbar, Other navigational elements will have their access lookup up in their respective files with a helper function. 

➖ Submenu access queries can't be ran in parallel I think like with extractors I think. (minor)
➖ Developer usability is kinda in-between, a little confusing perhaps for others why it was split up like this. (minor)
➕ No wasted queries

ToxicMushroom:
Other problems/questions:
- what should even the condition be to view the admin panel button?
- how do I distinguish scimreferences between users/group (only has uuid and spn) ?
- I think I want to show all attributes in a generic way instead of the hardcoding that I was doing right now, thoughts ?

server/core/templates/navbar.html
```diff
@@ -30,6 +30,12 @@
                     <a class="nav-link" href=((Urls::Profile))>
                         <span data-feather="file"></span>Profile</a>
                 </li>
+                (% if access_info.has_any_admin_access %)
```
Firstyear, line 33: Until we have a proper way to limit this by role, remove this from the navbar.

proto/Cargo.toml
```diff
@@ -38,6 +38,7 @@ uuid = { workspace = true, features = ["serde"] }
 webauthn-rs-proto = { workspace = true }
 sshkey-attest = { workspace = true }
 sshkeys = { workspace = true }
+log = "0.4.25"
```
Firstyear, line 41: Why are you adding the log crate? We have tracing already and it exposes a log interface.

proto/src/scim_v1/server.rs
```diff
@@ -3,8 +3,9 @@ use super::ScimOauth2ClaimMapJoinChar;
 use super::ScimSshPublicKey;
 use crate::attribute::Attribute;
 use crate::internal::UiHint;
+use log::debug;
```
Firstyear, line 6: Tracing?

server/core/src/https/extractors/mod.rs
Firstyear: Actually we already have this in the UAT we just need a way to expose it.

server/core/src/https/views/admin/accounts.rs
```diff
@@ -0,0 +1,216 @@
...
+        // TODO: Filtering away unsuccessful entries may not be desired.
+        .filter_map(scimentry_into_accountinfo)
+        .collect();
+
+    accounts.sort_by_key(|gi| gi.uuid);
+    accounts.reverse();
+    Ok(accounts)
+}
+
+fn scimentry_into_accountinfo(scim_entry: ScimEntryKanidm) -> Option<AccountInfo> {
```
Firstyear, line 192: This can probablyf be "Impl TryFrom<ScimEntryKanidm> on AccountInfo.". Though isn't there already a ScimPerson or something we crould use?

Firstyear (changes requested):

Firstyear:
I think to prevent this getting too large, we need to either break it up (just start with accounts perhaps) but also we need to disable it from the navbar by default so we can do smaller iterative merges.

Firstyear:
> For `limit navigation based on access` I wanted to discuss how to best implement that.
> 
> Please add your thoughts :)
> 
> For context: examples of items are the admin button on the navbar and actually can't think of any others at the moment but I'm sure it'll arise in the future.

You mean like this:

https://github.com/kanidm/kanidm/blob/master/proto/src/internal/token.rs#L47

This is extracted and set based on groups at login, so you can also set UiHints on access control groups which then influence what UI hints you get.

proto/Cargo.toml
ToxicMushroom: my IDE sometimes gets weird ideas (like adding crates) and I don't always check what its suggesting properly, sorry

server/core/src/https/views/admin/accounts.rs
ToxicMushroom: ScimSyncPerson exists but it doesn't contain other fun attributes like what groups they're a member of, and it doesn't use ScimMail for emails strangely.

Also kind of ties into my last question `I think I want to show all attributes in a generic way instead of the hardcoding that I was doing right now, thoughts ?`.
  with other words: after rebasing I started to feel the urge to remove AccountInfo and just pass the attribute map to askama. But that makes a nice layout harder I suppose.

ToxicMushroom:
> You mean like this:
> 
> https://github.com/kanidm/kanidm/blob/master/proto/src/internal/token.rs#L47
> 
> This is extracted and set based on groups at login, so you can also set UiHints on access control groups which then influence what UI hints you get.

Yep that solves it :)

I'll also further reduce this pr to just account viewing then

Firstyear:
Sounds good.

proto/src/scim_v1/server.rs
ToxicMushroom: I don't see tracing in proto, should I add it or should those functions not log

Firstyear: Add tracing because that's what we use everywhere.

server/core/src/https/views/admin/accounts.rs
Firstyear: Don't re-use scimSyncPerson, that's for sync accounts. 

Make something that is presentable as a person/account. I think that's a better way. That's why there is https://github.com/kanidm/kanidm/blob/master/proto/src/scim_v1/server.rs which is server to client entry types. Make a ScimPerson there.

Be-ing:
I just want to say a big thanks for working on this! I'm looking to deploy Kanidm for self-hosted infrastructure for a ~100 member volunteer organization and this is the biggest missing feature. I and a few other sysadmins can handle creating new users with CLI client, but we'd really like to delegate that to the people who onboard newcomers, and we don't want to ask them to use a CLI program that only runs on Linux and macOS.

Firstyear (commented):
Getting there, some more comments about polish especially re the server parts because I'm a bit fussy about how that should be put together. I think rename "accounts" to "people" and make the ui just about showing people, and that'll also help make it clearer. Otherwise I tested this and it looks good :)

Firstyear:
@ToxicMushroom Don't touch the PR, I'm working on it.

ToxicMushroom:
alright

Firstyear (approved):

PR #3036, opened 2024-09-11
implements additional traits for filter types

### LINKED ISSUE #3035: SCIM: Feature Request: make parse function public
Hi!

First off, thanks for making this crate! I appreciate the work that went into it!

If you're open to feature requests, I have a few small ones. I didn't see any specific guidance in the [Developer Guide](https://kanidm.github.io/kanidm/master/developers/index.html) on how to submit these, so if multiple issues is preferred, just let me know, and I can break it up accordingly. They're grouped here because they all seem like (relatively) small things.

1. Would it be possible to make the `scim_proto::filters::scimfilter::parse()` function public?
2. Since the crate already depends on `serde`, can `ScimFilter` implement `Serialize` and `Deserialize`?
  - I'm not sure how complicated this is with recursive types, so... please and thank you, if it's not too much work? 🙇 
3. Could `ScimFilter` implement `FromStr`? I assume this would be just calling `parse()` or something along those lines.
  - This would make it work out of the box with `clap::Arg`. Without this, it's still simple enough to just pass the `parse` function from (1) to `clap::Arg::value_parser`, so this is just a minor convenience, but it seems intuitive (to me) that `FromStr` and `parse` would... do the same thing... 🤷‍♀️ 

The goal in all of these is to make it easier for other crates to use this one.

If it helps illustrate things better, in my context, I've got an app that's taking SCIM filters in a CLI app (with `clap`) and including them in an HTTP request, and I'd just like to validate them before sending them, so that I can provide nice error messages when there are problems, rather than relying on the remote service to return a decent error message. 😉 

If there's any other information I can provide, please let me know.

Thanks!

colstrom:
# Change summary

This is mostly resolving my own feature request in https://github.com/kanidm/kanidm/issues/3035, with the addition of a `ToString` implementation.

- for each of `filter::AttrPath`, `filter::ScimFilter`, and `filter::ScimComplexFilter`:
  - implements `serde::Serialize` and `serde::Deserialize` using the typical `derive` macros.
  - implements `FromStr` by calling the parsing function for the corresponding type in the crate-private `scimfilter` module.
  - implements `ToString` by hand, so if I botched something, it's probably here. `match` can only ensure completeness, not correctness. 😅
    - Typically I would test something like this by round-tripping through the conversion, but the implementation here parenthesizes all expressions to preserve precedence, so the resulting string may differ from the original input.

Fixes #3035, for the most part. While `parse()` is public, the containing module is not, and changing the module visibility there exposes a lot more just `parse()`, which may or may not be desirable (thoughts?). The `FromStr` implementation seems like a reasonable compromise there.

This is essentially an updated version of [my previous pull request](https://github.com/kanidm/scim/pull/5) to the now-archived [kanidim/scim](https://github.com/kanidm/scim) repo. Quoting the relevant bits from that PR message, since it applies here as well:

> I checked the [Developer Guide](https://kanidm.github.io/kanidm/master/developers/index.html) for any rules or guidance around formatting and pull request structure, and made what I hope to be reasonable assumptions:
>
> - Since this repository does not contain a `rustfmt.toml`, it's formatted using the default formatting rules.
> - The changes are one commit per feature, in case any happen to be undesirable or incorrect.
> - I didn't see any [DCO](https://developercertificate.org/) requirements mentioned, but all commits are signed and include sign-off, just in case.

Notable differences from the original PR are omitting the version bump (per [this comment](https://github.com/kanidm/scim/pull/5#issuecomment-2342688678)), and including implementations for `filter::ScimComplexFilter`.

Checklist

Please let me know if I've missed anything.

Thanks!

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -42,6 +42,7 @@
 - fossdd
 - Wei Jian Gan (weijiangan)
 - adamcstephens
+- Chris Olstrom (colstrom)
 
 ## Acknowledgements
 
--- a/libs/scim_proto/src/filter.rs
+++ b/libs/scim_proto/src/filter.rs
@@ -63,6 +63,26 @@ pub enum ScimComplexFilter {
     LessOrEqual(String, Value),
 }
 
+impl ToString for ScimComplexFilter {
+    fn to_string(&self) -> String {
+        match self {
+            Self::And(this, that) => format!("({} and {})", this.to_string(), that.to_string()),
+            Self::Contains(attrname, value) => format!("({attrname} co {value})"),
+            Self::EndsWith(attrname, value) => format!("({attrname} ew {value})"),
+            Self::Equal(attrname, value) => format!("({attrname} eq {value})"),
+            Self::Greater(attrname, value) => format!("({attrname} gt {value})"),
+            Self::GreaterOrEqual(attrname, value) => format!("({attrname} ge {value})"),
+            Self::Less(attrname, value) => format!("({attrname} lt {value})"),
+            Self::LessOrEqual(attrname, value) => format!("({attrname} le {value})"),
+            Self::Not(expr) => format!("(not ({}))", expr.to_string()),
+            Self::NotEqual(attrname, value) => format!("({attrname} ne {value})"),
+            Self::Or(this, that) => format!("({} or {})", this.to_string(), that.to_string()),
+            Self::Present(attrname) => format!("({attrname} pr)"),
+            Self::StartsWith(attrname, value) => format!("({attrname} sw {value})"),
+        }
+    }
+}
+
 // separator()* "(" e:term() ")" separator()* { e }
 
 peg::parser! {
--- a/libs/scim_proto/src/filter.rs
+++ b/libs/scim_proto/src/filter.rs
@@ -271,6 +271,27 @@ peg::parser! {
     }
 }
 
+impl FromStr for AttrPath {
+    type Err = peg::error::ParseError<peg::str::LineCol>;
+    fn from_str(input: &str) -> Result<Self, Self::Err> {
+        scimfilter::attrpath(input)
+    }
+}
+
+impl FromStr for ScimFilter {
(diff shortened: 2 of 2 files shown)


Firstyear (approved):

Firstyear:
This looks great, thank you! Feel free to add yourself to contributors.md as well if you wish.

Firstyear (approved):

colstrom:
> This looks great, thank you! Feel free to add yourself to contributors.md as well if you wish.

Done. Thanks!

Also, not directly related to this PR, but while _using_ this, I noticed an `eprintln!` [here](https://github.com/kanidm/kanidm/blob/d3891e301f6382f8b77526cf88aa7718744197ea/libs/scim_proto/src/filter.rs#L203). Was that deliberate, or left over from some parser development context? I'm not sure, since it is actually helpful when there's an error, but it also shows up when there isn't, so... either seems plausible. 🤷‍♀️

On an even _more_ unrelated note... given the run time of the GHA workflows, it might be worth looking into something like [paths-filter](https://github.com/dorny/paths-filter) to skip the more expensive checks when no source files were changed (unless you have some sort of fancy workflow that validates markdown changes?). 🤔

yaleman:
> Also, not directly related to this PR, but while _using_ this, I noticed an `eprintln!` [here](https://github.com/kanidm/kanidm/blob/d3891e301f6382f8b77526cf88aa7718744197ea/libs/scim_proto/src/filter.rs#L203). Was that deliberate, or left over from some parser development context? I'm not sure, since it is actually helpful when there's an error, but it also shows up when there isn't, so... either seems plausible. 🤷‍♀️

Yeah, that'll be some leftover thing :S 
 
> On an even _more_ unrelated note... given the run time of the GHA workflows, it might be worth looking into something like [paths-filter](https://github.com/dorny/paths-filter) to skip the more expensive checks when no source files were changed (unless you have some sort of fancy workflow that validates markdown changes?). 🤔

I've pondered something like that but honestly GHA workflows are already flaky enough as it is, and there's so many cross-dependencies on things in the monorepo that it's difficult to find a balance.

PR #3050, opened 2024-09-20
Document basic user auth for self-hosted GitLab with Kanidm

micolous:
# Change summary

Document basic user auth for self-hosted GitLab with Kanidm

Fixes #

Checklist

- [x] This PR contains no AI generated code
- [x] book chapter included (if relevant)
- [ ] design document included (if relevant)

PR #3055, opened 2024-09-23
Fix landing and redirect URLs for GitLab, add some useful links

micolous:
# Change summary

The redirect URL added in #3050 was for an `oauth2_generic` provider rather than `openid_generic`, so wasn't actually used. I've fixed this to use the correct redirect URL.

This changes the landing URL to `/users/sign_in` which makes things a bit nicer when public access is allowed.

This adds some links to some useful sections in GitLab's documentation for disabling local password auth and making the sign-in process more streamlined.

Fixes #

Checklist

### DIFF
--- a/book/src/integrations/oauth2/examples.md
+++ b/book/src/integrations/oauth2/examples.md
@@ -78,7 +78,7 @@ To set up a self-managed GitLab instance to authenticate with Kanidm:
     and scope access to the `gitlab_users` group:
 
     ```sh
-    kanidm system oauth2 create gitlab GitLab https://gitlab.example.com
+    kanidm system oauth2 create gitlab GitLab https://gitlab.example.com/users/sign_in
     kanidm system oauth2 update-scope-map gitlab gitlab_users email openid profile groups
     ```
 
--- a/book/src/integrations/oauth2/examples.md
+++ b/book/src/integrations/oauth2/examples.md
@@ -155,6 +155,16 @@ To set up a self-managed GitLab instance to authenticate with Kanidm:
 Once GitLab is up and running, you should now see a "Kanidm" option on your
 GitLab sign-in page below the normal login form.
 
+Once you've got everything working, you may wish configure GitLab to:
+
+*   [Automatically redirect to the `openid_connect` provider at the login form](https://docs.gitlab.com/ee/integration/omniauth.html#sign-in-with-a-provider-automatically)
+
+*   [Disable password authentication in GitLab](https://docs.gitlab.com/ee/administration/settings/sign_in_restrictions.html#password-authentication-enabled)
+
+*   [Disable new sign-ups in GitLab](https://docs.gitlab.com/ee/administration/settings/sign_up_restrictions.html)
+
+More information about these features is available in GitLab's documentation.
+
 ## JetBrains Hub and YouTrack
 
 > These instructions were tested with the on-prem version of JetBrains YouTrack
--- a/book/src/integrations/oauth2/examples.md
+++ b/book/src/integrations/oauth2/examples.md
@@ -74,12 +74,11 @@ To set up a self-managed GitLab instance to authenticate with Kanidm:
     kanidm group add-members gitlab_users your_username
     ```
 
-3.  Create a new OAuth2 application configuration in Kanidm (`gitlab`),
-    configure the redirect URL, and scope access to the `gitlab_users` group:
+3.  Create a new OAuth2 application configuration in Kanidm (`gitlab`)
+    and scope access to the `gitlab_users` group:
 
     ```sh
     kanidm system oauth2 create gitlab GitLab https://gitlab.example.com
-    kanidm system oauth2 add-redirect-url gitlab https://gitlab.example.com/users/auth/oauth2_generic/callback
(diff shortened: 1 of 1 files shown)


Firstyear:
@micolous Can you test what happens with the following flow?

* logout of kanidm
* login to kanidm
* click link to gitlab
* go through oauth2 flow etc
* manually open a new tab with kanidm again
* select the link to gitlab

The intent of the two is that "landing" is a url that *triggers* the initialisation of the oauth2 flow *and* is a reasonable location as the landing page at the conclusion of the flow. The redirect url is there to validate the URL's used within the flow.

If you set the "landing" to the redirection URL, that is valid and works, but it also assumes that the redirecation url is also sensible as a landing page.

micolous:
On a GitLab instance where:

* There is no public access allowed – authentication is always required
* The login system is configured to allow multiple sign-in methods (this example config allows both local logins and OIDC via Kanidm)
* It is **not** [configured to automatically sign in with a single provider](https://docs.gitlab.com/ee/integration/omniauth.html#sign-in-with-a-provider-automatically)

With the landing page set to `https://gitlab.example.com`:

1. When logged out of GitLab, clicking the "GitLab" link in the Kanidm applications list redirects me to the login form.

2. If I then click the "Sign in with Kanidm" button, it signs me in using Kanidm, with no further prompts (as I'm signed in to Kanidm, and have already consented to transfer of data) – and GitLab sets its sign-in cookies.

3. If I then go back to Kanidm with GitLab already signed in, it drops me to GitLab's main page (Your work -> Projects).

If you configured GitLab to allow unauthenticated access, then if you weren't logged in, using `https://gitlab.example.com` as the landing URL would just take you to GitLab, and not actually try to log you in.

If I set the landing page to `https://gitlab.example.com/users/sign_in`:

* When logged out of GitLab, I'm sent to the log in form

* When logged in to GitLab, I'm sent to GitLab's main page (Your work -> Projects).

This seems like it might be a better option if the GitLab instance *also* allows unauthenticated access.

GitLab's "Sign in with Kanidm" button does a HTTP POST to `/users/auth/openid_connect` with a CSRF token (`authenticity_token`).

If I set the landing page in Kanidm to `https://gitlab.example.com/users/auth/openid_connect`, I get a HTTP 404 error – with no login form, regardless of whether I was already signed in to GitLab or not. The only link is a JavaScript back button.

If I set the landing page to `/users/auth/openid_connect/callback`:

* When logged out of GitLab, it returns an error "Invalid 'state' parameter", and drops me to the regular login form.

* When logged in to GitLab, I'm sent to GitLab's main page (Your work -> Projects).

I think it may be marginally better to jump to `/users/sign_in`, but there doesn't seem to be any way to have GitLab automatically attempt to use one of the OmniAuth providers with a URL parameter.

One *significant* caveat with all of this is that there's no way to signal to GitLab from Kanidm *which* user you're signed in to.  You could be signed in with a different account on GitLab, and GitLab would not re-authenticate.

But I think that's more of an OAuth2/OIDC design issue 😄 

> The redirect url is there to validate the URL's used within the flow.
>
> If you set the "landing" to the redirection URL, that is valid and works, but it also assumes that the redirecation url is also sensible as a landing page.

[Kanidm adds the landing URL to the set of allowed redirect URLs](https://github.com/kanidm/kanidm/blob/41ac21743a8001a7f38e141e1a4dcf722994ab8b/server/lib/src/idm/oauth2.rs#L462), so this seems unnecessary.

It looks like the OAuth2 application list page mixes terminology and calls the landing page a `redirect_url`; though that's "where to redirect the user when they select an application from the list", rather than some sort of allow list:

https://github.com/kanidm/kanidm/blob/41ac21743a8001a7f38e141e1a4dcf722994ab8b/server/lib/src/idm/applinks.rs#L46-L61

Firstyear:
> But I think that's more of an OAuth2/OIDC design issue 😄

Yeah, we can't do much about that sadly. That's very much an oauth2 issue. 

> It looks like the OAuth2 application list page mixes terminology and calls the landing page a `redirect_url`; though that's "where to redirect the user when they select an application from the list", rather than some sort of allow list:

Yes, we do currently add the "landing url" to the allowed redirection target list. In theory we *don't* need to do this, but for compatibility with our former (less secure) origin matching, we need to account for people that have only set the landing url. It was pretty much done as a "safety" measure to ease migration for people since this is a bit of an annoying change at the best of times. 

The language is also a bit scuffed here because of that - really we need to do a database migration and update all the key names of those values to be a bit clearer. I think mixing that with the current migration we expect people to do might cause some confusion, so we should target a cleanup of the variable names for 1.5.0 once admins have completed this.

micolous:
> Yes, we do currently add the "landing url" to the allowed redirection target list. In theory we _don't_ need to do this, but for compatibility with our former (less secure) origin matching, we need to account for people that have only set the landing url. It was pretty much done as a "safety" measure to ease migration for people since this is a bit of an annoying change at the best of times.

I've added a redirect URL back in again – but now it's the correct `openid_connect` one. 😄

Firstyear (approved):

micolous:
Rebased because #3058

Firstyear (approved):

PR #3067, opened 2024-09-26
Feat: Adding POSIX Password fallback

### LINKED ISSUE #287: kanidm pw-sync inside an account
In some cases we split passwords between roles (login, posix, radius). We could consider for accounts with few privileges to have these auto-synced.

CEbbinghaus:
# Change summary

Allow users to log in using the default credentials if the POSIX password is not set. This improves accessibility since remembering 2 passwords and which one to use when is not something that can be asked of everyone.

---

Fixes #287

Checklist


Firstyear:
> I dislike the option in the config and would prefer a more elegant solution if possible.

I dislike it too! 

The better place would be to add it as a boolean on account policy so that this can be selectively applied to accounts. 

Additionally, it shouldn't be part of the UserAccount::from_entry path. That's a big hammer, we should have the logic to check the account policy as part of the auth flow instead. 

Now, the question for you is if you want to make the changes, or if you want me to follow up and fix the PR to be acceptable?

Firstyear:
Okay, since you would like to do the changes, here is a longer response then to help get this sorted :) 

Some early advice here would be to do each of these numbered steps one at a time, and submit for review as you complete each one of them. That way it's going to be easier than going back over things too much, as each part depends on the earlier bits. 

And if at any time you want me to step in and help, let me know. 

--
The general approach to something like this is:

* Write a test demonstrating the expected user behaviour and the negative cases.
* Add the needed supporting elements (schema, migrations etc).
* Deduplicate code and refactor to minimise the scope of changes if possible
* Add the code needed for the feature.
* Add the cli bits to configure the feature. 

1. Write a test demonstrating the expected user behaviour and the negative cases.

We generally put tests in modules, so look at:

https://github.com/kanidm/kanidm/blob/master/server/lib/src/idm/server.rs#L3418 and https://github.com/kanidm/kanidm/blob/master/server/lib/src/idm/ldap.rs#L860

Here we will want a new test case in each that shows the set of:

* No account policy + no unix pw + acc pw == no auth
* No account policy + unix pw + acc pw == auth (w unix pw)
* Account policy (true) + no unix pw + acc pw == auth (w acc pw)
* Account policy (true) + unix pw + acc pw == auth (w unix pw)
* Account policy (false) + no unix pw + acc pw == no auth
* Account policy (false) + unix pw + acc pw == auth (w unix pw)

This shows us the full matrix of expected possibilities  from the account policy setting. 

If you run the tests at this point they should fail! 

2. Add the needed supporting elements (schema, migrations etc).

To do this we will want to use account policy - this way specific groups can be targetted with a rule that says to *allow* this behaviour, or to explicitly *deny* this behaviour. A benefit of this being in account policy is also that this is replicated, so it's nice an easy to configure. 

Account policy is structured with the class from https://github.com/kanidm/kanidm/blob/master/server/lib/src/constants/schema.rs#L918 

This will needed to be copied to make SCHEMA_CLASS_ACCOUNT_POLICY_DL8 and added here https://github.com/kanidm/kanidm/blob/master/server/lib/src/server/migrations.rs#L596 to trigger the migration.

In the DL8 version you will add a new systemMay attribute, something like `Attribute::AuthAllowPrimaryCredentialForUnix`. 

Then you need to define that attribute, similar to: https://github.com/kanidm/kanidm/blob/master/server/lib/src/constants/schema.rs#L755 - again should be suffixed as DL8, and added to the migrations as above. The UUID is added https://github.com/kanidm/kanidm/blob/master/server/lib/src/constants/uuids.rs#L321 sequentially. 

The Attribute name definition is added here: https://github.com/kanidm/kanidm/blob/master/proto/src/attribute.rs 


Similar to this, the access controls needed to be updated to support the new attribute so that an Account Policy admin can change it: As before, https://github.com/kanidm/kanidm/blob/master/server/lib/src/constants/acp.rs#L494 is copied to a DL8 version, and Attribute::AuthAllowPrimaryCredentialForUnix would be added to search / mod_removed / mod_present. 


Then account policy should be updated with a flag for allow_primary_credential_for_unix. https://github.com/kanidm/kanidm/blob/master/server/lib/src/idm/accountpolicy.rs#L7 

This should resolve in the following way: https://github.com/kanidm/kanidm/blob/master/server/lib/src/idm/accountpolicy.rs#L97 

* If AuthAllowPrimaryCredentialForUnix is NOT set on any of the groups, then the value should default to FALSE (not allowed).
* If AuthAllowPrimaryCredentialForUnix IS set, and *any* value is FALSE, then the value MUST become FALSE.
* If AuthAllowPrimaryCredentialForUnix IS set, and no values are FALSE, and at least ONE is true, then the value can become TRUE.

This logically means that if there is a "true" value in one accpol, then the user can use this, *unless* an accpol exists that strictly denies it by setting false. 


3. Deduplicate code and refactor to minimise the scope of changes if possible

There are currently two places where verify_unix_credential is called: https://github.com/kanidm/kanidm/blob/master/server/lib/src/idm/server.rs#L1341 and https://github.com/kanidm/kanidm/blob/master/server/lib/src/idm/server.rs#L1471 

These two locations are very similar, so we should consider refactoring to make a single fn that handles this path. 

4. Add the code needed for the feature.

The auth_unix call should be changed such that:

* If a unix_pw exists on the account, the unix pw is checked and used.
* If no unix_pw exists, and the account policy value is *true*, then the primary cred password is used. 

In both cases, we need to ensure that the softlock (slock) in the code is updated appropriately to prevent bruteforcing. 

At this point if you run your tests, you should find they start to pass! 

5. Add the cli bits to configure the feature. 

The CLI options are set for account policy here:

https://github.com/kanidm/kanidm/blob/master/tools/cli/src/opt/kanidm.rs#L166

You would add an option to *enable* or *disable* this feature. 

The logic of the command is executed here: https://github.com/kanidm/kanidm/blob/master/tools/cli/src/cli/group/account_policy.rs 

Which indirectly will call on client library features from https://github.com/kanidm/kanidm/blob/master/libs/client/src/group.rs#L40

Firstyear:
cc @ToxicMushroom ^ This might be a good guide for future features ^^:

Firstyear:
And of course, a lot of consideration here is how to ensure this is implemented safely :) I hope this helps!

CEbbinghaus:
I know the current single test is bad (and will break) since it is running all tests in a single IDMS. Planning to split it out into multiple named tests for the time being hoping to at some point use test cases to build the matrix

server/lib/src/idm/server.rs
```diff
@@ -4125,4 +4126,92 @@ mod tests {
...
+                (
+                    Attribute::PrimaryCredential,
+                    Value::Cred(
+                        "primary".to_string(),
+                        Credential::new_password_only(&p, "banana").unwrap()
+                    )
+                )
+            );
+
+            if let Some(allow_primary_cred_fallback) = allow_primary_cred_fallback {
```
Firstyear, line 4160: AllowPrimaryCredFallback is part of account policy - account policy goes on groups, not users. So you need to create a group, or attach this to IDM_ALL_ACCOUNTS which is an easy way to do this in tests https://github.com/kanidm/kanidm/blob/983135e353aa8e69385fa45fef5082992221f328/server/lib/src/idm/credupdatesession.rs#L3697

server/lib/src/idm/server.rs
```diff
@@ -4125,4 +4126,92 @@ mod tests {
...
+        let ct = Duration::from_secs(TEST_CURRENT_TIME);
+        let target_uuid = Uuid::new_v4();
+        let p = CryptoPolicy::minimum();
+
+        {
+            let mut idms_prox_write = idms.proxy_write(ct).await.unwrap();
+
+            let mut e = entry_init!(
+                (Attribute::Class, EntryClass::Object.to_value()),
+                (Attribute::Class, EntryClass::Account.to_value()),
```
Firstyear, line 4146: Needs `(Attribute::Class, EntryClass::PosixAccount.to_value()),`. Only posix accounts can use ldap/unix auth.

Firstyear:
> I know the current single test is bad (and will break) since it is running all tests in a single IDMS. Planning to split it out into multiple named tests for the time being hoping to at some point use test cases to build the matrix

Single test actually may be better here, because each time the test harness is stood up/torn down there are extended validations that take place which can be time consuming.

Firstyear:
Those tests get the big old stamp of approval from me. Great work!

ieugen:
It is my understanding that this will allow me to use the account password for LDAP auth without going through TOTP ?

That is great. I will be able to give my parents access to some of my services. 
I've read the discussion on Matrix and I am in a similar situation with @CEbbinghaus : Accessibility is more important than security in the case of my home setup. 

I would love to be able to disable mandatory Oauth TOTP via an account policy like it's done for LDAP.

CEbbinghaus:
Happy to help, This feature and the accessibility it provides is the only blocker for me running kanidm as my only idm provider. Right this moment I am still stuck with OpenLDAP until I get this merged.

CEbbinghaus:
Far from ready to be merged. First working version that passes the tests but realistically we want to refactor and create a single shared login path for both UNIX & LDAP.

Will do more work on this tomorrow. Need some sleep 💤

server/lib/src/idm/unix.rs
```diff
@@ -146,9 +146,22 @@ impl UnixUserAccount {
     pub(crate) fn try_from_entry_ro(
         value: &Entry<EntrySealed, EntryCommitted>,
         qs: &mut QueryServerReadTransaction,
+        allow_default_cred_fallback: bool,
```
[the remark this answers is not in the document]
CEbbinghaus: Did that (maybe a bit too much)

CEbbinghaus:
@Firstyear Refactored a ton, Left a bunch of comments throughout the code for you to answer. Personally I would like to refactor the rest of UnixUserAccount since it is currently only used for writing. Additionally it would be nice if the UnixGroup could be moved under the other Group definition and a top level method could be added that resolves all groups and filters out the unix ones into a second array so we only do 1 query and end up with both types of groups. Otherwise it may require refactoring of the macro (somehow). 

Most tests are passing except for this one which I just cannot figure out. Do you have any idea why login would still be succeeding despite the ldap bind being disabled?:
https://github.com/kanidm/kanidm/blob/master/server/lib/src/idm/ldap.rs#L916

At some point it may be worth considering typing Account to be `Account<Base>` & `Account<Unix>` so all of the unix methods can be moved into the base definition and we have some type safety around what properties of the account are avaliable or not.

Since this is quite a lot of big updates I suspect that this is pretty critical in terms of direction of this PR and how we want to tackle some of the problems. Please do let me know if you think this is heading in the wrong direction or what other changes you would like to see. 

Best Regards (time for 💤)

CEbbinghaus:
Had an itching to write more code so I preemptively moved the group and unified their loading behavior. I think it could be a bit more optimal if the common properties between Group & UnixGroup were resolved only once but as far as it goes this is miles better than what we had. 

The UnixGroup lookup looks a lot more complicated than the Group. Is there some complexity that I am missing? I decided to go with the KISS approach and just do the minimum. Tests are passing so if this implementation is wrong then that is a good sign more test are gonna help

server/lib/src/idm/server.rs
```diff
@@ -1275,127 +1276,133 @@ impl<'a> IdmServerAuthTransaction<'a> {
...
+        if !slock.is_valid() {
             security_info!("Account is softlocked.");
-            Ok(None)
+            //TODO: Did we want to return an error condition?
+            return Ok(None);
+        }
+
+        let Ok(password) = cred.password_ref() else {
+            // The credential should only ever be a password
+            unreachable!();
```
Firstyear, line 1349: I think we return an error here instead of unreachable!();

server/lib/src/idm/server.rs
```diff
@@ -1275,127 +1276,133 @@ impl<'a> IdmServerAuthTransaction<'a> {
...
-            // Account is slocked!
+        let mut slock = slock_ref.lock().await;
+
+        slock.apply_time_step(ct);
+
+        if !slock.is_valid() {
             security_info!("Account is softlocked.");
-            Ok(None)
+            //TODO: Did we want to return an error condition?
+            return Ok(None);
```
Firstyear, line 1344: ERr means "backend went bang" but Ok(None) indicates the auth failed and callers expect that.

server/lib/src/idm/server.rs
```diff
@@ -1275,127 +1276,133 @@ impl<'a> IdmServerAuthTransaction<'a> {
...
+        if !valid {
+            // Update it.
+            slock.record_failure(ct);
+
+            return Ok(None);
+        }
+
+        security_info!("Successfully authenticated with unix (or primary) password");
+        if password.requires_upgrade() {
+            //TODO: Do we really want to throw if pushing to the async queue fails?
```
Firstyear, line 1367: If the queue fails here it indicates a pretty catastrophic failure of the backend. It also will impact the ability for sessions to be created or other security data to be updated. So  Ithink that the error here is right, we should "fail noisy".

Firstyear (commented):
Just some minor comments and one change needed, but otherwise this looks fantastic. Thank you!

server/lib/src/idm/server.rs
CEbbinghaus: Right, So the queue is integral to the function of kanidm, I thought it was a async task list of "stuff we gotta do at some point" and therefore non critical

Firstyear: Yeah, it's for allowing concurrent things to write back "eventually", and potentially batched. A really good example here is auth session data. If a user logs in, we async write the session data to the DB in the bg. When the session token is issued there is a short grace window (5 mins) to account for replication delay etc, where the session token will still validate if the session isn't in the db yet. 

But after that if the session isn't in the db, the session is canned. 

So here if the session is sent for async write back and it fails, then any session this server issues will fail to work. That's a pretty big failure for us.

Similar, the async pw upgrade writeback is to "raise" the security level/quality of a password. If that fails, that means we could be holding on to insecure pw material when we shouldn't.

Anyway, better to error here. I think it's better to have "1 narrow path to success" where everything has to work, and to make it fragile to get there. And then we error at the slightest chance of something being wrong, because we don't want unintended consequences.

CEbbinghaus: Probably a smart choice given that this does authentication. I am still in the mindset of "make the user experience as pleasant as possible". 

pw upgrade queue could be handled gracefully since it would just requeue the next time that user logs in. But if the same queue is also responsible for writing sessions into the database, then its failure to write is a big enough deal to warrant stopping an execution flow.

server/lib/src/idm/server.rs
```diff
@@ -1340,13 +1338,13 @@ impl<'a> IdmServerAuthTransaction<'a> {
...
         if !slock.is_valid() {
             security_info!("Account is softlocked.");
-            //TODO: Did we want to return an error condition?
             return Ok(None);
         }
 
         let Ok(password) = cred.password_ref() else {
             // The credential should only ever be a password
-            unreachable!();
+            error!("user unix or primary credential is not a password");
```
yaleman, line 1346:
```suggestion
            error!("User's UNIX or primary credential is not a password, can't authenticate!");
```

Side note, perhaps for @Firstyear: I wonder if it's faster to check this first or the softlock with its read/write/change steps, given they can't auth at all...

server/lib/src/idm/account.rs
```diff
@@ -794,6 +809,34 @@ impl Account {
...
+
+    pub(crate) fn to_unixusertoken(&self, ct: Duration) -> Result<UnixUserToken, OperationError> {
+        let (gidnumber, shell, sshkeys, groups) = match &self.unix_extn {
+            Some(ue) => {
+                let sshkeys: Vec<String> = ue.sshkeys.keys().cloned().collect();
+                (ue.gidnumber, ue.shell.clone(), sshkeys, ue.groups.clone())
+            }
+            None => {
+                return Err(OperationError::InvalidAccountState(
+                    "Missing class: posixaccount".to_string(),
```
yaleman, line 821: Minor nit around string-typo-avoidance, you'll need to check `EntryClass` is imported.

```suggestion
                    format!("Missing class: {}", EntryClass::PosixAccount),
```

server/lib/src/idm/server.rs
```diff
@@ -1275,128 +1275,131 @@ impl<'a> IdmServerAuthTransaction<'a> {
...
-        let maybe_slock = if let Some(s) = maybe_slock_ref.as_ref() {
-            Some(s.lock().await)
-        } else {
-            None
+        let (cred, cred_id, cred_slock_policy) = match cred {
+            None => {
+                if acp.allow_primary_cred_fallback() == Some(true) {
+                    security_info!("Account does not have a posix or primary password configured.");
+                } else {
+                    security_info!("Account does not have a posix password configured.");
```
yaleman, line 1314:
```suggestion
                    security_info!("Account does not have a POSIX or primary password configured.");
                } else {
                    security_info!("Account does not have a POSIX password configured.");
```

server/lib/src/idm/server.rs
```diff
@@ -1422,85 +1425,58 @@ impl<'a> IdmServerAuthTransaction<'a> {
...
+                    session_id: uat.session_id,
+                    spn,
+                    effective_session: LdapSession::UserAuthToken(uat),
+                }))
+            }
+            Token::ApiToken(apit, entry) => {
+                let spn = entry
+                    .get_ava_single_proto_string(Attribute::Spn)
+                    .ok_or_else(|| {
+                        OperationError::InvalidAccountState("Missing attribute: spn".to_string())
```
yaleman, line 1472: Similar nit with strings, will need to ensure import's there.

```suggestion
                        OperationError::InvalidAccountState(format!("Missing attribute: {}", Attribute::Spn))
```

server/lib/src/idm/server.rs
```diff
@@ -1786,23 +1789,28 @@ impl<'a> IdmServerProxyWriteTransaction<'a> {
...
             .map_err(|e| {
                 admin_error!("Failed to start set unix account password {:?}", e);
                 e
             })?;
-        // Ask if tis all good - this step checks pwpolicy and such
+
+        // Account is not a unix account
+        if account.unix_extn().is_none() {
+            return Err(OperationError::InvalidAccountState(
+                "Missing class: posixaccount".to_string(),
```
yaleman, line 1802: See previous comments about using `EntryClass::PosixAccount` instead

server/lib/src/idm/server.rs
```diff
@@ -1969,24 +1977,36 @@ impl<'a> IdmServerProxyWriteTransaction<'a> {
...
             .map_err(|e| {
                 admin_error!("Failed to start unix pw upgrade -> {:?}", e);
                 e
             })?;
 
+        let cred = match account.unix_extn() {
+            Some(ue) => ue.ucred(),
+            None => {
+                return Err(OperationError::InvalidAccountState(
+                    "Missing class: posixaccount".to_string(),
```
yaleman, line 1991: See previous comments about using `EntryClass::PosixAccount` instead

yaleman (changes requested):
This is great work 🥳

server/lib/src/idm/server.rs
Firstyear: Faster yes, but not correct. In this case cred.password_ref() should in 99.99999% of cases always return a password. We just need to handle that fractional and unlikely case it doesn't since that's still a possibly until I undo some of the credential structure tech-debt. 

So I think in this case, we leave the logic as is because it makes more sense from the view of "okay, get the credential, are the softlocked, now lets check the password" and the unlikely case is ... well, unlikely.

Firstyear:
Agreed, once these little nits are done, I'm happy to merge :)

Firstyear (dismissed):
Thank you!

CEbbinghaus:
image

Firstyear (dismissed):

CEbbinghaus:
Okay, This is finally going to actually merge. I got it compiling locally so this time it shouldn't fail.
image

Firstyear (dismissed):

CEbbinghaus:
Found a bit that can be improved, and implemented Yaleman's suggestion to move up the password to be earlier and stop us from taking a slock on the credential if its not a password.

yaleman (approved):

yaleman (approved):

PR #3079, opened 2024-10-01
Scim add EntryReference 

ToxicMushroom:
# Change summary
Replaces the uuid references with a struct containing the uuid and name.

Checklist

So far added a test to check the goal behavior


Firstyear (commented):
As we talked about, I think we should return a different type than Option<ScimValue> which represents that more resolution is needed.

ToxicMushroom:
> As we talked about, I think we should return a different type than Option which represents that more resolution is needed.

Well the commit said not to look at it :)

proto/src/scim_v1/server.rs
```diff
@@ -179,6 +179,182 @@ pub struct ScimOAuth2ClaimMap {
...
+    }
+}
+
+impl From<Vec<ScimKeyInternal>> for ScimResolveStatus {
+    fn from(set: Vec<ScimKeyInternal>) -> Self {
+        Self::Resolved(set.into())
+    }
+}
+
+impl ScimResolveStatus {
```
Firstyear, line 340: block needs to be #![cfg(feature = "test")]

proto/src/scim_v1/server.rs
```diff
@@ -179,6 +179,182 @@ pub struct ScimOAuth2ClaimMap {
...
+    }
+}
+
+impl From<Vec<ScimApiToken>> for ScimResolveStatus {
+    fn from(set: Vec<ScimApiToken>) -> Self {
+        Self::Resolved(set.into())
+    }
+}
+
+impl From<Vec<ScimIntentToken>> for ScimResolveStatus {
```
Firstyear, line 316: Can't this be

```
impl From<T> for ScimResolveStatus
where T: Into<ScimValueKanidm> {
    fn from(v: T) -> Self {
        Self::Resolved(v.into())
    }
}
```

proto/src/scim_v1/server.rs
Firstyear: Or alternately, don't define it here - define it in ValueSet behind the cfg(test) as a private impl. We don't want panics in pub interfaces.

Firstyear (changes requested):

proto/src/scim_v1/server.rs
ToxicMushroom: yep ty

proto/src/scim_v1/server.rs
```diff
@@ -180,6 +180,21 @@ pub struct ScimOAuth2ClaimMap {
...
+#[derive(Serialize, Debug, Clone, PartialEq, Eq, ToSchema)]
+#[serde(rename_all = "camelCase")]
+pub struct ScimReference {
+    pub uuid: Uuid,
+    pub value: String,
+}
+
+#[serde_as]
+#[derive(Serialize, Debug, Clone, PartialEq, Eq)]
+pub enum ScimValueIntermediate {
```
Firstyear, line 193: Why is this in proto? We never release it publicly?

server/lib/src/entry.rs
```diff
@@ -2353,6 +2372,37 @@ impl Entry<EntryReduced, EntryCommitted> {
     }
 }
 
+fn resolve_interim(
```
Firstyear, line 2375: resolve_scim_interim

server/lib/src/entry.rs
```diff
@@ -29,12 +29,25 @@ pub use std::collections::BTreeSet as Set;
 use std::collections::{BTreeMap as Map, BTreeMap, BTreeSet};
 use std::sync::Arc;
 
+use crate::be::dbentry::{DbEntry, DbEntryVers};
+use crate::be::dbvalue::DbValueSetV2;
+use crate::be::{IdxKey, IdxSlope};
+use crate::credential::apppwd::ApplicationPassword;
+use crate::credential::Credential;
```
Firstyear, line 36: Was this cargo fmt re-arranging things? Or your editor?

server/lib/src/entry.rs
```diff
@@ -2230,15 +2230,34 @@ impl Entry<EntryReduced, EntryCommitted> {
...
+    ) -> Result<ScimEntryKanidm, OperationError> {
+        let result: Result<BTreeMap<Attribute, ScimValueKanidm>, OperationError> = self
             .attrs
             .iter()
             // We want to skip some attributes as they are already in the header.
             .filter(|(k, _vs)| **k != Attribute::Uuid)
-            .filter_map(|(k, vs)| vs.to_scim_value().map(|scim_value| (k.clone(), scim_value)))
+            .filter_map(|(k, vs)| {
+                let opt_resolve_status = vs.to_scim_value();
+                let res_opt_scim_value = match opt_resolve_status {
```
Firstyear, line 2244: look at 'transpose()' on Result.

server/lib/src/entry.rs
```diff
@@ -2353,6 +2372,37 @@ impl Entry<EntryReduced, EntryCommitted> {
...
+) -> Result<Option<ScimValueKanidm>, OperationError> {
+    match scim_value_intermediate {
+        ScimValueIntermediate::Refer(uuid) => {
+            if let Some(option) = read_txn.uuid_to_spn(uuid)? {
+                Ok(Some(ScimValueKanidm::EntryReference(ScimReference {
+                    uuid,
+                    value: option.to_proto_string_clone(),
+                })))
+            } else {
+                // TODO: didn't have spn, fallback to uuid.to_string ?
```
Firstyear, line 2387: Doesn't uuid_to_spn already do this?

Firstyear (changes requested):

server/lib/src/entry.rs
ToxicMushroom: Doesn't appear like that according to the test
https://github.com/ToxicMushroom/kanidm/blob/957aaabe5e6cb6a950ee9f8aef63d368e7577241/server/lib/src/server/mod.rs#L2362
      VV
 ```rust
      
        // Name doesn't exist
        let r1 = server_txn.uuid_to_spn(uuid!("bae3f507-e6c3-44ba-ad01-f8ff1083534a"));
        // There is nothing.
        assert_eq!(r1, Ok(None));
   ```

server/lib/src/entry.rs
ToxicMushroom: Should be cargo fmt's doings.
with toolchain: stable-x86_64-unknown-linux-gnu (default)
rustc 1.81.0

Firstyear (approved):
Thank you!

PR #3081, opened 2024-10-02
Add support for group extension

Firstyear:
# Change summary

- Support group extension in unixd

Checklist


unix_integration/resolver/src/resolver.rs
```diff
@@ -711,8 +722,26 @@ impl Resolver {
     }
 
     async fn get_nssgroup(&self, grp_id: Id) -> Result<Option<NssGroup>, ()> {
-        if let Some(nss_group) = self.system_provider.get_nssgroup(&grp_id).await {
+        if let Some(mut nss_group) = self.system_provider.get_nssgroup(&grp_id).await {
             debug!("system provider satisfied request");
+
+            for client in self.clients.iter() {
+                if let Some(extend_group_id) = client.has_map_group(&nss_group.name) {
+                    let token = self.get_grouptoken(extend_group_id.clone()).await?;
```
CEbbinghaus, line 730: Why does this one not use the cached token?

CEbbinghaus (approved):
Looks good. Just had a question and nit

unix_integration/resolver/src/resolver.rs
Firstyear: get_nssgroups is enumeration of everything on the system, so it smashes the cache. It's generally only used during "getent group" so we only want to shows what's locally available.

Where as get_nssgroup (singular) is for getting the defaults of a singular named group, so we want to call the backend to check if it's up-to-date or not.

CEbbinghaus (approved):
LGTM

examples/unixd
```diff
@@ -140,3 +140,14 @@ version = '2'
 
 pam_allowed_login_groups = ["posix_group"]
 
+# Allow extension (mapping) of a local system groups members with members from a
+# kanidm provided group. An example of this is that the local group
+# `libvirt` can have it's membership extended with the members from
+# `virt-admins`. This section can be repeated many times.
```
yaleman, line 146:
```suggestion
# Kanidm-provided group. An example of this is that the local group
# `libvirt` can have its membership extended with the members from
# `virt-admins`. This section can be repeated many times.
```

Given the "can be repeated many times" can we please include more than one in the example? I know I've tripped over things like this in the past where it's not clear on the exact syntax of *how* to repeat them, or weird edge cases etc. (And ensure it's tested so we can point at something and go "yes this matches what we document")

unix_integration/resolver/src/unix_config.rs
```diff
@@ -69,7 +69,7 @@ struct KanidmConfigV2 {
     conn_timeout: Option<u64>,
     request_timeout: Option<u64>,
     pam_allowed_login_groups: Option<Vec<String>>,
-    extend: Vec<GroupMap>,
+    map_group: Vec<GroupMap>,
```
yaleman, line 72: Why isn't this called group_maps or something? 😄 Makes my brain itch...

yaleman (changes requested):

unix_integration/resolver/src/unix_config.rs
Firstyear: Because of what it becomes in the config - it's a table of 

```
[[kanidm.map_group]]
local: "name"
with: "kanidm group"
```

So it kind of becomes english-like - kanidm maps group "name" with "remote".

That's why I chose it

examples/unixd
Firstyear: Yep can do

CEbbinghaus:
Did we want to consider the possibility of mapping 1 local group to multiple remote groups.
e.g `docker` maps to `docker-users@idm.example.org` & `docker-admins@idm.example.org`.

Its already possible to map multiple local groups to the same remote group but having the ability to do a many to many mapping could be helpful.

for the implementation it could be as simple as:
```
[[kanidm.map_group]]
local: "name"
with: "kanidm group1"

[[kanidm.map_group]]
local: "name"
with: "kanidm group2"
```

and during the resolution we just map to a `Map<string, Vec<Id>>`. 

Maybe that is something to consider for the future or just wait till someone asks for it

Firstyear:
> Did we want to consider the possibility of mapping 1 local group to multiple remote groups. e.g `docker` maps to `docker-users@idm.example.org` & `docker-admins@idm.example.org`.
> 
> Its already possible to map multiple local groups to the same remote group but having the ability to do a many to many mapping could be helpful.
> 
> for the implementation it could be as simple as:
> 
> ```
> [[kanidm.map_group]]
> local: "name"
> with: "kanidm group1"
> 
> [[kanidm.map_group]]
> local: "name"
> with: "kanidm group2"
> ```
> 
> and during the resolution we just map to a `Map<string, Vec<Id>>`.
> 
> Maybe that is something to consider for the future or just wait till someone asks for it

Since Kanidm can do group nesting, you can just have a the nesting done in the database rather than client side here.

yaleman (approved):

CEbbinghaus (approved):
LGTM 👍

PR #3083, opened 2024-10-02
Improve deb packaging, add aarch64

### LINKED ISSUE #3078: Debian ARM packaging
### I did this
- Bought a shiny arm64 computer running Debian bookworm (pi5)
 
### I expected the following
- Install kanidm-unixd on it
 
### This happened instead
- Alas we do not package debs for arm64.
- While a cargo install might succeed, even the pi5 is still memory constrained in lower end models, and then one misses out on systemd units, pam automation etc. 

### Kanidm version details
1.3.3 & HEAD

### Any other comments
- Logging this for posterity and discoverability, I'm already in the codebase with a machete trying to make this happen.
- It'll likely include as much refactoring of the crossbuild & deb toolchain as I can stomach.
- Draft PRs incoming once I get to a point where I _think_ it works.

jinnatar:
# Change summary
- The end goal is to build aarch64 debs. Sadly there be many dragons along the way.
- As per discussions, this PR specifically removes most of the debian packaging from the main repo, while it is being stood up as a repo of it's own. However, to maintain human ad-hoc deb building, the packaging repo is added in as a completely optional submodule.

- The remainder of these changes lives on as https://github.com/kanidm/kanidm_ppa_automation/ which is ready to go, pending this PR.
- The `kanidm/kanidm_ppa` repo will stop receiving newer nightly builds when this one is merged, but old ones remain still available until the new automation kicks it all out.

## WIP tasklist, roughly in priority order:
- [x] Purge this repo of most debian packaging artifacts. 
- [x] cargo-deb metadata for unixd and it's dependencies.
- [x] cargo-deb metadata for `cli`
- [x] Rewiggle the submodule once https://github.com/kanidm/kanidm_ppa_automation/pull/1 is merged.
- [x] Switch @yaleman to be the Maintainer instead of inheriting @Firstyear.
- [x] Test every permutation, at least for a clean `dpkg -i` if nothing else.

There's also a couple of further decisions to resolve, once I've marked this ready for review:
- [x] Consider if we want unixd to pre-create something at `/etc/kanidm` to make it friendlier. Say a `unixd.example`.
- [x] Decide if we're happy with `kanidm-pam` & `kanidm-nss` as package names, instead of say `libkanidm-{pam,nss}`. Now's the time to change that before they're in the wild.
- [x] Consider backporting this to 1.3.3 somehow so we get a stable deb in sooner than 1.4.0.

Fixes #3078

Checklist


unix_integration/resolver/Cargo.toml
```diff
@@ -110,3 +110,21 @@ kanidmd_testkit = { workspace = true }
 clap = { workspace = true, features = ["derive"] }
 clap_complete = { workspace = true }
 kanidm_build_profiles = { workspace = true }
+
+## Debian packaging
```
yaleman, line 114: Is this what powers cargo-deb ? Looks handy if it works 😄

jinnatar: Yup! This is the appeal, much less bespoke pipeline and instead relying one someone else.

Firstyear:
Wow, this is really good stuff! Keen to see it finished :)

jinnatar:
Updated the missing WIP in the top level comment. Got pretty far actually with the Aptly action but then realized I'd painted myself in a corner: you can't matrix create repos and then merge them since the signed package lists need to cover the entire repo and can't be merged. I think it's still doable, just need to inject the matrix logic within the action so it can create multiple repos, distros and components in one action. We could actually have separate `stable` and `dev` tracks in the same repo!

Where I got so far with the action before realizing the external matrix won't work:
https://github.com/jinnatar/actions-aptly-repo-demo/actions/runs/11169264652/job/31049567345

jinnatar:
Most of the discussed refactoring into a separate repo is now done. The portions of this PR that I've deleted here live on in https://github.com/jinnatar/kanidm_deb_packaging

A few steps forward, two backwards or so, so still plenty to iron out in crossbuild due to the ever blasted openssl.

jinnatar:
The companion PR is now also under review at: https://github.com/kanidm/kanidm_ppa_automation/pull/1

.. Still need a few tweaks on this side as well before this is ready for review, but it's not far now.

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "platform/debian/kanidm_ppa_automation"]
```
yaleman, line 1: git submodules are a bit cursed, is this just here for testing purposes?

book/src/packaging/debian_ubuntu_packaging.md
```diff
@@ -2,40 +2,83 @@
...
+   platform/debian/kanidm_ppa_automation/scripts/crossbuild.sh
+   # Launch a target:
+   platform/debian/kanidm_ppa_automation/scripts/crossbuild.sh debian-12-aarch64-unknown-linux-gnu
+   # You can also specify multiple targets within the same distribution:
+   platform/debian/kanidm_ppa_automation/scripts/crossbuild.sh debian-12-{aarch64,x86_64}-unknown-linux-gnu
+   ```
+1. Go get a drink of your choice while the build completes.
+1. Create a sacrificial deb builder container to avoid changing your own system:
+   ```shell
+   docker run --rm -it -e CI=true -v $PWD:/src --workdir /src rust:bookworm
```
yaleman, line 34: I hate `-v` mounts more than anything else in Docker 😱 

```suggestion
   docker run --rm -it -e CI=true \
       --mount "type=bind,src=$PWD,target=/src" \
       --workdir /src \
       rust:bookworm
```

unix_integration/resolver/Cargo.toml
```diff
@@ -110,3 +110,31 @@ kanidmd_testkit = { workspace = true }
...
+name = "kanidm-unixd"
+maintainer = "James Hodgkinson <james@terminaloutcomes.com>"
+# Can't use $auto depends because the name of libssl3 varies by distro and version
+depends = ["libc6", "tpm-udev", "libssl3 | libssl3t64"]
+section = "network"
+priority = "optional"
+recommends = [
+	"kanidm-pam",
+	"kanidm-nss"
+]
```
yaleman, line 126: Can we make these part of `requires`? It's not much use having `kanidm-unixd` without the other integrations IMO, and it makes the install process more usable for humans.

scripts/install_ubuntu_dependencies.sh
```diff
@@ -68,6 +79,14 @@ if [ $ERROR -eq 0 ] && [ -z "$(which wasm-bindgen)" ]; then
     echo "You don't have wasm-bindgen installed! Installing it now..."
     cargo install -f wasm-bindgen-cli
 fi
+if [ $ERROR -eq 0 ] && [ -z "$(which cross)" ]; then
+    echo "You don't have cargo-deb installed! Installing it now..."
```
yaleman, line 83:
```suggestion
    echo "You don't have cross installed! Installing it now..."
```

platform/debian/README.md
```diff
@@ -0,0 +1,2 @@
+Debian packaging has moved to it's own repo, see https://github.com/kanidm/kanidm_ppa_automation.
```
yaleman, line 1:
```suggestion
Debian packaging has moved to its own repo, see https://github.com/kanidm/kanidm_ppa_automation.
```

.gitmodules
yaleman: Ignore me, found ref later. Ew, submodules.. but I get it ^_^

yaleman (changes requested):
minor changes requested but 💟 love this 💟

yaleman:
>  Consider if we want unixd to pre-create something at /etc/kanidm to make it friendlier. Say a unixd.example.

If we're creating anything, it should be in the "right" spot and just have every value commented out - I don't see a reason to create a file in the "wrong" spot unless it's in /usr/share/doc or the appropriate "docs" dir.

> Decide if we're happy with kanidm-pam & kanidm-nss as package names, instead of say libkanidm-{pam,nss}. Now's the time to change that before they're in the wild.

Is there a standard? (Also... sad that we can't seem to just have kanidm-unixd with the necessary bits? 😞 )

> Consider backporting this to 1.3.3 somehow so we get a stable deb in sooner than 1.4.0.

Maaaayybe?

jinnatar:
> > Consider if we want unixd to pre-create something at /etc/kanidm to make it friendlier. Say a unixd.example.
> 
> If we're creating anything, it should be in the "right" spot and just have every value commented out - I don't see a reason to create a file in the "wrong" spot unless it's in /usr/share/doc or the appropriate "docs" dir.
> 
Worth a try, just worried unixd will throw bad things with that. It's already a bit error happy since it gets pam enabled but isn't running.

> > Decide if we're happy with kanidm-pam & kanidm-nss as package names, instead of say libkanidm-{pam,nss}. Now's the time to change that before they're in the wild.
> 
> Is there a standard? (Also... sad that we can't seem to just have kanidm-unixd with the necessary bits? 😞 )
> 

The two reasons for separate packaging are that they're separate packages in the codebase, and multiarch sez sdynlib containing packages shall not deliver binaries because there's no multiarch for binaries.
As far as conventions go, sure enough I found one:
> The package name should match the shared library SONAME. It can be deduced with
sed s/\.so\./-/ and removing ?-? later when removing it will not result in
consecutive numbers.

Unfortunately that makes zero sense for `pam_kanidm.so` but nss_kanidm would be `libnss_kanidm` to match the so name. So then for consistency, `libpam_kanidm` might make sense. :woman_shrugging: 
> > Consider backporting this to 1.3.3 somehow so we get a stable deb in sooner than 1.4.0.
> 
> Maaaayybe?

I imagine this would be cherry-picking the eventual squash merge in a v1.3.3 branch  and then updating the v1.3.3 tag to it.

jinnatar:
I think this is ready for a real review. I've tested the full chain (https://github.com/jinnatar/kanidm_ppa_automation/actions/runs/11314622787) and qemu tested the stable versions.

- Once everyone is happy with this one, the next to go is: https://github.com/kanidm/kanidm_ppa_automation/pull/4, it just needs final tweaks and tests once this PR is live first.
- After that point would be the backport for the 1.3.0 branch. I've been testing it and a cherry-pick of this PR on top of that branch should be enough, no need to shift the tag or release 1.3.4.
- And finally, if all goes well with all of the above, there's one more PR to do in the automation repo to enable keeping state.

book/src/packaging/debian_ubuntu_packaging.md
```diff
@@ -2,40 +2,86 @@
...
+1. Launch your desired crossbuild target. Do note the script assumes you use rustup!
+   ```shell
+   # See valid targets:
+   platform/debian/kanidm_ppa_automation/scripts/crossbuild.sh
+   # Launch a target:
+   platform/debian/kanidm_ppa_automation/scripts/crossbuild.sh debian-12-aarch64-unknown-linux-gnu
+   # You can also specify multiple targets within the same distribution:
+   platform/debian/kanidm_ppa_automation/scripts/crossbuild.sh debian-12-{aarch64,x86_64}-unknown-linux-gnu
+   ```
+1. Go get a drink of your choice while the build completes.
```
Firstyear, line 31: This is a very important step

examples/kanidm-safe-default
```diff
@@ -0,0 +1,7 @@
+## Kanidm minimal Service Configuration - /etc/kanidm/config
+# For a full example and documentation, see /usr/share/kanidm/kanidm
+# or `example/kanidm` in the source repository.
+
+# Replace this with your kanidmd URI and uncomment the line
+#uri = "https://idm.example.com"
+verify_ca = true
```
Firstyear, line 7: verify_ca defaults to true, so you can remove this line if you want.

Firstyear (approved):

Firstyear:
This is really impressive, fantastic work!

yaleman (approved):

PR #3086, opened 2024-10-03
Feat: Adding JSON/Authorized_keys api for SSH Publickeys

### LINKED ISSUE #2983: REST API: SSH key tags not gettable

CEbbinghaus:
# Change summary

Adds a new http endpoint (or modifies an existing one in a backwards compatible way) to retrieve the ssh keys in either a JSON kv mapping
```json
{
  "name1": "...",
  "name2": "..."
}
```
or in the `authorized_keys` format.

Fixes #2983

Checklist

### DIFF
--- a/libs/client/src/lib.rs
+++ b/libs/client/src/lib.rs
@@ -27,6 +27,7 @@ use std::time::Duration;
 
 use compact_jwt::Jwk;
 
+use http::HeaderMap;
 use kanidm_proto::constants::uri::V1_AUTH_VALID;
 use kanidm_proto::constants::{
     ATTR_DOMAIN_DISPLAY_NAME, ATTR_DOMAIN_LDAP_BASEDN, ATTR_DOMAIN_SSID, ATTR_ENTRY_MANAGED_BY,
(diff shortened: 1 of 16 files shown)


CEbbinghaus:
Closed in Favour of Firstyear taking over

Firstyear:
I think this isn't the direction I wanted this to go, I wanted to add the person get via scim as that's where we're going. So I think we close and re-work this with that in mind, I'm happy to do it.

PR #3089, opened 2024-10-04
Fix Increment Replication Post Upgrade

Firstyear:
During an upgrade, we often add new entries or touch existing ones. These entries, notably access control profiles often contain references.

Replication currently blocks until all nodes are on the same domain level and version. This means that all nodes will *internally* perform their migrations and then resume replication.

This leads to all nodes creating all new entries specified by a migration individually, and then a conflict occuring such that the first node to migrate will have it's copy of the system entries persisted by all nodes.

Ordinarily, this is not a problem as the retained system node from the earliest server will work on all nodes. The issue is that for security reasons we need to remove reference uuids of nodes that are marked into the conflict state. This is to prevent a security edge case where:

* a user on node A is added
* a user on node B is added
* user B is added to group X

Due to our conflict rules, the user from node A would be retained as it has the earliest creation CID, but now the user from node A was added to group X when they should not have been.

To prevent this, during replication conflict uuids automatically have all their references purged. While this would be annoying to admins, it prevents the security leak of a group membership.

This protection however is what caused the issue with incremental replication failing on node upgrades. All nodes would add the new system groups and access controls *but* as the migrated entries now conflict, they would then have their references removed. This subsequently triggered a schema violation, as access control profiles that relied on entry-managed-by would now fail to validate as a must attribute in schema was now missing.

To resolve this, we now exempt system uuids from this conflict handling. There are protections in place to prevent users creating uuids in this system range, and the existing conflict resolution process is sufficient to ensure that the earliest migrated node's content will persist as expected.

# Change summary

-

Fixes #

Checklist

### DIFF
--- a/server/lib/src/repl/consumer.rs
+++ b/server/lib/src/repl/consumer.rs
@@ -89,8 +89,32 @@ impl<'a> QueryServerWriteTransaction<'a> {
         // server for some conflicts, but we still need to know the UUID is IN the conflict
         // state for plugins. We also need to do this here before the conflict_update
         // set is consumed by later steps.
-        let mut conflict_uuids: BTreeSet<_> =
-            conflict_update.iter().map(|(_, e)| e.get_uuid()).collect();
+        //
+        // ⚠️  When we upgrade between two nodes, migrations will often create *new* system
+        // entries on both nodes. Until both nodes upgrade they can't replicate. This creates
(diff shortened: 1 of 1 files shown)


Firstyear:
@jinnatar Thanks for finding this!

yaleman (approved):

PR #3091, opened 2024-10-05
Cache buster buster

### LINKED ISSUE #3010: Javascript errors after server-side update blocking login. Fixed after cache invalidating
### I did this

I upgraded from 1.2.3 to 1.3.2 following the upgrade flow described [here](https://kanidm.github.io/kanidm/stable/server_updates.html)

### I expected the following

For users that have previously visited an application behind version 1.2.3 to be able to have their session stored, or at least be able to log back in again after the upgrade.

### This happened instead

Users are stuck on the "Kanidm is loading, please wait..." screen. Their Javascript console shows the following error:

```
wasmloader_login_flows.js:6 Uncaught (in promise) LinkError: WebAssembly.instantiate(): Import #164 module="wbg" function="__wbindgen_closure_wrapper1416" error: function import requires a callable
await (async)  
(anonymous) @ wasmloader_login_flows.js:7
```

It appears the browser is using outdated Javascript files; network traffic shows that they're served a 304 (Request not modified). Disabling the local browser cache seems to resolve the issue, and shows them a login screen.

### Kanidm version details

- Output of kanidm(d) version: 1.3.2
- Are you running it in a container? If so, which image/tag?: yes, :latest (version 1.3.2)
- Operating System / Version (On Unix please post the output of uname -a): 6.10.3-arch1-2 on the docker host

### Any other comments

I previously created #3009 where I wrongly diagnosed this to be related to cookies.

yaleman:
# Change summary

- adding cache-busting query-strings to the end of everything I could find to avoid issues with caching on updates
- fixing up a deprecation in the json schema validation while I'm there

Fixes #3010 (javascript was already mostly done)

Checklist


server/core/Cargo.toml
```diff
@@ -94,4 +94,5 @@ walkdir = { workspace = true }
 tempfile = { workspace = true }
 
 [build-dependencies]
+gix = { workspace = true }
```
Firstyear, line 97: There is an env var with this info that's derived from profiles.

server/core/build.rs
```diff
@@ -1,4 +1,40 @@
...
+        .arg("--message-format=plain")
+        .output()
+        .unwrap()
+        .stdout;
+    let cargo_path = Path::new(std::str::from_utf8(&output).unwrap().trim());
+    cargo_path.parent().unwrap().to_path_buf()
+}
+
+/// Determine the git rev, we do this here so it's only actually run and checked once at build time.
+fn determine_git_rev() -> Option<String> {
```
Firstyear, line 17: Not everyone will build out of git, so this will break things. We sohuld use the kanidm version instead here as it contains a commit hash on devel builds.

Firstyear: https://github.com/kanidm/kanidm/blob/master/libs/profiles/src/lib.rs#L86

Firstyear: I'd say rather than a direct commit rev, we just sha2 the current KANIDM_PKG_VERSION instead and use KANIDM_PKG_HASH. Potentially we should probably truncate that hash so that it's not the full 16 bytes or whatever of hex.

server/core/src/https/ui_html.html
```diff
@@ -5,35 +5,41 @@
...
         <meta name="viewport" content="width=device-width" />
-        <title>{}</title>
-
-        <link rel="icon" href="/pkg/img/favicon.png" />
+        <title>{display_name}</title>
         <link rel="manifest" href="/manifest.webmanifest" />
-        <link rel="apple-touch-icon" href="/pkg/img/logo-256.png" />
+
+        <link rel="icon"
+            href="/pkg/img/favicon.png?v={cache_buster_key}" />
```
Firstyear, line 12: Doesn't this cause the client to not cache *anything* at all? 

A cleaner option could just be to have /pkg/{pkg version hash}/img/.... that way the resource dir is stable, and we don't need cache busting at all, and when we change versions, the resource path changes causing the client to refetch the item. Also pretty easy to achieve with route matching.

Firstyear (changes requested):

server/core/src/https/ui_html.html
yaleman: No it'll cache based on the URL, pretty common way of doing it that we use elsewhere already for JavaScript - but we use the file hash for those since we are already calculating it

yaleman: It's the "don't have to make magic routers to update versions" equivalent

server/core/build.rs
yaleman: > Not everyone will build out of git, so this will break things. We sohuld use the kanidm version instead here as it contains a commit hash on devel builds.

It wouldn't break anything - it returns `None` if the git repo's not found and falls back to CARGO_PKG_VERSION in the function that grabs it. 

I'd forgotten about KANIDM_PKG_VERSION, I'll use the hash of that instead (I'd use it as-is but it has a space in it, which can't be used :smile:)

scripts/setup_dev_environment.sh
```diff
@@ -7,6 +7,8 @@
 # - set up a test oauth2 rp (https://kanidm.com)
 # - prompt to reset testuser's creds online
 
+set -e
```
ToxicMushroom, line 10: this seems unrelated, and idk what it'd do

ToxicMushroom (commented):
I'm not a fan of changing the indentation of the askama templating in the way you did.
e.g. having if, for and match bodies be on the same indentation level as themselves

scripts/setup_dev_environment.sh
yaleman: It's making the script fail on any errors, which is important.

Firstyear (approved):

PR #3093, opened 2024-10-05
Add nss testframework and fallback when daemon offline

Firstyear:
# Change summary

- Add nss fallback so that in the even the daemon is offline, accounts and groups can still resolve for system operation
- Add test framework to start to improve nss module testing.

Checklist


Firstyear:
Next  I'll add equivalents for pam.

Firstyear:
Refactors pam and nss so that we can test the .so behaviours more effectively. Test the fallback mode to ensure that we can rely on if it the daemon is offline for auth.

yaleman (approved):

Issue #1523, opened 2023-04-06
OAuth 2.0 Device Authorization Grant (RFC 8628)


sjoerdsimons:
**Is your feature request related to a problem? Please describe.**

Feature request

**Describe the solution you'd like**

 OAuth 2.0 Device Authorization Grant provide a nice way for e.g. command line apps and remote clients to trigger authorization against an oidc/oath service without having to require a local webbrowser.

yaleman:
I support this, it'd allow for things like passkey-backed SSH/CLI usage without having to build it into the client applications.

Firstyear:
This has certainly been on my radar for some time, I just hadn't made it a priority yet. I'm currently focused on refresh tokens and then some idm sync stuff, maybe once that's a bit further I'll come back to this.

Firstyear:
https://www.rfc-editor.org/rfc/rfc8628

Firstyear:
@yaleman Can you explain more about the ssh/cli use case for me? I'm curious about it.

yaleman:
> @yaleman Can you explain more about the ssh/cli use case for me? I'm curious about it.

Duo's module would be the one to look at for a comparison - https://duo.com/docs/duounix#test-your-setup

It prompts for auth as one of the factors and in our case we'd be starting the device auth flow and prompting the user to open a url to complete the auth.

Firstyear:
So it prompts for unix password I guess then does the device flow for the second/extra factors? 

I think given how device auth works, we'll have to "preauth" the flow somehow in this case with the unix password. There are some challenges to make device auth flows work nicely in a distributed system so I need to think more about this. But I think it certainly would have uses.

yaleman:
yeah, there's a bunch of ways you could implement it, for sure - and PAM would allow you to entirely replace the auth flow with our module if you really wanted to. IIRC the default they recommend is still a UNIX password with them for second factor, but was designed before "passwordless" flows were common.

Firstyear:
So I'm guessing the value here is for non-kanidm ssh clients/auth? Because in Kanidm we'll be able to do passwordless flows with attested ssh keys in the future so it's not super important on that front.

yaleman:
At some point you're going to have to authenticate, for things like sudo I guess?

yaleman:
> in Kanidm we'll be able to do passwordless flows with attested ssh keys in the future so it's not super important on that front.

I might need some info about what *this* looks like before I can really respond though 😄

[...]

PR #3098, opened 2024-10-11
OAuth2 Device flow

yaleman:
Built a tiny test device code client, and implementing it on the server side.

I might have also gone on a side quest of replacing strings with consts and building some helper functions for the OAuth2RS struct.

Screenshot 2024-10-12 at 09 03 37

TODO: 
- [ ] store the sessions somewhere, I'm too dumb to build an entry and need help with 😄 
- [ ] second half of the flow once the user pops a browser.
- [x] eat bacon 🥓

Refers #1523 

Checklist


proto/src/oauth2.rs
```diff
@@ -8,6 +8,11 @@ use serde_with::{base64, formats, serde_as, skip_serializing_none, StringWithSep
 use url::Url;
 use uuid::Uuid;
 
+/// How long a device code is valid for.
+pub const OAUTH2_DEVICE_CODE_EXPIRY: u64 = 300;
+/// How often a client device can query the status of the token
```
Firstyear, line 13: Should say these are seconds.

proto/src/oauth2.rs
```diff
@@ -504,6 +515,32 @@ pub struct ErrorResponse {
     pub error_uri: Option<Url>,
 }
 
+#[derive(Debug, Serialize, Deserialize)]
+pub struct DeviceAuthorizationResponse {
+    device_code: String,
+    user_code: String,
+    verification_uri: String,
```
Firstyear, line 522: Use the URI type.

proto/src/oauth2.rs
```diff
@@ -504,6 +515,32 @@ pub struct ErrorResponse {
     pub error_uri: Option<Url>,
 }
 
+#[derive(Debug, Serialize, Deserialize)]
```
Firstyear, line 518: ref https://www.rfc-editor.org/rfc/rfc8628#section-3.2

proto/src/oauth2.rs
```diff
@@ -102,6 +107,12 @@ pub enum GrantTypeReq {
         #[serde_as(as = "Option<StringWithSeparator::<SpaceSeparator, String>>")]
         scope: Option<BTreeSet<String>>,
     },
+    #[serde(rename = "urn:ietf:params:oauth:grant-type:device_code")]
+    DeviceCode {
+        device_code: String,
```
Firstyear, line 112: ref https://www.rfc-editor.org/rfc/rfc8628#section-3.4

server/lib/src/server/migrations.rs
```diff
@@ -563,10 +563,13 @@ impl<'a> QueryServerWriteTransaction<'a> {
 
         let results = self.internal_search(filter)?;
 
-        let affected_entries = results
-            .into_iter()
-            .map(|entry| entry.get_display_id())
-            .collect::<Vec<_>>();
+        // let affected_entries = results
```
Firstyear, line 566: Should be fixed now.

tools/device_flow/Cargo.toml
```diff
@@ -0,0 +1,33 @@
...
+
+[[bin]]
+name = "kanidm_device_flow"
+path = "src/main.rs"
+
+[dependencies]
+kanidm_proto = { workspace = true }
+anyhow = { workspace = true }
+oauth2 = "4.4.2"
+reqwest = { version = "0.12.8", default-features = false, features = [
```
Firstyear, line 24: If you have tokio, why do you need blocking?

tools/device_flow/src/main.rs
```diff
@@ -0,0 +1,140 @@
+use std::str::FromStr;
```
Firstyear, line 1: What's the intent of this tool?

server/lib/src/idm/oauth2.rs
```diff
@@ -926,7 +979,119 @@ impl<'a> IdmServerProxyWriteTransaction<'a> {
...
+        scope: &Option<BTreeSet<String>>,
+        eventid: Uuid,
+    ) -> Result<DeviceAuthorizationResponse, Oauth2Error> {
+        // DANGER: Why do we have to do this? During the use of qs for internal search
+        // and other operations we need qs to be mut. But when we borrow oauth2rs here we
+        // cause multiple borrows to occur on struct members that freaks rust out. This *IS*
+        // safe however because no element of the search or write process calls the oauth2rs
+        // excepting for this idm layer within a single thread, meaning that stripping the
+        // lifetime here is safe since we are the sole accessor.
+        let o2rs: &Oauth2RS = unsafe {
```
Firstyear, line 1002: I think we could make this a helper method.

server/lib/src/idm/oauth2.rs
```diff
@@ -926,7 +979,119 @@ impl<'a> IdmServerProxyWriteTransaction<'a> {
...
+        // device code is a combination of the inputs
+        fn gen_device_code(user_code: &str, expiry: &Duration) -> Result<String, Oauth2Error> {
+            let mut hasher = Sha256::new();
+            hasher.update(user_code.as_bytes());
+            hasher.update(expiry.as_secs().to_be_bytes());
+            let hash = hasher.finalize();
+            Ok(general_purpose::STANDARD_NO_PAD.encode(hash))
+        }
+
+        fn gen_user_code() -> Result<String, Oauth2Error> {
```
Firstyear, line 1039: I think we should limit to numeric only, and we should hyphenate on triples. This makes it easier to read the code and enter it on a pin pad, phone or similar. Per the rfc this gives 012-324-567 as the code style, and it has 10^9 - 1 bits of entropy, meaning that an attacker would need to sample and attempt 3 million codes per second to bruteforce the whole space. In the db I think we should store the code as a uint32 since we are doing numeric only.

server/lib/src/idm/oauth2.rs
```diff
@@ -926,7 +979,119 @@ impl<'a> IdmServerProxyWriteTransaction<'a> {
...
+                .map(|s| s.iter().cloned().collect::<Vec<_>>().into_iter().join(","))
+                .unwrap_or("[]".to_string()),
+            client_auth_info.source
+        );
+
+        let mut verification_uri = self.oauth2rs.inner.origin.clone();
+        verification_uri.set_path(uri::OAUTH2_DEVICE_LOGIN);
+
+        // device code is a combination of the inputs
+        fn gen_device_code(user_code: &str, expiry: &Duration) -> Result<String, Oauth2Error> {
```
Firstyear, line 1031: Realistically it doesn't have to be. I think a randomised byte array is enough here. I'd do something like a 16 byte value here.

Firstyear (changes requested):
I think the big thing here is we need to add the tests of the device code flow first. This will shape how the endpoints look and work too.

I'd say we focus on the tests in oauth2.rs next, with all the cases like wrong device code, incorrect user code, etc. 

I don't know if pkce here is actually needed - normally pkce is needed because of the redirections for the users browser to prevent theft of the exchange token, but in device code that's not an attack because the user never follows redirects and nor does the device. So I think pkce actually may not be a requirement of this flow. 

WRT to the entry that we need to create for the user to poll against and find, we need to:

Add a new class https://github.com/kanidm/kanidm/blob/master/server/lib/src/constants/schema.rs#L1366  and the attributes. 

Add the new schema to be migrated: https://github.com/kanidm/kanidm/blob/master/server/lib/src/server/migrations.rs#L581


An open question is here is how we want the oauth2 device session to look: i think re-using https://github.com/kanidm/kanidm/blob/master/server/lib/src/valueset/session.rs might not be right as we want to distinguish oauth2 devices from sessions, since they may need to have a disconnected parent. I might need to hurry up and add the stuff so we can do "referenced entries" so that sessions can be separate entries to the user they relate to, and we can handle them correctly. So I'll try to do a generic impl of that this week.

tools/device_flow/Cargo.toml
yaleman: Because that's what the oauth2 client uses for its "simple get things working" example code, and I'm only half done writing this :D

tools/device_flow/src/main.rs
yaleman: Just a test client, it'll probably go away or go into the examples folder at some point :)

Firstyear:
Focus on the oauth2 file and writing tests next I'd say, that's going to direct/guide a lot of this.

server/lib/src/idm/oauth2.rs
```diff
@@ -2701,12 +2815,38 @@ fn validate_scopes(req_scopes: &BTreeSet<String>) -> Result<(), Oauth2Error> {
...
+    let num: u32 = rng.gen_range(0..=999999999);
+    let result = format!("{:09}", num);
+    (
+        format!("{}-{}-{}", &result[0..3], &result[3..6], &result[6..9]),
+        num,
+    )
+}
+
 #[cfg(test)]
 mod tests {
```
Firstyear, line 2844: Still no tests here?

Firstyear: This is where the bulk of the logic and correctness testing needs to be, even if they are "failing" initially.

yaleman: There's e2e tests being built in testkit https://github.com/yaleman/kanidm/blob/754114d4fc9ac6c18fb59c0dc1c998a7a8ac05bd/server/testkit/tests/oauth2_device_flow.rs

yaleman: (But this is still very much a draft currently)

Firstyear: Normally the testkit tests are to check the api works "as we expect" but we put all the detailed logic tests in here because we can control parameters like time and replication for example :)

yaleman: I will be much more able to add them once we've figured out what the internals will look like, and there's something to test, but will add some stuff before this un-drafts 😄

Firstyear: Yeah, I think I just want to be clear the majority of the testing should be here, not in testkit. Testkit is just for api checking etc.

server/core/src/https/oauth2.rs
```diff
@@ -724,6 +737,37 @@ pub async fn oauth2_preflight_options() -> Response {
...
 }
 
+#[serde_as]
+#[derive(Deserialize, Debug, Serialize)]
+pub(crate) struct DeviceFlowForm {
+    client_id: String,
+    #[serde_as(as = "Option<StringWithSeparator::<CommaSeparator, String>>")]
+    scope: Option<BTreeSet<String>>,
+    #[serde(flatten)]
+    extra: HashMap<String, String>,
```
Firstyear, line 747: Should be BTreeMap as HashMap has extra overheads to build, we should only use HashMap for long-lived structures that benefit from the lookup performance.

server/lib/Cargo.toml
```diff
@@ -26,9 +26,10 @@ name = "image_benches"
 harness = false
 
 [features]
-# default = [ "libsqlite3-sys/bundled", "openssl/vendored" ]
+default = ["dev-oauth2-device-flow"]
```
Firstyear, line 29: Shouldn't be on by default until it's ready.

server/lib/src/idm/oauth2.rs
```diff
@@ -305,6 +334,38 @@ pub struct Oauth2RS {
...
+    }
+
+    pub fn is_pkce(&self) -> bool {
+        match self.type_ {
+            OauthRSType::Basic { .. } => false,
+            OauthRSType::Public { .. } => true,
+        }
+    }
+
+    /// Does this RS require PKCE?
```
Firstyear, line 357: Should be "Client" not RS.

server/lib/src/idm/oauth2.rs
```diff
@@ -687,6 +762,8 @@ impl<'a> Oauth2ResourceServersWriteTransaction<'a> {
                     prefer_short_username,
                     type_,
                     has_custom_image,
+                    #[cfg(feature = "dev-oauth2-device-flow")]
```
Firstyear, line 765: Since the cfg feature forces false on this if the feature isn't enabled, you don't need to have a cfg guard on the URL here since it will always be None.

server/lib/src/idm/oauth2.rs
```diff
@@ -925,8 +990,95 @@ impl<'a> IdmServerProxyWriteTransaction<'a> {
...
+        let device_code = gen_device_code()
+            .inspect_err(|err| error!("Failed to generate a device code! {:?}", err))?;
+
+        info!(
+            "verification_origin={} expiry={:?}",
+            verification_uri.to_string(),
+            expiry
+        );
+
+        // TODO: store user_code / expiry / client_id / device_code in the backend, needs to be checked on the token exchange.
```
Firstyear, line 1057: Until it's done, return an error?

server/lib/src/idm/oauth2.rs
```diff
@@ -2701,12 +2835,49 @@ fn validate_scopes(req_scopes: &BTreeSet<String>) -> Result<(), Oauth2Error> {
...
+    Ok(result)
+}
+
+#[inline]
+/// Returns (xxx-yyy-zzz, digits) where one's the human-facing code, the other is what we store in the DB.
+fn gen_user_code() -> (String, u32) {
+    use rand::Rng;
+    let mut rng = rand::thread_rng();
+    let num: u32 = rng.gen_range(0..=999999999);
+    let result = format!("{:09}", num);
```
Firstyear, line 2856: There is probably a modulo or subtract trick you could do here ...

Firstyear (changes requested):
Okay, some small comments, but the main one is default disable of the feature, then I think we can merge so we don't let this get out of control.

server/lib/Cargo.toml
yaleman: It's just while I'm testing 😄

server/lib/src/idm/oauth2.rs
yaleman: All I'm doing is zero-padding it for display purposes, why bother with tricks?

Firstyear: I meant so that you don't need to double format.

Firstyear: Like you could do format!("{:03}-{:03}-{:03}", num /  1000000,  (num / 1000) % 1000,  num % 1000  );

Firstyear: yes, I'm micro-optimising.

Firstyear (approved):

Firstyear (approved):

PR #3101, opened 2024-10-12
Add support for initgroups hooks in libnss_kanidm

Doridian:
# Change summary

This adds support for the `initgroups` NSS feature to Kanidm.

The reasoning behind this change is that [musl-nscd](https://github.com/pikhq/musl-nscd/blob/master/src/main.c#L181) can't use `getgrent` / `setgrent` / `endgrent` hooks to instead populate auxiliary groups. glibc's NSS has a solution for this: `libnss_compat`, but musl has no such feature.

This results in the current [experimental Alpine package](https://pkgs.alpinelinux.org/package/edge/testing/x86_64/kanidm) to not resolve any auxiliary groups, leaving users with only their primary group. This is of course suboptimal.

This introduces an implementation of the `initgroups` hooks therefor, which makes musl happy and populates all groups on Alpine.

I understand if this PR is not wanted, but since it is so little code I figured I'd submit it anyway :)

# Checklist

- ~[ ] book chapter included (if relevant)~
- ~[ ] design document included (if relevant)~

### DIFF
--- a/unix_integration/nss_kanidm/src/implementation.rs
+++ b/unix_integration/nss_kanidm/src/implementation.rs
@@ -5,6 +5,7 @@ use kanidm_unix_common::unix_proto::{ClientRequest, ClientResponse, NssGroup, Ns
 use libnss::group::{Group, GroupHooks};
 use libnss::interop::Response;
 use libnss::passwd::{Passwd, PasswdHooks};
+use libnss::initgroups::{InitgroupsHooks};
 
 struct KanidmPasswd;
 libnss_passwd_hooks!(kanidm, KanidmPasswd);
@@ -179,6 +180,22 @@ impl GroupHooks for KanidmGroup {
                 _ => Response::NotFound,
             })
             .unwrap_or_else(|_| Response::NotFound)
+    }
+}
+
+struct KanidmInitgroups;
+libnss_initgroups_hooks!(kanidm, KanidmInitgroups);
+
+impl InitgroupsHooks for KanidmInitgroups {
+    fn get_entries_by_user(user: String) -> Response<Vec<Group>> {
+        return match KanidmGroup::get_all_entries() {
+            Response::Success(g) => Response::Success(
+                g.into_iter()
+                    .filter(|g| g.members.contains(&user))
+                    .collect(),
+            ),
+            res => res,
+        };
     }
 }
 


Firstyear (changes requested):
Thanks for the PR! We do want the feature, but I think rather than filtering in the nss module to get only groups for that user, within the resolver daemon we actually have a call internally that can handle this. I can either help you through the process of adding that, or I can pick up and finish this PR for you.

Either way, thanks!

Doridian:
@Firstyear I'd love to learn how to do this right, but keep in mind this is the first time I've been touching Rust, so some patience may be required as I learn the quirks and patterns of a new language.
If that is not what you were looking for, feel free to take the PR and finish it, but i'd love to learn if you got the time!

yaleman:
And if it's musl-specific it probably should be filtered for that target? 🤔

Doridian:
> And if it's musl-specific it probably should be filtered for that target? 🤔

I wouldn't call it musl specific as the initgroups API existing is also beneficial for other platforms that have nss_compat, since it means no longer having to rely on that compatibility layer

Firstyear:
> @Firstyear I'd love to learn how to do this right, but keep in mind this is the first time I've been touching Rust, so some patience may be required as I learn the quirks and patterns of a new language. If that is not what you were looking for, feel free to take the PR and finish it, but i'd love to learn if you got the time!

We're very happy to be patient! 

Well it's also a challenge because not only are you learning a new language, but you're also stepping into a new project, which has all of it's own patterns and quirks. So I appreciate you wanting to help.

I'm sure by this point you already know that pam/nss is split into two layers - the pam/nss library which is a thin layer that connects to a unix socket, and the resolver that actually does the work and can cache information about users and groups.

The change you made is only in the nss layer. That layer should really only communicate requests to the resolver, which actually does the work to service them.


There are protocol bindings that the module uses are defined here: https://github.com/kanidm/kanidm/blob/master/unix_integration/common/src/unix_proto.rs#L113 

So I'd say step 1 is to add "NssGroupsByMember(String)" where String is the name of the user we want to fetch our groups from. Then you'd make a call from nss, similar to https://github.com/kanidm/kanidm/blob/master/unix_integration/nss_kanidm/src/core.rs#L181 which sends the ClientRequest of NssGroupsByMember you just defined. 

Once you do that, a quick compile/check will show you that you need to handle the call in the resolver here:

https://github.com/kanidm/kanidm/blob/master/unix_integration/resolver/src/bin/kanidm_unixd.rs#L220

This simply calls down to the actual backend which is https://github.com/kanidm/kanidm/blob/master/unix_integration/resolver/src/resolver.rs#L691


So you'd add a get_nssgroups_by_member() into resolver.rs, and wire the bits up that call it. At this point it's a good idea to add a test: Have a look at https://github.com/kanidm/kanidm/blob/master/unix_integration/resolver/tests/cache_layer_test.rs#L319 which is the "most similar" to what you'll need to do here. 

Now you have to add the logic in that gets the groups.

First you have to handle the possibility the user is in local groups, so you would copy and adjust https://github.com/kanidm/kanidm/blob/master/unix_integration/resolver/src/resolver.rs#L692 filtering on groups where the user is a member (after the group is extended). 

To actually get the remote groups you would need to add a "get groupmembers" using something like https://github.com/kanidm/kanidm/blob/master/unix_integration/resolver/src/resolver.rs#L564 and https://github.com/kanidm/kanidm/blob/master/unix_integration/resolver/src/db.rs#L795 as a template (but just query on account_t.uuid instead of memberof_t.g_uuid)


Anyway I know this seems like a lot, but we have to consider cases where systems have 10,000s of users and groups, so iterating over them all to get the groups with one member would be really slow. This will be more efficient and works for all backends. 

At anypoint feel free to update the PR with your code and ask questions, we're here to help :)

Doridian:
> First you have to handle the possibility the user is in local groups

From my (quite limited of course) looking into NSS stuff. Should we populate local groups for a user even if the user is a kandim user in initgroups? At least to my understanding, the "correct" way would be to just keep the local groups resolution in your `nsswitch.conf` (in `group` and `initgroups`) and NSS would then query both, and append the data together?

I'll definitely get reading on all the other stuff as well soon, when I find the time. Thanks for the very detailed explanation for sure.

Firstyear:
> > First you have to handle the possibility the user is in local groups
> 
> From my (quite limited of course) looking into NSS stuff. Should we populate local groups for a user even if the user is a kandim user in initgroups? At least to my understanding, the "correct" way would be to just keep the local groups resolution in your `nsswitch.conf` (in `group` and `initgroups`) and NSS would then query both, and append the data together?

Yes, but we also are making it so that pam_kanidm can replace pam_unix / nss files :) 

So we need to return when a remote user is a member of a local group too. 

> 
> I'll definitely get reading on all the other stuff as well soon, when I find the time. Thanks for the very detailed explanation for sure.

No problem at all. If you need any more help at all, lmk :)

Doridian:
@Firstyear Sorry, its taken me a while to get back to this. I have started implementing the `NssGroupsByMember` functionality. I think the call itself now. Not sure I'm happy with [the filtering I implemented](https://github.com/kanidm/kanidm/pull/3101/files#diff-323dced691870506ba1b019ae52d5c3dce1c42718a0ff99ad66aae11445da931R745) but the functionality especially dealing with system groups would probably have to be duplicated otherwise, or I might be missing something.

Firstyear:
> @Firstyear Sorry, its taken me a while to get back to this. I have started implementing the `NssGroupsByMember` functionality. I think the call itself now. Not sure I'm happy with [the filtering I implemented](https://github.com/kanidm/kanidm/pull/3101/files#diff-323dced691870506ba1b019ae52d5c3dce1c42718a0ff99ad66aae11445da931R744) but the functionality especially dealing with system groups would probably have to be duplicated otherwise, or I might be missing something.

No problem, I'll try to get to reviewing this in the coming days :) Really appreciate you putting in the effort!

unix_integration/resolver/src/resolver.rs
```diff
@@ -736,6 +736,15 @@ impl Resolver {
         Ok(r)
     }
 
+    pub async fn get_nssgroups_member_name(&self, account_id: &str) -> Result<Vec<NssGroup>, ()> {
+        let account_name = account_id.to_string();
+        Ok(self.get_nssgroups().await.
+            unwrap_or_else(|_| Vec::new())
+            .into_iter()
+            .filter(|g| g.members.contains(&account_name))
```
Firstyear, line 744: Here

Firstyear (commented):
This is looking really good. We need to handle two more cases at "Here" though.

The first, is we can actually use our DB more effectively here. Within our backend we have a memberof_t table: https://github.com/kanidm/kanidm/blob/master/unix_integration/resolver/src/db.rs#L724

You should be able to add a query that does something like getting the user's account uuid, then just select all memberof_t where account_uuid, then you have all the groups you need to retrieve. That will be much more efficient long term :) You could do it with a join, or simply a set of sequential queries, I'm not too fussed. 

The second is that in the resolver, like in "fallback" we need to handle etcgroups, so a similar "groups.filter()" call in https://github.com/kanidm/kanidm/blob/master/unix_integration/resolver/src/idprovider/system.rs#L329 will be perfect #3187 

Thank you again very much, this is looking really good.

Firstyear:
OH PS: add yourself to contributors.md as well :)

Doridian:
> You should be able to add a query that does something like getting the user's account uuid, then just select all memberof_t where account_uuid, then you have all the groups you need to retrieve. That will be much more efficient long term :) You could do it with a join, or simply a set of sequential queries, I'm not too fussed.

I'll look into that



> The second is that in the resolver, like in "fallback" we need to handle etcgroups, so a similar "groups.filter()" call in https://github.com/kanidm/kanidm/blob/master/unix_integration/resolver/src/idprovider/system.rs#L329 will be perfect #3187

Agreed, right now it wouldn't be necessary as we just call `get_nssgroups` which does the fallback but with a more optimized query this would no longer be the same.

I am confused about the link to #3187 though? Copy paste error or am I missing something?

Firstyear:
> > You should be able to add a query that does something like getting the user's account uuid, then just select all memberof_t where account_uuid, then you have all the groups you need to retrieve. That will be much more efficient long term :) You could do it with a join, or simply a set of sequential queries, I'm not too fussed.
> 
> I'll look into that
> 
> > The second is that in the resolver, like in "fallback" we need to handle etcgroups, so a similar "groups.filter()" call in https://github.com/kanidm/kanidm/blob/master/unix_integration/resolver/src/idprovider/system.rs#L329 will be perfect #3187
> 
> Agreed, right now it wouldn't be necessary as I just call `get_nssgroups` which does the fallback but with a more optimized query this would no longer be the case.
> 
> I am confused about the link to #3187 though? Copy paste error or am I missing something?

I meant, add the function there :)

Firstyear:
To clarify a bit extra, in the resolver we need to handle initgroups here for both system users (since we read and handle /etc/group and passwd now) as well as for other backends. You can see a bit of this "juggling" here https://github.com/kanidm/kanidm/blob/master/unix_integration/resolver/src/resolver.rs#L670 where we have to attempt to "get the user from system" first, then we check the provider backends. 

In this case, we'll need to check the provider backends *first*  (aka the call to the database I mentioned last comment) then we do the filter through system (/etc/groups ) since an admin may add a remote user to a local group.

PR #3121, opened 2024-10-18
more errors for the people

yaleman:
# Change summary

- more explainable errors coming back from things in unix-land because I'm getting errors but ... everything seems to be working? 😄 

Checklist

### DIFF
--- a/unix_integration/common/src/unix_proto.rs
+++ b/unix_integration/common/src/unix_proto.rs
@@ -177,7 +177,7 @@ pub enum ClientResponse {
     ProviderStatus(Vec<ProviderStatus>),
 
     Ok,
-    Error,
+    Error(String),
 }
 
 impl From<PamAuthResponse> for ClientResponse {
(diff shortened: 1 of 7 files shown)


unix_integration/common/src/unix_proto.rs
```diff
@@ -177,7 +177,7 @@ pub enum ClientResponse {
     ProviderStatus(Vec<ProviderStatus>),
 
     Ok,
-    Error,
+    Error(String),
```
Firstyear, line 180: I'm not a fan of String in error types. I think it was a mistake that we did it in the past, so we should probably add stronger types here I think.

yaleman: So swap to an enum?

yaleman: And in that case, should I be using the OperationError enum as the one true error place? 😄

yaleman: (done)

Firstyear (approved):

PR #3123, opened 2024-10-19
Fix passkey auth flow redirects

### LINKED ISSUE #3077: [htmx] Oauth flows don't redirect when needing to auth with passkey
### I did this
sign in via oauth on mealie, using mobile and passkey

### I expected the following
works

### This happened instead
stay stuck on kanidm

### Kanidm version details
1.4.0-dev

ToxicMushroom:
# Change summary

Swaps the js fetching into redirect logic with an html form

Fixes #3077 

Checklist

### DIFF
--- a/server/core/static/pkhtml.js
+++ b/server/core/static/pkhtml.js
@@ -20,7 +20,10 @@ function asskey_login() {
             },
         });
         document.getElementById("cred-form").submit();
-    })
+    }).catch((error) => {
+        console.error(`Failed to complete passkey authentication: ${error}`);
+        throw error;
+    });
 }
 
 try {
--- a/server/core/static/pkhtml.js
+++ b/server/core/static/pkhtml.js
@@ -1,5 +1,5 @@
 
-function asskey_login(target) {
+function asskey_login() {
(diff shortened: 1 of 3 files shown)


server/core/static/pkhtml.js
```diff
@@ -8,46 +8,32 @@ function asskey_login(target) {
...
-        fetch(myRequest).then((response) => {
-            if (response.redirected) {
-                window.location.replace(response.url);
-                return;
-            } else {
-                console.error("expected a redirect");
-            }
-        })
+        document.getElementById("cred-form").submit();
     })
```
yaleman, line 23: Please add a `catch()` in case of errors that ... at least `console.log`'s it? I ... think this'd be enough
```suggestion
    }).catch((error) => console.log(error))
```

yaleman (changes requested):

server/core/static/pkhtml.js
ToxicMushroom: good idea

Firstyear (approved):
Very clever solution :)

yaleman (approved):

PR #3135, opened 2024-10-22
yale's rabbit-hole-chasing-htmx-fixing-megapatch

### LINKED ISSUE #3134: [HTMX] When Kanidm prompts the user for auth unlock, it's not clear what's going on
When the user's unlocking their profile, they don't know why, all they get is a login prompt.

A message saying what's going on.

Normalising random prompts for auth.

### Additional context

This is what you get now:

Screenshot 2024-10-22 at 19 04 37

yaleman:
# Change summary

- adding the `with-axum` feature to askama so templates can be rendered by axum instead of all the extra error handling
- adding an explanatory message for when you're prompted for auth (fixes #3134 
- basic "nothing to see here" page for ui ssh_keys
- stripping out some unnecessary htmx magic and just using web pages as web pages.
- continuing my war against strings-everywhere by replacing a load of URLs with an enum

Checklist


server/core/src/https/views/login.rs
```diff
@@ -51,6 +52,8 @@ struct LoginView {
     domain_custom_image: bool,
     username: String,
     remember_me: bool,
+    /// Prompts the user with a message.
+    login_message: Option<String>,
```
Firstyear, line 56: Not sure I like the message, but as you said, you'll need to rebase on my PR.

server/core/src/https/views/profile.rs
```diff
@@ -36,40 +32,28 @@ struct ProfilePartialView {
...
-            .into_response()
-    } else {
-        HtmlTemplate(profile_view).into_response()
+    Ok(ProfileView {
+        profile_partial: ProfilePartialView {
+            menu_active_item: ProfileMenuItems::UserProfile,
+            can_rw,
+            account_name: uat.name().to_string(),
+            display_name: uat.displayname.clone(),
+            // TODO: this should be the legal name
```
Firstyear, line 52: Given today that it's currently just the uat, I think we rip out anything that isn't uat relevant. Just remove legal name imo.

Firstyear (changes requested):
I think we should just disable the ssh keys page and the edit features on profile for now.

Firstyear (approved):
You are the hero this project doesn't deserve <3

Issue #436, opened 2021-05-06
Redesign of Account/UnixAccount/RadiusAccount intermediates


Firstyear:
These structs in IDM are constructed from entries (at probably non-trivial cost), then used and thrown away. Not all elements would be used either. 

As well there are lot of duplications between these, so we could base these from account, and structure these up. Perhaps trait impls on entry or similar?

CEbbinghaus:
Its been a lil while since I did the last Refactor. Keen on finishing this up by moving the accounts but unsure what path you want to take. 

Personally I'm happy with how the Groups ended up and would be happy to tackle the accounts in the same way. However there are more shared properties within Accounts and given that we cannot inherit (the normal way at least) it would likely require a lot of nesting or duplication (e.g RadiusAccount & ServiceAccount share the `valid_from` member).

The alternative (which we could still add to the existing design) would be a mass of Traits that each define a subset of properties accessible through methods. So for example a `ValidFrom` trait would define `valid_from() -> Option<OffsetDateTime>` implemented on different classes.

Firstyear:
Honestly, I'd rather see them become a reference to an Entry, which just borrows and validates the needed parts of the entry. That way we can divide these up by usage a bit better like "AuthenticatingAccount" which has what need for auth, RadiusAccount which has just what's needed for radius etc. We should probably try to avoid clones here too for these. 

But there are some challenges with it and the need to resolve groups, I think some more thought needed. Mostly I really want to get away from so many macros/clones in these.

CEbbinghaus:
Okay. Let's make part of the goal reducing the number of clones. Happy to revisit the groups and improve them as part of this design. I think a wrapped entity might be a good option. I'll give it a bit more of a detailed thought but "reduce clones" is a good guiding principle

Firstyear:
Yeah, the original reason for this was to "validate the structure of an entry and then make it accessible" so that we didn't have to be doing option checks repeatedly (think getting credentials or groups etc). But yeah, it ended up with lots of clones and it's own mess. So I think a way to validate an entry has "some defined shape", and do it with less clones would be good.

PR #3136, opened 2024-10-22
Chore: Refactor Groups to be more generic

CEbbinghaus:
# Change summary

Refactored the groups to be generic over T which lets us implement different kinds of groups without having to duplicate as much logic. Cleaned up the code while at it.

It did require adding 2 extra traits to allow Narrowing down Entries into only valid states for us to read.

Precursor to fixing #436

Checklist


server/lib/src/idm/group.rs
```diff
@@ -1,401 +1,275 @@
...
-                let unix_group = UnixGroup::try_from_entry(group_entry.as_ref())?;
-                unix_groups.push(unix_group);
-            }
-        }
+use super::accountpolicy::{AccountPolicy, ResolvedAccountPolicy};
 
-        (groups, unix_groups)
-    }};
-}
+// I hate that rust is forcing this to be public
```
yaleman, line 13: It's fine, it's a trait 🤷🏻

server/lib/src/idm/account.rs
```diff
@@ -235,7 +232,8 @@ impl Account {
         value: &Entry<EntrySealed, EntryCommitted>,
         qs: &mut QueryServerReadTransaction,
     ) -> Result<Self, OperationError> {
-        let (groups, unix_groups) = load_all_groups_from_account_entry(value, qs)?;
+        let groups = Group::<()>::try_from_account(value, qs)?;
+        let unix_groups = Group::<Unix>::try_from_account(value, qs).unwrap_or_default();
```
CEbbinghaus, line 236: I do dislike that doing it properly requires this double query and iteration of the groups. If we believe this may become a performance bottleneck or just want it optimized I can write a helper function to replicate the previous single iteration behavior.

yaleman: I'd look at doing that, especially since that call's in the path for oauth2/unix/ldap auth.

server/lib/src/idm/account.rs
```diff
@@ -248,8 +246,10 @@ impl Account {
     where
         TXN: QueryServerTransaction<'a>,
     {
-        let ((groups, unix_groups), rap) =
-            load_all_groups_from_account_entry_with_policy(value, qs)?;
+        let groups = Group::<()>::try_from_account(value, qs)?;
+        let unix_groups: Vec<Group<Unix>> =
+            Group::<Unix>::try_from_account(value, qs).unwrap_or_default();
```
yaleman, line 251: Another double-iterate here.

CEbbinghaus: there are about 5 lying around. I'll fix that up with a single call that does both

Firstyear: I wonder if the answer is something like we did before with account and account_unix_extn. Where we iterate over groups once and they optionally have unix extn. 

Saying that I do wonder if these "intermediate" types need to just be removed and re-thought somehow.

CEbbinghaus: I mean, The second part of this PR is giving the same treatment to the Account. We want the properties to be known at compile time rather than runtime. The fact we have to check `unix_extn` every time is what is killing me. We should know during compilation what properties this struct has.

Firstyear (approved):

Firstyear:
@CEbbinghaus It looks like there is a failing test?

CEbbinghaus:
Yeah, There are a few failing tests including the one I wrote. It looks like I messed something up in the double loading. I need to fix that but had to call it a night at 1:40 am. Will try and get it fixed tonight. Will depend on how intoxicated I am when I get back from work :3

server/lib/src/idm/group.rs
```diff
@@ -280,99 +280,44 @@ pub(crate) fn load_all_groups_from_account<'a, E, TXN>(
...
     let entries = qs.internal_search(f).map_err(|e| {
         admin_error!(?e, "internal search failed");
         e
     })?;
 
-    println!("entries: {:?}", entries);
+    println!("{}", entries.len());
 
-    let mut unix_groups = vec![];
+    // Do we need the user group for non unix?
```
Firstyear, line 305: For non-unix, no.

CEbbinghaus: Easy. Had a feeling. I'll remove that comment and then it's ready to go

Firstyear: Still reading :)

Firstyear (approved):

Firstyear (approved):

PR #3143, opened 2024-10-25
examples: add ExecReload for server

0xC0ncord:
With cert reloading available to the server, add an `ExecReload=` directive to the example systemd unit providing this functionality.

# Change summary
- Add an `ExecReload=` directive to the server example systemd unit

Checklist


Firstyear:
This is a good suggestion, but I think there may be a nicer way - it looks like systemd supports a "notify-reload" type which automatically handles sighup, so I think I'll close this PR of implementing that quickly :)

PR #3165, opened 2024-10-31
chore: Made scopes required

CEbbinghaus:
# Change summary

the cli now requires the scopes to be provided. This ensures that empty scope maps aren't accidentally created and helps remove confusion

Fixes: My Sanity

Checklist

### DIFF
--- a/tools/cli/src/opt/kanidm.rs
+++ b/tools/cli/src/opt/kanidm.rs
@@ -943,7 +943,7 @@ pub struct Oauth2CreateScopeMapOpt {
     nopt: Named,
     #[clap(name = "group")]
     group: String,
-    #[clap(name = "scopes", num_args = 1..)]
+    #[clap(name = "scopes", required = true)]
     scopes: Vec<String>,
 }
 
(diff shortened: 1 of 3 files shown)


tools/cli/src/opt/kanidm.rs
```diff
@@ -943,7 +943,7 @@ pub struct Oauth2CreateScopeMapOpt {
     nopt: Named,
     #[clap(name = "group")]
     group: String,
-    #[clap(name = "scopes")]
+    #[clap(name = "scopes", num_args = 1..)]
```
Firstyear, line 946: Num args doesn't set a minimum on how many args, it's how many to collect per occurance. 

https://docs.rs/clap/latest/clap/struct.Arg.html#method.num_args

I think you actually want "required".

CEbbinghaus: > For example, if you had a -f <file> argument where you wanted exactly 3 ‘files’ you would set .num_args(3), and this argument wouldn’t be satisfied unless the user provided 3 and only 3 values.

Sounded like it was for specifying the required count. But required is exactly the property needed

Firstyear (approved):

PR #3170, opened 2024-11-01
chore: Remove empty scopemaps

### LINKED ISSUE #3167: Ensure scope-maps are deleted when empty
Currently the scope maps continue to exist when empty. This isn't ideal as they serve no purpose and clutter the info on the oauth2 service.

Scope maps get deleted when they are set to empty

### LINKED ISSUE #3168: Prevent empty scope list in `update-scope-map`
Currently it is possible to create an empty scope map with `update-scope-map` just by providing an empty array. This will in future cause the scope map to be deleted so we will want to block creation of an empty scope map at the API level to stop someone accidentally deleting a scope-map by "updating" it.

CEbbinghaus:
# Change summary

This change removes the ability to create or update a scope map with a list of scopes that is empty. It also deletes a scope map that is inserted and empty.

Fixes #3167
Fixes #3168

Checklist

### DIFF
--- a/server/core/src/https/v1_oauth2.rs
+++ b/server/core/src/https/v1_oauth2.rs
@@ -201,10 +201,6 @@ pub(crate) async fn oauth2_id_scopemap_post(
 ) -> Result<Json<()>, WebError> {
     let filter = oauth2_id(&rs_name);
 
-    if scopes.is_empty() {
-        return Err(WebError::OperationError(OperationError::EmptyRequest));
-    }
-
     state
         .qe_w_ref
         .handle_oauth2_scopemap_update(client_auth_info, group, scopes, filter, kopid.eventid)
--- a/server/core/src/https/v1_oauth2.rs
+++ b/server/core/src/https/v1_oauth2.rs
@@ -200,6 +200,11 @@ pub(crate) async fn oauth2_id_scopemap_post(
     Json(scopes): Json<Vec<String>>,
 ) -> Result<Json<()>, WebError> {
     let filter = oauth2_id(&rs_name);
+
+    if scopes.is_empty() {
+        return Err(WebError::OperationError(OperationError::EmptyRequest));
+    }
+
     state
         .qe_w_ref
         .handle_oauth2_scopemap_update(client_auth_info, group, scopes, filter, kopid.eventid)
--- a/server/lib/src/valueset/oauth.rs
+++ b/server/lib/src/valueset/oauth.rs
@@ -279,6 +279,7 @@ impl ValueSetT for ValueSetOauthScopeMap {
         match value {
             Value::OauthScopeMap(u, m) => {
                 match self.map.entry(u) {
+                    // We are going to assume that a vacant entry will not be set to empty.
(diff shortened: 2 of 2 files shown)


yaleman:
If an empty change is inherently a delete in the internal interface then why not match this in the API?

CEbbinghaus:
Wouldn't want someone to run `kanidm system oauth2 update-scope-map service group` and run `kanidm system oauth2 get service` and not see `group`. An empty scope is useless so they should be deleted but creating an empty scope map is pointless too so it should be disallowed from the API. 

It should be very clear when a scope map is created / updated or deleted and we wouldn't want someone to accidentally create a scope map just to find it missing because it was empty and therefore deleted.

server/core/src/https/v1_oauth2.rs
```diff
@@ -200,6 +200,11 @@ pub(crate) async fn oauth2_id_scopemap_post(
     Json(scopes): Json<Vec<String>>,
 ) -> Result<Json<()>, WebError> {
     let filter = oauth2_id(&rs_name);
+
```
Firstyear, line 203: Yeah, I think we don't need this error here, if you set an empty scope map via the api, you're saying "it doesn't need to be set". I think just the changes to valueset suffice here since the cli has some other ux guards now.

Firstyear (changes requested):

server/core/src/https/v1_oauth2.rs
CEbbinghaus: Did we want to delete the entire delete endpoint then as a delete would become a empty update? Depricate the api and switch over the cli?

Firstyear:
@CEbbinghaus Any issues if I finish this PR for you? I reckon I could smash it out tomorrow.

CEbbinghaus:
Oh shit. I forgot I still had this open. Let me check but I'd like to finish what I started. Gotta get my contrib score up somehow

server/core/src/https/v1_oauth2.rs
CEbbinghaus: This remains unanswered. I would go either:
* Delete is an update operation with an empty scopemap (we depricate the delete endpoint)
* Delete is its own operation not to be conflated with update. (We ensure in the API that every update contains a body, and keep using the delete endpoint)

Firstyear: In a batch operation we still need to know the difference and this matters in scim too. Leave it as is, we still have a delete operation, and it also happens that setting empty set acts as a delete. Let's not get too tricky that "oh you can delete every attribute except a scope map where you set it to empty.". That's just confusing.

CEbbinghaus: hmm, I'd prefer to have single responsibility so deletion should either be its own endpoint or implicit behaviors. Not a big fan of keeping both but I can see why it may be important for scim when other services.

Will remove the warning

CEbbinghaus:
This is API internal so I don't think it needs any changes to docs, If there is nothing else this should do

Firstyear (approved):

PR #3194, opened 2024-11-09
Prevent Invalid MFA Reg States

### LINKED ISSUE #2940: Invalid MFARegStates should not be achievable during credential update
### I did this
Write new webui #2897 

### I expected the following
Everything works first try /j

### This happened instead
Invalid MFARegState state during certain credential update flows such ass adding a passkey, adding a totp code if they were interrupted previously.

### Solution
Internally only 1 of these MFA flows maybe be ongoing, a potential solution is to just cancel the old MFA flow when the user attempts to start a new one.

This could lead to the user having a cancelled flow opened on another tab, further actions on a cancelled flow should result in a descriptive error message if possible.

### LINKED ISSUE #2977: Improve cred-update session cleanup
To assist @weijiangan and @ToxicMushroom with the credential update, we need to make auto-cleanup of sessions a bit more seamless to prevent complex ux/ui callbacks to try to trim things. 

- If an MFA reg is already inflight and we start another, cancel the first
- we should have the cred update session be *shorter* but renew itself while the user is interacting with it

This way if the user navigates away and stops interaciton, we clean up the session automatically.

Firstyear:
* If an MFA reg was abandoned, we just overwrite it rather than error
* Make cred max_ttl validity checking a bit clearer in the flow
* Review session timeout handling

Fixes #2940 Fixes #2977 

Checklist

### DIFF
--- a/server/lib/src/idm/credupdatesession.rs
+++ b/server/lib/src/idm/credupdatesession.rs
@@ -31,6 +31,8 @@ use compact_jwt::jwe::JweBuilder;
 
 use super::accountpolicy::ResolvedAccountPolicy;
 
+// A user can take up to 15 minutes to update their credentials before we automatically
+// cancel on them.
 const MAXIMUM_CRED_UPDATE_TTL: Duration = Duration::from_secs(900);
 // Minimum 5 minutes.
 const MINIMUM_INTENT_TTL: Duration = Duration::from_secs(300);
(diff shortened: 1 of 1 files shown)


server/lib/src/idm/credupdatesession.rs
```diff
@@ -1848,11 +1847,9 @@ impl<'a> IdmServerCredUpdateTransaction<'a> {
             return Err(OperationError::AccessDenied);
         };
 
-        // Is there something else in progress?
-        // Or should this just cancel it ....
+        // Is there something else in progress? Cancel it if so.
         if !matches!(session.mfaregstate, MfaRegState::None) {
-            admin_info!("Invalid TOTP state, another update is in progress");
-            return Err(OperationError::InvalidState);
+            debug!("Cancelling abandoned mfareg");
```
yaleman, line 1852: For clarity we are allowing the regeneration rather than taking by a specific "cancel" action, maybe make this "ignoring abandoned/incomplete"?

yaleman (changes requested):

yaleman (approved):

PR #3195, opened 2024-11-09
Brain Dumps

Firstyear:
Brain dumping about some upcoming ideas. 

Fixes #

Checklist


book/src/developers/designs/ephemeral_entries.md
```diff
@@ -0,0 +1,24 @@
...
+
+## Class
+
+A new class `EphemeralObject` will be added. It will have a must attribute of `removedAt`
+which will contain a time at which the entry will be deleted.
+
+## Automatic Deletion
+
+A new interval task similar to the recycle/tombstone tasks will be added that checks for
+and deletes ephemeral objects once removedAt has past.
```
CEbbinghaus, line 18: How about making the deletion lazy? and having the interval run more rarely (every hour, day or week). So a ephemeral entry queried after its `removeAt` would be deleted during the query and a negative result would be returned and those that do not get queried would be cleaned up during the next run. That would ensure no query would succeed on an expired ephemeral while also keeping the search task rare (which may remove the need for a index entirely but its probably better to have the index in any case.

Firstyear: Index or not, the issue is that during that prune operation we need to hold a write txn, and without an index it takes longer blocking user writes etc. So it's going to be better to have that index anyway, because there are plenty of other use cases for it outside of just this :) 

Most of these I think after thinking about it are bigger tasks than I initially thought in the first place.

Firstyear: Though I think the initial version won't have the sorting index, and provided its small entry sets it's okay.

CEbbinghaus: Could do the same with the ephemeral as we do with the weak password. If its past its due date send a message to get it deleted and reject the request. That way we don't do the write in place, still reject the read / auth and queuing a deletion of the expired record which would should run quickly since it would be removing a single entry (so small lock). Yeah I agree that a index would be required given that a deletion would require a write lock

Firstyear: That's true, it could be an async delayed task. I'll have to think about it more. But also I think sometimes it's better to go "correct, simple, fast". Correct is just a task that checks for things that need deleting on an interval, and it's not that complex. And initially the index won't be needed like I said. It's later we'll need it.

CEbbinghaus: This discussion was resolved in our community channel yesterday with the conclusion that the entry isn't guaranteed to be invalid after the `removedAt` timestamp. It should instead read `removedAfter` and only guarantees an entry to be removed at some point after a certain timestamp. It remains valid up until the point at which it is deleted which simplifies the logic since all that is required is a periodic cleanup task.

This also tracks given that this ephemeral record may also exist in another instance, and it would take some time to mirror the deletion out anyways.

book/src/developers/designs/subentries.md
```diff
@@ -0,0 +1,131 @@
...
+The existing referential integrity code is designed to ensure that values from one entry are removed
+cleanly if the referenced entry is deleted. As an example, a group with a member "ellie" should have
+the reference deleted when the entry "ellie" is deleted.
+
+If the group were deleted, this has no impact on ellie, since the reference is defining a weak
+relationship - the user is a member of a group.
+
+## What Is Required
+
+What we need in a new reference type are the following properties.
```
yaleman, line 25: A lot of this section should be simply handled with foreign keys in the database, we shouldn't reinvent the wheel on this one.

Firstyear: @yaleman We can't, because we have replication and replication conflicts. *We* are the database and *we* are implementing foreign keys here, so we need to be aware of things.

PR #3232, opened 2024-11-22
client: read attestation CA list JSON from file

Be-ing:
# Change summary
Read the attestation CA list JSON from a file instead of passing a giant blob of JSON as a command argument. Before, it was not possible to allow all valid authenticators certified by the FIDO Alliance because

```bash
fido-mds-list query -o "status gte valid"
```

outputs a JSON string longer than Linux allows for command arguments.

Checklist

- [x] book chapter included: will update #3231 after this is merged


yaleman:
This makes sense in context but the docs need updating please.

yaleman:
Oh I didn't see you did it in two PR's, that's fine. Thanks!

tools/cli/src/opt/kanidm.rs
```diff
@@ -212,7 +212,7 @@ pub enum GroupAccountPolicyOpt {
     #[clap(name = "webauthn-attestation-ca-list")]
     WebauthnAttestationCaList {
         name: String,
-        attestation_ca_list_json: String,
+        attestation_ca_list_json_file: PathBuf,
```
CEbbinghaus, line 215: Yeah, nah this is a backwards breaking change. We try to keep backwards breaking changes to a minimum so ideally we keep allowing the old json string while also allowing the user to provide a file.

```suggestion
        attestation_ca_list_json: Option<String>,
        attestation_ca_list_json_file: Option<PathBuf>,
```

I think the above change would allow the user to keep providing the json through the cli. Might require some clap macro flags to make it work. And will probably need to check that at least one of the arguments was provided in the logic that handles this command.

CEbbinghaus (changes requested):
Good start, I think it can be improved a bit however. We don't want to break people's existing workflow unless we absolutely have to.

tools/cli/src/opt/kanidm.rs
CEbbinghaus: Did not realise the cli wasn't considered stable. Suppose @Firstyear would need to chime in on what he prefers

Firstyear: The CLI is not a stable interface, and this change improves usability, so I think I'm happy with it.

CEbbinghaus (approved):
Happy with it otherwise, @Firstyear you got anything against this getting merged?

Firstyear (approved):

PR #3252, opened 2024-11-30
Allow OAuth2 loopback redirects if the path matches

yaleman:
# Change summary

- Per BCP212 (RFC8252) 7.3 and 8.3, fitting the requirement that IDPs MUST allow a loopback address with an ephemeral port. 
- Require that localhost URIs are specified in the URI list for OAuth2 RSen (where we match the path)
- Update the docs to reflect that

Refers #2390

Checklist

### DIFF
--- a/server/lib/src/idm/oauth2.rs
+++ b/server/lib/src/idm/oauth2.rs
@@ -46,7 +46,7 @@ use serde_with::{formats, serde_as};
 use time::OffsetDateTime;
 use tracing::trace;
 use uri::{OAUTH2_TOKEN_INTROSPECT_ENDPOINT, OAUTH2_TOKEN_REVOKE_ENDPOINT};
-use url::{Origin, Url};
+use url::{Host, Origin, Url};
 
 use crate::idm::account::Account;
 use crate::idm::server::{
@@ -1826,12 +1826,6 @@ impl<'a> IdmServerProxyReadTransaction<'a> {
             } => *allow_localhost_redirect,
         };
 
-        let localhost_redirect = auth_req
-            .redirect_uri
-            .domain()
-            .map(|domain| domain == "localhost")
-            .unwrap_or_default();
-
         // Strict uri validation is in use.
         let strict_redirect_uri_matched =
             o2rs.strict_redirect_uri && o2rs.redirect_uris.contains(&auth_req.redirect_uri);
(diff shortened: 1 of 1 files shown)


server/lib/src/idm/oauth2.rs
```diff
@@ -1819,67 +1830,57 @@ impl<'a> IdmServerProxyReadTransaction<'a> {
...
+            let origin_uri_matched =
+                !o2rs.strict_redirect_uri && o2rs.origins.contains(&auth_req.redirect_uri.origin());
+
+            // Strict uri validation is in use.
+            let strict_redirect_uri_matched =
+                o2rs.strict_redirect_uri && o2rs.redirect_uris.contains(&auth_req.redirect_uri);
+            // Allow opaque origins such as app uris.
+            let opaque_origin_matched = o2rs.opaque_origins.contains(&auth_req.redirect_uri);
+
+            // At least one of these conditions must hold true to proceed.
```
Firstyear, line 1850: I don't like this re-arrangement, it makes it harder to follow the logic. Before it was split up to check what conditions might be true, then we assert the combination we need.

book/src/integrations/oauth2.md
```diff
@@ -328,14 +329,29 @@ kanidm system oauth2 create-public <name> <displayname> <origin>
...
 
 ```bash
 kanidm system oauth2 enable-localhost-redirects <name>
 kanidm system oauth2 disable-localhost-redirects <name>
 kanidm system oauth2 enable-localhost-redirects mywebapp
 ```
 
+Then configure a local URL. We're going to match on "is it a loopback address" and "does the path match" so specify something similar to the following:
+
+```bash
```
Firstyear, line 348: This is a lot of configuration to make this work. It also enshrines magic which is that we have to handle the possible localhost variants, and we set ports etc. There also likely isn't a guarantee the path is stable.

We also need to be clear that these are *only* allowed on public clients which enforce pkce.

Firstyear (changes requested):
I think we need a simpler approach, I think this adds a lot of configuration complexity, and it enshrines some magic that I don't like (such as localhost never actually checking ports etc). 

Where we went wrong was poor explanation of what the problem could be in our error messages. 

Because we have the preconditions that localhost redirects can only exist on a public client, and public clients enforce PKCE, then I don't think we should enforce strict URI matching on localhost redirection. localhost redirection due to it's use of an ephemeral port and possibly even a dynamic path, combined with PKCE gives a very strong assurance that the actual exchange is legitimate and can't be altered. 

I think that we should undo most of this PR, and change the localhost redirection check to correctly handle ports, and we should improve the error so that if the host is localhost, we explain the conditions required for the redirect to work. I'm happy to redo the PR because I'm being a complainy clown about it.

server/lib/src/idm/oauth2.rs
yaleman: I'll try another run at it - it got pretty convoluted hence not doing the checks until they were needed.

server/lib/src/idm/oauth2.rs
```diff
@@ -1819,67 +1831,49 @@ impl<'a> IdmServerProxyReadTransaction<'a> {
...
+                    warn!(
+                        "Invalid OAuth2 redirect_uri (must be related to origin) - got {:?}",
+                        auth_req.redirect_uri.origin()
+                    );
+                }
+                return Err(Oauth2Error::InvalidOrigin);
+            }
+            // We have to specifically match on http here because non-http origins may be exempt from this
+            // enforcement.
+            if o2rs.origin_https_required && auth_req.redirect_uri.scheme() == "http" {
```
Firstyear, line 1865: Shoud be != https here instead of == http (for example, ftp:// )

Firstyear (changes requested):
One minor comment but looks good !

server/lib/src/idm/oauth2.rs
yaleman: Hah, I nearly made that change and then was like ... weeeellllll

Firstyear (dismissed):

Firstyear (approved):

PR #3262, opened 2024-12-03
Work around systemd race condition

Firstyear:
Systemd reload can't handle us reloading so quickly which causes "reload or restart" to always "restart" kanidm incorrectly.

# Change summary

-

Fixes #

Checklist


yaleman (approved):
thanks `systemd` we love you (my love language involves wet haddock)

PR #3266, opened 2024-12-04
Cleanup of println and other outputs

### LINKED ISSUE #3199: Output the server version while starting up

If you are looking at server logs of a broken kanidm, neither *where* the kanidm came from (`argv[0]`) nor the version of the kanidm are printed out during server startup. Ideally both would be printed to make it clear which kanidm is being run without needing to do external diagnostics.

I would like for kanidm to print out the server version on startup.

Are there any alternative solutions or features you've considered.

Firstyear:
# Change summary

- While debugging something I noticed a println that made it to production. Went and chased it down, ended up on a bit of a deep dive to make the reindex print process nice. Also adds the server version at startup. 

Fixes #3199 

Checklist

### DIFF
--- a/server/lib/src/be/mod.rs
+++ b/server/lib/src/be/mod.rs
@@ -1684,7 +1684,7 @@ impl<'a> BackendWriteTransaction<'a> {
         let dbv = self.get_db_index_version()?;
         admin_debug!(?dbv, ?v, "upgrade_reindex");
         if dbv < v {
-            self.reindex()?;
+            self.reindex(false)?;
             self.set_db_index_version(v)
         } else {
             Ok(())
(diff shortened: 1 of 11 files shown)


server/lib/src/be/mod.rs
```diff
@@ -1704,39 +1710,46 @@ impl<'a> BackendWriteTransaction<'a> {
...
+                    }
                 }
+
                 self.entry_index(None, Some(e))
             })
-            .map_err(|e| {
-                admin_error!("reindex failed -> {:?}", e);
-                e
+            .inspect_err(|err| {
+                error!(?err, "reindex failed");
```
yaleman, line 1735: I'm thinking we need to print a new line before this so the dots don't end on the logline or just look ugly in general

yaleman (approved):

server/lib/src/be/mod.rs
Firstyear: I tested this, and surprisingly, we don't. It has a new line properly there.

PR #3274, opened 2024-12-08
Always advertise PKCE code challenges methods supported even if PKCE is disabled in kanidm for a client.

gbw:
# Change summary
Per RFC8414 the authorization server should include metadata of supported code challenge methods. _If omitted, the authorization server does not support PKCE._

Clients and libraries should be able to check the well known endpoint and attempt PKCE if the auth server supports it.

Lastly this is a minor info leak if not always shown.

Checklist

### DIFF
--- a/server/lib/src/idm/oauth2.rs
+++ b/server/lib/src/idm/oauth2.rs
@@ -2512,11 +2512,7 @@ impl IdmServerProxyReadTransaction<'_> {
 
         let service_documentation = Some(URL_SERVICE_DOCUMENTATION.clone());
 
-        let code_challenge_methods_supported = if o2rs.require_pkce() {
-            vec![PkceAlg::S256]
-        } else {
-            Vec::with_capacity(0)
-        };
+        let code_challenge_methods_supported = vec![PkceAlg::S256];
 
         Ok(Oauth2Rfc8414MetadataResponse {
             issuer,
@@ -2587,11 +2583,7 @@ impl IdmServerProxyReadTransaction<'_> {
         let claims_supported = None;
         let service_documentation = Some(URL_SERVICE_DOCUMENTATION.clone());
 
-        let code_challenge_methods_supported = if o2rs.require_pkce() {
-            vec![PkceAlg::S256]
-        } else {
-            Vec::with_capacity(0)
-        };
+        let code_challenge_methods_supported = vec![PkceAlg::S256];
 
         // The following are extensions allowed by the oidc specification.
 


yaleman:
But if you don't enable PKCE for  client then it's *not* supported/going to work?

gbw:
The way I thought about it is that the well known endpoint basically describes the capabilities that the auth server in this case Kanidm supports. The client uses this info to negotiate the "best" possible flow. It's analogous to server and client negotiating SSL protocols/ciphers.

Perhaps my understanding is incorrect though!

yaleman:
@Firstyear thoughts on our interpretation of the RFC here?

Firstyear:
The specific section from RFC8414 that matters here is 

```
code_challenge_methods_supported
      OPTIONAL.  JSON array containing a list of Proof Key for Code
      Exchange (PKCE) [[RFC7636](https://www.rfc-editor.org/rfc/rfc7636)] code challenge methods supported by this
      authorization server.  Code challenge method values are used in
      the "code_challenge_method" parameter defined in Section 4.3 of
      [RFC7636].  The valid code challenge method values are those
      registered in the IANA "PKCE Code Challenge Methods" registry
      [[IANA.OAuth.Parameters](https://www.rfc-editor.org/rfc/rfc8414#ref-IANA.OAuth.Parameters)].  If omitted, the authorization server
      does not support PKCE.
```

When we read this, we need to consider the relationship between the authorisation server *and* the client. So I think we should interpret this as "JSON array containing a list of PKCE code challenge methods supported by this authorisation server __when this specific client interacts with the AS__". 

This is especially true as our authorisation server metadata is *per client* rather than domain-wide.

Were we sharing a single metadata statement between multiple oauth2 client definitions, then I would agree here that we should always advertise PKCE. However, given each metadata is per client, we should advertise precisely what the authorisation server expects from *this* client.

To take this further, if we advertised PKCE to a client when the oauth2 client config in Kanidm had PKCE disabled, then we would actually warn here https://github.com/kanidm/kanidm/blob/master/server/lib/src/idm/oauth2.rs#L1868 since we don't want users to be mixing/matching having PKCE sometimes - it should be always or never. Consider that an attacker may attempt to strip the PKCE request if we didn't have this set to "required".

The other side of this is if we advertised PKCE to a client, but the client doesn't support PKCE, then it may refuse to interoperate with Kanidm even if in this context the oauth2 client definition had PKCE requirements disabled. 

In conclusion, I think our current metadata statement is correct - we advertise PKCE for client configurations when it is required, and we do not advertise it when it is *not* required to prevent client confusion and inconsistency. Each metadata statement accurately reflects the capabilities of the authorisation server *for that client*. 

Hope that helps :)

gbw:
Thanks it does help my understanding. The other side of the coin was that disabling PKCE on Kanidm's side didn't actually disable PKCE. It is like you said which just changes it to a "flexible" (warn) mode vs. a "strict" (or required as you say) mode.

I would personally think that since it's encouraged for clients/libraries to implement PKCE that a flexible mode is desired and the well known endpoint would reflect that. 

I don't feel strongly about this, so message received and closing!

Firstyear:
Thanks for asking the question - if there is a compelling reason to change it in the future, I'm not going to say no, but I think today it's "clearer" this way :)

PR #3282, opened 2024-12-10
Trigger Rust GitHub Action workflows when Rust files are changed and …

gbw:
…similar for Python files.

# Change summary

Selectively run GitHub workflows based on files changed. 

Checklist

### DIFF
--- a/.github/workflows/clippy.yml
+++ b/.github/workflows/clippy.yml
@@ -4,7 +4,21 @@ name: Linting checks
 # Trigger the workflow on push
 "on":
   push:
+    paths:
+      - 'libs/**'
+      - 'platform/**'
+      - 'proto/**'
+      - 'server/**'
+      - 'tools/**'
+      - 'unix_integration/**'
   pull_request:
+    paths:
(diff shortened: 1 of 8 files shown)


yaleman:
I'd rather leave it as is - what's the *actual* issue with the current setup?

A few issues:

- the intention is to (soon) have proper integration testing so that rust/python code has to interact with a live instance to pass checks, which would require reverting this change
- the current PR can't pass the branch-merge requirements (because the rust checks won't be run)...
- nor can we configure Github to have merge requirements that match the "only these files changed"

Firstyear:
@yaleman I think there are a lot of things that run in our CI/CD that don't need to run every commit. It can take a looooooong time to get things merged. So being more selective about what runs and when would be useful.

yaleman:
Sure, but with the changes as-is any commit that doesn't change rust won't be able to be merged as the required tasks won't run.

gbw:
I think caching builds would solve both our concerns. Closing in favor of that.

PR #3296, opened 2024-12-16
Limit OAuth2 resumption to session

### LINKED ISSUE #3292: InvalidState when accessing user credentials through web UI
### I did this
I wanted to add a second passkey to my account and decided to add it through the web UI. After authenticating with my YubiKey to access the credentials page instead of sending me back to the credentials page it gave me an error.

### I expected the following
To be sent right to the user credential page.

### This happened instead
I got an error message indicating "InvalidState". The page URL ends with "/ui/oauth2/resume".
image
Refreshing the page would keep the error open but change the Operation ID. I was able to bypass the error by going back to the apps page, then navigating to the credentials page. This seems to indicate that I authenticated successfully, but somehow an error comes up after I first authenticate with my passkey.

### Kanidm version details

- Output of `kanidm(d) version`: kanidm 1.4.4
- If not a container, how'd you install it: bare metal
- Operating System / Version: NixOS 24.11 (Linux 6.8.12-5-pve)

### Any other comments
I am able to replicate this every time I need to reauth for the credentials menu, so I can get additional logs.
I do get the same "InvalidState" error page when trying to use OAuth with a new Gitea server I am setting up. I am unsure if this is related, but I believe that may be due to bad configuration I still need to iron out on my end.

Firstyear:
OAuth2 session resumption was accidentally made a permanent cookie which led to continuing issues with it causing invalid redirections after login. Make this a session only cookie.

Fixes #3292 

Checklist


server/core/src/https/views/login.rs
```diff
@@ -899,22 +899,25 @@ async fn view_login_step(
...
                         // Important - this can be make unsigned as token_str has it's own
                         // signatures.
-                        let bearer_cookie = cookies::make_unsigned(
+                        let mut bearer_cookie = cookies::make_unsigned(
                             &state,
                             COOKIE_BEARER_TOKEN,
                             token_str.clone(),
                             "/",
                         );
+                        // Important - can be permanent as the token has it's own expiration time internally
```
yaleman, line 908:
```suggestion
                        // Important - can be permanent as the token has its own expiration time internally
```

yaleman (dismissed):

CEbbinghaus (approved):
LGTM, Unrelated question but do we have any configuration option to set cookie timeout? I have to enter my username what feels like very 2 hours which kinda sucks. I would like my username to be saved for at least a month and just need to reauth with the security key every now and then (ideally that would be configurable too)

Firstyear:
> LGTM, Unrelated question but do we have any configuration option to set cookie timeout? I have to enter my username what feels like very 2 hours which kinda sucks. I would like my username to be saved for at least a month and just need to reauth with the security key every now and then (ideally that would be configurable too)

Username save should be in a permanent cookie I think it was a bug previously that we made it session only. Should have already been fixed though?

CEbbinghaus:
> Should have already been fixed though?

I'm still on 1.4.0, Was it fixed since?

Firstyear:
> > Should have already been fixed though?
> 
> I'm still on 1.4.0, Was it fixed since?

Yep :)

yaleman (approved):

PR #3297, opened 2024-12-16
Ignore system users for UPG synthesiseation

Firstyear:
Our unix resolver would attempt the right thing to synthesise user private groups on linux as these are an important security boundary. However, it turns out that almost every distro has botched their default system user accounts, and many are installed with numeric-only UPGs that don't resolve. In the case that later the user does attempt to fix that, because we synthesised as UPG for the system account, the user trying to add the UPG would now fail. In some cases this could cause system updates to be prevented from installing.

This change limits UPG synth to user accounts only (uid > 1000) which is the common uid boundary on unix-like platforms.

cc @tacerus 

Checklist

### DIFF
--- a/unix_integration/resolver/src/idprovider/system.rs
+++ b/unix_integration/resolver/src/idprovider/system.rs
@@ -8,6 +8,8 @@ use kanidm_unix_common::unix_passwd::{CryptPw, EtcGroup, EtcShadow, EtcUser};
 use kanidm_unix_common::unix_proto::PamAuthRequest;
 use kanidm_unix_common::unix_proto::{NssGroup, NssUser};
 
+const SYSTEM_GID_BOUNDARY: u32 = 1000;
+
 pub struct SystemProviderInternal {
     users: HashMap<Id, Arc<EtcUser>>,
     user_list: Vec<Arc<EtcUser>>,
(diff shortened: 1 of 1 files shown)


tacerus:
Thanks for this, should also reduce some of the warnings on startup for system users.

CEbbinghaus (approved):
Beauty

Firstyear:
> Thanks for this, should also reduce some of the warnings on startup for system users.

It's a pain because opensuse seems to be particularly bad with botched system user account uid/gid allocations, and I doubt we'll ever be able to fix it.

CEbbinghaus (approved):
Much prefer this <3

unix_integration/resolver/src/idprovider/system.rs
```diff
@@ -8,6 +8,8 @@ use kanidm_unix_common::unix_passwd::{CryptPw, EtcGroup, EtcShadow, EtcUser};
 use kanidm_unix_common::unix_proto::PamAuthRequest;
 use kanidm_unix_common::unix_proto::{NssGroup, NssUser};
 
+const SYSTEM_GID_BOUNDARY: u32 = 1000;
```
yaleman, line 11: Just so it's clear without having to dig into logic

```suggestion
// The minimum GID that Kanidm will consider for creating a UPG
const SYSTEM_GID_BOUNDARY: u32 = 1000;
```

yaleman (changes requested):
Minor nit

yaleman (approved):

Issue #3299, opened 2024-12-17
Rejecting attribute members during LDAP sync


mweinelt:
### I did this

```
sudo -u kanidm kanidm-ldap-sync -l ldapsync.conf -d 
```

### I expected the following

LDAP syncing to kanidm

### This happened instead

```
2024-12-17T00:40:46.668146Z DEBUG kanidm_ldap_sync: LdapSyncReplEntry {
    entry_uuid: fb8108ce-87e0-103e-8c1d-09fda87104ce,
    state: Add,
    entry: LdapEntry {
        dn: "uid=ccccccccc,ou=users,ou=people,ou=accounts,dc=example.com",
        attrs: {
            "cn": {
                "Aaaaaaa Bbbbbbb",
            },
            "createtimestamp": {
                "20240405214119Z",
            },
            "creatorsname": {
                "gidNumber=0+uidNumber=0,cn=peercred,cn=external,cn=auth",
            },
            "displayname": {
                "ccccccccc",
            },
            "entrycsn": {
                "20240405214757.380391Z#000000#001#000000",
            },
            "entrydn": {
                "uid=ccccccccc,ou=users,ou=people,ou=accounts,dc=example.com",
            },
            "entryuuid": {
                "fb8108ce-87e0-103e-8c1d-09fda87104ce",
            },
            "gecos": {
                "Aaaaaaa Bbbbbbb",
            },
            "gidnumber": {
                "100",
            },
            "givenname": {
                "Aaaaaaa",
            },
            "hassubordinates": {
                "FALSE",
            },
            "homedirectory": {
                "/home/ccccccccc",
            },
            "loginshell": {
                "/bin/bash",
            },
            "mail": {
                "ccccccccc@example.com",
            },
            "memberof": {
                "cn=davical,ou=groups,ou=people,ou=accounts,dc=example.com",
                "cn=git,ou=groups,ou=people,ou=accounts,dc=example.com",
                "cn=mail,ou=groups,ou=people,ou=accounts,dc=example.com",
                "cn=shell,ou=groups,ou=people,ou=accounts,dc=example.com",
            },
            "modifiersname": {
                "uid=ccccccccc,ou=users,ou=people,ou=accounts,dc=example.com",
            },
            "modifytimestamp": {
                "20240405214757Z",
            },
            "objectclass": {
                "inetorgperson",
                "ldappublickey",
                "organizationalperson",
                "person",
                "posixaccount",
                "shadowaccount",
                "top",
            },
[...]

Issue #3298, opened 2024-12-16
LDAP sync fails on group to scim_sync_group


mweinelt:
### I did this

```
sudo -u kanidm kanidm-ldap-sync -l ldapsync.conf -d 
```

### I expected the following

LDAP to sync into Kanidm.

### This happened instead

```
2024-12-16T22:24:58.894775Z DEBUG kanidm_ldap_sync: LdapSyncReplEntry {
    entry_uuid: 7b8bfa7c-132c-1038-804b-759e7035fbdf,
    state: Add,
    entry: LdapEntry {
        dn: "cn=foo,ou=groups,ou=people,ou=accounts,dc=example,dc=com",
        attrs: {
            "cn": {
                "foo",
            },
            "createtimestamp": {
                "20180703164714Z",
            },
            "creatorsname": {
                "cn=admin,dc=example,dc=com",
            },
            "entrycsn": {
                "20230615225636.625437Z#000000#002#000000",
            },
            "entrydn": {
                "cn=foo,ou=groups,ou=people,ou=accounts,dc=example,dc=com",
            },
            "entryuuid": {
                "7b8bfa7c-132c-1038-804b-759e7035fbdf",
            },
            "gidnumber": {
                "5016",
            },
            "hassubordinates": {
                "FALSE",
            },
            "member": {
                "uid=a,ou=users,ou=people,ou=accounts,dc=example,dc=com",
                "uid=b,ou=users,ou=people,ou=accounts,dc=example,dc=com",
                "uid=c,ou=users,ou=people,ou=accounts,dc=example,dc=com",
                "uid=d,ou=users,ou=people,ou=accounts,dc=example,dc=com",
                "uid=e,ou=users,ou=people,ou=accounts,dc=example,dc=com",
            },
            "modifiersname": {
                "cn=admin,dc=example,dc=com",
            },
            "modifytimestamp": {
                "20230615225636Z",
            },
            "objectclass": {
                "groupofnames",
                "posixgroup",
                "top",
            },
            "structuralobjectclass": {
                "groupOfNames",
            },
            "subschemasubentry": {
                "cn=Subschema",
            },
        },
    },
}
2024-12-16T22:24:58.894835Z ERROR kanidm_ldap_sync: Unable to convert group to scim_sync_group json_err=Error("data did not match any variant of untagged enum ScimValue", line: 0, column: 0)
[...]

PR #3305, opened 2024-12-18
Further SCIM sync testing, minor fixes

Firstyear:
This adds further testing of SCIM sync, especially around conversion of the SCIM Sync Person and Group types into SCIM Entry. This test would have prevented #3298 and #3299 from occurring.

During testing two more fixes were found. external_id should have been required (not optional) and a group with no members would cause a serialisation issue.

Checklist


proto/src/scim_v1/synch.rs
```diff
@@ -128,7 +133,7 @@ impl ScimSyncPerson {
                         SCIM_SCHEMA_SYNC_PERSON.to_string(),
                     ],
                     id,
-                    external_id: None,
+                    external_id: Some(external_id),
```
CEbbinghaus, line 136: Not 100% certain on whether we expect the `ScimSyncPerson` to be sent to us without `external_id` (or really anything scim related for that matter), But if not do we need the `external_id` field to still be `Option<String>`?

CEbbinghaus (approved):
Just a question. Looks nice otherwise

proto/src/scim_v1/synch.rs
Firstyear: SCIM rfc states that external ID is optional, but in our usage of it, it's required. That's why there is a disparity here.

CEbbinghaus: No stress, Just thought to ask because I wasn't sure. Good to go

yaleman (approved):
More testing more better nice work 🥳

Issue #1596, opened 2023-05-05
Seasonally themed icons


Firstyear:
We have seasonal images already available - I think we should display them automatically in certain date ranges such as the christmas Kani.

yaleman:
As long as there's an "out" for folks without a sense of humor or particular cultural sensitivities... or both.

Firstyear:
Yeah, this was an "future fun idea" so I'm not in a rush to do it. We'll need a way to allow custom icons in future too anyway.

tumbl3w33d:
> We'll need a way to allow custom icons in future too

Will this be the issue about theming/customizing the look in general (logo, title, headline, link to data privacy terms,…)? Was taking a look around but didn't see any other issue about it. #1955 is somewhat related since it's all about the enduser experience, building trust, make them feel like they're entering the right place.

yaleman:
Please create a new issue specifically for those things, #1955 is specific to the OAuth2 login flow. 

What it looks like you're asking for is admin-configurable theming, which is separate again to both this issue and that one.

PR #3308, opened 2024-12-18
Allow opt-in of easter eggs

Firstyear:
So that we can start to add some more easter eggs to the server, we also need to respect user preferences that may not want them.

This adds a configuration setting to the domain allowing a release build to opt-in to easter eggs, and development builds to opt-out of them.

Relates #1596

Checklist


yaleman (approved):
At some point we're going to have to convert these kinds of things to a feature-flag type so we don't have to have a whole ACP just to manage an admin-level on/off switch 😄

CEbbinghaus (approved):
I agree that feature flag would be good to reduce the amount of code needed to turn on/off anything like this (holly moly that is a lot of lines of code for `bool allow_easter_eggs`). 

However that GUID is now taken so there is no difference whether we keep this attribute or turn it into a feature flag right? (in terms of GUID assignments)

Firstyear:
> At some point we're going to have to convert these kinds of things to a feature-flag type so we don't have to have a whole ACP just to manage an admin-level on/off switch 😄

"replication"

The problem with feature flags is they need to be equivalent on all nodes of the replication topology. Be it allow easter eggs, or allow unix pw binds etc. Every node must have the same set of feature flags.

So then we say "okay, is that up to the admin to get right in config files?". Well that leaves people open to mistakes and changing the flag becomes a lot of work on the admin.

Or do we make it something replicated in our database? It's set once, and we make sure every node gets all the correct flags. 

So I think the current approach, which not perfect, is correct wrt to admin experience. 

From a code view, if we wanted to *reduce* the complexity to add feature flags, We could do something like a bit-mask where we have a feature flag attribute, and when the admin "changes" a flag we are a bit twiddle of the setting. We can make it nicer to the admin with enums for the variants and then how they get sent can be checked etc. That means to add more flags is just defining a new variant and we already have the room in the flag set for it. Means there would be no changes to acp etc either since domain admin just can write the flags.

yaleman:
> > At some point we're going to have to convert these kinds of things to a feature-flag type so we don't have to have a whole ACP just to manage an admin-level on/off switch 😄
> 
> "replication"
> 
> The problem with feature flags is they need to be equivalent on all nodes of the replication topology. Be it allow easter eggs, or allow unix pw binds etc. Every node must have the same set of feature flags.
> 
> So then we say "okay, is that up to the admin to get right in config files?". Well that leaves people open to mistakes and changing the flag becomes a lot of work on the admin.
> 
> Or do we make it something replicated in our database? It's set once, and we make sure every node gets all the correct flags.
> 
> So I think the current approach, which not perfect, is correct wrt to admin experience.
> 
> From a code view, if we wanted to _reduce_ the complexity to add feature flags, We could do something like a bit-mask where we have a feature flag attribute, and when the admin "changes" a flag we are a bit twiddle of the setting. We can make it nicer to the admin with enums for the variants and then how they get sent can be checked etc. That means to add more flags is just defining a new variant and we already have the room in the flag set for it. Means there would be no changes to acp etc either since domain admin just can write the flags.

That seems like an overly complicated solution, how about instead... for each flag we want to create as coders in the platform, we have (for example) an _Entry_ with the class `FeatureFlag`.

Which then has a name, ie, the flag that we're setting, and an attribute of `FeatureFlagSet` which when it exists is explicitly enabled or is treated as disabled if it doesn't have the attribute (ie, fail closed).

When we add a feature in code the migration can automatically create the entry (either enabled or disabled) just like any other thing we as the coders create... and then it's replicated because.. replication.

No bitmasks needed, just use the data store exactly as you're doing with the explicit attribute type of `DomainAllowEasterEggs`, but don't have a new Attribute type for it. If the flag entry's not there, then it's not enabled and won't affect nodes that haven't gotten it yet (or will automatically disable the feature, whatever it is). If the entry's there and the nodes aren't running a version that cares about it, then they'll also effectively ignore it.

If Kanidm wants to check for a flag (eg `EasterEggOptIn`) it can look for a `FeatureFlag` entry called that, and if it's found and has the enable attribute then we enable the feature. If it's either missing the `enable` attribute or the entry isn't there, we treat it as disabled.

For names of flags we _should_ use an enum, it just doesn't have to be hard coded into the schema layer.

Firstyear:
We need a way to generate those feature flag entries from a template, but otherwise in theory this does work yes.

PR #3309, opened 2024-12-20
nss/pam resolver should reauth faster

Firstyear:
This can have visible impacts on accounts that don't have a pam password cached yet, but then appear to "stall" for a minute or two until it works due to the fact that the provider was offline and waiting to reauth.

When we are still connected but our provider auth session has expired we should reconnect faster. This reduces the timeout for reauthentication for the provider so that it can return to the online state sooner. We also loop when we detect the provider session is no longer authenticated so that we can reauth immediately, rather than causing a noticable interuption.

Checklist


unix_integration/resolver/src/idprovider/kanidm.rs
```diff
@@ -255,23 +257,35 @@ impl KanidmProviderInternal {
...
+        while max_attempts > 0 {
+            max_attempts -= 1;
+            match self.client.auth_anonymous().await {
+                Ok(_uat) => {
+                    debug!("provider is now online");
+                    self.state = CacheState::Online;
+                    return true;
+                }
+                Err(ClientError::Http(StatusCode::UNAUTHORIZED, reason, opid)) => {
+                    error!(?reason, ?opid, "Provider authentication failed");
```
yaleman, line 272:
```suggestion
                    error!(?reason, ?opid, "Provider authentication returned unauthorized, {} attempts remaining.");
```

yaleman (changes requested):
Miiinor nit

yaleman (approved):

PR #3332, opened 2025-01-03
make everything cookie consistent

### LINKED ISSUE #3269: Web UI logins fail/have unexpected results after an oauth2 request
### I did this

In an unauthenticated browser session, bail on or fail an oauth2 authorization flow, then attempt to log in to the kanidm web ui.

### I expected the following

Expected to reach the kanidm web ui.

### This happened instead

The previous oauth2 flow attempts to complete instead, redirecting to /ui/oauth2/resume, and possibly resulting in error or an unexpected destination.

### Kanidm version details

- Output of `kanidm(d) version`: kanidmd 1.4.4
- If not a container, how'd you install it: "bare metal", debian 12 lxc
Linux idm 6.8.12-4-pve \# 1 SMP PREEMPT_DYNAMIC PMX 6.8.12-4 (2024-11-06T15:04Z) x86_64 GNU/Linux

### Any other comments

discovered doing something like this:

1) i have not previously signed into the kanidm web ui, no kanidm cookies are set.

2) i want to sign into an oauth2 app as "user1" but do not complete the request.  the o2-authreq cookie created for that request is set.

3) i want to use the kanidm web ui for something (not much to do there yet, but it's coming along!). depending on where i left the previous oauth2 request, i may or may not be signed into kanidm aleady as user1. if so, i sign out so i can get the login form again to demonstrate the issue. again, the o2-authreq cookie remains.

4) i log in as user2 (can be any user including the previous user1), but the login is hijacked by the previous oauth2 request and ultimately redirects to /ui/oauth2/resume. i may get access denied or otherwise end up somewhere i didn't expect to (or shouldn't) be.

since the login did succeed, i can manually re-enter the base kanidm web ui address in the browser to get to the ui, but i will continue to get the oauth2 request hijack for subsequent logins. signing out from the web ui does not remove the o2-authreq cookie. the only solution appears to be manually deleting the cookie or waiting for it to expire (hey does this cookie need a 30 day expiration?).

i'm testing this with oauth2 for desktop/cli apps, which still has some issues we're discussing in #2390 so i haven't actually completed an oauth2 authorization, but i wonder if this may affect web ui logins even when the oauth2 request succeeds since the current issue with native app requests looks to already be at the redirect-url point.

### LINKED ISSUE #3330: Cookies with a path other than / are not removed on sign out
### I did this
Using Firefox > Developer Tools > Storage > Cookies, I injected a cookie with name `o2-authreq`, a dummy value, an expiration in the future, and the path `/ui`. Upon logging in I get the expected `UI0003InvalidOauth2Resume` error. So far this is expected. 

Go to apps page and sign out. Observe logout request includes a set header to expire the o2-authreq cookie with path `/`. Also observe in cookies storage the `o2-authreq` cookie remains. 

### I expected the following
The cookie to be deleted on sign out. 

### This happened instead
The cookie stayed causing the error above on sign in until expiration. 

### Kanidm version details

- Output of `kanidm(d) version`: 1.4.5
- Are you running it in a container? If so, which image/tag?: 1.4.5
- Operating System / Version (On Unix please post the output of `uname -a`): Debian 12

### Any other comments
It seems expiring a cookie requires the path to match. Not sure if there's a simple way to do this.

Firstyear:
Fixes #3269 Fixes #3330 

Seems that trying to clear a cookie from alternate paths doesn't work how we want. This makes all cookies consistent to "/" to allow values to be cleared at any point in the flow. 

Checklist


Firstyear:
Waiting for @mweinelt to test as they have a solid reproducer

server/core/src/https/views/login.rs
```diff
@@ -917,12 +917,8 @@ async fn view_login_step(
 
                         // Important - this can be make unsigned as token_str has it's own
```
yaleman, line 918:
```suggestion
                        // Important - this can be make unsigned as token_str has its own
```

yaleman (approved):

PR #3335, opened 2025-01-03
Add OAuth2 `response_mode=fragment`

### LINKED ISSUE #3334: Support `response_mode=fragment` in OAuth 2.0 Auth Code requests

I want to be able to authenticate [a Browser-Based OAuth 2.0 (public) client](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-browser-based-apps#section-6.3) (ie: an SPA which handles the OAuth 2.0 token acquisition and refresh process on the client side) to Kanidm to be able to access a resource server.

This is already possible today, but the `redirect_uri` gets the `code` and `state` parameters back as query parameters (`?code=...`), which are generally logged by commonly-used HTTP servers. A good PKCE implementation makes it impossible to actually *use* these to get an access or refresh token, but PKCE can be disabled.

It would be nicer to be able to send these as a fragment instead (`#code=...`), because then the authorisation code is never transmitted in the HTTP request to the application.

[OAuth 2.0 Multiple Response Type Encoding Practices](https://openid.net/specs/oauth-v2-multiple-response-types-1_0.html) adds an optional `response_mode` parameter for Auth Code (`response_type=code`) and Implicit Grant (`response_type=token`) flows.

By default, this is `query` for Auth Code, and `fragment` for Implicit Grant.

The [Form Post response mode](https://openid.net/specs/oauth-v2-form-post-response-mode-1_0.html) could be useful for server-side confidential clients, as then you don't have a non-idempotent `GET` request (per OAuth 2.0 spec) and the `POST` body is less likely to be logged by typical web servers.

But fragment mode is easier to implement, and doesn't depend on having some extra JS to automatically submit a hidden form.

### Additional context

I'm toying around with an implementation in a branch at the moment: #3335

micolous:
# Change summary

Adds support for the `response_mode` parameter from [OAuth 2.0 Multiple Response Type Encoding Practices: Response Modes](https://openid.net/specs/oauth-v2-multiple-response-types-1_0.html#ResponseModes), allowing Authorisation Code flows to use `response_mode=fragment`.

This *does not* implement the "multiple response type" functionality described by that document.

- Change `AuthorisationRequest::response_type` from `String` to `ResponseType`, so that can be plumbed through the stack more sensibly.

  This changes the way an invalid `response_type` is reported to the client, as that'll now error out on a deserialisation layer with:

  ```
  Failed to deserialize query string: unknown variant `hello`, expected one of `code`, `token`, `id_token`
  ```

  If you use `response_type=token` or `response_type=id_token`, Kanidm responds with "Unrecoverable error; Error Code: InvalidState", the same as it does with other invalid `response_type` values before this change.

- Deduplicate `AuthorisePermitSuccess` URL construction, and use the `response_mode`.

- Make `check_oauth2_authorise_reject` work similarly to the success case, and use the `response_mode`.

- Add `response_mode=fragment` to OAuth 2.0 and OpenID discovery documents.

Fixes #3334

Checklist

### DIFF
--- a/server/lib/src/idm/oauth2.rs
+++ b/server/lib/src/idm/oauth2.rs
@@ -2560,7 +2560,7 @@ impl IdmServerProxyReadTransaction<'_> {
         let jwks_uri = Some(o2rs.jwks_uri.clone());
         let scopes_supported = Some(o2rs.scopes_supported.iter().cloned().collect());
         let response_types_supported = vec![ResponseType::Code];
-        let response_modes_supported = vec![ResponseMode::Query];
+        let response_modes_supported = vec![ResponseMode::Query, ResponseMode::Fragment];
         let grant_types_supported = vec![GrantType::AuthorisationCode];
 
         let token_endpoint_auth_methods_supported = vec![
@@ -2631,7 +2631,7 @@ impl IdmServerProxyReadTransaction<'_> {
         let jwks_uri = o2rs.jwks_uri.clone();
         let scopes_supported = Some(o2rs.scopes_supported.iter().cloned().collect());
         let response_types_supported = vec![ResponseType::Code];
-        let response_modes_supported = vec![ResponseMode::Query];
+        let response_modes_supported = vec![ResponseMode::Query, ResponseMode::Fragment];
 
         // TODO: add device code if the rs supports it per <https://www.rfc-editor.org/rfc/rfc8628#section-4>
         // `urn:ietf:params:oauth:grant-type:device_code`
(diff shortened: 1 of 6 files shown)


server/lib/src/idm/oauth2.rs
```diff
@@ -236,6 +238,66 @@ pub struct AuthorisePermitSuccess {
...
+            ResponseMode::Fragment => redirect_uri.set_fragment(Some(&encoded)),
+        }
+
+        redirect_uri
+    }
+}
+
+#[derive(Debug)]
+pub struct AuthoriseReject {
+    // Where the RS wants us to go back to.
```
Firstyear, line 272: Client, not RS. It's a language mistake from the past.

micolous: Fixed.

Firstyear (approved):

Firstyear:
Thank you!

Issue #3319, opened 2024-12-28
Allow unixd with anonymous disabled


johansmitsnl:
### Is your feature request related to a problem? Please describe

Have the unixd integration with the anonymous account disabled. 

### Describe the solution you'd like

Include a service token or something else to make a authorized connection. Like this you can also limit the devices that have access or if they are compromised to prevent users from logging in.

### Describe alternatives you've considered

No

### Additional context

By default with anonymous you can get a lot of data (ssh, groups, login shell, gidnumber) that can be used to gain specific knowledge of the users environment. And use this information to build phishing attacks. I don't want this information to be in the open.

jinnatar:
First, from previous discussions, I'll add two things:
1. The enumeration problem is a bit worse than just users, the anonymous account can by default also enumerate for example enabled oauth2 clients. Ergo, one can (and I've done it as a PoC) find publicly exposed Kanidm instances on Shodan and query exactly what apps are gated by said instance. So not only can one find users to target, you can also likely identify apps they are actually using to enhance the realism of a campaign, or to find attack vectors. Nothing you can't already do, but it's served on a very convenient platter.
2. The alternative mitigation in addition to disabling or impairing the anonymous account is to use a WAF or capable reverse proxy to limit what API endpoints are accessible from the public internet.
   - It's best to explicitly allow only known needed endpoints externally.
   -  An example is provided below that allows oauth flows & user account management to work from the outside. It does not allow most management calls, meaning that also the kanidm cli tool won't work from behind the rules.
   - Allowlist in regex form applied on the path: `^/v1/(auth|self|credential|jwk|reauth|oauth2)`, everything else under `^/v1/` blocked.

### My 2c on a proper solution:
I agree that personally for me empowering unixd with credentials of some sort generated by Kanidm would be a low effort solution from an admin perspective in a small to moderate environment. Most likely I would create, securely store in a vault & distribute at runtime a single token to all trusted server clients, or even perhaps bother with per-host tokens.

Any kind of more automated or less-trusted environment would benefit from a cert dispenser style strategy, say for example for using a secrets vault to mint & dispense them. My knowledge on that front goes only so far, the canonical example that comes to mind would be using PKI, but that would require Kani to accept certs for auth. The benefit being that PKI is a well tooled solution for large/complex/over-engineered infra, even if quite obtuse. Just mentioning it as an example of how trust distribution has been solved elsewhere, happy to hear of any easier alternatives!

Firstyear:
[...]

PR #3336, opened 2025-01-04
Ignore anonymous in oauth2 read allow access

Firstyear:
Administrators will sometimes configure oauth2 clients with `idm_all_accounts` as an allowed scope group. Despite anonymous being *unable* to interact with oauth2, this still allowed oauth2 clients to be read by anonymous in this configuration. For some users, this may be considered a public info disclosure.

Relates #3319 

Checklist

### DIFF
--- a/server/lib/src/server/access/search.rs
+++ b/server/lib/src/server/access/search.rs
@@ -168,6 +168,11 @@ fn search_oauth2_filter_entry(ident: &Identity, entry: &Arc<EntrySealedCommitted
     match &ident.origin {
         IdentType::Internal | IdentType::Synch(_) => AccessResult::Ignore,
         IdentType::User(iuser) => {
+            if iuser.entry.get_uuid() == UUID_ANONYMOUS {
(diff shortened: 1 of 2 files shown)


server/lib/src/server/access/search.rs
```diff
@@ -168,6 +168,11 @@ fn search_oauth2_filter_entry(ident: &Identity, entry: &Arc<EntrySealedCommitted
     match &ident.origin {
         IdentType::Internal | IdentType::Synch(_) => AccessResult::Ignore,
         IdentType::User(iuser) => {
+            if iuser.entry.get_uuid() == UUID_ANONYMOUS {
+                trace!("Ignore UUID_ANONYMOUS for oauth2 filter checking");
```
yaleman, line 172:
```suggestion
                debug!("Anonymous can't access OAuth2 entries, ignoring");
```

yaleman (changes requested):
Just a nit so folks can find it down the line, but otherwise 🥳

yaleman (approved):

PR #3343, opened 2025-01-07
Resolve passkey regression

Firstyear:
During other testing I noticed that passkeys no longer worked on a reauthentication. This was due to a regression in you guessed it, cookies, where the auth session id wasn't being removed properly.

Checklist

### DIFF
--- a/server/core/src/https/views/login.rs
+++ b/server/core/src/https/views/login.rs
@@ -12,7 +12,7 @@ use axum::{
     response::{IntoResponse, Redirect, Response},
     Extension, Form, Json,
 };
-use axum_extra::extract::cookie::{Cookie, CookieJar, SameSite};
+use axum_extra::extract::cookie::{CookieJar, SameSite};
 use kanidm_proto::internal::{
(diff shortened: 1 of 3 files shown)


server/core/src/https/views/login.rs
```diff
@@ -835,10 +835,8 @@ async fn view_login_step(
                 break res;
             }
             AuthState::Continue(allowed) => {
-                // Reauth inits its session here so we need to be able to add cookie here ig.
-                if jar.get(COOKIE_AUTH_SESSION_ID).is_none() {
-                    jar = add_session_cookie(&state, jar, &session_context)?;
-                }
+                // Reauth inits its session here so we need to be able to add it's cookie here.
+                jar = add_session_cookie(&state, jar, &session_context)?;
```
yaleman, line 839: `add_session_cookie` sets `SameSite::Strict`, do we really want to keep using that?

Firstyear: Yes. This is the auth-session-id, so you'd only be getting it within the samesite.

Like if you navigate away then back to Kanidm, you should get a new auth-session.

yaleman (approved):

Firstyear:
Only master right now.

PR #3346, opened 2025-01-08
Add ssh_publickeys as a claim for oauth2

### LINKED ISSUE #3331: SSH Public key claim

A claim type for the OAuth2 JWT's that holds SSH Public Key

Having the public key in the service I'm building

### Additional context

I'm currently slowly planning an SSH CA infrastructure where the signing server (PKI??) uses kanidm and other compatible IdP's. The signed SSH public key, I want to be from the JWT token, in hopes that it'll make the system easier to manage.

I don't know if this is the only usecase for such claim type however.

Firstyear:
Allow ssh_publickeys to be exposed as a claim for oauth2 and oidc applications so that they can consume these keys for various uses. An example could be something like gitlab which can then associate the public keys with the users account.

Fixes #3331

Checklist


server/lib/src/idm/oauth2.rs
```diff
@@ -2070,6 +2070,10 @@ impl IdmServerProxyReadTransaction<'_> {
                 }
             };
 
+            if granted_scopes.contains(OAUTH2_SCOPE_SSH_PUBLICKEYS) {
+                pii_scopes.insert(OAUTH2_SCOPE_SSH_PUBLICKEYS.to_string());
```
yaleman, line 2074: Are public keys really PII? Or is this just funny naming

Firstyear: IMO yes ssh publickeys are PII and we should tell the user when we plan to share them to another service.

yaleman: I disagree on the PII bit but warning them's sensible

yaleman (approved):

PR #3350, opened 2025-01-10
fix(server/config): reduce string allocations

jalil-salame:
# Change summary

- Replace `String::replace()` with `str::strip_prefix()` remove a source of allocations.

Previously the code would do `key.replace("KANIDM_", "")`, this allocates a new string, which is unnecessary, as we can simply call `strip_prefix("KANIDM_")`.

This removes the `KANIDM_` prefix from a bunch of places, and doubles as a check that the variable is prefixed with `KANIDM_`. Overall I believe this change makes the code more robust and slightly reduces allocations, speeding up an admittedly cold function (only called very infrequently).

Checklist

### DIFF
--- a/server/core/src/config.rs
+++ b/server/core/src/config.rs
@@ -250,31 +250,28 @@ impl ServerConfig {
     /// Updates the ServerConfig from environment variables starting with `KANIDM_`
     fn try_from_env(mut self) -> Result<Self, String> {
         for (key, value) in std::env::vars() {
-            if !key.starts_with("KANIDM_") {
+            let Some(key) = key.strip_prefix("KANIDM_") else {
                 continue;
-            }
+            };
 
             let ignorable_build_fields = [
-                "KANIDM_CPU_FLAGS",
-                "KANIDM_CPU_FLAGS",
-                "KANIDM_DEFAULT_CONFIG_PATH",
-                "KANIDM_DEFAULT_CONFIG_PATH",
-                "KANIDM_DEFAULT_UNIX_SHELL_PATH",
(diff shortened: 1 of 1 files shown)


server/core/src/config.rs
```diff
@@ -250,31 +250,31 @@ impl ServerConfig {
...
-                "KANIDM_PKG_VERSION_HASH",
-                "KANIDM_PKG_VERSION",
-                "KANIDM_PRE_RELEASE",
-                "KANIDM_PROFILE_NAME",
+                "CPU_FLAGS",
+                "CPU_FLAGS",
+                "DEFAULT_CONFIG_PATH",
+                "DEFAULT_CONFIG_PATH",
+                "DEFAULT_UNIX_SHELL_PATH",
+                "DEFAULT_UNIX_SHELL_PATH",
```
yaleman, line 263: I know you didn't put them there but these repetitions are clearly a mistake too 😂

yaleman (changes requested):

server/core/src/config.rs
jalil-salame: Whoops, I'll fix them too c:

jalil-salame: Sorted them while I was at it c:

yaleman (approved):
This is my level of yak-shaving, thanks 😄

Firstyear:
> Overall I believe this change makes the code more robust and slightly reduces allocations, speeding up an admittedly cold function (only called very infrequently).

Still an improvement, still good :) 

Thank you!

PR #3352, opened 2025-01-11
fix: unrecoverable error page doesn't include logo or domain name

yaleman:
# Change summary

- unrecoverable error page doesn't include logo or domain name, adding it where we have access

Checklist


server/core/src/https/views/mod.rs
```diff
@@ -8,8 +8,12 @@ use axum::{
 
 use axum_htmx::HxRequestGuardLayer;
 
+use concread::cowcell::CowCellReadTxn;
```
Firstyear, line 11: There is an exposed type for this https://github.com/kanidm/kanidm/blob/master/server/core/src/https/extractors/mod.rs#L18 which already has the concread wrapper, so you don't need concread as a dep in server/views here at all.

server/core/src/https/views/mod.rs
```diff
@@ -29,6 +33,8 @@ mod reset;
 struct UnrecoverableErrorView {
     err_code: OperationError,
     operation_id: Uuid,
+    // This is an option because it's not always present in an "unrecoverable" situation
+    domain_info: Option<CowCellReadTxn<DomainInfo>>,
```
Firstyear, line 37: Option<DomainInfoRead>

server/core/src/https/views/login.rs
```diff
@@ -981,3 +997,35 @@ fn add_session_cookie(
         })
         .ok_or(OperationError::InvalidSessionState)
 }
+
+#[cfg(test)]
+mod tests {
```
Firstyear, line 1002: It's good to have tests, but this test doesn't seem to be exercising anything that could be that fragile? Is the purpose of the test to find if we break the body http or something? I can't see a scenario where this could ever fail ....

server/core/Cargo.toml
```diff
@@ -30,6 +30,7 @@ axum-server = { version = "0.7.1", default-features = false }
 bytes = { workspace = true }
 chrono = { workspace = true }
 compact_jwt = { workspace = true }
+concread = { workspace = true }
```
Firstyear, line 33: See below, not needed.

server/core/src/https/views/login.rs
```diff
@@ -59,7 +60,7 @@ impl fmt::Display for ReauthPurpose {
         }
     }
 }
-
+#[derive(Clone)]
```
Firstyear, line 63: Why do these need clone?

server/core/src/https/views/errors.rs
```diff
@@ -58,6 +58,7 @@ impl IntoResponse for HtmxError {
                         UnrecoverableErrorView {
                             err_code: inner,
                             operation_id: kopid,
+                            domain_info: None,
```
Firstyear, line 61: I don't know what I think about this one, because this is the only place we use domain_info as a "none". So could we give this a unique error view instead? But also if the point of this change is to render domain info on errors, this seems like the place where a lot of errors pass through, so we'd want the domain info here too wouldn't we?

Firstyear (changes requested):

server/core/Cargo.toml
```diff
@@ -88,6 +89,8 @@ webauthn-rs = { workspace = true, features = [
 [dev-dependencies]
 walkdir = { workspace = true }
 tempfile = { workspace = true }
+http-body-util = "0.1.2"
```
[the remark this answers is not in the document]
Firstyear: Especially since this dep relates to axum, and axum is bad enough to upgrade without having a pile of other deps that depend with it. I think we should only be adding deps if they really provide a massive value to us in some way, we have too many as is and we need to cut down on them, not add more.

server/core/src/https/views/login.rs
```diff
@@ -981,3 +997,35 @@ fn add_session_cookie(
...
+            operation_id: Uuid::new_v4(),
+            domain_info: Some(domain_info.read()),
+        };
+
+        let response = view.into_response();
+
+        // TODO: this really should be an error code :(
+        assert_eq!(response.status(), 200);
+
+        let body = BodyExt::collect(response.into_body())
```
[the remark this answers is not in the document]
yaleman: I'd gladly see a way to do it if anyone can point it out but... I looked through all the axum example code and couldn't find how to get the body from a `Response` object without it.

server/core/src/https/views/login.rs
yaleman: Because `LoginDisplayCtx` has `domain_info` [here](https://github.com/kanidm/kanidm/pull/3352/files#diff-7886cf3d77b5287a765c6414392df5fa1e66ce34f033459147258dfc7df1af38R239) and the error from that needs to be able to access domain_info...

server/core/src/https/views/login.rs
yaleman: Mainly just making sure it renders and includes something we expect it to, because actually triggering the error while human-testing is unreliable/hard/bordering on impossible without finding another bug.

server/core/src/https/views/errors.rs
yaleman: Sorted

yaleman:
@Firstyear fixes in, removed a lot of the things. Oh look selinux is being an asshole in a whole new way! In the tests!

yaleman:
(Works fine locally)

Firstyear:
So I had more of a think about this overnight and something about this pr isn't sitting right, I'm wondering what it "fixes". If we are showing the unrecoverable error page a lot, that means we probably need some actual dedicated error handling in the places this pops up. I'm also not sure what we get by branding that error page, when we should be avoiding showing this page unless something really really bad happens. 

What do you think this pr fixes? Like what is it trying to achieve?

yaleman:
This one had a clearly broken theme (screenshot from a resized desktop browser below) with the text jammed up in the corner, so that needed fixing, so I tried to make it consistent with the other error messages we show - users should know what to expect, even if the text can be confusing to them.

We sure should be avoiding showing it, but given it's come up a *lot* lately for folks with the cookie error and others, and the number of places I had to modify - I think it's important that all error pages look as consistent as we possibly can.

Given we've had lots of changes around wording for things like "TOTP" etc, and included the custom system name / logo on other branding, this fits that general theme.

Screenshot_20250110_100846

server/core/src/https/views/login.rs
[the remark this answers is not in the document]
yaleman: Found it and removed it.

Firstyear (approved):

PR #3355, opened 2025-01-11
Repair systemd reload notifications

tacerus:
# Change summary

In order for the RELOAD and the subsequent READY notifications to be correctly processed, the RELOAD notification must be accompanied with a MONOTONIC_USEC one.
Unfortunately, the sd-notify crate does not seem to support this, and I have not yet looked into implementing it.
Hence this patch is comprised of:
- replacement of sd-notify with systemd for the daemon component to support the MONOTONIC_USEC notification
- adding of the MONOTONIC_USEC notification to repair the RELOAD => READY notifications
- adding of STATUS notifications for both easier debugging of whether notifications arrive with systemd and easier user facing feedback about the current daemon status
- removal of the artificial delay, which is no longer beneficial
- disabling of unset_environment, as the preservation of $NOTIFY_SOCKET
is crucial

Fixes https://progress.opensuse.org/issues/173533 (can create accompanying issue here if you'd like).

Checklist

### DIFF
--- a/.github/workflows/clippy.yml
+++ b/.github/workflows/clippy.yml
@@ -27,12 +27,13 @@ jobs:
           sudo apt-get update && \
           sudo apt-get install -y \
             libpam0g-dev \
-            libudev-dev \
+            libselinux1-dev \
             libssl-dev \
             libsystemd-dev \
+            libtss2-dev \
+            libudev-dev \
             pkg-config \
-            tpm-udev \
-            libtss2-dev
+            tpm-udev
       - name: "Run clippy"
         run: cargo clippy --lib --bins --examples --all-features
   fmt:
--- a/.github/workflows/clippy.yml
+++ b/.github/workflows/clippy.yml
@@ -29,6 +29,7 @@ jobs:
             libpam0g-dev \
             libudev-dev \
             libssl-dev \
+            libsystemd-dev \
             pkg-config \
             tpm-udev \
             libtss2-dev
--- a/server/daemon/Cargo.toml
+++ b/server/daemon/Cargo.toml
@@ -38,7 +38,6 @@ tokio = { workspace = true, features = ["rt-multi-thread", "macros", "signal"] }
 tokio-util = { workspace = true, features = ["codec"] }
 tracing = { workspace = true }
 serde_json.workspace = true
-systemd = { version = "0.10.0", default-features = false }
 
 [target.'cfg(target_os = "linux")'.dependencies]
 sd-notify.workspace = true
--- a/server/daemon/Cargo.toml
+++ b/server/daemon/Cargo.toml
@@ -38,6 +38,7 @@ tokio = { workspace = true, features = ["rt-multi-thread", "macros", "signal"] }
 tokio-util = { workspace = true, features = ["codec"] }
 tracing = { workspace = true }
 serde_json.workspace = true
+systemd = { version = "0.10.0", default-features = false }
 
 [target.'cfg(target_os = "linux")'.dependencies]
 sd-notify.workspace = true
--- a/server/daemon/src/main.rs
+++ b/server/daemon/src/main.rs
@@ -751,7 +751,17 @@ async fn kanidm_main(
             if !config_test {
                 // On linux, notify systemd.
                 #[cfg(target_os = "linux")]
-                let _ = sd_notify::notify(true, &[sd_notify::NotifyState::Ready]);
+                {
+                    let _ = sd_notify::notify(true, &[sd_notify::NotifyState::Ready]);
+                    // Undocumented systemd feature - all messages should have a monotonic usec sent
+                    // with them. In some cases like "reloading" messages, it is undocumented but
+                    // failure to send this message causes the reload to fail.
+                    let _ = sd_notify::notify(true, &[sd_notify::NotifyState::MonotonicUsec]);
+                    let _ = sd_notify::notify(
+                        true,
+                        &[sd_notify::NotifyState::Status("Started Kanidm 🦀")],
+                    );
+                };
 
                 match sctx {
                     Ok(mut sctx) => {
@@ -785,7 +795,13 @@ async fn kanidm_main(
                                         // Reload TLS certificates
                                         // systemd has a special reload handler for this.
                                         #[cfg(target_os = "linux")]
+                                        {
                                         let _ = sd_notify::notify(true, &[sd_notify::NotifyState::Reloading]);
+                                        // CRITICAL - if you do not send a monotonic usec message after a reloading
+                                        // message, your service WILL BE KILLED.
+                                        let _ = sd_notify::notify(true, &[sd_notify::NotifyState::MonotonicUsec]);
+                                        let _ = sd_notify::notify(true, &[sd_notify::NotifyState::Status("Reloading ...")]);
+                                        }
 
                                         sctx.tls_acceptor_reload().await;
 
@@ -794,7 +810,11 @@ async fn kanidm_main(
                                         tokio::time::sleep(std::time::Duration::from_secs(5)).await;
 
                                         #[cfg(target_os = "linux")]
+                                        {
                                         let _ = sd_notify::notify(true, &[sd_notify::NotifyState::Ready]);
+                                        let _ = sd_notify::notify(true, &[sd_notify::NotifyState::MonotonicUsec]);
+                                        let _ = sd_notify::notify(true, &[sd_notify::NotifyState::Status("Reload Success")]);
+                                        }
 
                                         info!("Reload complete");
                                     }
--- a/server/daemon/src/main.rs
+++ b/server/daemon/src/main.rs
@@ -751,14 +751,7 @@ async fn kanidm_main(
             if !config_test {
                 // On linux, notify systemd.
                 #[cfg(target_os = "linux")]
-                let _ = systemd::daemon::notify(
-                    false,
-                    [
-                        (systemd::daemon::STATE_READY, "1"),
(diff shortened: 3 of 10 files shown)


tacerus:
I'm not sure the replacement of the library is acceptable, after all the "systemd" one is bigger with other features we do not have a use for.
Should we try to implement MONOTONIC_USEC in "sd-notify" instead?

server/daemon/src/main.rs
```diff
@@ -785,18 +792,33 @@ async fn kanidm_main(
                                         // Reload TLS certificates
                                         // systemd has a special reload handler for this.
                                         #[cfg(target_os = "linux")]
-                                        let _ = sd_notify::notify(true, &[sd_notify::NotifyState::Reloading]);
+                                        // https://stackoverflow.com/a/70337205
+                                        // is this close enough to CLOCK_MONOTONIC in clock_gettime(2) ?
```
yaleman, line 796: For this use case it should be fine, we're not worried about filestomping and things like that.

yaleman (changes requested):

server/daemon/src/main.rs
```diff
@@ -785,18 +792,33 @@ async fn kanidm_main(
...
+                                        let duration_since_epoch = match std::time::SystemTime::now().duration_since(std::time::SystemTime::UNIX_EPOCH) {
+                                            Ok(val) => val,
+                                            Err(err) => {
+                                                error!(
+                                                  "Failed to get UNIX_EPOCH!\nError: {:?}",
```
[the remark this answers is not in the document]
tacerus: Ok, I was mislead by `"Failed to parse {:?} into CA chain!\nError: {:?}"` in the same file. ;-)

server/daemon/Cargo.toml
```diff
@@ -38,6 +38,7 @@ tokio = { workspace = true, features = ["rt-multi-thread", "macros", "signal"] }
 tokio-util = { workspace = true, features = ["codec"] }
 tracing = { workspace = true }
 serde_json.workspace = true
+systemd = "0.10.0"
```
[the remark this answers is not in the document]
tacerus: Thanks, added! I didn't realize this was possible.

server/daemon/src/main.rs
tacerus: Thanks for confirming!

tacerus:
Any idea what's missing for the test suite? Seems the clippy test is still missing systemd headers (`cargo clippy --lib --bins --examples --all-features` on my machine worked fine).

yaleman:
Er welp, that was supposed to go to my account not your PR, but hopefully that sorts clippy 😄

tacerus:
Thanks! I missed the other workflow file.

yaleman:
@Firstyear if you can give this a second look (being much more the systemd ninja) then this should be mergeable 😄

yaleman:
I promise I'll learn to computers one day

tacerus:
Careful with promising things beyond what a human is capable of

tacerus:
I think it was missing from the Dockerfile as well, amended that now.

yaleman (approved):

Firstyear:
So I think the PR is good and you've done good work here, but I am beyond over systemd and this kind of half baked behaviour that they keep pushing. 

The docs clearly state:

https://www.freedesktop.org/software/systemd/man/latest/sd_notify.html

"Tells the service manager that the service is beginning to reload its configuration. This is useful to allow the service manager to track the service's internal state, and present it to the user. Note that a service that sends this notification must also send a "READY=1" notification when it completed reloading its configuration. Reloads the service manager is notified about with this mechanisms are propagated in the same way as they are when originally initiated through the service manager. This message is particularly relevant for Type=notify-reload services, to inform the service manager that the request to reload the service has been received and is now being processed."

This was added in version 217.

Then later we see:

"A field carrying the monotonic timestamp (as per CLOCK_MONOTONIC) formatted in decimal in μs, when the notification message was generated by the client. This is typically used in combination with "RELOADING=1", to allow the service manager to properly synchronize reload cycles. See [systemd.service(5)](https://www.freedesktop.org/software/systemd/man/latest/systemd.service.html#) for details, specifically "Type=notify-reload"."

Added in version 253.

So effectively from 217 reloading worked, and then after 253, systemd *regressed and broke reloading notifications* for any compliant user who followed the docs, then didn't even do the basics to actually update the docs around RELOADING=1 to say "oh yeah, if you don't use MONOTONIC_USEC then everything breaks.

**This means we are likely not the only service affected by this regression.**

IMO we should:

1)  raise a bug in systemd that they have regressed, and monotonic_usec should not be needed. 
2) If systemd refuse to fix it (which is the likely outcome), then we fix the crate https://crates.io/crates/sd-notify since this has 5 million downloads, meaning this fix should not just be limited to us and needs to be available to the broader ecosystem.

Firstyear:
I have submitted a pr to sd-notify for this.

server/daemon/src/main.rs
```diff
@@ -785,18 +792,33 @@ async fn kanidm_main(
...
+                                            Ok(val) => val,
+                                            Err(err) => {
+                                                error!(
+                                                  "Failed to get UNIX_EPOCH: {:?}",
+                                                  err
+                                                );
+                                                return ExitCode::FAILURE;
+                                            }
+                                        };
+                                        let timestamp_nanos = duration_since_epoch.as_nanos().to_string();
```
Firstyear, line 806: Should be microseconds not nanos btw.

tacerus:
Thanks for the improvement to sd-notify! I agree the documentation made this change not very obvious, I only accidentally found about it upon finding a similar issue in `named`: https://gitlab.isc.org/isc-projects/bind9/-/issues/4377.

server/daemon/src/main.rs
tacerus: Corrected, thanks.

Firstyear:
I should have said @tacerus I don't want to link to libsystemd, so I want to fix this in sd-notify.

Firstyear:
Which as a reminder, linking to libsystemd is what caused the xz backdoor to occur.

.github/workflows/clippy.yml
```diff
@@ -27,11 +27,13 @@ jobs:
           sudo apt-get update && \
           sudo apt-get install -y \
             libpam0g-dev \
-            libudev-dev \
+            libselinux1-dev \
             libssl-dev \
+            libsystemd-dev \
```
Firstyear, line 32: I think we don't need this now? But I know there is some relationship to udev sometimes ....

tacerus:
I know, I was also not a fan of the new linking to libsystemd for that reason, if we can do without it and keep sd-notify that's of course much better. :)

Firstyear:
> I know, I was also not a fan of the new linking to libsystemd for that reason, if we can do without it and keep sd-notify that's of course much better. :)

Already on it :)

.github/workflows/clippy.yml
tacerus: I think before it built without it, so it might not be needed anymore.

server/daemon/src/main.rs
```diff
@@ -760,66 +774,86 @@ async fn kanidm_main(
...
+                                    error!("CRITICAL!!! Unable to access clock monotonic time. SYSTEMD WILL KILL US.");
+                                };
+                                                    let _ = sd_notify::notify(true, &[sd_notify::NotifyState::Status("Reloading ...")]);
+                                                    }
+
+                                                    sctx.tls_acceptor_reload().await;
+
+                                                    // Systemd freaks out if you send the ready state too fast after the
+                                                    // reload state and can kill Kanidmd as a result.
+                                                    tokio::time::sleep(std::time::Duration::from_secs(5)).await;
```
tacerus, line 819: Is this really needed again?

yaleman:
> Which as a reminder, linking to libsystemd is what caused the xz backdoor to occur.

Eeehhh that was more about including big binary “test” blobs in the m4? build script.

Firstyear:
> > Which as a reminder, linking to libsystemd is what caused the xz backdoor to occur.
> 
> Eeehhh that was more about including big binary “test” blobs in the m4? build script.

The binary blob in the test, was only activated *because* openssh linked to libsystemd which then triggeder liblzma to be in the process memory so that the blob could trigger.

libsystemd brings in *a lot* of garbage.

lnicola:
> libsystemd brings in a lot of garbage.

It no longer does, see https://mastodon.social/@pid_eins/112256363180973672. It's now using `dlopen` when it needs something:

```
libsystemd.so => /usr/lib64/libsystemd.so (interpreter => none)
    libcap.so.2 => /lib64/libcap.so.2
    libgcc_s.so.1 => /lib64/libgcc_s.so.1
    libc.so.6 => /lib64/libc.so.6
        ld-linux-x86-64.so.2 => /lib64/ld-linux-x86-64.so.2
```

Firstyear (approved):

yaleman:
> > libsystemd brings in a lot of garbage.
> 
> It no longer does, see https://mastodon.social/@pid_eins/112256363180973672. It's now using `dlopen` when it needs something:
> 
> ```
> libsystemd.so => /usr/lib64/libsystemd.so (interpreter => none)
>     libcap.so.2 => /lib64/libcap.so.2
>     libgcc_s.so.1 => /lib64/libgcc_s.so.1
>     libc.so.6 => /lib64/libc.so.6
>         ld-linux-x86-64.so.2 => /lib64/ld-linux-x86-64.so.2
> ```

That’s WHY the xz thing got pushed forward by the threat actor, they were going to lose their access 😅

PR #3359, opened 2025-01-14
20250114 3325 SCIM access control

### LINKED ISSUE #3325: SCIM Access Control Policy support
Currently 
 - https://github.com/kanidm/kanidm/pull/2994
 - https://github.com/kanidm/kanidm/pull/3019
Are waiting for SCIM ACP so they don't need to be refactored later.

Iirc it should allow us to use a standard api to get info on whether or not access is available to certain actions.
E.g. 
 - changing you email address. 
 - editing others users their email address
 - creating groups
 ...

/

Firstyear:
# Change summary

- Adds support for querying effective access on entries. This is needed for our UI development so that forms can correctly enable or disable fields based on your permissions, so that you can't edit something that will fail to commit for example.

Fixes #3325

Checklist


Firstyear:
@ToxicMushroom The big question here is the "shape" of the effective access. If you have a look, right now it's "if I am to do a search, what attributes can I see". https://github.com/kanidm/kanidm/pull/3359/files#diff-abd0e43d3bb132b22bea3b0871568b3674bb4c8cb6ea47519b29d5109265626eR51

We can invert this so that it's "for this attribute, can I search?"

An example of how this would look if inverted is:

```
attribute_access: BTreeMap<Attribute, EffectivePermission>;

struct EffectivePermission {
    search: bool,
    modify_present: bool,
    modify_remove: bool,
}
```

So the difference is how you check the effective access basicly - in the "inverted" method you query an attribute, and it says "you can search but not modify this". In the other way you are asking "I want to search, the attribute named X, can I do that?".

I'm happy to do the inversion, I think it might make ui dev a bit easier wrt to rendering forms. What do you think?

ToxicMushroom (changes requested):
Seems to work well (after I remembered that reauth was a thing)
image

ToxicMushroom:
I don't think the inversion is very useful, getting entries from a map means I must handle the case of them not being present.

If we were to add helper functions I'd prefer something like `self_access.can_modify(attribute)` which just forwards it to `self_access.modify_present.check(&attribute)`

here is an askama example that lists a couple attributes and their modifiability using the current system:
```html
    (% for attr in [Attribute::Name, Attribute::DisplayName, Attribute::Mail, Attribute::Uuid] %)
        (% if permissions.modify_present.check(attr|as_ref) %)
            <div class="mb-2 row">Can modify (( attr ))</div>
        (% else %)
            <div class="mb-2 row">Cannot modify (( attr ))</div>
        (% endif %)
    (% endfor %)
```

Firstyear:
> I think the inversion would work well in forms/complex ui's where we need to show a lot of conditional things based on viewing/modify/delete permissions yes :) . But it'd also need some helpers since I don't want to be handling Option<> when getting values from the map.
> 
> If we were to add helper functions, I'd also like something like `self_access.can_modify(attribute)` which just forwards it to `self_access.modify_present.check(&attribute)` ect.
> 
> here is an askama example that lists a couple attributes and their modifiability using the current system:
> 
> ```
>     (% for attr in [Attribute::Name, Attribute::DisplayName, Attribute::Mail, Attribute::Uuid] %)
>         (% if permissions.modify_present.check(attr|as_ref) %)
>             <div class="mb-2 row">Can modify (( attr ))</div>
>         (% else %)
>             <div class="mb-2 row">Cannot modify (( attr ))</div>
>         (% endif %)
>     (% endfor %)
> ```

We can always also just add helpers that can return the inverted form too for any attribute you query. So maybe that solves that problem.

server/lib/src/server/access/mod.rs
```diff
@@ -318,33 +345,61 @@ pub trait AccessControlsTransaction<'a> {
...
-            // If the acp doesn't overlap with our requested attrs, there is no point in
-            // testing it!
-            related_acp
-                .into_iter()
-                .filter(|acs| !acs.acp.attrs.is_disjoint(r_attrs))
-                .collect()
-        } else {
-            related_acp
-        };
+        let search_related_acp = self.search_related_acp(&se.ident, se.attrs.as_ref());
```
CEbbinghaus, line 378: since this isn't used until line 402 wouldn't it make sense to move it under line 396? Because it currently gives the impression that it relates to the block in 380 which it does not.

server/lib/src/server/access/mod.rs
```diff
@@ -892,6 +911,57 @@ pub trait AccessControlsTransaction<'a> {
...
+                ModifyResult::Grant => (Access::Grant, Access::Grant, AccessClass::Grant),
+                ModifyResult::Allow { pres, rem, cls } => (
+                    Access::Allow(pres.into_iter().collect()),
+                    Access::Allow(rem.into_iter().collect()),
+                    AccessClass::Allow(cls.into_iter().map(|s| s.into()).collect()),
+                ),
+            };
+
+        // == delete ==
+        let delete_status = apply_delete_access(ident, delete_related_acp, entry);
```
CEbbinghaus, line 948: any reason we assign delete_status to a variable rather than matching on the method result? or is this just trying to keep in line with the above variables?

CEbbinghaus (dismissed):
Looks good, Purely stylistic feedback. 
<sub>p.s sorry it took me so long to get to the review</sub>

server/lib/src/server/access/mod.rs
```diff
@@ -79,7 +92,7 @@ pub struct AccessEffectivePermission {
 pub enum AccessResult {
     // Deny this operation unconditionally.
     Denied,
-    // Unbounded allow, provided no denied exists.
+    // Unbounded allow, pro -vided no denied exists.
```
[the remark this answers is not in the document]
Firstyear: Yes, new keyboard and I keep making mistakes without noticing

server/lib/src/server/access/mod.rs
Firstyear: Technically it doesn't matter either way, since there is no data dependence, the compiler can move it above or below this, but I'll move it.

server/lib/src/server/access/mod.rs
Firstyear: It's because match apply_delete_access(...) gets pretty long, this just helps to break it up visually.

server/lib/src/server/access/mod.rs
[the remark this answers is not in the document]
CEbbinghaus: Thought so :P

CEbbinghaus (approved):
Lovely, Appreciate it <3

PR #3362, opened 2025-01-17
Added shell.nix to create dev environment

CEbbinghaus:
# Change summary
Added shell.nix which allows for a development environment to be created by running `nix-shell` in the root directory.

Checklist

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -1,0 +1,53 @@
+{ pkgs ? import <nixpkgs> {} }:
+	let
+	overrides = (builtins.fromTOML (builtins.readFile ./rust-toolchain.toml));
+	libPath = with pkgs; lib.makeLibraryPath [
+		# load external libraries that you need in your rust project here
+	];
+in
+pkgs.mkShell rec {
+	# Kanidm dependencies
+	buildInputs = with pkgs; [
+		pkg-config
+		
+		git
+		rustup
+
+		clang
+		llvmPackages.bintools
+		
+		openssl
+		systemd
+		linux-pam
+	];
+	
+	RUSTC_VERSION = overrides.toolchain.channel;
+	# https://github.com/rust-lang/rust-bindgen#environment-variables
+	LIBCLANG_PATH = pkgs.lib.makeLibraryPath [ pkgs.llvmPackages_latest.libclang.lib ];
+	shellHook = ''
+		export PATH=$PATH:''${CARGO_HOME:-~/.cargo}/bin
+		export PATH=$PATH:''${RUSTUP_HOME:-~/.rustup}/toolchains/$RUSTC_VERSION-x86_64-unknown-linux-gnu/bin/
+		'';
+
+	# Add precompiled library to rustc search path
+	RUSTFLAGS = (builtins.map (a: ''-L ${a}/lib'') [
+		# add libraries here (e.g. pkgs.libvmi)
+	]);
+	LD_LIBRARY_PATH = libPath;
+	PKG_CONFIG_PATH = "${pkgs.openssl.dev}/lib/pkgconfig";
+
+	# Add glibc, clang, glib, and other headers to bindgen search path
+	BINDGEN_EXTRA_CLANG_ARGS =
+	# Includes normal include path
+	(builtins.map (a: ''-I"${a}/include"'') [
+		# Dev Libraries
+		pkgs.glibc.dev
+		pkgs.openssl.dev
+	])
+	# Includes with special directory paths
+	++ [
+		''-I"${pkgs.llvmPackages_latest.libclang.lib}/lib/clang/${pkgs.llvmPackages_latest.libclang.version}/include"''
+		''-I"${pkgs.glib.dev}/include/glib-2.0"''
+		''-I${pkgs.glib.out}/lib/glib-2.0/include/''
+	];
+}
--- a/shell.nix
+++ b/shell.nix
@@ -1,22 +1,20 @@
 { pkgs ? import <nixpkgs> {} }:
 	let
 	overrides = (builtins.fromTOML (builtins.readFile ./rust-toolchain.toml));
-	libPath = with pkgs; lib.makeLibraryPath [
-		# load external libraries that you need in your rust project here
-	];
 in
-pkgs.mkShell rec {
+pkgs.mkShellNoCC rec {
 	# Kanidm dependencies
 	buildInputs = with pkgs; [
 		pkg-config
 		
-		git
-		rustup
+		cargo
+		rustc
 
 		clang
 		llvmPackages.bintools
 		
 		openssl
+	] ++ pkgs.lib.optionals (pkgs.stdenv.isLinux) [
 		systemd
 		linux-pam
 	];
@@ -24,30 +22,4 @@ pkgs.mkShell rec {
 	RUSTC_VERSION = overrides.toolchain.channel;
 	# https://github.com/rust-lang/rust-bindgen#environment-variables
 	LIBCLANG_PATH = pkgs.lib.makeLibraryPath [ pkgs.llvmPackages_latest.libclang.lib ];
-	shellHook = ''
-		export PATH=$PATH:''${CARGO_HOME:-~/.cargo}/bin
-		export PATH=$PATH:''${RUSTUP_HOME:-~/.rustup}/toolchains/$RUSTC_VERSION-x86_64-unknown-linux-gnu/bin/
-		'';
-
-	# Add precompiled library to rustc search path
-	RUSTFLAGS = (builtins.map (a: ''-L ${a}/lib'') [
-		# add libraries here (e.g. pkgs.libvmi)
-	]);
-	LD_LIBRARY_PATH = libPath;
-	PKG_CONFIG_PATH = "${pkgs.openssl.dev}/lib/pkgconfig";
-
-	# Add glibc, clang, glib, and other headers to bindgen search path
-	BINDGEN_EXTRA_CLANG_ARGS =
-	# Includes normal include path
-	(builtins.map (a: ''-I"${a}/include"'') [
-		# Dev Libraries
-		pkgs.glibc.dev
-		pkgs.openssl.dev
-	])
-	# Includes with special directory paths
-	++ [
-		''-I"${pkgs.llvmPackages_latest.libclang.lib}/lib/clang/${pkgs.llvmPackages_latest.libclang.version}/include"''
-		''-I"${pkgs.glib.dev}/include/glib-2.0"''
-		''-I${pkgs.glib.out}/lib/glib-2.0/include/''
-	];
 }
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,2 +1,4 @@
 # yale-mistakes were made
 /pykanidm/* @yaleman
+# Least qualified nix guy :P
+shell.nix @cebbinghaus
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,0 +1,2 @@
+[toolchain]
+channel = "stable"


shell.nix
```diff
@@ -0,0 +1,53 @@
...
+		systemd
+		linux-pam
+	];
+	
+	RUSTC_VERSION = overrides.toolchain.channel;
+	# https://github.com/rust-lang/rust-bindgen#environment-variables
+	LIBCLANG_PATH = pkgs.lib.makeLibraryPath [ pkgs.llvmPackages_latest.libclang.lib ];
+	shellHook = ''
+		export PATH=$PATH:''${CARGO_HOME:-~/.cargo}/bin
+		export PATH=$PATH:''${RUSTUP_HOME:-~/.rustup}/toolchains/$RUSTC_VERSION-x86_64-unknown-linux-gnu/bin/
```
yaleman, line 29: Does it really need this arch-specific path? That seems … bad

yaleman (commented):
Please add yourself to the codeowners file for this one 😊

/me resets the timer

shell.nix
```diff
@@ -0,0 +1,53 @@
...
+		# load external libraries that you need in your rust project here
+	];
+in
+pkgs.mkShell rec {
+	# Kanidm dependencies
+	buildInputs = with pkgs; [
+		pkg-config
+		
+		git
+		rustup
```
2xsaiko, line 14: Why? Either use user's preinstalled rustup and don't do anything here, or add rustc and cargo here to use the compiler from nixpkgs. I'd do the latter because that way everything is managed by Nix.

Those two variants also don't require you to mess with PATH like that.

shell.nix
```diff
@@ -0,0 +1,53 @@
+{ pkgs ? import <nixpkgs> {} }:
+	let
+	overrides = (builtins.fromTOML (builtins.readFile ./rust-toolchain.toml));
+	libPath = with pkgs; lib.makeLibraryPath [
+		# load external libraries that you need in your rust project here
+	];
```
2xsaiko, line 6: Is this copied from some template? It does nothing, you can remove it

shell.nix
```diff
@@ -0,0 +1,53 @@
...
+	LIBCLANG_PATH = pkgs.lib.makeLibraryPath [ pkgs.llvmPackages_latest.libclang.lib ];
+	shellHook = ''
+		export PATH=$PATH:''${CARGO_HOME:-~/.cargo}/bin
+		export PATH=$PATH:''${RUSTUP_HOME:-~/.rustup}/toolchains/$RUSTC_VERSION-x86_64-unknown-linux-gnu/bin/
+		'';
+
+	# Add precompiled library to rustc search path
+	RUSTFLAGS = (builtins.map (a: ''-L ${a}/lib'') [
+		# add libraries here (e.g. pkgs.libvmi)
+	]);
```
2xsaiko, line 35: This also does nothing

shell.nix
```diff
@@ -0,0 +1,53 @@
...
+	LD_LIBRARY_PATH = libPath;
+	PKG_CONFIG_PATH = "${pkgs.openssl.dev}/lib/pkgconfig";
```
2xsaiko, line 37: You shouldn't and shouldn't need to override PKG_CONFIG_PATH like that. You might need to add openssl.dev to buildInputs instead if just openssl doesn't work

shell.nix
```diff
@@ -0,0 +1,53 @@
...
+		# Dev Libraries
+		pkgs.glibc.dev
+		pkgs.openssl.dev
+	])
+	# Includes with special directory paths
+	++ [
+		''-I"${pkgs.llvmPackages_latest.libclang.lib}/lib/clang/${pkgs.llvmPackages_latest.libclang.version}/include"''
+		''-I"${pkgs.glib.dev}/include/glib-2.0"''
+		''-I${pkgs.glib.out}/lib/glib-2.0/include/''
+	];
```
2xsaiko, line 52: This is also only correct for Linux (glibc and glib). The [nixpkgs package](https://github.com/NixOS/nixpkgs/blob/nixos-unstable/pkgs/by-name/ka/kanidm/generic.nix) also doesn't have this at all, is this only recently necessary?

shell.nix
```diff
@@ -0,0 +1,53 @@
...
+	libPath = with pkgs; lib.makeLibraryPath [
+		# load external libraries that you need in your rust project here
+	];
+in
+pkgs.mkShell rec {
+	# Kanidm dependencies
+	buildInputs = with pkgs; [
+		pkg-config
+		
+		git
```
2xsaiko, line 13: Why replace the user's git?

shell.nix
```diff
@@ -0,0 +1,53 @@
...
+		rustup
+
+		clang
+		llvmPackages.bintools
+		
+		openssl
+		systemd
+		linux-pam
+	];
```
2xsaiko, line 22:
```suggestion
    ] ++ pkgs.lib.optionals (pkgs.stdenv.isLinux) [
		systemd
		linux-pam
	];
```

Only valid on Linux

shell.nix
2xsaiko: That is probably pkgs.stdenv.targetPlatform.rust.rustcTarget but this is bad regardless, see comment about rustup

2xsaiko (changes requested):

mart-w:
I’d suggest also adding a simple `.envrc` to allow users of [Direnv](https://direnv.net/) to switch nix shell environments automatically. It only takes  a simple text file called `.envrc` in the project root containing a single line:

```
use nix;
```

Direnv will then pick up on that automatically and work its magic.

yaleman:
> I’d suggest also adding a simple `.envrc` to allow users of [Direnv](https://direnv.net/) to switch nix shell environments automatically. It only takes a simple text file called `.envrc` in the project root containing a single line:
> 
> ```
> use nix;
> ```
> 
> Direnv will then pick up on that automatically and work its magic.
> 
> Edit: You’ll also have to add `.direnv/` to `.gitignore` as that’s where Direnv keeps its project-specific state.

That blasts every other use of direnv or other state managers, it’s not only used by nix folks, and it’s why `.envrc` is already in the gitignore. Please leave it that way.

Firstyear:
@mweinelt I think you may be qualified to help review this too :)

shell.nix
```diff
@@ -0,0 +1,53 @@
+{ pkgs ? import <nixpkgs> {} }:
```
mweinelt, line 1: This ties the versioning of everything to the locally configured nixpkgs channel. For different users on different channel versions this will lead to different outcomes. The alternative would be to include a shared lock file and use e.g. a flake devshell instead.

shell.nix
```diff
@@ -0,0 +1,53 @@
...
+	let
+	overrides = (builtins.fromTOML (builtins.readFile ./rust-toolchain.toml));
+	libPath = with pkgs; lib.makeLibraryPath [
+		# load external libraries that you need in your rust project here
+	];
+in
+pkgs.mkShell rec {
```
mweinelt, line 8:
```suggestion
pkgs.mkShellNoCC rec {
```

Smaller footprint, if you don't need a C compiler.

yaleman: It's still needed for the c dependencies like OpenSSL and friends isn't it?

mweinelt: Maybe, if in doubt just try and error.

2xsaiko: Rust also needs ld and that isn't in the no-cc shell, right? I recall having problems with something like that I think.

CEbbinghaus:
Whoo boy, that's a lot of feedback. Probably won't get to all of it tonight but I wanted to say a huge thank you for all the comments.

shell.nix
CEbbinghaus: Rustup is the idiomatic way to work with Rust in nixos. We are able to include RustC & cargo at a specific version using nix but that removes control for the user to be able to run a different toolchain. I thought it best to stick with the standard and better documented case of just using rustup. Happy to do more investigation though if we think that it will be cleaner in the end doing it ourselves.

shell.nix
CEbbinghaus: indeed it is. Could remove it but I thought to leave it to make future contributions and fixes easier (don't have to go searching for how to do X). Equally happy to remove it since functionally it doesn't do anything (yet)

shell.nix
CEbbinghaus: I'll give that a go but when I was trying to get this shell to work it started working after including this ENV variable so I'll have to test without it to make sure it was unrelated.

shell.nix
CEbbinghaus: can BSD even run nix. This shell.nix only makes sense within the nix context and afaik that focuses on linux.

shell.nix
2xsaiko: aarch64-* is probably more popular than x86 at this point (at least nixpkgs has better support for it). But yeah

shell.nix
2xsaiko: I don't know, usually I see people use either nixpkgs rustc, or fenix/rust-overlay. All three of these are options that "just work" (just need to run nix develop/nix-shell), and only place files in the Nix store.

Mostly it feels wrong to me to use Nix to just install an updater program instead of getting the real compiler straight away, I suppose.

If someone does want to use a different version, they can always comment it out here locally or run another inner nix shell or whatever.

shell.nix
2xsaiko:
```
nix-repl> :p builtins.readFile (runCommand "" {buildInputs = [pkg-config openssl];} "exec > $out; echo $PKG_CONFIG_PATH_FOR_TARGET; pkg-config --libs openssl")
/nix/store/y940g0f3as2crh7frkb4p9f99wxzyzkm-openssl-3.3.2-dev/lib/pkgconfig
-L/nix/store/1a4j0c5xn6xl53z77sifkncg3fing1qj-openssl-3.3.2/lib -lssl -lcrypto
```

It should be setting PKG_CONFIG_PATH_FOR_TARGET (see https://github.com/NixOS/nixpkgs/issues/64530#issuecomment-1030313034).
If it's not working, what is reading this env var if not pkg-config that isn't respecting this?

shell.nix
2xsaiko: The big other platform nixpkgs supports (and probably the only one it's feasible to do Rust dev with Nix on right now tbh) which is pretty popular is macOS.

shell.nix
2xsaiko: If it's not used for building, then I would remove it here. If someone downloaded a tarball release and doesn't already have git preinstalled, they won't need it anyway. I thought it might at least be used for getting the version from a git tag or something.

shell.nix
CEbbinghaus: Ah, I didn't even realize nix ran on macos. That is indeed a consideration. I feel like the main demographic is still linux so if we have the choice between it working and being cross platform I'll take the former. But if we can get it working without limiting it to only linux then all the better

shell.nix
CEbbinghaus: The shell uses the same version as the system git as long as its resolved to the same version. Removing it in any case

shell.nix
mweinelt: The following is a `flake.nix`, that consumes

- nixpkgs
- git-hooks.nix (for pre-commit hooks, optional)
- flake-utils (for the `eachSystem` helper)

When running `nix flake update --commit-lock-file` the `./flake.lock` file will get updated and pinned to the latest versions of these repositories.

In the `devShell` you'll see `mkShell` and you can use it like you already do in the `shell.nix`.

To use that you do need to have flakes and the nix-command features enabled, which [this installer](https://github.com/DeterminateSystems/nix-installer) does by default, so I would recommend documenting that one.

```nix
{
  description = "Kanidm development environment";

  inputs = {
    nixpkgs.url = "github:NixOS/nixpkgs/nixpkgs-unstable";
    git-hooks = {
      url = "github:cachix/git-hooks.nix";
      inputs.nixpkgs.follows = "nixpkgs";
    };
    flake-utils.url = "github:numtide/flake-utils";
  };

  outputs =
    {
      self,
      nixpkgs,
      git-hooks,
      flake-utils,
    }:

    flake-utils.lib.eachSystem
      [
        "aarch64-linux"
        "aarch64-darwin"
        "x86_64-linux"
      ]
      (
        system:
        let
          pkgs = nixpkgs.legacyPackages.${system};
          lib = pkgs.lib;
        in
        {
          checks = {
            pre-commit = git-hooks.lib.${system}.run {
              src = ./.;
              hooks = {
                # https://github.com/cachix/git-hooks.nix#rust
              };
            };
          };
          devShells.default =
            with pkgs;
            mkShell {
              buildInputs = [
                # TODO: dependencies here
              ];

              env = {
                # TODO: any environment variables here
                FOO = "bar";
              };

              shellHook = ''
                # TODO: any shell code that gets executed on load
              '' + self.checks.${system}.pre-commit.shellHook;
            };

          formatter = pkgs.nixfmt-rfc-style;
        }
      );
}
```

shell.nix
CEbbinghaus: @yaleman Any preference re keeping it or removing it?

shell.nix
CEbbinghaus: I don't think we can build the pam module on mac? I'll fix it up but running `cargo build` at the root would fail on macos because it relies on linux-pam

shell.nix
yaleman: Meh, leave it until the rest of the bike-shedding in this PR is done, if it’s not doing anything at that time it can be removed.

shell.nix
yaleman: It’ll be ignored, there’s a bunch of work we did to make this OK.

yaleman: It’ll be ignored, there’s a bunch of work we did to make this OK.

yaleman: But also don’t try and install Linux PAM support libs on macOS 😳

shell.nix
CEbbinghaus: hmmm, @Firstyear what is your opinion on this? Personally I am happy with either and as it is right now it works. In theory all packages within a particular version should be compatible so this is only really there to pin packages between different developers so there is no problems with it working on one machine but not another. This however comes at the cost of us pinning the version ourselves which means we need to regularly update the lock to ensure we are on a recent version. This feel like a lot of maintenance for something very few people will use, And something that we already have to deal with and hasn't provided massive problems thus far (as far as I can tell).

It does increase the complexity by a non insignificant amount (especially compared to the version I managed to whittle it down to.

shell.nix
CEbbinghaus: that is what this suggestion is for. Implemented it since it just makes sense in any case

shell.nix
CEbbinghaus: This has been removed entirely

shell.nix
CEbbinghaus: Changed to cargo & rustc

shell.nix
CEbbinghaus: This has all been removed too

shell.nix
CEbbinghaus: Tested it. Works with NoCC so I changed to that

2xsaiko (approved):

shell.nix
yaleman: > The alternative would be to include a shared lock file and use e.g. a flake devshell instead.

nooooooooo not the cult of reproducible builds. doesn’t nix already have a dev mode for the packages if you’re getting to that point?

mweinelt: Neither of these approaches provide [reproducible builds](https://reproducible-builds.org/). The only choice is between a floating nixpkgs reference vs. a locked one, with which you can return to the exact previous version of the build environment.

Both approaches do use `pkgs.mkShellNoCC`, resulting in a shell environment with some dependencies made available and environment variables set. This is, in my understanding, a development environment, and unless you want to expose a Kanidm package in the `flake.nix` there is no dev/prod code to share.

CEbbinghaus: I'm gonna go ahead and vote strongly for keeping as is. As much as I agree that "reproducible builds" (same version between devs) is a good thing, it's also something that Rust does much better than most other programming toolchains. It's easier to once a year say "you need to update your nix packages" than maintaining a lock file by updating it every month or so for the 1 guy that uses this nix shell.

Its functional and the packages will work, only a matter of how old they are which we already have to deal with, given people currently install the toolchain via rustup by hand.

yaleman (approved):

PR #3365, opened 2025-01-19
Book: Added small section on primary cred fallback

CEbbinghaus:
# Change summary
Forgot to add a book chapter when raising the PR for the primary credential feedback. This makes good on that mistake.

Amends #3067

Checklist

### DIFF
--- a/book/src/accounts/account_policy.md
+++ b/book/src/accounts/account_policy.md
@@ -193,7 +193,7 @@ kanidm group account-policy webauthn-attestation-ca-list idm_all_persons trusted
 The primary credential fallback enables behavior which allows authenticating
 using the primary account password when logging in via LDAP.
 
-If both a LDAP and primary password are specified it will only accept the LDAP password.
+If both an LDAP and primary password are specified, Kanidm will only accept the LDAP password.
 
 ```bash
 kanidm group account-policy allow-primary-cred-fallback <group name> <enabled>
--- a/book/src/accounts/account_policy.md
+++ b/book/src/accounts/account_policy.md
@@ -188,6 +188,17 @@ account policy for a group. For example, to set the allowlist for all persons, r
 kanidm group account-policy webauthn-attestation-ca-list idm_all_persons trusted-authenticators
 ```
 
+### Setting Primary Credential Fallback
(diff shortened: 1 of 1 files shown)


book/src/accounts/account_policy.md
```diff
@@ -188,6 +188,17 @@ account policy for a group. For example, to set the allowlist for all persons, r
 kanidm group account-policy webauthn-attestation-ca-list idm_all_persons trusted-authenticators
 ```
 
+### Setting Primary Credential Fallback
+
+The primary credential fallback enables behavior which allows authenticating
+using the primary account password when logging in via LDAP.
+
+If both a LDAP and primary password are specified it will only accept the LDAP password.
```
yaleman, line 196:
```suggestion
If both an LDAP and primary password are specified, Kanidm will only accept the LDAP password.
```

book/src/accounts/account_policy.md
```diff
@@ -188,6 +188,17 @@ account policy for a group. For example, to set the allowlist for all persons, r
...
+
+```bash
```
yaleman, line 198: Is there "disable" option? How does one complete that step?

yaleman (changes requested):

book/src/accounts/account_policy.md
CEbbinghaus: Default is set to false. If its enabled and it should be disabled, just running:
`kanidm group account-policy allow-primary-cred-fallback <group name> false` will work.

Firstyear (approved):

book/src/accounts/account_policy.md
CEbbinghaus: No stress, can do

CEbbinghaus: @yaleman sorry for the delay. (Got very busy with work). This good?

yaleman: No worries, I'm on holiday and this is a fun thing not super urgent 😃 thanks for the update

yaleman (approved):

yaleman (approved):

PR #3380, opened 2025-01-29
Docker images

JadedBlueEyes:
# Change summary

- This changes the dockerfiles to be built on a "scratch" image, making images ~15% smaller (the vast majority of the size is the kanidmd binary)
- Adds support for cross-compiling images (tested both amd64->arm64 and arm64->amd64), although this is not built in the pipeline yet
- Adds SBOMs to built images, to enable dependency scanning of built images
- Binaries are now built with cross-language LTO support
- Rust build artefacts are cached using cache mounts, speeding up local builds

- Pipelines now all correctly set labels and annotations
- Tags remain unchanged
- Pipeline builds now set a fixed timestamp based on the commit to ease reproducibility
- Docker build layers are cached to the github actions cache
- Cache mounts are automatically swapped in and out of container images as a hack to cache rust builds
  - Unfortunately, due to the large size of these directories and the slow final compile, this only seems to save ~5m on average. 
- Pipelines will now run in forked repos

- Deleted an outdated and unused build script

Future improvements:
- Cross-compile arm versions of the image in the pipeline
- Possibly use GitHub's build attestations (although this would tie the pipeline to GitHub's infrastructure)
- Test the pipeline with Forgejo Actions
- Use the docker-metadata-action tagging system to allow uploading and tagging images on different branches, and per-commit tags 
- Add similar labels / annotations to the make-based release process, or move the release process to the pipeline
- Add a renovate configuration to allow auto-updating the versions in the dockerfile
- Build OpenSSL from source (like sqlite is) to take advantage of the cross-language LTO (Possibly add a build feature both are controlled by?)
- Cache mounts should have IDs based on the target arch and RUSTFLAGS to reduce cache thrashing when switching between target configurations on the local machine
- Build artefacts for the separated fido-mds-tool build should be cached in Github Actions
- If possible, stop cargo from invalidating unchanged crates in the workspace in each Github Actions run.
- Don't copy build artefacts *out* of the image when the rust cache action doesn't want them - this could save a couple minutes as it seems to be one of the slower steps.

I haven't touched the Orca image. This could be very similar to the tools image, or it could be merged into it?

Checklist


JadedBlueEyes:
I've squished the commits, makes the history a bit more manageable

jinnatar:
As a drive-by comment, it might be beneficial for the original author(s) to summarize what the driving motivation is for these changes to make it easier to judge the added complexity etc against the merits of said motivations.

At a glance I'm reading on and between the lines the following and their potential drawbacks:
1. Smaller absolute image size. (Weigh against layer cached total size, complexity)
2. SBOMs for vuln scanning. (Weigh against extra dependencies)
3. Faster build time on GitHub Actions. (Weigh against complexity, practical effects on dev velocity)
4. Steps towards/achieving reproducible builds. (Weigh against complexity, extra dependencies, long term maintenance load)

Any motivations I missed or misunderstood would be good to add and your take on the caveats introduced by your solutions. I'm personally not against any of the motivations I inferred above, just conscious about being able to make informed choices and changes for the long term.

JadedBlueEyes:
> As a drive-by comment, it might be beneficial for the original author(s) to summarize what the driving motivation is for these changes to make it easier to judge the added complexity etc against the merits of said motivations.

That's a pretty good idea. Some of this is buried in the matrix room a bit, but this is relying on knowledge I've built up over months and I forget that a lot of it needs explaining 😅

> At a glance I'm reading on and between the lines the following and their potential drawbacks:
> 1. Smaller absolute image size. (Weigh against layer cached total size, complexity)
> 2. SBOMs for vuln scanning. (Weigh against extra dependencies)
> 3. Faster build time on GitHub Actions. (Weigh against complexity, practical effects on dev velocity)
> 4. Steps towards/achieving reproducible builds. (Weigh against complexity, extra dependencies, long term maintenance load)

You've pretty much got the gist of it here! Some things I'll add:

- Cross compilation is driving a lot of the complexity here, but it makes having non-x86 images possible (on another project, emulated cross compiles took over 3 hours in GitHub Actions from my tests)
- Most of the image size is now directly from the binaries - all the other stuff in total is probably under a megabyte. I would hope docker deduplicates these layers properly, but it's small potatoes.
- that is to say, we're now shipping iirc
  1. The binaries themselves
  2. Libraries that they link (and the dynamic linker)
  3. SBOMs
  4. Root certificates 
  5. The web interface dist
- SBOMs are essentially free, and pretty nice to have. The dependency can be swapped out, there are a couple other options (originally I used cargo-auditable, but that had issues with cross compiles) or it can be easily removed if issues are found.
- the CI caching comes in two parts. Instructing BuildKit to use the GitHub actions cache, which is a couple lines and quite nice, and the 'cache dance' stuff used for caching the rust build artifacts. The second one is a hack due to a missing feature in BuildKit that I hope to be able to remove, and looks pretty ugly in the pipelines, but doesn't have that much code impact otherwise.
- The cache mounts in the dockerfile help *a lot* for local builds, without adding much to the complexity.
- the 'steps towards achieving reproducible builds' are just using a documented environment variable to set a specific timestamp. I'm not sure it does much, but it doesn't take that much and it's and officially documented build option is as far as my justification goes.

Thanks for taking a look at it :)

yaleman:
Thanks for putting this in, I am going to look over it and poke around once I have some spare cycles. It won't be _soon_ but I am definitely liking the look of things so far 😊

Firstyear:
> Cross compilation is driving a lot of the complexity here, but it makes having non-x86 images possible (on another project, emulated cross compiles took over 3 hours in GitHub Actions from my tests)

The only catch here is that we only support x86_64 and aarch64 though, mostly because we want to be able to test on native machines if an issue arises. So this doesn't seem hugely important today.

yaleman:
Aarch64's the cross-compile target, so finding a way to actually complete the task is kinda important 😄 

With GitHub finally releasing ARM runners I'll fork this and look at using that; it worked in another project wonderfully for me today.

yaleman:
Uh, shit. Fat-fingered my git and pushed to this branch, sorry 😅

yaleman:
I've spent some time looking at this, and this PR's changing far too much at one time, I'd suggest we step back a bit and look at just doing small chunks please.

All the cross-compile stuff's not necessary if you either use native runners (which we have on GitHub actions) or QEMU, which should be part of any `docker buildx` invocation.

We'd prefer see this being a few different PR's, at the very least.

- strip it back to a scratch image
- rip out ORAS, get back to docker buildx and enable the caching, which speeds up builds in general (but can smash artifact storage on GHA, which is its own problem)
- SBOM generation

I appreciate trying to do it all at once, but there's a lot that needs to be reworked before I'd accept this to be merged, and as you said there's a lot of complexity in the cross-compile phases, which aren't at all necessary.

PR #3381, opened 2025-02-01
OpenBSD support

arthepsy:
# Implement OpenBSD support

With this patch compiles and runs fine on OpenBSD 7.6, rust 1.81.0. Did basic tests, all good, no issues.
Had to install `openssl-3.3.2`, because there were multiple failures (in various dependencies) using OpenBSD's `libressl`, but that's about it.

Only exception is `unix_integrations/pam_kanidm` (OpenBSD doesn't have built-in PAM), which I just commented out in `Cargo.toml`, when building and everything was build successfully.

Regarding this, I had an idea to rework `Cargo.toml` to exclude this while building on OpenBSD, but it seems that it isn't supported by cargo (read multiple issues by people requesting it, checked ignored pull requests, etc.) or my knowledge how to do is lacking. Maybe this can be reworked some-how (ab)using `features`, but that's to be researched. Anyhow, I think the best case would be to rework `pam_kanidm` for OpenBSD to use OpenPAM, which is available as a package. 

Please, give feedback about how to handle `pam_kanidm` for OpenBSD, so I can work on this.

Checklist

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -45,6 +45,7 @@
 - Chris Olstrom (colstrom)
 - Christopher-Robin (cebbinghaus)
 - Fabian Kammel (datosh)
+- Andris Raugulis (arthepsy)
 
 ## Acknowledgements
 
--- a/libs/file_permissions/src/unix.rs
+++ b/libs/file_permissions/src/unix.rs
@@ -3,6 +3,9 @@ use std::fs::Metadata;
 #[cfg(target_os = "freebsd")]
 use std::os::freebsd::fs::MetadataExt;
 
+#[cfg(target_os = "openbsd")]
+use std::os::openbsd::fs::MetadataExt;
+
 #[cfg(target_os = "linux")]
 use std::os::linux::fs::MetadataExt;
 
--- a/tools/cli/Cargo.toml
+++ b/tools/cli/Cargo.toml
@@ -93,6 +93,10 @@ features = ["mozilla"]
 workspace = true
 features = ["mozilla"]
 
+[target."cfg(target_os = \"openbsd\")".dependencies.webauthn-authenticator-rs]
+workspace = true
+features = ["mozilla"]
+
 ## Debian packaging
 [package.metadata.deb]
 name = "kanidm"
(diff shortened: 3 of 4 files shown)


yaleman:
I'd say stub them out using similar cfg flags to what we use to avoid trying to build on windows etc, eg:

https://github.com/kanidm/kanidm/blob/1453ba5d74663b4ab4dcb16006baf04425eb87b4/unix_integration/pam_kanidm/src/lib.rs#L14-L15

arthepsy:
> I'd say stub them out using similar cfg flags to what we use to avoid trying to build on windows etc, eg:
> 
> https://github.com/kanidm/kanidm/blob/1453ba5d74663b4ab4dcb16006baf04425eb87b4/unix_integration/pam_kanidm/src/lib.rs#L14-L15

Yeah, tried that, but then `core.rs` generates multiple Pam* reference errors, therefore, as I suggested, either the whole package should be "disabled" or reworked to support OpenPAM. Up to You guys to decide and I'll try to make it happen and test.

yaleman:
Hey if you can make it work with OpenPAM (on just OpenBSD), and want to maintain it, you're welcome to 😄

arthepsy:
> Hey if you can make it work with OpenPAM (on just OpenBSD), and want to maintain it, you're welcome to 😄

Sure, can do, will work on that next. Anyhow, is there something else required for this pull request to go forward? I'd like to have those things separate, as this adds OpenBSD support regardless of what is decided with pam module.

yaleman:
Looks good to me, can you rebase it and add yourself to the contributors file and we can merge it.

yaleman (approved):

Firstyear:
@arthepsy Sounds awesome, we'd love to see openbsd support here. Perhaps open an issue so we can track the work and what's needed? Thank you!

arthepsy:
> @arthepsy Sounds awesome, we'd love to see openbsd support here. Perhaps open an issue so we can track the work and what's needed? Thank you!

Great! And I'm willing to help make this happen. What kind of issue You had in mind (just so that I do everything correctly) - labels, tags, etc.?

Firstyear:
@arthepsy Done, made a bsd label for you. but we're pretty chill about it, just as long as we can see what's going on :) 

You may be interested in: https://github.com/kanidm/kanidm/pull/3404 which resolves some nss issues on freebsd that it looks like openbsd would have hit also. I was looking into it and I think we'd need a login_-kanidm next for login.conf on openbsd?

PR #3398, opened 2025-02-04
Fix the password reset form and possible resolver issue

Firstyear:
While testing for everything open I noticed two possible issues. This PR fixes both.

The first is a possible recursion in the resolver. I think I need to fix up it's transactions a bit in another PR.

The second was that the submit button on the reset form doesn't work. This fixes that as well as post reset redirecting to the correct location.

Checklist


server/core/src/https/views/reset.rs
```diff
@@ -204,11 +205,40 @@ impl Display for PasskeyClass {
...
+            "",
+        )
+            .into_response())
+    } else {
+        Ok((
+            jar,
+            HxLocation::from(Uri::from_static(Urls::Login.as_ref())),
+            "",
+        )
+            .into_response())
```
yaleman, line 233: Makes it a little cleaner 😄 
```suggestion
    let url = if state
        .qe_r_ref
        .handle_auth_valid(client_auth_info, kopid.eventid)
        .await
        .is_ok()
    {
        Urls::Profile.as_ref()
    } else {
        Urls::Login.as_ref()
    };

    Ok((jar, HxLocation::from(Uri::from_static(url)), "").into_response())
```

yaleman (approved):

PR #3400, opened 2025-02-04
Feat: Allowing spn query with non-spn structured data in LDAP

### LINKED ISSUE #3390: Support fuzzy matching (username/spn) for ldap queries (not just binds)

Currently LDAP binds support "fuzzy" matching of the user based on either the username `cebbinghaus` or their spn `cebbinghaus@idm.example.com`. This is quite useful given that a user can choose to log in with either their username or their "email" (the unique spn). 

However this seems to only be supported for binds and is not possible for queries where `uid=cebbinghaus@idm.cerver.au` leads to a ldap error 80 "NoMatchingEntries".

Any query for `uid=?` to also fuzzy match on both the spn & username. Especially given that `uid` is not a kanidm attribute and instead exists only as a virtual mapping. when querying the value of `uid` it should keep returning the username, But still allow the spn as a valid query.

Not being able to log into my services using the spn.

I mean, there isn't really much else I can do. Some services might also have bind support but every single service that does also runs a query which would still fail...

### Additional context

//Todo

CEbbinghaus:
# Change summary

- Adding ability to provide non spn formatted ([user]@[host]) text when query ldap for the spn
- Adding some additional documentation / cleaning up

Fixes #3390

Checklist

### DIFF
--- a/book/src/integrations/ldap.md
+++ b/book/src/integrations/ldap.md
@@ -69,6 +69,11 @@ As the POSIX password is not equivalent in strength to the primary credentials o
 most cases is multi-factor authentication), the LDAP bind does not grant rights to elevated read
 permissions. All binds have the permissions of "anonymous" even if the anonymous account is locked.
 
+> As the POSIX password is not equivalent in strength to the primary credentials of Kanidm
+
+This is no longer necessarily true (thanks to the fallback). Would it be possible to  enable elevated read permissions
+specifically for the the account being logged into?
+
 The exception is service accounts which can use api-tokens during an LDAP bind for elevated read
 permissions.
 
@@ -145,7 +150,8 @@ with a dn of `dn=token` and provide the api token in the password.
 > [!NOTE]
 >
 > The `dn=token` keyword is guaranteed to not be used by any other entry, which is why it was chosen
-> as the keyword to initiate api token binds.
+> as the keyword to initiate api token binds. Additionally it is not required, leaving the field empty
+> will fall back to the service-account if a "password" is provided
 
 ```bash
 ldapwhoami -H ldaps://URL -x -D "dn=token" -w "TOKEN"
--- a/book/src/integrations/ldap.md
+++ b/book/src/integrations/ldap.md
@@ -69,11 +69,6 @@ As the POSIX password is not equivalent in strength to the primary credentials o
 most cases is multi-factor authentication), the LDAP bind does not grant rights to elevated read
 permissions. All binds have the permissions of "anonymous" even if the anonymous account is locked.
 
-> As the POSIX password is not equivalent in strength to the primary credentials of Kanidm
-
-This is no longer necessarily true (thanks to the fallback). Would it be possible to  enable elevated read permissions
-specifically for the the account being logged into?
-
 The exception is service accounts which can use api-tokens during an LDAP bind for elevated read
 permissions.
 
@@ -240,6 +235,7 @@ ldapwhoami ... -x -D '22a65b6c-80c8-4e1a-9b76-3f3afdff8400'
 ldapwhoami ... -x -D 'spn=test1@idm.example.com,dc=idm,dc=example,dc=com'
 ldapwhoami ... -x -D 'name=test1,dc=idm,dc=example,dc=com'
 ```
+<sub>in fact, the key of the bind isn't used at all so `googoogaaga=test1` is entirely valid</sub> ;)
 
 ## Troubleshooting
 
--- a/server/lib/src/be/mod.rs
+++ b/server/lib/src/be/mod.rs
@@ -2381,6 +2381,40 @@ mod tests {
     }
 
     #[test]
+    fn test_be_search_with_invalid() {
+        run_test!(|be: &mut BackendWriteTransaction| {
+            trace!("Simple Search");
+
+            let mut e: Entry<EntryInit, EntryNew> = Entry::new();
+            e.add_ava(Attribute::UserId, Value::from("bagel"));
+            e.add_ava(
+                Attribute::Uuid,
+                Value::from("db237e8a-0079-4b8c-8a56-593b22aa44d1"),
+            );
+            let e = e.into_sealed_new();
+
+            let single_result = be.create(&CID_ZERO, vec![e]);
+            assert!(single_result.is_ok());
+            
+            // Test Search with or condition including invalid attribute
+            let filt = filter_resolved!(f_or(vec![f_eq(Attribute::UserId, PartialValue::new_utf8s("bagel")), f_invalid(Attribute::UserId)]));
+
+            let lims = Limits::unlimited();
+
+            let r = be.search(&lims, &filt);
+            assert!(r.expect("Search failed!").len() == 1);
+
+            // Test Search with or condition including invalid attribute
+            let filt = filter_resolved!(f_and(vec![f_eq(Attribute::UserId, PartialValue::new_utf8s("bagel")), f_invalid(Attribute::UserId)]));
+
+            let lims = Limits::unlimited();
+
+            let r = be.search(&lims, &filt);
+            assert!(r.expect("Search failed!").len() == 0);
+        });
+    }
+
+    #[test]
     fn test_be_simple_modify() {
         run_test!(|be: &mut BackendWriteTransaction| {
             trace!("Simple Modify");
--- a/server/lib/src/be/mod.rs
+++ b/server/lib/src/be/mod.rs
@@ -572,7 +572,7 @@ pub trait BackendTransaction {
                 (IdList::Indexed(IDLBitRange::new()), FilterPlan::Invalid)
             }
             FilterResolved::Invalid(_) => {
-                // Partial is the only one that doesn't run the risk of being cancelled due to resource constraint
+                // Indexed since it is always false and we don't want to influence filter testing
                 (IdList::Indexed(IDLBitRange::new()), FilterPlan::Invalid)
             }
         })
--- a/server/lib/src/be/mod.rs
+++ b/server/lib/src/be/mod.rs
@@ -573,7 +573,7 @@ pub trait BackendTransaction {
             }
             FilterResolved::Invalid(_) => {
                 // Partial is the only one that doesn't run the risk of being cancelled due to resource constraint
-                (IdList::Partial(IDLBitRange::new()), FilterPlan::Invalid)
+                (IdList::Indexed(IDLBitRange::new()), FilterPlan::Invalid)
             }
         })
     }
--- a/server/lib/src/be/mod.rs
+++ b/server/lib/src/be/mod.rs
@@ -571,6 +571,10 @@ pub trait BackendTransaction {
                 filter_error!("Requested a top level or isolated AndNot, returning empty");
                 (IdList::Indexed(IDLBitRange::new()), FilterPlan::Invalid)
             }
+            FilterResolved::Invalid(_) => {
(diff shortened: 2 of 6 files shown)


CEbbinghaus:
@Firstyear can you take a quick gander & offer some input?

book/src/integrations/ldap.md
```diff
@@ -69,6 +69,11 @@ As the POSIX password is not equivalent in strength to the primary credentials o
 most cases is multi-factor authentication), the LDAP bind does not grant rights to elevated read
 permissions. All binds have the permissions of "anonymous" even if the anonymous account is locked.
 
+> As the POSIX password is not equivalent in strength to the primary credentials of Kanidm
+
+This is no longer necessarily true (thanks to the fallback). Would it be possible to  enable elevated read permissions
```
Firstyear, line 74: If the primary cred is pw+totp, then ldap is pw only so it's still not equal security wise.

server/lib/src/filter.rs
```diff
@@ -1097,8 +1097,19 @@ impl FilterComp {
             }
             LdapFilter::Equality(a, v) => {
                 let a = ldap_attr_filter_map(a);
-                let v = qs.clone_partialvalue(&a, v)?;
-                FilterComp::Eq(a, v)
+                let pv = qs.clone_partialvalue(&a, v);
+                let Ok(pv) = pv else {
```
Firstyear, line 1101: I think there is an easier way here. 

```
match a {
    Attribute::Uid => {
        let maybe_spn = qs.clone_partialvalue(Attribute::Spn, v).ok(); //makes it an option
        let maybe_uuid = qs.clone_partialvalue(Attribute::Uuid, v).ok();
        let name = qs.clone_partialvalue(Attribute::Name, v)?;
        FilterComp::Or( .... some iterator magic to combine things here. ) 
    }
    _ => {
        qs.clone_partialvalue(&a, v)?;
        FilterComp::Eq(a, v)
    }
}
```

Firstyear:
Remember this isn't about a bind, but about a way to expand the query to include name and spn and other unique ids. So I think the suggestion above should work.

CEbbinghaus:
@Firstyear note as per [this comment](https://github.com/kanidm/kanidm/issues/3390#issuecomment-2633838102) on #3390 I amended the goal to be "allow queries against spn with invalid formats" rather than "query against 1 property for both spn & uid". This way we can implement the query in ldap as `(|(name=?)(spn=?))`. Right now that crashes when `?` is anything but the full spn given the error.

Firstyear:
> @Firstyear note as per [this comment](https://github.com/kanidm/kanidm/issues/3390#issuecomment-2633838102) on #3390 I amended the goal to be "allow queries against spn with invalid formats" rather than "query against 1 property for both spn & uid". This way we can implement the query in ldap as `(|(name=?)(spn=?))`. Right now that crashes when `?` is anything but the full spn given the error.

I think this is pretty easy to achieve, we need to add a filterComp variant which is "Always False". That way on error to convert to spn, we transform to AlwaysFalse, and then the rest of the query proceeds as usual.

The alternate is a PartialValue type which is "Invalid" that will never match.

CEbbinghaus:
That was exactly what I was thinking. either `FilterComp::Invalid` for nop or as per my comment in code:
> I'd rather be solving this by creating a `FilterComp::Constant(bool)` To allow us to embed an always false value given we need to return a filter comp.

which would just always return the constant when evaluated. Both approaches achieve the same goal in execution its just a matter of preference. You have any opinions on which one @Firstyear?

Firstyear:
We already have an "Always True" which is Pres Class, so I think we just make this a FilterComp::Invalid since that much more clearly indicates what's going on here that the requested attr was invalid. We could even do FilterComp::Invalid(Attr) just so we have the log/audit of "oh, the query wanted to search spn, but it was invalid, so that's what went on".

server/lib/src/filter.rs
```diff
@@ -1643,6 +1666,7 @@ impl FilterResolved {
             | FilterResolved::And(_, sf)
             | FilterResolved::Inclusion(_, sf)
             | FilterResolved::AndNot(_, sf) => *sf,
+            FilterResolved::Invalid(_) => Some(NonZeroU8::new(1).unwrap()),
```
Firstyear, line 1669: Comment that because this always returns false there is no index cost so it's always 1.

server/lib/src/filter.rs
```diff
@@ -1504,6 +1526,7 @@ impl FilterResolved {
                 FilterResolved::resolve_no_idx((*f).clone(), ev)
                     .map(|fi| FilterResolved::AndNot(Box::new(fi), None))
             }
+            FilterComp::Invalid(_) => None,
```
Firstyear, line 1529: Needs to return Some(FilterResolved::Invalid ) else this causes the filter element to be dropped from the query.

Firstyear: See line 539 in filter.rs.

server/lib/src/filter.rs
```diff
@@ -1278,6 +1295,7 @@ impl FilterResolved {
                 FilterResolved::Eq(a, v, idx)
             }
             FilterComp::SelfUuid => panic!("Not possible to resolve SelfUuid in from_invalid!"),
+            FilterComp::Invalid(_) => panic!("Not possible to resolve Invalid in from_invalid!"),
```
Firstyear, line 1298: This can be resolved, just return the same as you would in the other resolve handlers.

server/lib/src/filter.rs
```diff
@@ -1099,11 +1118,9 @@ impl FilterComp {
                 let a = ldap_attr_filter_map(a);
                 let pv = qs.clone_partialvalue(&a, v);
                 let Ok(pv) = pv else {
-                    // This is a massive hack to get around the fact that we can't to do a equality check on the spn
-                    // without providing a valid spn string `e.g spn=username`. I'd rather be solving this by creating a `FilterComp::Constant(bool)`
-                    // To allow us to embed an always false value given we need to return a filter comp.
+                    // This is to prevent a badly formed spn from causing the query to fail
                     if a == Attribute::Spn {
-                        return Ok(FilterComp::Eq(a, PartialValue::Spn("invalid".into(), "spn".into())));
+                        return Ok(FilterComp::Invalid(a));
```
Firstyear, line 1123:
```
match pv {
    Ok(pv) => FilterComp::Eq(a, pv),
    Err(_) if a == Attribute::Spn => FilterComp::Invalid(a),
    Err(err) => return Err(err), 
}
```

server/lib/src/filter.rs
```diff
@@ -952,6 +967,10 @@ impl FilterComp {
                 // Pretty hard to mess this one up ;)
                 Ok(FilterComp::SelfUuid)
             }
+            FilterComp::Invalid(attr) => {
+                // This is a special case, we can't validate it.
```
Firstyear, line 971: Technically this is valid, since we are saying that the partial value was invalid, so the pv validation is done and failed, so we need a valid filter part that will always fail.

server/lib/src/be/mod.rs
```diff
@@ -571,6 +571,10 @@ pub trait BackendTransaction {
                 filter_error!("Requested a top level or isolated AndNot, returning empty");
                 (IdList::Indexed(IDLBitRange::new()), FilterPlan::Invalid)
             }
+            FilterResolved::Invalid(_) => {
+                // Partial is the only one that doesn't run the risk of being cancelled due to resource constraint
+                (IdList::Partial(IDLBitRange::new()), FilterPlan::Invalid)
```
Firstyear, line 576: I think this should be IdList::Indexed, since Invalid can never match, so we don't need to trigger a filter test when this term returns. The issue with Partial is if this returns Partial and it's in an Or clause, then the neighbour in the Or condition also becomes partial and forces filter testing. So this assertion being "never matches" can be Indexed so that the index state of the other IdList's can be used.

Firstyear (commented):
Doing really good so far, we need some tests in here I think but mostly looks good. You're touching a really hard area of the codebase as well, so don't feel bad either if it seems like a lot.

server/lib/src/filter.rs
CEbbinghaus: Wouldn't that be desirable? An invalid filter shouldn't make it through and be dropped saving us a few CPU cycles. Or do we want to keep the query perfectly in tact to preserve "correctness"

server/lib/src/filter.rs
CEbbinghaus: Less so that it can't but that I didn't want to. I'll add it (probably good for testing)

server/lib/src/filter.rs
Firstyear: This component may be invalid, but the rest is valid. Consider the original reason for this issue - `|( (uid=?)(spn=?) )`

spn would be invalid, and then with your suggestion, the whole question would instant fail because the spn couldn't "resolve" the uuid. 

When the real answer is that spn will never match (invalid) but uid= can match, and then causes the whole OR condition to be true.

server/lib/src/be/mod.rs
CEbbinghaus: Indexed can cause the query to be cancelled due to overuse of resources however, Which is why I opted to go with the Partial. I'll trust your judement on this one. just thought it would be better to ensure this doesn't push the query ove the limit

server/lib/src/filter.rs
CEbbinghaus: Strangely the observed behavior was that it worked correctly in any case no matter if I returned None or Some. So I am unsure to what degree this is actually applicable but I changed the code just to be safe.

server/lib/src/be/mod.rs
```diff
@@ -571,6 +571,10 @@ pub trait BackendTransaction {
                 filter_error!("Requested a top level or isolated AndNot, returning empty");
                 (IdList::Indexed(IDLBitRange::new()), FilterPlan::Invalid)
             }
+            FilterResolved::Invalid(_) => {
+                // Partial is the only one that doesn't run the risk of being cancelled due to resource constraint
```
Firstyear, line 575: Stale comment. Perhaps "This is considered indexed as it can never evaluate to try and shouldn't influence filter testing"

server/lib/src/filter.rs
Firstyear: In otherwords, change the comment to say "While the filter component is named invalid, the invalid here only means that the partial value content was invalid. However, it's still *valid* to proceed with this search since the assertion can never match. This allows queries with invalid pvs to proceed and still evaluate correctly such as in an OR condition. "

server/lib/src/idm/ldap.rs
```diff
@@ -240,6 +240,7 @@ impl LdapServer {
             let mut all_op_attrs = false;
 
             // TODO #67: limit the number of attributes here!
+            // Is this still needed?
```
Firstyear, line 243: Yeah, probably, we should have an issue for it. It would be pretty easy to add IMO.

Firstyear (changes requested):
Pretty much just need some comments updated.

server/lib/src/be/mod.rs
CEbbinghaus: whoops, Thought I had deleted it :(

CEbbinghaus:
Should be ready to merge now. Since this would count as a unexpected behavior I don't think we need to explicitly document that querying against the spn will no longer crash the query.

Firstyear (approved):

PR #3402, opened 2025-02-05
Correctly return that uuid2spn changed on domain rename

### LINKED ISSUE #3360: Visual domain mixup after domain migration
### I did this

Migrated a kanidm domain from `example.com` to `id.example.com`.

### I expected the following

All entities should now be under the `id.example.com` domain name.

### This happened instead

The `spn` for users was (`name: hexa@id.example.com`, but e.g. the `member` attribute of a group used the old spn (`member: hexa@example.com` to reference that user. 

### Kanidm version details

- Output of `kanidm(d) version`: 51a976fed559aa4c6097b2dd2e6b6cb71b7498b9
- Operating System / Version (On Unix please post the output of `uname -a`): NixOS 24.11

### Any other comments

We tried updating the members with the explicit new spns, but that did not change anything.

The visual glitch was however corrected after reindexing the database.

I have the logs for the domain migration and the reindex stored and can share them privately if that helps.

Firstyear:
Due to a missing equality check in value, when a domain rename occured, the uuid2spn index differential function did not correctly detect that the domain name had updated which meant that the uuid2spn index was not updated. Only this index was affected, and a manual reindex would resolve.

Fixes #3360

Checklist

### DIFF
--- a/server/lib/src/plugins/spn.rs
+++ b/server/lib/src/plugins/spn.rs
@@ -208,10 +208,7 @@ impl Spn {
         // All we do is purge spn, and allow the plugin to recreate. Neat! It's also all still
         // within the transaction, just in case!
         qs.internal_modify(
-            &filter!(f_or!([
-                f_eq(Attribute::Class, EntryClass::Group.into()),
-                f_eq(Attribute::Class, EntryClass::Account.into()),
-            ])),
+            &filter!(f_pres(Attribute::Spn)),
             &modlist!([m_purge(Attribute::Spn)]),
         )
     }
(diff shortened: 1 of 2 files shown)


examples/insecure_server.toml
```diff
@@ -18,8 +18,8 @@ log_level = "info"
 
 # otel_grpc_url = "http://localhost:4317"
 
-domain = "localhost"
-origin = "https://localhost:8443"
+domain = "id.localhost"
+origin = "https://id.localhost:8443"
```
yaleman, line 22: nope!

Firstyear: Opps.

yaleman (approved):

PR #3410, opened 2025-02-09
feat: Added webfinger implementation

### LINKED ISSUE #3162: Add support for Webfinger - RFC 7033
### Why Webfinger

Webfinger is a pretty useful thing for other services that want to discover information about some URI, one such services is [Tailscale](https://tailscale.com). Since Kanidm supports OIDC Webfinger would allow services to easily discover the configuration for verifying a user.

### What could be done

Kanidm could have a Webfinger specific public client that could either be enabled or disabled. This would indeed conflict with the philosophy of having one client id for each application, but I think It's an interesting compromise for some services and could be circumvented with scopes.

### Current Solution

I've been maintaining a secondary service [Maronato/go-finger](https://github.com/Maronato/go-finger) but having it directly supported in Kanidm makes things much simpler and exclude the need of having two edited configurations

### Additional context

https://datatracker.ietf.org/doc/html/rfc7033
https://webfinger.net/

CEbbinghaus:
# Change summary

Adds new webfinger endpoint on every oauth2 client to allow users to set up webfinger however they want.

Fixes #3162

Checklist

### DIFF
--- a/book/src/integrations/oauth2.md
+++ b/book/src/integrations/oauth2.md
@@ -70,6 +70,36 @@ anything special for Kanidm (or another provider).
 **Note:** some apps automatically append `/.well-known/openid-configuration` to
 the end of an OIDC Discovery URL, so you may need to omit that.
 
+
+<dl>
+
+<dt>
+
+[Webfinger](https://datatracker.ietf.org/doc/html/rfc7033)
+URL
+
+</dt>
+<dd>
+
+`https://idm.example.com/oauth2/openid/:client_id:/.well-known/webfinger`
+
+The webfinger url is implemented for each openid client, under its specific url, giving full control to the administrator regarding which to use.
+
+To make this webfinger useful it **MUST** be served at the very root under the host (e.g `example.com/.well-known/webfinger`). How that is accomplished is left up to the administrator as kani has no opinion.
+
+One example would be dedicating one client as the `"primary" or "default" and redirecting all requests to that. Alternatively source ip or other request metadata could be used to decide which client to forward the request to.
+
+### Caddy
+`Caddyfile`
+```caddy
+# assuming a kanidm service with domain "example.com"
+example.com {
+  redir /.well-known/webfinger https://idm.example.com/oauth2/openid/:client_id:{uri} 307
+}
+```
+**Note:** the `{uri}` is important as it preserves the original request past the redirect.
+
+
 </dd>
 
 <dt>
--- a/book/src/integrations/oauth2.md
+++ b/book/src/integrations/oauth2.md
@@ -72,22 +72,17 @@ the end of an OIDC Discovery URL, so you may need to omit that.
 
 
 <dl>
+<dt>[Webfinger](https://datatracker.ietf.org/doc/html/rfc7033) URL</dt>
 
-<dt>
-
-[Webfinger](https://datatracker.ietf.org/doc/html/rfc7033)
-URL
-
-</dt>
 <dd>
 
 `https://idm.example.com/oauth2/openid/:client_id:/.well-known/webfinger`
 
-The webfinger url is implemented for each openid client, under its specific url, giving full control to the administrator regarding which to use.
+The webfinger URL is implemented for each OpenID client, under its specific endpoint, giving full control to the administrator regarding which to use.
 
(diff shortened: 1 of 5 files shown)


proto/src/oauth2.rs
```diff
@@ -443,6 +443,21 @@ fn require_request_uri_parameter_supported_default() -> bool {
...
 }
 
+#[derive(Serialize, Deserialize, Debug)]
+pub struct OidcWebfingerRel {
+    pub rel: String,
+    pub href: String,
+}
+
+/// The response to an Webfinger request. Only a subset of the body is defined here.
+/// <https://datatracker.ietf.org/doc/html/rfc7033>
```
yaleman, line 453:
```suggestion
/// <https://datatracker.ietf.org/doc/html/rfc7033#section-4.4>
```

book/src/integrations/oauth2.md
```diff
@@ -70,6 +70,36 @@ anything special for Kanidm (or another provider).
...
+
+<dl>
+
+<dt>
+
+[Webfinger](https://datatracker.ietf.org/doc/html/rfc7033)
+URL
+
+</dt>
+<dd>
```
yaleman, line 82:
```suggestion
<dl>
<dt>[Webfinger](https://datatracker.ietf.org/doc/html/rfc7033) URL</dt>

<dd>
```

book/src/integrations/oauth2.md
```diff
@@ -70,6 +70,36 @@ anything special for Kanidm (or another provider).
...
+
+`https://idm.example.com/oauth2/openid/:client_id:/.well-known/webfinger`
+
+The webfinger url is implemented for each openid client, under its specific url, giving full control to the administrator regarding which to use.
```
yaleman, line 86:
```suggestion
The webfinger URL is implemented for each OpenID client, under its specific endpoint, giving full control to the administrator regarding which to use.
```

book/src/integrations/oauth2.md
```diff
@@ -70,6 +70,36 @@ anything special for Kanidm (or another provider).
...
+
+To make this webfinger useful it **MUST** be served at the very root under the host (e.g `example.com/.well-known/webfinger`). How that is accomplished is left up to the administrator as kani has no opinion.
```
yaleman, line 88:
```suggestion
To make this compliant with the standard, it must be made available under the correct [well-known endpoint](https://datatracker.ietf.org/doc/html/rfc7033#section-10.1) (e.g `example.com/.well-known/webfinger`), typically via a reverse proxy or similar. Kanidm doesn't currently provide a mechanism for this URI rewrite.
```

book/src/integrations/oauth2.md
```diff
@@ -70,6 +70,36 @@ anything special for Kanidm (or another provider).
...
+
+One example would be dedicating one client as the `"primary" or "default" and redirecting all requests to that. Alternatively source ip or other request metadata could be used to decide which client to forward the request to.
```
yaleman, line 90:
```suggestion
One example would be dedicating one client as the "primary" or "default" and redirecting all requests to that. Alternatively, source IP or other request metadata could be used to decide which client to forward the request to.
```

book/src/integrations/oauth2.md
CEbbinghaus: I wrote the "must" as `**MUST**` given that is how it is written within the standard (since must has special meaning). but happy to make it boring :P

server/core/src/https/oauth2.rs
```diff
@@ -538,6 +538,49 @@ pub async fn oauth2_openid_discovery_get(
...
+    resource: String,
+}
+
+pub async fn oauth2_openid_webfinger_get(
+    State(state): State<ServerState>,
+    Path(client_id): Path<String>,
+    Query(query): Query<Oauth2OpenIdWebfingerQuery>,
+
+    // For the moment this is implemented ignoring the rel's
+    // Query(rel): Query<Vec<String>>,
```
yaleman, line 552:
```suggestion
```

CEbbinghaus: Forgor I had that. Needs to be deleted one way or another. <3

yaleman (approved):

PR #3419, opened 2025-02-11
TOTP label verification

### LINKED ISSUE #2920: [TOTP] empty and same name entries don't error
Currently neither the Web UI or API restricts or notifies about:
- setting an **empty name**.
- setting an already **existing name**

for a TOTP entry. 

We should probably protect the users a bit here :)

CEbbinghaus:
# Change summary

Adds the verification for TOTP Labels. Ensuring they are both not empty (or blank), as well as unique.

Fixes #2920

Checklist

### DIFF
--- a/server/core/src/https/v1.rs
+++ b/server/core/src/https/v1.rs
@@ -1396,6 +1396,14 @@ pub async fn credential_update_update(
             return Err(WebError::InternalServerError(errmsg));
         }
     };
+
+    if matches!(scr, CURequest::TotpVerify(_, ref label) if label.trim().is_empty()) {
+        // I'd really rather have a 400 here. Do I add a WebError::HttpError(u16, String)?
+        return Err(WebError::InternalServerError(
+            "Label must not be empty".into(),
+        ));
+    }
+
     let session_token = match serde_json::from_value(cubody[1].clone()) {
         Ok(val) => val,
         Err(err) => {
@@ -1406,6 +1414,7 @@ pub async fn credential_update_update(
     };
     debug!("session_token: {:?}", session_token);
     debug!("scr: {:?}", scr);
+
     state
         .qe_r_ref
         .handle_idmcredentialupdate(session_token, scr, kopid.eventid)
--- a/server/core/src/https/v1.rs
+++ b/server/core/src/https/v1.rs
@@ -1397,13 +1397,6 @@ pub async fn credential_update_update(
         }
     };
 
-    if matches!(scr, CURequest::TotpVerify(_, ref label) if label.trim().is_empty()) {
-        // I'd really rather have a 400 here. Do I add a WebError::HttpError(u16, String)?
-        return Err(WebError::InternalServerError(
-            "Label must not be empty".into(),
-        ));
-    }
-
     let session_token = match serde_json::from_value(cubody[1].clone()) {
         Ok(val) => val,
         Err(err) => {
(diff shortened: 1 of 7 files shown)


CEbbinghaus:
Given that the SecurityKeys are also in a Dictionary I am a little surprised that we can have more than 1 security key with the same name without running into the same problems as with TOTP.
image

Will likely leave the security key names to be solved by a followup PR given that they don't crash the credential update :P

server/lib/src/idm/credupdatesession.rs
```diff
@@ -1899,7 +1907,20 @@ impl IdmServerCredUpdateTransaction<'_> {
         match &session.mfaregstate {
             MfaRegState::TotpInit(totp_token)
             | MfaRegState::TotpTryAgain(totp_token)
+            | MfaRegState::TotpNameTryAgain(totp_token, _)
```
Firstyear, line 1910: Testcase in this file to also assert that the duplicate name gives the correct error etc. 

I think we should potentially also consider limiting *new* labels with https://github.com/kanidm/kanidm/blob/master/server/lib/src/value.rs#L2213

CEbbinghaus: Good point. Yeah last night got a bit late so I didn't end up getting to adding tests but I will. Won't get to it tonight but should have it done soon:tm:

server/lib/src/credential/mod.rs
```diff
@@ -505,6 +505,15 @@ impl Credential {
         })
     }
 
+    //TODO: Add validation for securitykey label
```
Firstyear, line 508: These will be removed soon, so don't worry about it.

server/core/src/https/v1.rs
```diff
@@ -1396,6 +1396,14 @@ pub async fn credential_update_update(
             return Err(WebError::InternalServerError(errmsg));
         }
     };
+
+    if matches!(scr, CURequest::TotpVerify(_, ref label) if label.trim().is_empty()) {
+        // I'd really rather have a 400 here. Do I add a WebError::HttpError(u16, String)?
```
Firstyear, line 1401: You're trying to double validate here then, that's the issue and why it feels "off". Because the validation oef the data is actually done in the idmcredentuialupdate which is better positioned to serve a contextual and informed response.

So remove this block, it's not needed.

Firstyear (changes requested):

server/lib/src/credential/mod.rs
CEbbinghaus: Ah, alright I'll remove that method then.

server/core/src/https/v1.rs
CEbbinghaus: It is double validating. **But** that is for additional contextual information in terms of Http response. We should be returning an invalid input early as 400, Yet currently a "InvalidName" is returned as a 200 json response.

I'd be nice if we made it a little more complaint to how HTTP expects it to work.

CEbbinghaus: Happy enough with that. And given that is already the precedent its not worth trying to change now. 

Given I have moved the empty check further in anyways this is no longer the only place checking it

Firstyear (approved):

Firstyear:
@CEbbinghaus You need to rebase and merge :)

PR #3422, opened 2025-02-12
Support redirect uris with query parameters

### LINKED ISSUE #3420: [Oauth2] Support retaining redirect urls with their query strings.
I tried using [kanboard](https://kanboard.org/) today with their oauth2 [plugin](https://github.com/kanboard/plugin-oauth2).
- The callback url that the plugin uses is: `https://kanboard.example.com/?controller=OAuthController&action=handler&plugin=OAuth2`
I added the callback url as redirect url to kanboard's kanidm resource provider.

I also see this redirect url being passed in the initial redirect from kanboard -> kanidm.
But kanidm redirects to `https://kanboard.example.com/?state=...&code=...` , dropping all the other query strings.

Kanidm should not drop query strings if they're part of the registered redirect url.

/

### Additional context
Kanboard is in maintenance mode, it'll likely never change on their side.

Firstyear:
RFC 6749 once again reminds us that given the room to do silly things, RFC authors absolutely will. In this case, it's query parameters in redirection uris which are absolutely horrifying and yet, here we are.

We strictly match the query pairs during the redirection to ensure that if a query pair did allow open redirection, then we prevent it.

Fixes #3420

Checklist

### DIFF
--- a/server/lib/src/idm/oauth2.rs
+++ b/server/lib/src/idm/oauth2.rs
@@ -248,24 +248,32 @@ impl AuthorisePermitSuccess {
     pub fn build_redirect_uri(&self) -> Url {
         let mut redirect_uri = self.redirect_uri.clone();
 
-        // Always clear query and fragment, regardless of the response mode
-        redirect_uri.set_query(None);
+        // Always clear the fragment per RFC
         redirect_uri.set_fragment(None);
 
-        // We can't set query pairs on fragments, only query.
-        let mut uri_builder = url::form_urlencoded::Serializer::new(String::new());
-
-        uri_builder.append_pair("code", &self.code);
-
-        if let Some(state) = self.state.as_ref() {
-            uri_builder.append_pair("state", state);
-        };
-
-        let encoded = uri_builder.finish();
-
         match self.response_mode {
-            ResponseMode::Query => redirect_uri.set_query(Some(&encoded)),
-            ResponseMode::Fragment => redirect_uri.set_fragment(Some(&encoded)),
+            ResponseMode::Query => {
+                redirect_uri
+                    .query_pairs_mut()
+                    .append_pair("code", &self.code);
+
+                if let Some(state) = self.state.as_ref() {
+                    redirect_uri.query_pairs_mut().append_pair("state", state);
+                };
+            }
+            ResponseMode::Fragment => {
+                redirect_uri.set_query(None);
+
+                // We can't set query pairs on fragments, only query.
+                let mut uri_builder = url::form_urlencoded::Serializer::new(String::new());
+                uri_builder.append_pair("code", &self.code);
+                if let Some(state) = self.state.as_ref() {
+                    uri_builder.append_pair("state", state);
+                };
+                let encoded = uri_builder.finish();
+
+                redirect_uri.set_fragment(Some(&encoded))
+            }
         }
 
         redirect_uri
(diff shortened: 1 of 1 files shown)


server/lib/src/idm/oauth2.rs
```diff
@@ -248,24 +248,32 @@ impl AuthorisePermitSuccess {
...
-        uri_builder.append_pair("code", &self.code);
-
-        if let Some(state) = self.state.as_ref() {
-            uri_builder.append_pair("state", state);
-        };
+        match self.response_mode {
+            ResponseMode::Query => {
+                redirect_uri
+                    .query_pairs_mut()
+                    .append_pair("code", &self.code);
```
CEbbinghaus, line 258: Will this conflict if the query was sent with a `?code=...` query parameter?

server/lib/src/idm/oauth2.rs
```diff
@@ -3680,6 +3688,70 @@ mod tests {
                 == Oauth2Error::InvalidOrigin
         );
 
+        // * invalid uri (doesn't match query params)
```
CEbbinghaus, line 3691: Does the URL have to match query params? I would think its base url only

CEbbinghaus (approved):
> The endpoint URI MAY include an
   "application/x-www-form-urlencoded" formatted (per [Appendix B](https://www.rfc-editor.org/rfc/rfc6749#appendix-B)) query
   component ([[RFC3986] Section 3.4](https://www.rfc-editor.org/rfc/rfc3986#section-3.4)), which MUST be retained when adding
   additional query parameters.
   
   Based on this it sounds like the query parameters should only be kept **if** the content-type is `application/x-www-form-urlencoded`

server/lib/src/idm/oauth2.rs
Firstyear: If the query param may contain an open redirect, we can't allow it. We want to be secure.

server/lib/src/idm/oauth2.rs
Firstyear: We stomp it in that case, and they can get in the bin.

CEbbinghaus: Just making sure we don't return `? code=somecode&code=someothercode`

server/lib/src/idm/oauth2.rs
CEbbinghaus: Are there any provisions in the rfc for this? I'd have to read through it again

Firstyear:
> Based on this it sounds like the query parameters should only be kept **when** the content-type is `application/x-www-form-urlencoded`

Rust URL parser only deals in form encoded by default, so queries must be in that encoding or they are invalid during parse.

CEbbinghaus:
Cool. That is all the confirmation I needed. Fire away ❤️

server/lib/src/idm/oauth2.rs
```diff
@@ -248,24 +248,32 @@ impl AuthorisePermitSuccess {
...
 
-        let encoded = uri_builder.finish();
+                if let Some(state) = self.state.as_ref() {
+                    redirect_uri.query_pairs_mut().append_pair("state", state);
+                };
+            }
+            ResponseMode::Fragment => {
+                redirect_uri.set_query(None);
+
+                // We can't set query pairs on fragments, only query.
```
yaleman, line 267:
```suggestion
                // Per [the RFC](https://www.rfc-editor.org/rfc/rfc6749#section-3.1.2), we can't set query pairs on fragment-containing redirects, only query ones.
```

mainly 'cause it was confusing to read this comment without the discussion context below

yaleman (changes requested):
one minor nit for future-coders, but woo

yaleman (approved):

Issue #3413, opened 2025-02-09
domain change isn't reflected in sign in page until after a restart


xyhhx:
# I did this
* Set up kanidm with idm.test.lan as the domain
* Shut off the service with `systemctl stop kanidm`
* Changed the origin and domain settings in the derivation (see below)
* Ran `sudo kanidmd domain rename` (the config file is not owned by an accessible user, so root was used to read it)
* Started the service with `systemctl start kanidm`
# I expected the following
After the service restart, the webui for kanidm should reflect the new domain.
# This happened instead
Groups and internal names did get changed, the domain in the webui did not.
# Kanidm version details
* kanidm 1.4.6 on NixOS Unstable (2025-02-09)
* No.
* The following derivation:
```nix
{pkgs, ...}: {
  services.kanidm = {
    enableServer = true;
    serverSettings = {
      bindaddress = "0.0.0.0:8443";
      origin = "https://idm.test.com";
      domain = "idm.test.com";
      # The directory is different because this module makes a read only bind for these files
      # https://github.com/NixOS/nixpkgs/issues/312012
      tls_chain = /var/lib/kanidm-certs/cert.pem;
      tls_key = /var/lib/kanidm-certs/key.pem;
    };
  };
}
```
* $ uname -a
* Linux node2 6.12.12 #1-NixOS SMP PREEMPT_DYNAMIC Sat Feb  1 17:39:40 UTC 2025 x86_64 GNU/Linux

# Other comments

(i'm opening this on behalf of another user from the matrix/gitter chat who does not have a github accouont)

Firstyear:
This isn't a bug - when we first initialise the server we have to create a default display name for the domain https://github.com/kanidm/kanidm/blob/master/server/lib/src/plugins/domain.rs#L112 

So post rename, the admin needs to update the displayname on their own. `kanidm system domain set-displayname`.

Additionally, domain renames are a large intrusive process that the admin needs to plan for with care, so asking them to set a new displayname post domain rename doesn't seem like a big deal imo.

yaleman:
I kinda feel like if the displayname isn't set (or is the same as the domain name by default) then it could be auto-updated from a "least surprise" situation.

Firstyear:
The problem is that it *is* set, it's set at first run. So we don't know if it was set by the human or not. The alternative is we make domainDisplayName optional and then synth it into the domain_info struct. Which could work here, but I also don't know if it's worth all the effort since a rename is rare and an admin can change this easily.

[...]

Issue #3270, opened 2024-12-06
idm_admin cannot modify entry-mananged-by attribute


pando85:
### I did this

- Set up a new Kanidm instance.
- Create a new Group with an `entry-managed-by` person with `idm_admin`.
- Try to update `entry -managed-by` for that group with `idm_admin`.

### I expected the following

It updates it correctly to the new `entry-managed-by`.

### This happened instead

```
INFO     request [ 245µs | 13.50% / 100.00% ] method: PUT | uri: /v1/group/managed/_attr/entry_managed_by | version: HTTP/1.1
INFO     ┝━ set_attribute [ 212µs | 52.51% / 86.50% ]
INFO     │  ┝━ validate_client_auth_info_to_ident [ 83.3µs | 33.99% ]
INFO     │  │  ┕━ ｉ [info]: A valid limited session value exists for this token | event_tag_id: 10
INFO     │  ┝━ ｉ [info]: modify initiator | event_tag_id: 10 | name: User( idm_admin@idm.localhost, 00000000-0000-0000-0000-000000000018 ) (c2efdc60-6426-4365-8b79-85cdf73baac0, read write)
INFO     │  ┝━ ｉ [info]: search | event_tag_id: 10 | initiator: User( idm_admin@idm.localhost, 00000000-0000-0000-0000-000000000018 ) (c2efdc60-6426-4365-8b79-85cdf73baac0, read write)
ERROR    │  ┝━ 🚨 [error]: requested_pres is not a subset of allowed | event_tag_id: 12
ERROR    │  ┝━ 🚨 [error]: requested_pres: {EntryManagedBy} !⊆ allowed: {AuthSessionExpiry, AuthPasswordMinimumLength, Class, CredentialTypeMinimum, Description, GidNumber, LimitSearchMaxResults, LimitSearchMaxFilterTest, Mail, Member, Name, PrivilegeExpiry, WebauthnAttestationCaList, AllowPrimaryCredFallback} | event_tag_id: 12
INFO     │  ┕━ ｉ [info]: denied ❌ - modify may not proceed | event_tag_id: 11
WARN     ┕━ 🚧 [warn]:  | latency: 244.982µs | status_code: 403 | kopid: "a6bc0016-7140-4dd8-9efa-09f1e7c5595f" | msg: "client error"
```

### Kanidm version details

- Output of `kanidm(d) version: `kanidmd 1.4.4`
- Are you running it in a container? If so, which image/tag?: `kanidm/server:latest`
[...]

PR #3425, opened 2025-02-13
20240213 3413 domain displayname

Firstyear:
This removes the mandatory requirement for the domain display name to exist and be generated.

However, due to how migrations work today, the domain display name will *still be generated*. This is because during the domain.rs overlay/plugin, we check domain_version and if it's 0 we are bootstrapping the server. Since older schema versions DO expect the displayname to still be set WE HAVE TO STILL SET IT. 

As a result, we still need to set the displayname *for now*, but come version 1.8 or so, when our min remigration level raises, we can drop this. 

This trims two migration levels out to assist in that process, but in future we also should reconsider how we apply migrations. Today they are somewhat error prone, and I think we need a different approach. As much as @yaleman will hate it, we should have migration dirs which list every entry in the server, and each version we copy-paste these and modify those entries. That way we have a complete snapshot of exactly how the server entries of that version must look, and we can apply them directly. That will allow us to run the migration to target level directly, rather than the current process where we have to apply the migration of the oldest version then step up. It will also mean that it's easier to see what changed, and do these clean ups, even if it's more verbose by copying things. 

The current system with the handlers in constants for builtin entries doesn't work well with migrations, but will work better with the copy-paste approach. Similar for setup. But most of all it means we can skip the DL0 implying a bootstrap, and at the moment we start we can immediately set target level so that plugins like the domain overlay can actually just see the precise version off the rip. 

Relates #3413

Checklist

### DIFF
--- a/server/lib/src/constants/mod.rs
+++ b/server/lib/src/constants/mod.rs
@@ -54,10 +54,6 @@ pub type DomainVersion = u32;
 /// previously.
 pub const DOMAIN_LEVEL_0: DomainVersion = 0;
 
-/// Domain Level introduced with 1.2.0.
-/// Deprecated as of 1.4.0
-pub const DOMAIN_LEVEL_6: DomainVersion = 6;
-
 /// Domain Level introduced with 1.3.0.
 /// Deprecated as of 1.5.0
 pub const DOMAIN_LEVEL_7: DomainVersion = 7;
@@ -81,7 +77,7 @@ pub const DOMAIN_LEVEL_11: DomainVersion = 11;
 
 // The minimum level that we can re-migrate from.
 // This should be DOMAIN_TGT_LEVEL minus 2
-pub const DOMAIN_MIN_REMIGRATION_LEVEL: DomainVersion = DOMAIN_LEVEL_6;
+pub const DOMAIN_MIN_REMIGRATION_LEVEL: DomainVersion = DOMAIN_LEVEL_8;
 // The minimum supported domain functional level (for replication)
 pub const DOMAIN_MIN_LEVEL: DomainVersion = DOMAIN_TGT_LEVEL;
 // The previous releases domain functional level
--- a/server/lib/src/constants/mod.rs
+++ b/server/lib/src/constants/mod.rs
@@ -54,13 +54,9 @@ pub type DomainVersion = u32;
 /// previously.
 pub const DOMAIN_LEVEL_0: DomainVersion = 0;
 
-/// Deprecated as of 1.3.0
-pub const DOMAIN_LEVEL_5: DomainVersion = 5;
-
 /// Domain Level introduced with 1.2.0.
 /// Deprecated as of 1.4.0
 pub const DOMAIN_LEVEL_6: DomainVersion = 6;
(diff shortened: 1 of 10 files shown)


Firstyear:
Blocks #3270 and #3426

Firstyear:
(Mostly because I need to do other migration things that build on this)

yaleman (approved):

server/lib/src/constants/mod.rs
```diff
@@ -54,14 +54,6 @@ pub type DomainVersion = u32;
...
-pub const DOMAIN_LEVEL_5: DomainVersion = 5;
-
-/// Domain Level introduced with 1.2.0.
-/// Deprecated as of 1.4.0
-pub const DOMAIN_LEVEL_6: DomainVersion = 6;
-pub const PATCH_LEVEL_1: u32 = 1;
-
 /// Domain Level introduced with 1.3.0.
 /// Deprecated as of 1.5.0
 pub const DOMAIN_LEVEL_7: DomainVersion = 7;
```
CEbbinghaus, line 59: keeping 7 even though its Deprecated?

server/lib/src/server/migrations.rs
```diff
@@ -928,22 +614,33 @@ impl QueryServerWriteTransaction<'_> {
             // DL4
             SCHEMA_CLASS_OAUTH2_RS_PUBLIC_DL4.clone().into(),
             // DL5
-            SCHEMA_CLASS_PERSON_DL5.clone().into(),
+            // SCHEMA_CLASS_PERSON_DL5.clone().into(),
```
CEbbinghaus, line 617: why comment them out instead of deleting?

server/lib/src/server/migrations.rs
```diff
@@ -1300,7 +909,7 @@ impl QueryServerReadTransaction<'_> {
 
 #[cfg(test)]
 mod tests {
-    use super::{ProtoDomainUpgradeCheckItem, ProtoDomainUpgradeCheckStatus};
+    // use super::{ProtoDomainUpgradeCheckItem, ProtoDomainUpgradeCheckStatus};
```
CEbbinghaus, line 912: again, why not just delete?

CEbbinghaus (approved):
image
I can already tell this is gonna be a good one.

Surprised with how little was actually removed. Would have expected more schema to go.

All nits nothing I consider important enough not to merge.

server/lib/src/constants/mod.rs
Firstyear: It's still checked in some places as a minimum.

server/lib/src/server/migrations.rs
Firstyear: Because I'm about to do a second PR to clean it up and wanted to validate some things.

server/lib/src/server/migrations.rs
Firstyear: Laziness.

Firstyear: Which really means I've left it there to uncomment once I put back something that uses the DomainUpgradeCheckItem in future.

PR #3429, opened 2025-02-14
Exempt idm_admin and admin from denied names.

### LINKED ISSUE #3426: Adding 'admin' to denied-names prevents account recovery of build-in admin account
### I did this

I added `admin` to the list of system denied-names like so:

```
kanidm system denied-names append admin
```

It succeeded!

### I expected the following

I probably shouldn't have been allowed to shoot myself in the foot like that. Or should I?

If you try to add `root` as a denied-name, it fails. Which is probably sensible.

Perhaps a warning inline that this is probably not a good idea before allowing someone to try to lock themselves out of the system like this.

### This happened instead

I was allowed to add `admin` as a denied name.

If you then try to do

```
kanidmd recover-account admin
```

It will fail, like this:

```
docker exec -i -t kanidm  kanidmd recover-account admin
00000000-0000-0000-0000-000000000000 INFO     ｉ [info]: Running account recovery ...
00000000-0000-0000-0000-000000000000 INFO     ｉ [info]: Error - you should inspect the logs.
Logging pipeline completed shutdown
```

The logs show

```
00000000-0000-0000-0000-000000000000 INFO     ｉ [info]: Allowing admin socket access | pid: Some(1407)
e65941c1-81cf-4f67-82ed-06837a7139be INFO     handle_admin_client_request [ 17.6ms | 0.08% / 100.00% ]
e65941c1-81cf-4f67-82ed-06837a7139be INFO     ┝━ handle_admin_recover_account [ 17.6ms | 99.92% ] name: "admin"
e65941c1-81cf-4f67-82ed-06837a7139be ERROR    │  ┝━ 🚨 [error]: name denied by system configuration | name: "admin"
e65941c1-81cf-4f67-82ed-06837a7139be ERROR    │  ┝━ 🚨 [error]: Pre-Modify operation failed (plugin), ValueDenyName | event_tag_id: 1
e65941c1-81cf-4f67-82ed-06837a7139be ERROR    │  ┕━ 🚨 [error]:  | event_tag_id: 4 | error: ValueDenyName
e65941c1-81cf-4f67-82ed-06837a7139be ERROR    ┕━ 🚨 [error]: error during recover-account | err: ValueDenyName
```

which does at least help you figure out which foot you shot yourself in.

Kanidm is doing what I technically asked it to do, but it's not really what I _wanted_ to do. I made a mistake because I didn't properly understand the danger.

You can reverse the failure with the `idm_admin` account by removing `admin` from denied-names:

```
kanidm system denied-names remove admin
```

and the account recovery will then succeed.

### Kanidm version details

- version 1.4.6
- Are you running it in a container? If so, which image/tag?: Yes, built from kanidm:1.4.6

### Any other comments

It is my own silly fault, but perhaps we should gently try to dissuade people from blowing off their own feet like this.

Firstyear:
idm_admin and admin should be exempted from the denied names process, as these values will already be denied due to attribute uniqueness. Additionally improved the denied names check to only validate the name during a change, not during a modifification. This way entries that become denied can get themself out of the pickle.

Fixes #3426

Checklist

### DIFF
--- a/server/lib/src/plugins/valuedeny.rs
+++ b/server/lib/src/plugins/valuedeny.rs
@@ -10,7 +10,7 @@ impl Plugin for ValueDeny {
         "plugin_value_deny"
     }
 
-    #[instrument(level = "debug", name = "base_pre_create_transform", skip_all)]
+    #[instrument(level = "debug", name = "denied_names_pre_create_transform", skip_all)]
     #[allow(clippy::cognitive_complexity)]
     fn pre_create_transform(
         qs: &mut QueryServerWriteTransaction,
@@ -19,9 +19,24 @@ impl Plugin for ValueDeny {
     ) -> Result<(), OperationError> {
         let denied_names = qs.denied_names();
 
+        if denied_names.is_empty() {
+            // Nothing to check.
+            return Ok(());
+        }
+
         let mut pass = true;
 
         for entry in cand {
+            // If the entry doesn't have a uuid, it's invalid anyway and will fail schema.
+            if let Some(e_uuid) = entry.get_uuid() {
+                // SAFETY - Thanks to JpWarren blowing his nipper clean off, we need to
+                // assert that the break glass accounts are NOT subject to this process.
+                if e_uuid == UUID_ADMIN || e_uuid == UUID_IDM_ADMIN {
+                    // These entries are exempt
+                    continue;
+                }
+            }
+
             if let Some(name) = entry.get_ava_single_iname(Attribute::Name) {
                 if denied_names.contains(name) {
                     pass = false;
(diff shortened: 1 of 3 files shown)


server/lib/src/plugins/valuedeny.rs
```diff
@@ -19,9 +19,24 @@ impl Plugin for ValueDeny {
...
 
         for entry in cand {
+            // If the entry doesn't have a uuid, it's invalid anyway and will fail schema.
+            if let Some(e_uuid) = entry.get_uuid() {
+                // SAFETY - Thanks to JpWarren blowing his nipper clean off, we need to
+                // assert that the break glass accounts are NOT subject to this process.
+                if e_uuid == UUID_ADMIN || e_uuid == UUID_IDM_ADMIN {
+                    // These entries are exempt
+                    continue;
+                }
```
CEbbinghaus, line 37: I'd be nice if we also did a lookup on the account attributes to check in a more generic way whether they are system accounts or not. Happy to keep both as this guard is very cheap but It'd be nice to not run into this problem again when someone adds a third system account and forgets to update these lines

server/lib/src/plugins/valuedeny.rs
```diff
@@ -37,34 +52,37 @@ impl Plugin for ValueDeny {
...
         let mut results = Vec::with_capacity(0);
 
         for denied_name in denied_names {
-            let filt = filter!(f_eq(Attribute::Name, PartialValue::new_iname(&denied_name)));
+            let filt = filter!(f_and(vec![
+                f_eq(Attribute::Name, PartialValue::new_iname(&denied_name)),
+                f_andnot(f_or(vec![
+                    f_eq(Attribute::Uuid, PartialValue::Uuid(UUID_ADMIN)),
+                    f_eq(Attribute::Uuid, PartialValue::Uuid(UUID_IDM_ADMIN)),
+                ])),
```
CEbbinghaus, line 84: This one I'd replace with a pure Attribute check for system account

server/lib/src/plugins/valuedeny.rs
```diff
@@ -83,17 +101,37 @@ impl Plugin for ValueDeny {
...
-                if denied_names.contains(name) {
+        for (pre_entry, post_entry) in pre_cand.iter().zip(cand.iter()) {
+            // If the entry doesn't have a uuid, it's invalid anyway and will fail schema.
+            let e_uuid = pre_entry.get_uuid();
+            // SAFETY - Thanks to JpWarren blowing his nipper clean off, we need to
+            // assert that the break glass accounts are NOT subject to this process.
+            if e_uuid == UUID_ADMIN || e_uuid == UUID_IDM_ADMIN {
+                // These entries are exempt
+                continue;
+            }
```
CEbbinghaus, line 127: 🎗️ Another attribute check here plox

CEbbinghaus (changes requested):
Just the attribute check for type = systemaccount. Rest looks 👌

Firstyear:
@CEbbinghaus Good to go.

yaleman (approved):

CEbbinghaus (approved):
LGTM 👍

PR #3430, opened 2025-02-14
ci: uniform Docker builds

JadedBlueEyes:
# Change summary

This is the first part of breaking down https://github.com/kanidm/kanidm/pull/3380, as suggested. It's a few small changes to the docker build pipelines. From the top:

- Harden repo clones by setting `persist-credentials: false` as recommended by zizmor
- Consistently use the same `docker/metadata-action` invocation for each pipeline
- Apply labels & annotations from the metadata action
- Instruct build to build from the locally cloned repo, instead of cloning a second time
- Allow `push` job to run on forks
- Use the lowercase repo name in the `push` job

Checklist


.github/workflows/docker_build_kanidm.yml
```diff
@@ -78,4 +87,4 @@ jobs:
           echo "${{ secrets.GITHUB_TOKEN }}" | \
             oras login -u "${{ github.actor }}" --password-stdin ghcr.io
           oras copy --from-oci-layout "/tmp/kanidm-docker.tar:devel" \
-            "ghcr.io/${{ github.repository_owner }}/kanidm:devel"
+            "ghcr.io/${{ needs.set_tag_values.outputs.owner_lc }}/kanidm:devel"
```
CEbbinghaus, line 90: These are guaranteed to be the same?

CEbbinghaus (changes requested):
Looks pretty good overall. I'll let @yaleman chime in to approve.

.github/workflows/docker_build_kanidm.yml
yaleman: No they're not, `repository_owner` can have capitals - in fact your account would fail to build this without the lower-casing that the step does to it.

CEbbinghaus: So they are not the same but in a good way. Love this change ❤️

PR #3431, opened 2025-02-14
Fixes #3406: add configurable maximum queryable attributes for LDAP

### LINKED ISSUE #3406: Limit amount of attributes that can be queried to stop overuse of resources
Following on #67, specifically for [ldap.rs:242](https://github.com/kanidm/kanidm/blob/master/server/lib/src/idm/ldap.rs#L242).

We will want to limit the amount of attributes that can be queried.

Seba-T:
# Change summary

Added a new configuration setting to limit the number of LDAP attributes that can be queried in a single operation. 
The parameter falls back to the default value of 100 unless specified in the server configuration file. Perhaps a lower / higher default value would be better, as I have no clue what a regular LDAP query looks like :p. 
Also added a test to ensure the proper functionality of the feature. 

Fixes #3406 

Checklist


Firstyear:
@Seba-T This config needs to be an attribute on the domain info/system config. Not in a config file.

Think about if we have 3 replicas, you need to set this value on all of them. If they are different you get different behaviours. 

So that's why it has to be a value in the databaes, so the config replicates.

Firstyear (changes requested):
As above.

Seba-T:
Damn it did feel too easy. Well then database will be!

Seba-T:
Wait tho, how are people gonna set it if it's not *somewhere* in the config file?

Firstyear:
> Wait tho, how are people gonna set it if it's not _somewhere_ in the config file?

`kanidm system domain --help ` Might give you some ideas.

Seba-T:
Ohhh, I see.

Firstyear:
@Seba-T Yeah, maybe try setting up the server and administering a bit to get a feel of how things have already been done :)

CEbbinghaus:
Adding extra options into the schema is a complex process. We are always happy to receive contributions & @Firstyear will likely provide some guidance on how to implement it. But it's not a simple 5 line fix so if you aren't looking for that level of investment it may be better to hand over the PR. If you're up for the challenge though we'd gladly accept the help 😊

server/lib/src/idm/ldap.rs
```diff
@@ -267,6 +275,13 @@ impl LdapServer {
                         }
                     }
                 })
+            } else {
+                admin_error!(
+                    "Too many LDAP attributes requested. Maximum allowed is {}, while your search query had {}",
+                    self.max_queryable_attrs, attrs_len
+                );
+                return Err(OperationError::ResourceLimit);
+                // TODO: Should we return ResourceLimit or InvalidRequestState here?
```
Firstyear, line 284: ResourceLimit is correct.

server/testkit/tests/integration.rs
```diff
@@ -231,6 +231,19 @@ async fn test_idm_domain_set_ldap_basedn(rsclient: KanidmClient) {
...
 
+#[kanidmd_testkit::test]
+async fn test_idm_domain_set_ldap_max_queryable_attrs(rsclient: KanidmClient) {
+    login_put_admin_idm_admins(&rsclient).await;
+    assert!(rsclient
+        .idm_domain_set_ldap_max_queryable_attrs(20)
+        .await
+        .is_ok());
+    assert!(rsclient
+        .idm_domain_set_ldap_max_queryable_attrs(10) //TODO: Help, Rust's type safety is so good I can't come up with a way to pass an invalid value
```
Firstyear, line 242: You don't have to, invalid value parsing is checked extensively elsewhere.

server/lib/src/constants/schema.rs
```diff
@@ -1227,6 +1242,7 @@ pub static ref SCHEMA_CLASS_DOMAIN_INFO_DL10: SchemaClass = SchemaClass {
     systemmay: vec![
         Attribute::DomainSsid,
         Attribute::DomainLdapBasedn,
+        Attribute::LdapMaxQueryableAttrs,
```
Firstyear, line 1245: This is the only one you need, the other versions are *past* schema versions of the domain info, and you shouldn't touch them. You only should touch the DL10 version. This will be clearer after another PR I have open will merge.

server/lib/src/constants/schema.rs
```diff
@@ -1203,6 +1217,7 @@ pub static ref SCHEMA_CLASS_DOMAIN_INFO_DL9: SchemaClass = SchemaClass {
...
```
Firstyear, line 1220: In other words remove this and the above ^

Firstyear (changes requested):
Looks really good, almost there!

server/lib/src/constants/schema.rs
Seba-T: oki!

server/lib/src/server/migrations.rs
```diff
@@ -519,6 +519,7 @@ impl QueryServerWriteTransaction<'_> {
             // SCHEMA_ATTR_DISPLAYNAME.clone().into(),
             SCHEMA_ATTR_DOMAIN_DISPLAY_NAME.clone().into(),
             SCHEMA_ATTR_DOMAIN_LDAP_BASEDN.clone().into(),
+            SCHEMA_ATTR_LDAP_MAXIMUM_QUERYABLE_ATTRIBUTES.clone().into(),
```
Firstyear, line 522: Wrong place! See the bit that says DO NOT MODIFY? 

Needs to be in "migrate_domain_9_to_10".

proto/src/constants.rs
```diff
@@ -39,6 +39,8 @@ pub const DEFAULT_SERVER_ADDRESS: &str = "127.0.0.1:8443";
 pub const DEFAULT_SERVER_LOCALHOST: &str = "localhost:8443";
 /// The default LDAP bind address for the Kanidm client
 pub const DEFAULT_LDAP_LOCALHOST: &str = "localhost:636";
+/// The default amount of attributes that can be queried in LDAP
+pub const DEFAULT_LDAP_MAXIMUM_QUERYABLE_ATTRIBUTES: usize = 100;
```
Firstyear, line 43: I feel like 16 is a more reasonable default, we can always bump it up later.

Firstyear (changes requested):

proto/src/constants.rs
CEbbinghaus: I recon it should be over 9000

server/lib/src/server/migrations.rs
Seba-T: > See the bit that says DO NOT MODIFY?

Definitely not :p 
But now I do and everything should be in the appropriate place.

proto/src/constants.rs
Seba-T: sure!

Firstyear (approved):

PR #3432, opened 2025-02-15
20250213 patch used wrong acp

### LINKED ISSUE #3270: idm_admin cannot modify entry-mananged-by attribute

Firstyear:
Major refactor of migrations to improve durability.

    Migrations and server bootstrap are very interconnected processes
    and in this we'll be addressing and improving both.

    Server bootstrap was performed by creating base entries in phases,
    eventually bringing up enough of the *oldest* supported server
    minimum remigration level, to then allow triggering of migrations.

    Migrations then applied "patches" effectively ontop of this minimum
    level to update entries to what they should be in newer versions of
    the server.

    This scheme has it's pros and cons, but the major con was that to
    remove a migration meant squashing it's content back into the
    minimum remigration level, and this was a human process that was
    quite error prone and difficult to automate. As well, this scheme
    also led to cases where the patch migrations would sometimes *not*
    reflect all the needed changes or content, or in one case was actually
    undone by a patchlevel fix up that was required to address a bug.

    Invariably this led to issues, and cases where a new server may have
    different content to a migrated one - not exactly what we want!

    This is a new migration scheme that addresses this fragility. However
    what it trades is verbosity of the content.

    Rather than having a base set of entries and patching/updating small
    sections ontop, we have migration data folders that contain the full
    set of entries as they should appear at that migration level. This
    makes the bootstrap process easier as we can just apply the migration
    level as a whole, and targetted to what precise version we want.

    This also makes migrations more durable as the content is explicitly
    copied and all entries fully applied, so there is no risk that a
    migration or data change can be forgotten or applied incorrectly. We
    are expressing the full state of what our builtin and provided entries
    should be.

    Finally this rips out a number of places where migration data was being
    used as test case data. Not all of these have been replaced (notably
    in authsession with Account), but the majority have and have been replaced
    with clearer use of constants rather than building whole entries just to
    access the name and throw them away for example.

Fixes #3270

Checklist


Firstyear:
Two bonus commits that make test execution ~10% faster, and significantly faster to compile.

server/lib/src/migration_data/dl8/access.rs
```diff
@@ -0,0 +1,2637 @@
...
+    /// is sufficient for you to receive the access control.
+    Group(Vec<Uuid>),
+    EntryManager,
+}
+
+#[derive(Clone, Debug, Default)]
+/// Objects that are affected by the rules of this ACP.
+pub enum BuiltinAcpTarget {
+    #[default]
+    None,
```
CEbbinghaus, line 56: Just wondering. Wouldn't we want a invalid target to be unrepresentative? Why define None

server/lib/src/migration_data/dl8/schema.rs
```diff
@@ -0,0 +1,1506 @@
...
+    name: Attribute::DisplayName,
+    description: "The publicly visible display name of this person".to_string(),
+
+    index: vec![IndexType::Equality],
+    sync_allowed: true,
+    syntax: SyntaxType::Utf8String,
+    ..Default::default()
+};
+
+pub static ref SCHEMA_ATTR_DISPLAYNAME_DL7: SchemaAttribute = SchemaAttribute {
```
CEbbinghaus, line 21: `DL8` if anything. But now that they have been moved into seperate modules we shouldn't need the naming right? would be `dl8::SCHEMA_ATTR_DISPLAYNAME`

This also goes for the other Domain Levels

server/lib/src/migration_data/dl8/system_config.rs
```diff
@@ -0,0 +1,1073 @@
...
+            Attribute::BadlistPassword,
+            Value::new_iutf8("100preteamare")
+        ),
+        (Attribute::BadlistPassword, Value::new_iutf8("14defebrero")),
+        (Attribute::BadlistPassword, Value::new_iutf8("1life1love")),
+        (Attribute::BadlistPassword, Value::new_iutf8("1life2live")),
+        (Attribute::BadlistPassword, Value::new_iutf8("1love1life")),
+        (Attribute::BadlistPassword, Value::new_iutf8("1love4life")),
+        (Attribute::BadlistPassword, Value::new_iutf8("212224236248")),
+        (Attribute::BadlistPassword, Value::new_iutf8("2813308004")),
```
CEbbinghaus, line 65: Can we codegen these from a common file? during build read all lines from the file and turn each into 1 attribute?

server/lib/src/server/migrations.rs
```diff
@@ -182,9 +176,6 @@ impl QueryServer {
         // we would have skipped the patch level fix which needs to have occurred *first*.
         if reload_required {
             write_txn.reload()?;
-            // We are not yet at the schema phase where reindexes will auto-trigger
-            // so if one was required, do it now.
-            write_txn.reindex(false)?;
```
CEbbinghaus, line 187: we no long er need to reindex?

server/lib/src/server/migrations.rs
```diff
@@ -294,6 +302,75 @@ impl QueryServerWriteTransaction<'_> {
         }
     }
 
+    /// Migration domain level 7 to 8 (1.4.0)
+    #[instrument(level = "info", skip_all)]
+    pub(crate) fn migrate_domain_7_to_8(&mut self) -> Result<(), OperationError> {
```
CEbbinghaus, line 307: Not really for this one but I wonder if we can't codegen these methods via a macro rule. would take a little bit of the repetition out especially given that all the methods within the module have the same name

server/testkit/defunct_examples/enumerating_access.rs
```diff
@@ -4,7 +4,9 @@
 //!
 
 use kanidmd_lib::constants::entries::Attribute;
-use kanidmd_lib::constants::groups::{idm_builtin_admin_groups, idm_builtin_non_admin_groups};
+use kanidmd_lib::migration_data::current::{
```
CEbbinghaus, line 7: I feel like I'm crazy, but I can't find where the 'current' module is
image

server/testkit/src/lib.rs
```diff
@@ -385,7 +385,7 @@ pub async fn login_put_admin_idm_admins(rsclient: &KanidmClient) {
 
     #[allow(clippy::expect_used)]
     rsclient
-        .idm_group_add_members(BUILTIN_GROUP_IDM_ADMINS_V1.name, &[ADMIN_TEST_USER])
+        .idm_group_add_members("system_admins", &[ADMIN_TEST_USER])
```
CEbbinghaus, line 388: why not use the `NAME_SYSTEM_ADMINS` constant?

CEbbinghaus (approved):
Oh boy that +-. I like where we are heading but I think I'd like to codegen a little more to get rid of some of the boilerplate.

Nothing stopping the PR except some parts I'm unsure about and some magic strings that I would prefer to just pull in from constants.

Note that I did not verify the correctness of the schema and am instead relying on you having copy/pasted all the correct properties for each DL. This was mostly verifying the soundness of the logic surrounding that change.

server/lib/src/migration_data/dl8/schema.rs
Firstyear: yes, but I just didn't prune all these. They'll age out naturally in time :)

server/lib/src/migration_data/dl8/system_config.rs
Firstyear: Probably but that should be it's own PR I think. That's it's own block of work.

server/lib/src/server/migrations.rs
Firstyear: No, we don't :) I carefully checked this flow, and we don't need it.

server/lib/src/server/migrations.rs
Firstyear: Yes, we could, but then if we change one migration we have to then duplicate the macro again, and it has it's own issues. I think for now we do it the correct way, then we simplify, then optimise. It may be stupid, but it's correct.

server/testkit/defunct_examples/enumerating_access.rs
Firstyear: The example is defunct, the current module doesn't exist. But this example was just too broken in the current work..

server/lib/src/migration_data/dl8/access.rs
Firstyear: Because we model this in the ACP code underneath, I don't remember why I added it like this, but there was probably a reason, most likely so we could have a safe fallback in a parse error of some kind such that the acp will never apply.

Firstyear: I seem to recall some migration related reason for it too ....

Firstyear: Ahh found it. It's so that if a filter in the access control has no matches then we map it to a none so that it can clearly indicate no one actually gains the access, and we can shortcut out.

Firstyear: And that then allows optimisation/caching of resolved access profiles which only map to conditions that apply to someone/something.

server/lib/src/migration_data/dl8/schema.rs
CEbbinghaus: They'll age like milk. Just want to make sure that we don't set a bad precident for the next set of schema stuff :P

Firstyear: I already cleaned up dl10 is more the point.

server/lib/src/migration_data/dl8/system_config.rs
CEbbinghaus: Yeah easy. Not wanting it in this PR. Just want it done eventually

CEbbinghaus (approved):
LGTM 👍

PR #3434, opened 2025-02-15
book: add OAuth2 Proxy example

newAM:
# Change summary

Adds an example for using Kanidm with [OAuth2 Proxy](https://github.com/oauth2-proxy/oauth2-proxy).

Checklist

### DIFF
--- a/book/src/integrations/oauth2/examples.md
+++ b/book/src/integrations/oauth2/examples.md
@@ -556,6 +556,65 @@ php occ config:app:set --value=0 user_oidc allow_multiple_user_backends
 You can login directly by appending `?direct=1` to your login page. You can re-enable other backends
 by setting the value to `1`
 
+## OAuth2 Proxy
+
+OAuth2 Proxy is a reverse proxy that provides authentication with OpenID Connect identity providers.
+It is typically used to secure web applications without native OpenID Connect support.
+
+Prepare the environment.
+Due to a [lack of public client support](https://github.com/oauth2-proxy/oauth2-proxy/issues/1714) we have to set it up as a basic client. 
+
+```bash
+kanidm system oauth2 create webapp 'webapp.example.com' 'https://webapp.example.com'
+kanidm system add-redirect-url webapp 'https://webapp.example.com/oauth2/callback'
+kanidm system oauth2 update-scope-map webapp email openid
+kanidm system oauth2 get webapp
+kanidm system oauth2 show-basic-secret webapp
+<SECRET>
+```
+
+Create a user group.
+
+```bash
+kanidm group create 'webapp_admin'
+```
+
+Setup the claim-map to add `webapp_group` to the userinfo claim.
+
+```bash
+kanidm system oauth2 update-claim-map-join 'webapp' 'webapp_group' array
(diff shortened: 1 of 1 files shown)


book/src/integrations/oauth2/examples.md
```diff
@@ -556,6 +556,64 @@ php occ config:app:set --value=0 user_oidc allow_multiple_user_backends
 You can login directly by appending `?direct=1` to your login page. You can re-enable other backends
 by setting the value to `1`
 
+## OAuth2 Proxy
+
+OAuth2 Proxy is a reverse proxy that provides authentication with OpenID Connect identity providers.
+It is typically used to secure web applications without native OpenID Connect support.
+
+Prepare the environment.
```
yaleman, line 564:
```suggestion
Prepare the environment. Due to a [lack of public client support](https://github.com/oauth2-proxy/oauth2-proxy/issues/1714) we have to set it up as a basic client. 
```

yaleman (changes requested):
Minor tweak to link/explain why we can't follow best practice.

yaleman (approved):

newAM:
> Minor tweak to link/explain why we can't follow best practice.

At the time this made sense to me, because I thought PKCE was only for public clients. That was a misunderstanding on my part; PKCE is required for public clients, and optional (but recommended) for confidential clients.

---

I have some confusion about best practice because the Kanidm CLI states confidential clients should be preferred:

https://github.com/kanidm/kanidm/blob/9bf17c4846ba682ffa778db0b919fa338168170d/tools/cli/src/opt/kanidm.rs#L1063-L1067

---

I read the [OAuth 2.0 Security Best Current Practice RFC](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-security-topics-16)  looking for answers, but it doesn't have recommendations on public vs confidential.

I did find that PKCE is best practice for confidential clients which makes sense:

> Clients MUST prevent injection (replay) of authorization codes into the authorization response by attackers. Public clients MUST use PKCE [[RFC7636](https://www.rfc-editor.org/info/rfc7636)] to this end. For confidential clients, the use of PKCE [[RFC7636](https://www.rfc-editor.org/info/rfc7636)] is RECOMMENDED.

And some attacks are more difficult to execute against confidential clients:

> This attack will not work as easily for confidential clients, since the code exchange requires authentication with the legitimate client's secret. The attacker can, however, use the legitimate confidential client to redeem the code by performing an authorization code injection attack

---

Is there a reason public clients are preferred in this case?
Sorry for the long followup question to a simple PR :sweat_smile:

Firstyear:
I don't think they were "preferred" in this case, I think it's just that the software in question doesn't support them so we can't give people the choice between the public vs confidential client.

iridial:
@newAM any tips on how to best handle user sign out? in oauth2-proxy docs there is the `backend_logout_url` config
https://oauth2-proxy.github.io/oauth2-proxy/configuration/overview#general-provider-options

or maybe with a redirect to kanidm in the url?  https://oauth2-proxy.github.io/oauth2-proxy/features/endpoints/#sign-out

newAM:
I haven't looked into that yet.  I'm only using OA2P to secure a Prometheus WebUI, which is only accessed on trusted devices.

Firstyear:
We have short-lived tokens, as backchannel logout is a bit of a mess.

PR #3442, opened 2025-02-19
doc: fix formatting of URL table, remove Caddyfile instructions

micolous:
# Change summary

Fixes broken formatting of OAuth2 URLs definition list:

Screenshot 2025-02-19 at 10 24 38

This also removes the `Caddyfile`-specific instructions for rewriting URLs, because it bloats the rest of the definition list.

I have some other thoughts about WebFinger – it makes a number of assumptions about OAuth2 / OpenID client configuration which don't hold true for Kanidm. But I will leave that for a later PR. 😄 

## Checklist

### DIFF
--- a/book/src/integrations/oauth2.md
+++ b/book/src/integrations/oauth2.md
@@ -70,30 +70,21 @@ anything special for Kanidm (or another provider).
 **Note:** some apps automatically append `/.well-known/openid-configuration` to
 the end of an OIDC Discovery URL, so you may need to omit that.
 
+<dt>
 
-<dl>
-<dt>[Webfinger](https://datatracker.ietf.org/doc/html/rfc7033) URL</dt>
+[WebFinger](https://datatracker.ietf.org/doc/html/rfc7033) URL
+
+</dt>
 
 <dd>
 
 `https://idm.example.com/oauth2/openid/:client_id:/.well-known/webfinger`
 
-The webfinger URL is implemented for each OpenID client, under its specific endpoint, giving full control to the administrator regarding which to use.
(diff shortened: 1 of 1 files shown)


Firstyear:
Yes, webfinger seems like a half baked mess, and the users of it tend also to the same.

Firstyear (approved):

book/src/integrations/oauth2.md
```diff
@@ -70,31 +70,22 @@ anything special for Kanidm (or another provider).
...
-The webfinger URL is implemented for each OpenID client, under its specific endpoint, giving full control to the administrator regarding which to use.
+The WebFinger URL is implemented for each OpenID client, under its specific endpoint, giving full control to the administrator regarding which to use.
 
 To make this compliant with the standard, it must be made available under the correct [well-known endpoint](https://datatracker.ietf.org/doc/html/rfc7033#section-10.1) (e.g `example.com/.well-known/webfinger`), typically via a reverse proxy or similar. Kanidm doesn't currently provide a mechanism for this URI rewrite.
 
 One example would be dedicating one client as the "primary" or "default" and redirecting all requests to that. Alternatively, source IP or other request metadata could be used to decide which client to forward the request to.
 
-### Caddy
-`Caddyfile`
-```caddy
```
yaleman, line 89: The CaddyFile is here to explain around the whole "it's in the wrong base URL, here's how to make it so" thing.

micolous: This breaks the visual flow of the rest of the definition list, which is why I removed that. There's also several limitations with the spec which means there's no one size fits all approach to the problem.

I'm working on a more extensive rewrite of this section, which will address this.

PR #3443, opened 2025-02-19
Rewrite WebFinger docs

micolous:
# Change summary

Follow-up from #3442

This rewrites the WebFinger docs to go over the various limitations of the protocol when using Kanidm, and explicitly discourage its use. Because there are a lot of them, I've split it out into a separate section.

Checklist

### DIFF
--- a/book/src/integrations/oauth2.md
+++ b/book/src/integrations/oauth2.md
@@ -72,7 +72,20 @@ the end of an OIDC Discovery URL, so you may need to omit that.
 
 <dt>
 
-[WebFinger](https://datatracker.ietf.org/doc/html/rfc7033) URL
+[RFC 8414 OAuth 2.0 Authorisation Server Metadata](https://datatracker.ietf.org/doc/html/rfc8414)
+URL **(recommended)**
+
+</dt>
+
+<dd>
+
+`https://idm.example.com/oauth2/openid/:client_id:/.well-known/oauth-authorization-server`
+
+</dd>
+
+<dt>
+
+[WebFinger URL **(discouraged)**](#webfinger)
 
 </dt>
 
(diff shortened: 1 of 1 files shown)


Firstyear (approved):

book/src/integrations/oauth2.md
```diff
@@ -457,3 +455,61 @@ kanidm system oauth2 reset-secrets
...
+* Kanidm responds to *all* WebFinger queries with
+  [an Identity Provider Discovery for OIDC URL](https://datatracker.ietf.org/doc/html/rfc7033#section-3.1),
+  **regardless** of what
+  [`rel` parameter](https://datatracker.ietf.org/doc/html/rfc7033#section-4.4.4.1)
+  was specified.
+
+  This is to work around
+  [a broken client](https://tailscale.com/kb/1240/sso-custom-oidc) which doesn't
+  send a `rel` parameter, but expects an Identity Provider Discovery issuer URL
+  in response.
```
CEbbinghaus, line 506: It is actually not. It is part of the spec that `rel` parameters can be ignored as long as the server responds with all `rel`'s it supports. This simplifies the design greatly and allows us to remain compliant. (If no rel parameters are specified the server is supposed to send all rel's it supports anyways so Tailscale isn't broken)

book/src/integrations/oauth2.md
```diff
@@ -457,3 +455,61 @@ kanidm system oauth2 reset-secrets
...
+
+  If you want to use WebFinger in any *other* context on Kanidm's hostname,
+  you'll need a load balancer in front of Kanidm which matches on some property
+  of the request.
+
+Because of the flaws of the WebFinger specification and the deployment
+difficulties they introduce, we recommend that applications use OpenID Connect
+Discovery or OAuth 2.0 Authorisation Server Metadata for client configuration
+instead of WebFinger.
```
CEbbinghaus, line 515: Very true, but unfortunately our admins rarely (if ever) have ownership over the client.

CEbbinghaus:
Additionally, It would be nice to retain the caddy example (especially the `{uri}` part). 

The RFC does not explicitly say clients must support redirects, but it does have an example returning a 307 redirect.

micolous:
Will address in a follow-up. :)

book/src/integrations/oauth2.md
micolous: You're correct, but [per the spec](https://datatracker.ietf.org/doc/html/rfc7033#section-4.3), it's sub-optimal client behaviour:

> The purpose of the "rel" parameter is to return a subset of "link relation objects" (see [Section 4.4.4](https://datatracker.ietf.org/doc/html/rfc7033#section-4.4.4)) that would otherwise be returned in the resource descriptor.  Use of the parameter might reduce processing requirements on either the client or server, and it might also reduce the bandwidth required to convey the partial resource descriptor, especially if there are numerous link relation values to convey for a given "resource" value.

If a client only really cares about a subset of related entities, it should request the subset.  😄

book/src/integrations/oauth2.md
micolous: TBH, after looking at Tailscale too closely, I'm not a fan of it. For OIDC, it makes many incorrect assumptions that are _explicitly called out_ in the core specification, which depending on how you use or configure Tailscale and which IdP you use, could lead to exploitable security weaknesses.

That said, I don't know of anything better. 🤷

book/src/integrations/oauth2.md
CEbbinghaus: Not saying tailscale wrote the best client. Just that it's not breaking spec and, not something we are working around.

PR #3446, opened 2025-02-19
address webfinger feedbacks

micolous:
# Change summary

Follow up from #3442 #3443 

Checklist

### DIFF
--- a/book/src/integrations/oauth2.md
+++ b/book/src/integrations/oauth2.md
@@ -460,7 +460,7 @@ Each client has unique signing keys and access secrets, so this is limited to ea
 
 [WebFinger](https://datatracker.ietf.org/doc/html/rfc7033) provides a mechanism
 for discovering information about entities at a well-known URL
-(`http://example.com/.well-known/webfinger`).
+(`https://{hostname}/.well-known/webfinger`).
 
 It can be used by a WebFinger client to
 [discover the OIDC issuer URL](https://datatracker.ietf.org/doc/html/rfc7033#section-3.1)
@@ -482,11 +482,11 @@ difficult to use with Kanidm:
 
   Kanidm uses *different* `iss` (issuer), signing keys, and some client-specific
   endpoint URLs, which ensures that tokens can only be used with their intended
-  service. *Changing this behaviour would reduce Kanidm's security.*
+  service.
 
 * WebFinger endpoints must be served at the *root* of the domain of a user's
   SPN (ie: information about the user with SPN `user@idm.example.com` is at
-  `https://idm.example.com/.well-known/webfinger?resource=acct%3Auser%40idm.example.com&rel=...`).
+  `https://idm.example.com/.well-known/webfinger?resource=acct%3Auser%40idm.example.com`).
 
   Unlike OIDC Discovery, WebFinger clients do not report their OAuth 2.0/OIDC
   client ID in the request, so there is no way to tell them apart.
--- a/book/src/integrations/oauth2.md
+++ b/book/src/integrations/oauth2.md
@@ -458,20 +458,19 @@ Each client has unique signing keys and access secrets, so this is limited to ea
 
 ## WebFinger
 
-[WebFinger](https://datatracker.ietf.org/doc/html/rfc7033) provides a mechanism
-for discovering information about entities at a well-known URL
-(`https://{hostname}/.well-known/webfinger`).
+[WebFinger][webfinger] provides a mechanism for discovering information about
+entities at a well-known URL (`https://{hostname}/.well-known/webfinger`).
 
 It can be used by a WebFinger client to
-[discover the OIDC issuer URL](https://datatracker.ietf.org/doc/html/rfc7033#section-3.1)
-of an identity provider from the hostname alone, and seems to be intended to
-support dynamic client registration flows for large public identity providers.
+[discover the OIDC issuer URL][webfinger-oidc] of an identity provider from the
+hostname alone, and seems to be intended to support dynamic client registration
+flows for large public identity providers.
 
 Kanidm v1.5.1 and later can respond to WebFinger requests, using a user's SPN as
-[an `acct` URI][rfc7565] (eg: `acct:user@idm.example.com`).
-[As per RFC 7565][rfc7565s4], there is no guarantee that account IDs are valid
-for any particular application protocol (such as email), unless an administrator
-explicitly provides it.
+part of [an `acct` URI][rfc7565] (eg: `acct:user@idm.example.com`). While SPNs
+and `acct` URIs look like email addresses, [as per RFC 7565][rfc7565s4], there
+is no guarantee that it is valid for any particular application protocol, unless
+an administrator explicitly provides for it.
 
 When setting up an application to authenticate with Kanidm, WebFinger **does not
 add any security** over configuring an OpenID Discovery URL directly. In an OIDC
@@ -515,22 +514,28 @@ difficult to use with Kanidm:
   to a client ID, and redirect to the appropriate WebFinger URL for that client.
 
 * Kanidm responds to *all* WebFinger queries with
-  [an Identity Provider Discovery for OIDC URL](https://datatracker.ietf.org/doc/html/rfc7033#section-3.1),
-  **ignoring** any supplied
-  [`rel` parameter](https://datatracker.ietf.org/doc/html/rfc7033#section-4.3).
+  [an Identity Provider Discovery for OIDC URL][webfinger-oidc], **ignoring**
+  any supplied [`rel` parameter][webfinger-rel].
 
   If you want to use WebFinger in any *other* context on Kanidm's hostname,
   you'll need a load balancer in front of Kanidm which matches on some property
   of the request.
 
-  WebFinger clients *may* omit the `rel=` parameter, so if another service has
-  relations for an `acct:` entity and a client *does not* supply the `rel=`
-  parameter, your load balancer will need to merge JSON responses from Kanidm
-  and the other service(s).
+  WebFinger clients *may* omit the `rel=` parameter, so if you host another
+  service with relations for a Kanidm [`acct:` entity][rfc7565s4] and a client
+  *does not* supply the `rel=` parameter, your load balancer will need to merge
+  JSON responses from Kanidm and the other service(s).
 
 Because of these issues, we recommend that applications support *directly*
 configuring OIDC using a Discovery URL or OAuth 2.0 Authorisation Server
 Metadata URL instead of WebFinger.
 
+If a WebFinger client only checks WebFinger once during setup, you may wish to
+temporarily serve an appropriate static WebFinger document for that client
+instead.
+
 [rfc7565]: https://datatracker.ietf.org/doc/html/rfc7565
 [rfc7565s4]: https://datatracker.ietf.org/doc/html/rfc7565#section-4
+[webfinger]: https://datatracker.ietf.org/doc/html/rfc7033
+[webfinger-oidc]: https://datatracker.ietf.org/doc/html/rfc7033#section-3.1
+[webfinger-rel]: https://datatracker.ietf.org/doc/html/rfc7033#section-4.3
--- a/book/src/integrations/oauth2.md
+++ b/book/src/integrations/oauth2.md
@@ -162,7 +162,7 @@ Token endpoint
 
 <dt>
 
-OpenID Connect issuer URI
+OpenID Connect Issuer URL
 
 </dt>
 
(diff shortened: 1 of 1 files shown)


book/src/integrations/oauth2.md
```diff
@@ -458,58 +458,78 @@ Each client has unique signing keys and access secrets, so this is limited to ea
...
-response to all incoming WebFinger requests, using a user's SPN as their account
-ID. This does not match on email addresses as they are not guaranteed to be
-unique.
+Kanidm v1.5.1 and later can respond to WebFinger requests, using a user's SPN as
+the account (eg: `user@idm.example.com`). This *does not* match on email
+addresses, because they are not required by Kanidm nor guaranteed to be unique.
 
-However, WebFinger has a number of flaws which make it difficult to use with
-Kanidm:
+When setting up an (enterprise) application to authenticate with Kanidm,
```
yaleman, line 478:
```suggestion
When setting up an application to authenticate with Kanidm,
```

Doesn't matter what kind of application.

book/src/integrations/oauth2.md
```diff
@@ -458,58 +458,78 @@ Each client has unique signing keys and access secrets, so this is limited to ea
...
-Because of the flaws of the WebFinger specification and the deployment
-difficulties they introduce, we recommend that applications use OpenID Connect
-Discovery or OAuth 2.0 Authorisation Server Metadata for client configuration
-instead of WebFinger.
+  WebFinger clients *may* omit the `rel=` parameter, so if another service has
+  relations for an `acct:` entity and a client *does not* supply the `rel=`
+  parameter, your load balancer will need to merge JSON responses from Kanidm
+  and the other service(s).
+
+Because of these issues, we recommend that (enterprise) applications support
```
yaleman, line 533:
```suggestion
Because of these issues, we recommend that applications support
```

yaleman (changes requested):
Mentioning the required version twice in the same screenful, when the docs don't do it anywhere else, for any other feature seems unnecessary - I'd say remove the big note block.

book/src/integrations/oauth2.md
```diff
@@ -458,58 +458,78 @@ Each client has unique signing keys and access secrets, so this is limited to ea
...
 
+> [!NOTE]
+>
+> WebFinger support requires Kanidm v1.5.1 or later.
+
 [WebFinger](https://datatracker.ietf.org/doc/html/rfc7033) provides a mechanism
-for discovering information about people or other entities. It can be used by an
-identity provider to supply OpenID Connect discovery information.
+for discovering information about entities at a well-known URL
+(`http://example.com/.well-known/webfinger`).
```
CEbbinghaus, line 467: [Hostname] instead of example.com. else we are implying the well known URL has to be under example.com

book/src/integrations/oauth2.md
```diff
@@ -458,58 +458,78 @@ Each client has unique signing keys and access secrets, so this is limited to ea
...
+[discover the OIDC issuer URL](https://datatracker.ietf.org/doc/html/rfc7033#section-3.1)
+of an identity provider from the hostname alone, and seems to be intended to
+support dynamic client registration flows for large public identity providers.
 
-Kanidm provides
-[an Identity Provider Discovery for OIDC URL](https://datatracker.ietf.org/doc/html/rfc7033#section-3.1)
-response to all incoming WebFinger requests, using a user's SPN as their account
-ID. This does not match on email addresses as they are not guaranteed to be
-unique.
+Kanidm v1.5.1 and later can respond to WebFinger requests, using a user's SPN as
```
CEbbinghaus, line 474: Agree with yaleman. No need to add the version a second time. We already point it out above

book/src/integrations/oauth2.md
```diff
@@ -458,58 +458,78 @@ Each client has unique signing keys and access secrets, so this is limited to ea
...
+the account (eg: `user@idm.example.com`). This *does not* match on email
+addresses, because they are not required by Kanidm nor guaranteed to be unique.
```
CEbbinghaus, line 476:
```suggestion
the account (eg: `user@idm.example.com`). This may not match their email
address possibly leading to problems in the client. Because the email attribute is not required by Kanidm nor guaranteed to be unique, it is up to the administrator to ensure that it matches the spn if the service requires that.
```

book/src/integrations/oauth2.md
```diff
@@ -458,58 +458,78 @@ Each client has unique signing keys and access secrets, so this is limited to ea
...
 * WebFinger assumes that the identity provider will give the same `iss`
-  (Issuer) for every OAuth 2.0/OIDC client, and there is no standard way for a
-  WebFinger client to report its client ID.
+  (issuer) and OpenID Discovery document, including all URLs and signing keys,
+  for *all* OAuth 2.0/OIDC clients.
 
-  Kanidm uses a *different* `iss` (Issuer) value for each client.
+  Kanidm uses *different* `iss` (issuer), signing keys, and some client-specific
+  endpoint URLs, which ensures that tokens can only be used with their intended
+  service. *Changing this behaviour would reduce Kanidm's security.*
```
CEbbinghaus, line 489:
```suggestion
  service.
```

There is no way kani is changing this so I wouldn't even mention changing it.

CEbbinghaus (changes requested):
Looks good overall. Just some small comments

book/src/integrations/oauth2.md
```diff
@@ -458,58 +458,78 @@ Each client has unique signing keys and access secrets, so this is limited to ea
...
 
-Because of the flaws of the WebFinger specification and the deployment
-difficulties they introduce, we recommend that applications use OpenID Connect
-Discovery or OAuth 2.0 Authorisation Server Metadata for client configuration
-instead of WebFinger.
+  WebFinger clients *may* omit the `rel=` parameter, so if another service has
+  relations for an `acct:` entity and a client *does not* supply the `rel=`
+  parameter, your load balancer will need to merge JSON responses from Kanidm
+  and the other service(s).
+
```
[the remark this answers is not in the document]
micolous: Depends on the client, really.

In the Tailscale case, it seems like they only check WebFinger during setup, then assume that a once-valid WebFinger claim for an account ID is a valid claim for the whole domain for all time. So you could absolutely host a static file for a few minutes, then delete it.

If a client checks WebFinger on each login attempt (eg: it has a login form which prompts for an email-like username, then attempts to pick an authentication methods based on a WebFinger response), then it'll need to stay up, and provide a correct account ID in the response.

And yes, this sucks!

book/src/integrations/oauth2.md
micolous: I did some more reading, I was incorrect to suggest that's because of how Kanidm works with email addresses. [`acct` is explicitly "maybe not email address"](https://datatracker.ietf.org/doc/html/rfc7565):

> It is not assumed that an entity will necessarily be able to interact with a user's account using any particular application protocol, such as email

However, I strongly suspect that some clients will assume a vaguely email address shaped account ID *is* a valid email address, so I'll update this.

book/src/integrations/oauth2.md
micolous: Done

book/src/integrations/oauth2.md
micolous: Done

book/src/integrations/oauth2.md
micolous: Done

book/src/integrations/oauth2.md
```diff
@@ -458,58 +458,78 @@ Each client has unique signing keys and access secrets, so this is limited to ea
...
-  Kanidm uses a *different* `iss` (Issuer) value for each client.
+  Kanidm uses *different* `iss` (issuer), signing keys, and some client-specific
+  endpoint URLs, which ensures that tokens can only be used with their intended
+  service. *Changing this behaviour would reduce Kanidm's security.*
 
-* WebFinger requires that this be served at the *root* of the domain of a user's
+* WebFinger endpoints must be served at the *root* of the domain of a user's
   SPN (ie: information about the user with SPN `user@idm.example.com` is at
-  `https://idm.example.com/.well-known/webfinger`).
+  `https://idm.example.com/.well-known/webfinger?resource=acct%3Auser%40idm.example.com&rel=...`).
```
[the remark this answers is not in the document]
micolous: Done

book/src/integrations/oauth2.md
micolous: Done

book/src/integrations/oauth2.md
micolous: Done

book/src/integrations/oauth2.md
micolous: Now reworded in light of the above.

book/src/integrations/oauth2.md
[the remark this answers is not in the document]
CEbbinghaus: Isn't mean temporarily hosting a static file. I was referring to a permanent static file containing the webfinger URLs for all kinds of rels including profile picture, blog & contact.

Its pretty unconventional so its probably best not to include it.

book/src/integrations/oauth2.md
```diff
@@ -459,57 +459,78 @@ Each client has unique signing keys and access secrets, so this is limited to ea
...
-ID. This does not match on email addresses as they are not guaranteed to be
-unique.
+It can be used by a WebFinger client to
+[discover the OIDC issuer URL](https://datatracker.ietf.org/doc/html/rfc7033#section-3.1)
+of an identity provider from the hostname alone, and seems to be intended to
+support dynamic client registration flows for large public identity providers.
 
-However, WebFinger has a number of flaws which make it difficult to use with
-Kanidm:
+Kanidm v1.5.1 and later can respond to WebFinger requests, using a user's SPN as
```
CEbbinghaus, line 469: I preferred the version in a little note at the top so admins don't need to read the entire text to know what version they need to update to

book/src/integrations/oauth2.md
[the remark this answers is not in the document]
micolous: Added in a suggestion to serve statically.

book/src/integrations/oauth2.md
```diff
@@ -458,58 +458,84 @@ Each client has unique signing keys and access secrets, so this is limited to ea
...
-  was specified.
-
-  This is to work around
-  [a broken client](https://tailscale.com/kb/1240/sso-custom-oidc) which doesn't
-  send a `rel` parameter, but expects an Identity Provider Discovery issuer URL
-  in response.
+  [an Identity Provider Discovery for OIDC URL][webfinger-oidc], **ignoring**
+  any supplied [`rel` parameter][webfinger-rel].
 
   If you want to use WebFinger in any *other* context on Kanidm's hostname,
```
CEbbinghaus, line 520:
```suggestion
  [`rel` parameters](https://datatracker.ietf.org/doc/html/rfc7033#section-4.3).
```

CEbbinghaus (commented):
Looks good.

book/src/integrations/oauth2.md
```diff
@@ -458,58 +458,84 @@ Each client has unique signing keys and access secrets, so this is limited to ea
...
+difficult to use with Kanidm:
 
 * WebFinger assumes that the identity provider will give the same `iss`
-  (Issuer) for every OAuth 2.0/OIDC client, and there is no standard way for a
-  WebFinger client to report its client ID.
+  (issuer) and OpenID Discovery document, including all URLs and signing keys,
+  for *all* OAuth 2.0/OIDC clients.
 
-  Kanidm uses a *different* `iss` (Issuer) value for each client.
+  Kanidm uses *different* `iss` (issuer), signing keys, and some client-specific
```
Firstyear, line 484: Different client signing keys may be better.

Firstyear (changes requested):

book/src/integrations/oauth2.md
micolous: I don't think your suggestion is showing up GitHub as you intended:

Screenshot 2025-02-21 at 14 02 17

I think you actually want me to replace "any supplied `rel` parameter" with "`rel` parameters" (less words, also the spec allows it to be specified multiple times).

book/src/integrations/oauth2.md
micolous: @yaleman any opinions on this? 😅

book/src/integrations/oauth2.md
micolous: I reworded this to "uses *client-specific* ... token signing keys"

book/src/integrations/oauth2.md
micolous: I've added the above change.

book/src/integrations/oauth2.md
Firstyear: It's 1.5.0, but also I think we don't need it because docs are published per-server-version, so when we merge this it won't show until 1.6 iirc.

Firstyear (approved):

Firstyear (approved):

PR #3452, opened 2025-02-23
build: Create daemon image from scratch

JadedBlueEyes:
# Change summary

The daemon's final image is now based on a scratch image. This moves the compressed size from 196.74 MB to 141.67 MB (the vast majority being the kanidmd binary)

This doesn't contain any special support for cross compilation, and it relies on ldd to detect dynamic library dependencies (which works by kind of pseudo-running the binary, and therefore doesn't work for non-emulated crossbuilds). 

#3380 pulled in [lddtree](https://github.com/messense/lddtree-rs), a rust CLI using goblin, which does work for cross-compilation, but isn't packaged anywhere. There are also shell scripts that use scanelf that work in a similar way. However, considering cross-compilation isn't a priority, this was simpler for now.

I've not touched the tools image or the Orca image, as they rely on the fido tool. After some thinking, I think that that should be built in its own docker image and then the final result from that should be copied into orca and tools - however, I'm not sure if that should live in this repo or the webauthn-rs repo.

Checklist

### DIFF
--- a/server/Dockerfile
+++ b/server/Dockerfile
@@ -54,7 +54,7 @@ RUN --mount=type=cache,id=cargo,target=/cargo \
         --release; \
     sccache -s
 
-# Find and copy dynalcially linked libraries using ldd
+# Find and copy dynamically linked libraries using ldd
 # caveat: this actually partially runs the binary, so it doesn't work for cross-compilation
 RUN <<EOF
     mkdir -p /out/libs
--- a/server/Dockerfile
+++ b/server/Dockerfile
@@ -54,20 +54,43 @@ RUN --mount=type=cache,id=cargo,target=/cargo \
         --release; \
     sccache -s
 
+# Find and copy dynalcially linked libraries using ldd
+# caveat: this actually partially runs the binary, so it doesn't work for cross-compilation
+RUN <<EOF
+    mkdir -p /out/libs
+    mkdir -p /out/libs-root
+    ldd /usr/src/kanidm/target/release/kanidmd
+    ldd /usr/src/kanidm/target/release/kanidmd | grep -v 'linux-vdso.so' | awk '{print $(NF-1) " " $1}' | sort -u -k 1,1 | awk '{print "install", "-D", $1, (($2 ~ /^\//) ? "/out/libs-root" $2 : "/out/libs/" $2)}' | xargs -I {} sh -c {}
+    ls -Rla /out/libs
+    ls -Rla /out/libs-root
+EOF
+
 # ======================
 
-FROM repos
-RUN \
-    --mount=type=cache,id=zypp,target=/var/cache/zypp \
-    zypper install -y \
-        timezone \
-        openssl-3 \
-        sqlite3 \
-        pam
+FROM scratch
 
-COPY --from=builder /usr/src/kanidm/target/release/kanidmd /sbin/
-COPY --from=builder /usr/src/kanidm/server/core/static /hpkg
-RUN chmod +x /sbin/kanidmd
+WORKDIR /
+
+# Copy root certs for tls into image
+# You can also mount the certs from the host
+# --volume /etc/ssl/certs:/etc/ssl/certs:ro
+COPY --from=repos /etc/ssl/certs /etc/ssl/certs
+
+# Copy our build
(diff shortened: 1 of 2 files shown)


server/Dockerfile
```diff
@@ -54,20 +54,43 @@ RUN --mount=type=cache,id=cargo,target=/cargo \
         --release; \
     sccache -s
 
+# Find and copy dynalcially linked libraries using ldd
```
CEbbinghaus, line 57:
```suggestion
# Find and copy dynamically linked libraries using ldd
```

CEbbinghaus:
This may be ignorance but as I understand it, when building a distroless image it is always preferrable to statically link all dependencies. Given that docker doesn't allow us to update dynamic dependencies it wouldn't make sense to copy the libraries into the output image if we can just statically build them into the kanidmd binary.

yaleman:
Rust isn't really into static binaries. 

From memory, we should only really be pulling in openssl until we can excise that from our code - we could base our build on `gcr.io/distroless/cc` which saves chasing things automagically and works cross-arch, and includes libssl.

CEbbinghaus:
Perfectly possible if kanidmd can run using `musl libc`. Would be nice to get rid of openssl

Firstyear:
openssl is a work in progress.

yaleman:
> Perfectly possible if kanidmd can run using `musl libc`. 

Doesn't need to be musl for that

JadedBlueEyes:
glibc (used on the default linux-gnu target) shouldn't be statically linked.

I'd prefer not to use musl, due to the fairly large performance penalty. That can be mitigated by using a different allocator, but that requires nontrivial code changes.

It looks like `distroless/cc` includes libcrypto and libssl, which are the OpenSSL bits that kanidm uses. 
OpenSSL could also be vendored into the binary. Vendoring doesn't look *too* difficult though, and we're already doing it with sqlite iirc: https://docs.rs/openssl/latest/openssl/#vendored

If using the distroless base is the preferred path, I can get that sorted fairly easily.

yaleman:
Let's try the distroless thing to avoid having to mess around with guessing about library includes

Firstyear:
> glibc (used on the default linux-gnu target) shouldn't be statically linked.
> 
> I'd prefer not to use musl, due to the fairly large performance penalty. That can be mitigated by using a different allocator, but that requires nontrivial code changes.

We already use a different allocator because glibc malloc is terrible. So this is a non issue. 

> 
> It looks like `distroless/cc` includes libcrypto and libssl, which are the OpenSSL bits that kanidm uses. OpenSSL could also be vendored into the binary. Vendoring doesn't look _too_ difficult though, and we're already doing it with sqlite iirc: https://docs.rs/openssl/latest/openssl/#vendored

We can't vendor openssl as it vendors openssl 1, and we have features that require 3. libressl also can't be used as it's missing some things. We are working to remove openssl as a dep, but that takes time. 

> 
> If using the distroless base is the preferred path, I can get that sorted fairly easily.

I don't know what distroless is - my major concern there is if they have a proper security team and have been audited etc.

While making the image smaller is an important goal, having a base image that actually is backed by an org with a security team does matter. I think we can find ways to shrink our image in incremental steps within what we already have, and then later we can see about swapping the base container image if needed.

yaleman:
> I don't know what distroless is - my major concern there is if they have a proper security team and have been audited etc.

It's Google, so yeah. https://github.com/GoogleContainerTools/distroless/

> While making the image smaller is an important goal, having a base image that actually is backed by an org with a security team does matter. I think we can find ways to shrink our image in incremental steps within what we already have, and then later we can see about swapping the base container image if needed.

Firstyear:
> It's Google, so yeah. https://github.com/GoogleContainerTools/distroless/

Then I trust it even less - google have a history of publish stuff that works for them and their needs, but not anyone else. See also boringssl.

yaleman:
🤷🏻 it's a basically-scratch image with minimal dependencies that I've been using for ~years and includes the requirements without the ldd-guessing. Feel free to continue to bike shed on this one, I'm out.

Firstyear:
Well in this case, because we're using the scratch image and copying the libs from the base suse image, it's fine.

Firstyear (approved):

PR #3459, opened 2025-02-25
20250225 improve test performance

Firstyear:
Improves test performance by a large factor. Waiting on #3432 to merge. 

Checklist


server/lib/src/be/idl_sqlite.rs
```diff
@@ -205,12 +205,15 @@ pub(crate) trait IdlSqliteTransaction {
         let mut stmt = self
             .get_conn()?
             .prepare(&format!(
```
yaleman, line 207: This really should be 

```rust
.prepare(r#"SELECT DISTINCT NAME from sqlite_master where type="table" AND name = :tname"#)

**Review by CEbbinghaus — COMMENTED** [review]
I love the http client change.

File: server/lib/src/idm/server.rs [hunk 2]
```diff
@@ -143,8 +143,9 @@ impl IdmServer {
     pub async fn new(
         qs: QueryServer,
         origin: &str,
+        is_integration_test: bool,
```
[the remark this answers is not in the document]
Firstyear: This has been there for ages, and the reason is so that we can control certain flags such that they only are possible in test cases.
The issue is cfg test flag doesn't always pass through, we could improve it now with features but that has it's own complextiy.

server/lib/src/be/idl_sqlite.rs
Firstyear: rowid is a faster lookup, but you're correct to select on type=table. I've made that change.

yaleman (approved):

PR #3466, opened 2025-02-26
Support rfc2307 memberUid in sync operations.

### LINKED ISSUE #3464: Allow ldap sync group_member_attr to support memberUid or warn if not using full DN

I have an issue syncing posixGroup objects with members as `memberUid: username`. Our openldap `memberUid` matches to a users `uid`, in our case a unique posix username rather than a full DN like `uid=user,ou=People,dc=...`

Example group object:

```ldif
dn: cn=sa_group1,ou=Groups,dc=ldap,dc=example,dc=com
gidNumber: 10001
memberUid: user1
```

As of 1.6.0-dev `cccf3f4b9baa` the server debug shows `Could not convert external_id to reference - user1`: 

```
622bbfc1-c7c8-4564-8113-95ae77f040b0 DEBUG    │        ┝━ 🐛 [debug]: Schemas valid - Proceeding with entry 0093b368-7d5f-103c-9ee6-c7b825f17e5b
622bbfc1-c7c8-4564-8113-95ae77f040b0 DEBUG    │        ┝━ 🐛 [debug]:  | requested_classes: {"group": SchemaClass { name: "group", uuid: 00000000-0000-0000-0000-ffff00000045, description: "
Object representation of a group", sync_allowed: true, systemmay: [Description, GrantUiHint, Mail, Member], may: [], systemmust: [Name, Spn], must: [], systemsupplements: [], supplements: [
], systemexcludes: [], excludes: [] }, "posixgroup": SchemaClass { name: "posixgroup", uuid: 00000000-0000-0000-0000-ffff00000058, description: "Object representation of a posix group, requ
ires group", sync_allowed: true, systemmay: [], may: [], systemmust: [GidNumber], must: [], systemsupplements: ["group"], supplements: [], systemexcludes: [], excludes: [] }}
622bbfc1-c7c8-4564-8113-95ae77f040b0 DEBUG    │        ┝━ 🐛 [debug]:  | sync_owned_attrs: {Description, GidNumber, Mail, Member, Name, PasswordImport, TotpImport, UnixPasswordImport}
622bbfc1-c7c8-4564-8113-95ae77f040b0 DEBUG    │        ┝━ 🐛 [debug]: Could not convert external_id to reference - user1
```

with reference to the group object . Both the group and user individually sync, but not the members of the group.

Caveat: I don't know how much effort is worth putting in to support syncing from old, poorly configured LDAPs.

Sync can map the `group_attr_member` against the `name` of the user object rather the `externalId` in the case of `memberUid` instead of `member`.

Or alternatively at least a warning emitted when this happens if this is not a supported workflow.

One possible change I can make on my end is to export the ~100 groups with `memberUid` into an ldif format that replaces `memberUid` with the `member` field. I'm not sure of a way to do this programmatically in one step besides handling each organizational unit individually.

For anyone attempting this, here is the search filter I crafted to generate the base ldif:
```ldif
ldapsearch -LLL -Y EXTERNAL -H ldapi:/// -b "dc=ldap,dc=example,dc=com" \
    '(& (|(objectClass=groupOfNames)(objectClass=posixGroup)) (!(ou:dn:=PendingAccounts)) (!(ou:dn:=DisabledAccounts)))' dn gidNumber objectClass memberUid \
    > group_membership.ldif

# gnu sed 
# Change field from memberUid to member and switch to full DN, specific to that OU
sed -i -e 's/memberUid: \(.*\)/member: uid=\1,ou=People,dc=ldap,dc=example,dc=com/' group_membership.ldif
```
Converts to
```ldif
dn: cn=sa_group1,ou=Groups,dc=ldap,dc=example,dc=com
gidNumber:  10001
member: uid=user1,ou=Users,ou=People,dc=ldap,dc=example,dc=com
objectClass: posixGroup
objectClass: top
```

Exercise left to the reader to convert to an ldapmodify format.

### Additional context

Same sync setup as in https://github.com/kanidm/kanidm/issues/3457

The ldap-sync-config matches groups with the below.
```toml
group_objectclass = "posixgroup"

group_attr_name = "cn"
#group_attr_description = "description"
group_attr_member = "memberuid"
group_attr_gidnumber = "gidnumber"
```

Below is the proto-dump for the user and group objects.

```json
{
      "schemas": [
        "urn:ietf:params:scim:schemas:kanidm:sync:1:account",
        "urn:ietf:params:scim:schemas:kanidm:sync:1:person",
        "urn:ietf:params:scim:schemas:kanidm:sync:1:posixaccount"
      ],
      "id": "48c4c166-b1f0-103c-80a4-c7b825f17e5b",
      "externalId": "uid=user1,ou=People,dc=ldap,dc=example,dc=com",
      "displayname": "User One",
      "gidnumber": 1000,
      "loginshell": "/bin/bash",
      "mail": [
        {
          "value": "user1@example.com"
        }
      ],
      "name": "user1",
      "password_import": "{crypt}$1$xYpcyTrT$fmX8O/snip...."
    },
...snip  
  {
      "schemas": [
        "urn:ietf:params:scim:schemas:kanidm:sync:1:group",
        "urn:ietf:params:scim:schemas:kanidm:sync:1:posixgroup"
      ],
      "id": "fd779974-dabf-103c-81d5-c7b825f17e5b",
      "externalId": "cn=sa_group1,ou=Groups,dc=ldap,dc=example,dc=com",
      "gidnumber": 10001,
      "member": [
        {
          "external_id": "user1"
        }
      ],
      "name": "sa_group1"
    },
```

Looking at the group object, it is trying to match against only `user1` which does not match the `externalId` on the user object. It seems even though we can change the member attribute mapping, its still expected to be the full DN and not a uid/username.

Thanks for bearing the conversion from openldap to kanidm patiently!

Firstyear:
A lot of legacy directory servers will use rfc2307 schema where members of groups are stored as the uid instead of a dn. Within kani, we absolutely need this to be a dn, else we risk accidentally adding kanidm entries into ldap synced groups which isn't what we want.

If we have an rfc2307 schema, then we pre-resolve the uid to the member dn so that kanidm gets the correct information.

Fixes #3464

Checklist


Firstyear:
@pmarr Are you able to test this branch with your ldap server to see if it helps you? I have done some testing, but would really want to hear back from you about it.

Firstyear:
Thank you for testing this! Really appreciate your feedback on the process, it's making Kanidm better.

yaleman (approved):

PR #3467, opened 2025-02-26
Handle form-post as a response mode

### LINKED ISSUE #3461: OIDC error with 1.5.0
### I did this

Updated kanidm from 1.4.6 to 1.5.0

### I expected the following

All my oauth2 clients work like before.

### This happened instead

If i login to one of my Clients([Bitwarden](https://bitwarden.com/help/configure-sso-oidc/#configuration-reference-materials)) i get a error from kanidm:

> Failed to deserialize query string: unknown variant `form_post`, expected `query` or `fragment`

Bitwarden OIDC Redirect Behavior was and is Redirect GET.

My Other Clients work fine.

### Kanidm version details

- Output of `kanidm(d) version`: 1.5.0
- Are you running it in a container? yes If so, which image/tag?: 1.5.0

Firstyear:
Some oauth2 clients apparently ignore what we tell them and request response modes we don't support.

First, we should deserialise these and error correctly.

Second, to maintain temporary compatibility, we remap form-post to query. This will be removed in future.

Fixes #3461

Checklist


server/lib/src/idm/oauth2.rs
```diff
@@ -1873,15 +1878,31 @@ impl IdmServerProxyReadTransaction<'_> {
...
         };
 
+        let response_mode = match response_mode {
+            ResponseMode::Query => SupportedResponseMode::Query,
+            ResponseMode::Fragment => SupportedResponseMode::Fragment,
+            ResponseMode::FormPost => {
+                warn!(
+                    "Invalid response mode form_post requested - many clients request this incorrectly but proceed with response_mode=query. Remapping to query."
+                );
+                warn!("This behaviour WILL BE REMOVED in a future release.");
```
CEbbinghaus, line 1897: sorry to necro but this is gonna be a hard sell. Once we support it removing support for it becomes significantly harder given that we would be breaking people's existing working setups.

Firstyear: We've done it before - before we remove it, we can convert it to a compatibility boolean, and then gauge affected users.

PR #3471, opened 2025-02-26
Chore: Add locked to build

CEbbinghaus:
# Change summary
Changes all CI/CD (and the release) builds to utilize `--locked`. This ensures that cargo won't automatically update the lock file and will instead exit with an error condition. This is to ensure that we know exactly what dependencies were included in which build rather than relying on every developer having run a build after updating dependencies (which we can assume but isn't guaranteed).

also changed the bang to `/bin/sh` since `/bin/bash` doesn't exist on nixos :(

Fixes: some theoretical issue that could happen in future

Checklist

### DIFF
--- a/server/daemon/run_insecure_dev_server.sh
+++ b/server/daemon/run_insecure_dev_server.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/sh
 
 set -e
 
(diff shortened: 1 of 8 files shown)


server/daemon/run_insecure_dev_server.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/sh
```
yaleman, line 1: wot

yaleman (approved):

PR #3490, opened 2025-03-06
Add missing lld dependency and fix syntax typo

Foosec:
# Change summary
Added lld dependency and syntax fix in devcontainer_postcreate.sh, changed shell to bash for run_insecure_dev_server.sh because of pushd and popd 

-

Fixes #

Checklist

### DIFF
--- a/server/daemon/run_insecure_dev_server.sh
+++ b/server/daemon/run_insecure_dev_server.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 
 set -e
 
--- a/server/daemon/run_insecure_dev_server.sh
+++ b/server/daemon/run_insecure_dev_server.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/sh
 
 set -e
 
@@ -29,7 +29,9 @@ if [ ! -f "${CONFIG_FILE}" ]; then
     exit 1
 fi
 
-pushd "${SCRIPT_DIR}" > /dev/null 2>&1
+# Save current directory and change to script directory without pushd
+OLD_DIR=$(pwd)
+cd "${SCRIPT_DIR}" > /dev/null 2>&1
 if [ -n "${1}" ]; then
     COMMAND=$*
     #shellcheck disable=SC2086
@@ -40,4 +42,4 @@ else
     #shellcheck disable=SC2086
     cargo run ${KANI_CARGO_OPTS} --bin kanidmd -- server -c "${CONFIG_FILE}"
 fi
-popd > /dev/null 2>&1
+cd $OLD_DIR > /dev/null 2>&1
--- a/server/daemon/run_insecure_dev_server.sh
+++ b/server/daemon/run_insecure_dev_server.sh
@@ -31,7 +31,7 @@ fi
 
 # Save current directory and change to script directory without pushd
 OLD_DIR=$(pwd)
-cd "${SCRIPT_DIR}" > /dev/null 2>&1
+cd "${SCRIPT_DIR}" || exit 1
 if [ -n "${1}" ]; then
     COMMAND=$*
     #shellcheck disable=SC2086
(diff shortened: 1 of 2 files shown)


yaleman (approved):
Thanks!

server/daemon/run_insecure_dev_server.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh
```
CEbbinghaus, line 1: does it not have sh installed? how come this change?

CEbbinghaus (changes requested):
Just a quick question.

server/daemon/run_insecure_dev_server.sh
yaleman: `sh` is aliased to different things based on operating system, if it's not bash-compatible then pushd/popd won't work.

hence

> changed shell to bash for run_insecure_dev_server.sh because of pushd and popd

Foosec: This is the reason, i ran into it in the devcontainer

CEbbinghaus: I'd rather remove pushd and rewrite it as
```sh
dir=$(pwd)
cd ... 

...

cd $dir
````

And keep compatibility with all shells rather than requiring bash. Yes /bin/sh is aliased but as long as we keep the script running in sh it should work in all shells. I cannot use that script whatsoever if it is /bin/bash because nixos won't have that shortcut and only ever provides /bin/sh

Comment (yaleman) [reply server/daemon/run_insecure_dev_server.sh:1]: That's a better fix for sure

Comment (Foosec) [reply server/daemon/run_insecure_dev_server.sh:1]: Makes sense to me, want me to just change it in this pr?

Comment (CEbbinghaus) [reply server/daemon/run_insecure_dev_server.sh:1]: Please do. I'd rather that than change it back to `/bin/bash`

File: server/daemon/run_insecure_dev_server.sh [hunk 2]
```diff
@@ -29,7 +29,9 @@ if [ ! -f "${CONFIG_FILE}" ]; then
     exit 1
 fi
 
-pushd "${SCRIPT_DIR}" > /dev/null 2>&1
+# Save current directory and change to script directory without pushd
+OLD_DIR=$(pwd)
+cd "${SCRIPT_DIR}" > /dev/null 2>&1
```
CEbbinghaus, line 34: I don't think the `> /dev/null 2>&1` is required here. The only time cd should print anything is if the directory doesn't exist in which case swallowing the error is probably not great for debugging.

server/daemon/run_insecure_dev_server.sh
```diff
@@ -40,4 +42,4 @@ else
     #shellcheck disable=SC2086
     cargo run ${KANI_CARGO_OPTS} --bin kanidmd -- server -c "${CONFIG_FILE}"
 fi
-popd > /dev/null 2>&1
+cd $OLD_DIR > /dev/null 2>&1
```
CEbbinghaus, line 45: Same with this one

CEbbinghaus (changes requested):
@yaleman what do you recon with the stdout/stderr redirection?

Nothing functionally wrong with this PR just doesn't seem as clean as it can be

server/daemon/run_insecure_dev_server.sh
yaleman: That probably should be a `cd "${SCRIPT_DIR}" || exit 1`

server/daemon/run_insecure_dev_server.sh
yaleman:
```suggestion
cd "${OLD_DIR}" || exit 1
```

Foosec: I wasn't really sure why it was there in the first place, to what i could gather pushd and popd will also only ever print stuff out if something goes wrong;

CEbbinghaus: Well that sounds like a code smell having gone unsmelled for too long. Better fixed now than never :3

CEbbinghaus (approved):
LGTM

yaleman (approved):
Looks good, thanks for the fixes!

PR #3501, opened 2025-03-13
20250313 unixd system cache

### LINKED ISSUE #3249: Separation of components that require access to shadow
I maintain a COPR build of kanidm for Fedora, and I would *like* to get this to a level of quality where I would recommend it to other people and upstream the build materials into this repo, but I've been a bit time and effort limited. Now that I've got a holiday weekend though I'm looking to clean up everything around packaging 1.4 so that I can get my machines upgraded.

Here is a current challenge: kanidm's OpenSuSE service file (which the Fedora package currently uses verbatim, but I am about to change that for a couple of reasons) adds the shadow group as a supplementarygroup to permit access to /etc/shadow. On Red Hat-like distributions, this group does not exist, and /etc/shadow is root:root and 0000. that means that uid=0 and CAP_DAC_OVERRIDE are required to read it, a very strict situation but one that's not too odd for such sensitive files and comports with the history of tools that access shadowfiles being setuid root. Of course kanidm sensibly does not want to run as root.

As a temporary solution, I may just leave the Fedora package unable to read these files, or I might experiment with setting up an ACL in a post-install script. Those are probably good enough for a COPR but a little hacky.

I think similar issues are going to come up with packaging in multiple distributions, though, and I'd like a method I feel better about. The best thing that occurs to me (and when I asked in fedora-devel Lennart seemed to agree) is separating out reading /etc/shadow into an independent binary that could run setuid. There are some other options as well but they seem more architecturally complicated.

I wanted to start a discussion on this and see if there are other ideas as well.

### LINKED ISSUE #3407: Script to generate directories for `kanidm-unix` during development

It would be nice if, when developing `kanidm-unix`, it was easier for the required directories to be generated automatically, or though a script.

Either one of:
- A script like `generate_unixd_directories.sh` to make directories and symbolic links so we can run that prior to running `cargo run --bin kanidm_unix`
- A script like `run_insecure_dev_server.sh` that passes the config files in repo, and also generates the necessary directories

### LINKED ISSUE #3499: unixd ends up caching system groups past their validity
### I did this
- With unixd 1.5.0 installed and configured for PAM & NSS resolution as per the PPA package default configs.
- With the following sysusers.d(5) config `kanidmd.conf`:
```
g kanidmd -
```
Ran the following sequence:
```shell
%> sudo systemd-sysusers kandimd.conf 
Creating group 'kanidmd' with GID 991.

%> getent group kanidmd                                  
kanidmd:x:991:

%> sudo groupdel kanidmd
```

### I expected the following
The group has been deleted from `/etc/{group,gshadow}` which originally provided it, thus getent & NSS should no longer return it. For example systemd-sysusers checks group existence via NSS before attempting to re-create it.

### This happened instead
```
%> getent group kanidmd 
kanidmd:x:991:
```

The result remains the same until I either take kanidm out of the `/etc/nsswitch.conf` group chain, or restart kanidm-unixd. `kanidm-unix cache-invalidate` & `cache-clear --really` do not clear the cached result, only a restart.

After a unixd restart, one can do the repro steps again just like before.

### Kanidm version details

- Output of `kanidm(d) version`: `kanidm-unix 1.5.0 d9f4dbdd3e`
- If not a container, how'd you install it: PPA `kanidm-unixd=1.5.0-202502091034+d9f4dbd`
- Operating System / Version (On Unix please post the output of `uname -a`): `Linux host 6.1.0-28-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.1.119-1 (2024-11-22) x86_64 GNU/Linux`

### Any other comments
NSS & PAM configs:

```
%> cat /etc/nsswitch.conf 
passwd:         files systemd kanidm
group:          files systemd kanidm
shadow:         files
gshadow:        files
hosts:          files dns
networks:       files
protocols:      db files
services:       db files
ethers:         db files
rpc:            db files
netgroup:       nis

%> cat /etc/pam.d/common-auth | grep -vE '^#'

auth    [success=2 default=ignore]      pam_unix.so nullok
auth    [success=1 new_authtok_reqd=done default=ignore]    pam_kanidm.so ignore_unknown_user use_first_pass
auth    requisite                       pam_deny.so
auth    required                        pam_permit.so
```

Firstyear:
# Change summary

The implementation of the unixd cache relies on inotify to detect changes to files in /etc so that we know when to reload the data for nss/passwd. However, the way that groupadd/del and other tools work is they copy the file, change it, and then move it into place. It turns out that william of the past didn't realise that inotify works on *inodes* not *paths* like other tools do (auditctl for example).

As a result, when something modified /etc/group or another related file, the removal was seen, but this breaks notifications on any *future* change until you reload unixd.

To resolve this we need to recursively watch /etc with inotify - yep, that's correct. We have to watch everything in /etc for changes because it's the only way to pick up on the add/remove of files. But because we have to watch everything, we need *permissions* to watch everything. 

This forces us to move the parsing of the etc passwd/group/shadow files to the unixd tasks daemon - arguably, this is the correct place to read these anyway since that is a high priv (and locked down) daemon. Because of this, we actually end up solving the missing "shadow" group on debian issue, and probably similar on the BSD's in future. 

In order to make my life easier while testing I also threw in a makefile that symlinks the files to needed locations for testing. It has plenty of warnings as it should. 

Fixes #3499
Fixes #3407
Fixes #3249

Checklist

### DIFF
--- a/unix_integration/Makefile
+++ b/unix_integration/Makefile
@@ -1,0 +1,19 @@
+
+
+
+dev_install:
+	@ echo "WARNING: THIS WILL BREAK EXISTING UNIXD INSTALLS"
+	@ echo "ctrl-c now if this is not what you want"
+	read
+	@ echo "LAST CHANCE"
+	sleep 5
+	ln -s -f ../platform/opensuse/kanidm-unixd.service /etc/systemd/system/kanidm-unixd.service
+	ln -s -f ../platform/opensuse/kanidm-unixd-tasks.service /etc/systemd/system/kanidm-unixd-tasks.service
+	ln -s -f ../target/debug/kanidm-unix /usr/sbin/kanidm-unix
+	ln -s -f ../target/debug/kanidm_ssh_authorizedkeys /usr/sbin/kanidm_ssh_authorizedkeys
+	ln -s -f ../target/debug/kanidm_unixd_tasks /usr/sbin/kanidm_unixd_tasks
+	ln -s -f ../target/debug/kanidm_unixd /usr/sbin/kanidm_unixd
+	ln -s -f ../target/debug/libpam_kanidm.so /lib64/security/pam_kanidm.so
+	ln -s -f ../target/debug/libnss_kanidm.so /usr/lib64/libnss_kanidm.so.2
+
+
--- a/unix_integration/Makefile
+++ b/unix_integration/Makefile
@@ -1,5 +1,5 @@
 
-
+current_dir = $(shell pwd)
 
 dev_install:
 	@ echo "WARNING: THIS WILL BREAK EXISTING UNIXD INSTALLS"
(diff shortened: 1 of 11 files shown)


unix_integration/Makefile
```diff
@@ -0,0 +1,19 @@
+
+current_dir = $(shell pwd)
+
```
yaleman, line 3:
```suggestion
.PHONY: dev_install
```

unix_integration/resolver/src/idprovider/system.rs
```diff
@@ -147,65 +147,58 @@ impl SystemProvider {
...
-                }
-            });
+                    Arc::new(Shadow {
+                        crypt_pw: password,
+                        aging_policy,
+                        expiration_date,
+                    }),
+                ))
+            } else {
+                // Invalid password, skip the account
```
yaleman, line 196: Need to add a debug message here so folks can figure it out

yaleman (changes requested):
Also, buncha tests failing 😄

Firstyear:
> Also, buncha tests failing 😄

Whaaaaaaa they weren't failing before. I'll have a look soon.

unix_integration/resolver/src/idprovider/system.rs
Firstyear: This is more common than you may think, it's normal to lock an account by replacing the pw hash with an x or ! so that they become locked. So I think this could be noisy. I'm adding the message, but only at a debug level.

yaleman (approved):

CEbbinghaus (commented):
RIP, Missed submitting the comment before the merge 😅

Issue #3505, opened 2025-03-15
kanidmd helper functions cannot be run without a db


jinnatar:
### I did this
```shell
sudo kanidmd configtest
sudo kanidmd cert-generate
```

### I expected the following
Testing to be performed on the config, and testing certs to be generated as per said config.

### This happened instead
> %> sudo kanidmd configtest   
📜 Using config file: "/etc/kanidmd/server.toml"
00000000-0000-0000-0000-000000000000 INFO     ｉ [info]: Starting Kanidmd | version: 1.5.0 08768249ee
00000000-0000-0000-0000-000000000000 WARN     🚧 [warn]: This is running as uid == 0 (root) which may be a security risk.
00000000-0000-0000-0000-000000000000 WARN     🚧 [warn]: DB folder /var/lib/private/kanidmd may not exist, server startup may FAIL!
00000000-0000-0000-0000-000000000000 INFO     ｉ [info]:  | diag: -- diagnosis for path: /var/lib/private/kanidmd/kanidm.db
unable to canonicalise path
running as: 0:0
path permissions
>
>  /                                  : DIR access: true owner: 0 group: 0 mode: rwxr-xr-x
  /var                               : DIR access: true owner: 0 group: 0 mode: rwxr-xr-x
  /var/lib                           : DIR access: true owner: 0 group: 0 mode: rwxr-xr-x
  /var/lib/private                   : DIR access: true owner: 0 group: 0 mode: rwx------
  /var/lib/private/kanidmd           : ERROR: Os { code: 2, kind: NotFound, message: "No such file or directory" }
  /var/lib/private/kanidmd/kanidm.db : ERROR: Os { code: 2, kind: NotFound, message: "No such file or directory" }
>
>  note that accessibility does not account for ACL's or MAC
-- end diagnosis
>
> 00000000-0000-0000-0000-000000000000 ERROR    🚨 [error]: Unable to read metadata for database folder '/var/lib/private/kanidmd' - Os { code: 2, kind: NotFound, message: "No such file or directory" }
Logging pipeline completed shutdown

### Kanidm version details

- Output of `kanidm(d) version`: `1.5.0 08768249ee` (experimental build based on the 1.5.0 branch)
- Are you running it in a container? If so, which image/tag?:
- If not a container, how'd you install it: experimental deb based on extension of current PPA build pipeline
[...]

PR #3506, opened 2025-03-15
packaging: Add kanidmd deb package, update documentation

### LINKED ISSUE #3481: Kanidm server Debian package

Currently, I can't seem to find a way to install the Kanidm server on a Debian system without Docker, except for compiling it by hand and writing a systemd unit file and the like manually. A PPA already exists, but it only seems to include the client tools.

I'd love for a package for the server to be added to the PPA, so I don't have to manually keep everything up to date.

I could use Docker, but I have a personal grudge against containerization so the server I'm attempting to deploy Kanidm on doesn't have it installed. This is a fantastic-looking project and I'm excited to use it, but being forced to choose between Docker and compiling it myself has put a bit of a damper on my mood.

jinnatar:
# Change summary
The main change is adding cargo-deb metadata to build kanidmd. The intent is to keep the security level as high as we can provide on supported Debian systemd based distros, while balancing ease of maintenance. Thus, the security level is roughly equivalent to that of kanidm-unixd, with some deviations:
- kanidmd is run as a systemd DynamicUser, this implies a bunch of systemd controls intended for isolation vs. running as a dedicated user where only standard filesystem access controls apply.
- A SupplementalGroup is created and used: `kanidmd`. This enables keeping `server.toml` & TLS key material in a sysadmin friendly `/etc/kanidmd` directory. Both files can be chowned to `:kanidmd` with `g+r` permissions to allow kanidmd only read access.
- No attempt is made to make the service able to start up without config. Providing TLS  materials is left as an "implementation detail". However, the default config does allow successful use of `kanidmd cert-generate` for evaluation purposes, sans the caveat of issue #3505.
- There are no book documentation changes to in any way imply that this is a possible nor supported way to run kanidmd. That's probably something that needs to be discussed if any mention should be made, perhaps only to discourage it as a production quality install? I did my best to actually make it safe, but it's a different kind of attack surface than what's been considered before.

Changes in support of the kanidmd deb release are:
- Fork a separate `release_debian` build profile. This allows changing the static asset path to a per-package one. There may be more "correct" ways of doing this, but based on a chat with @yaleman this may be the least disruptive option.

Tangential changes are:
- Doc updates to manual packaging guidance to match already published 1.5.0 tooling.
- Fix an internal mistake in systemd service unit naming, cargo-deb is patching this error but doing it right eliminates warnings. (No need to backport.)
- Implement dynlib multiarch builds via new cargo-deb feature instead of rolling our own variant based sillyness. (Support already merged on the PPA automation side, no need to backport)

Fixes #3481

The PR is against master, but all testing has been done against 1.5.0 due to a breaking kanidmd bug that made it hard to test against master. So should be trivial to backport to 1.5. Ultimately final testing before unleashing it to the masses is integration testing for a PPA 1.5.1 / 1.6.0 release. I haven't built that qemu test set yet.

Checklist


server/daemon/debian/kanidmd.service
```diff
@@ -10,13 +10,16 @@ Before=radiusd.service
 [Service]
 Type=notify
 DynamicUser=yes
-StateDirectory=kanidm
+User=kanidmd_dyn
+Group=kanidmd_dyn
```
yaleman, line 14: What's the effective owner/group of things like the backups folder and its contents when the daemon is running with this config? I'd be concerned that it makes it impossible to get to backups without adding the backup user to the kanidmd_dyn group (which would be impossible). Probably makes more sense to keep the user as kanidmd and systemd will limit its access outside the given folders with its controls

yaleman (changes requested):

server/daemon/debian/kanidmd.service
jinnatar: Yeaaah, I didn't consider the backup scenario and I don't think there's a clean way to apply the `kanidmd` group to them.

The next best I can think of is trying to stay within DynamicUser, but using the `kanidmd` group as the effective group instead of supplemental. That would automatically infer group privileges to a known GID. But need to test if DynamicUser allows a static group without losing the dynamic user. If that fails, then we replicate the implied controls of DynamicUser but without the actual dynamic UID/GID.

jinnatar: It _seems_ to work but I'm not sure how exactly. What the systemd NSS module reports doesn't fully match reality, i.e. the dynamic user definitely is getting the permissions of the static group, but from the perspective of NSS it is not a member of the group, and it's basegroup is still created matching the value given to User. So this might be undefined behavior that while seemingly beneficial, may turn out later to be a bad idea.

jinnatar:
(rebased against master again, forgot it against 1.5.0 due to testing)

server/daemon/debian/kanidmd.service
yaleman: Yep, that seems like the best of all possible options

yaleman (approved):

yaleman:
Yay! Thanks for all the testing and work on this 🥳

Firstyear:
Amazing work, thank you so much!

Issue #2417, opened 2024-01-06
PIV/Smartcard Authentication


Firstyear:
Now that we have attested passkeys, there is one other possible type of stricter authentication level for enterprise users - smartcards. This uses mTLS with x509 certificates that are backed by hardware. This gives a very high level of assurance to the service that the user is physically at the machine, and completely nullifies session exfiltration since the session is bound to the TLS connection. 

The foundations are inplace, the final part is user mapping, policy, and exposing this. 

In addition, when a user is mapped to a smartcard, it should be possible to mark features of the smartcard such as session rw/ro, or if the card can be used for ssh authentication.

Firstyear:
https://github.com/kanidm/kanidm/issues/755

PR #3513, opened 2025-03-18
Automatic Certificate Management Environment Proposal

sevki:
# Automatic Certificate Management Environment Proposal

Hi all, I wanted to share something that I would like to work on, but before dumping a huge PR I wanted to ask
firstly if anyone is actually interested in such a contribution and if so, since this would be my first contribution if anyone would
like to offer any feedback in-terms of how to design things that align with this project. 

## Security

In terms of security, we would not need to add any more functionality to the crypto library, as both JWS and x509 minting code
exists. (ACME could also be used to issue mTLS certificates for replication). 

## Open questions

- Should the certificate issuance be separated into their own service or be configurable so people can bring their own CA? 
- What account permissions required to create/revoke certificates. 
- How to handle cert revocation? (Short lived certificates? CT log?)

I don't really have all that much right now, I really wanted some very early feedback. 

Checklist

### DIFF
--- a/book/src/developers/designs/acme.md
+++ b/book/src/developers/designs/acme.md
@@ -1,0 +1,144 @@
+# Automatic Certificate Management Environment (ACME)
+
+> [!WARNING]  
+> This is very much a working progress. This is a very minimal possible implementation to get a discussion going. 
+
+## Rationale
+
+Kanidm already provides support for several types of public key systems, such as passkeys and ssh, and 
+mTLS altough only for replication.
+
+ACME is the standard protocol for obtaining and managing TLS certificates.
+
+Implementing ACME protocol support in Kanidm will enable person and system accounts to easily obtain and
+manage TLS certificates for their services, improving security and convenience.
+
+## Design
+
+### Add a new enrollment method for creating an External Account Binding
+
+#### `/v2/acme/eab` 
+
+which would return
+```json
+{"hmac_key": "deadbeef", "kid": "1"}
+```
+### Create a new Account
+
+#### `/v2/acme/directory` 
+
+```rust
+/// ACME Directory Endpoint - The entry point for the ACME protocol
+///
+/// As defined in https://www.rfc-editor.org/rfc/rfc8555#section-7.1.1
+#[utoipa::path(
+    get,
+    path = "/v2/acme/directory",
+    responses(
+        (status = 200, description = "ACME directory information", body = Directory)
+    ),
+    tag = "acme"
+)]
+pub async fn acme_directory(State(state): State<ServerState>) -> impl IntoResponse {
+    // Build the base URL from the server's origin
+    let base_url = format!("{}acme", state.origin);
+
+    Json(Directory {
+        new_nonce: format!("{}/new-nonce", base_url),
+        new_account: format!("{}/new-account", base_url),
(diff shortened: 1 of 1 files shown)


Firstyear:
I'm not 100% clear on what the proposal is that you're making? Are you suggesting that Kanidm have a built in ACME client for requesting certificates? Or are you suggesting that Kanidm becomes an ACME issuer of certificates?

Firstyear:
Okay, so this actually then is a relation to https://github.com/kanidm/kanidm/issues/2417 as this is the issuance side if we choose to become a CA.

Firstyear:
Realistically the proposal you're making is "implement a CA". Which we do want to do, but there is a long road ahead before we are there. 

For now, the other side of this is "improve our support for certificate authentication". Which is a much more tangible and possible goal today. It turns out today we already *secretly* support mTLS for auth, and that support just needs improvement and polish. So perhaps we look there initially.

Firstyear:
> Have we considered integrating with another self-hosted CA server such as [Smallstep](https://smallstep.com/docs/step-ca/) instead of Kanidm re-implementing that functionality?
> 

The problem is that goes against: 

https://github.com/kanidm/kanidm/blob/master/book/src/developers/faq.md#why-dont-you-use-libraryproject-x

If we depend on step-ca, suddenly step-ca is now a dependency and has to be setup so we can test against it. Within Kani, we can control the whole stack so we can test everything *extensively*

Second, is step-ca specifically doesn't allow some features that community members have requested (authenticated acme I believe). 

> I am currently using `step-ca`, and use Kanidm as an provider so the OIDC data (groups, policies, etc.) are available in a [certificate templating engine](https://smallstep.com/docs/step-ca/templates/#template-syntax-overview), which makes it very flexible. I use it for SSH and x509 certificates, and it also supports the ACME API, which I use to issue mTLS certs. I'm happy to provide examples if anyone is interested.

If step-ca works for you, then great! But that also means Kanidm doesn't need to do anything at all - you already have a CA! We don't need to implement ACME as a server because step-ca does it for you.

Firstyear:
> > If Kanidm does become a CA it will be optional, not mandatory, we aren't going to be forcing it on anyone.
> 
> functionally speaking kanidm is already a [CA](https://en.wikipedia.org/wiki/Certificate_authority)
> 
> > In [cryptography](https://en.wikipedia.org/wiki/Cryptography), a certificate authority or certification authority (CA) is an entity that stores, signs, and issues [[public key certificate]](https://en.wikipedia.org/wiki/Public_key_certificate).

The key word there is "stores, signs, AND issues". We only store certificates - we do not yet sign or issue them. So we are not a CA. 


> > Not sure I understand this?
> 
> I think about these things in the dumbest way possible; big prime number services. I really don't want to be maintaining two big prime number services.
> 
> We already have kanidm that does store and deal with big prime numbers, lets just have it also speak acme so it can also sign and store other types of certificates with various other clients sounds much less work than, let's try to keep two different systems work together, if that makes sense?

But we aren't a CA. So you aren't maintaining two CA's. You are maintaining a CA and an IDM? 

If Kanidm were a CA I would understand this. But we aren't. Regardless of if we proxy your ACME protocol bits, you still need the external CA. 

So I think we don't need to implement ACME until we are a CA.

PR #3514, opened 2025-03-18
Update shell.nix to work with stable nixpkgs

qenya:
# Change summary

The existing shell.nix uses whatever versions of rustc and cargo are in the system nixpkgs. In the current stable nixpkgs version (24.11), this is rustc 1.82.0. Unfortunately, we depend on the `strict_provenance` feature, which was unstable before 1.84.0. (See: kanidm/concread#132)

This patch makes minimal changes to shell.nix to overlay nixpkgs with the rustc version defined in rust-toolchain.toml, enabling Kanidm to build locally on stable versions of NixOS.

Checklist

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -1,14 +1,13 @@
-{ pkgs ? import <nixpkgs> {} }:
-	let
-	overrides = (builtins.fromTOML (builtins.readFile ./rust-toolchain.toml));
+let
+	rust-overlay = (import (builtins.fetchTarball "https://github.com/oxalica/rust-overlay/archive/master.tar.gz"));
 in
-pkgs.mkShellNoCC rec {
+{ pkgs ? import <nixpkgs> { overlays = [ rust-overlay ]; } }:
+pkgs.mkShellNoCC {
 	# Kanidm dependencies
 	buildInputs = with pkgs; [
 		pkg-config
 		
-		cargo
-		rustc
+		(rust-bin.fromRustupToolchainFile ./rust-toolchain.toml)
 
 		clang
 		llvmPackages.bintools
@@ -19,7 +18,6 @@ pkgs.mkShellNoCC rec {
 		linux-pam
 	];
 	
-	RUSTC_VERSION = overrides.toolchain.channel;
 	# https://github.com/rust-lang/rust-bindgen#environment-variables
 	LIBCLANG_PATH = pkgs.lib.makeLibraryPath [ pkgs.llvmPackages_latest.libclang.lib ];
 }


CEbbinghaus (approved):
LVTM

shell.nix
```diff
@@ -1,14 +1,13 @@
-{ pkgs ? import <nixpkgs> {} }:
-	let
-	overrides = (builtins.fromTOML (builtins.readFile ./rust-toolchain.toml));
+let
+	rust-overlay = (import (builtins.fetchTarball "https://github.com/oxalica/rust-overlay/archive/master.tar.gz"));
```
Firstyear, line 2: @mweinelt Can you confirm if this is the correct rust overlay to use? I don't know about this URL is all ....

Firstyear (dismissed):

shell.nix
yaleman: I'm definitely not a fan of arbitrarily including random github repos in configs we suggest.

qenya: I don't altogether disagree - it would be useful to have a wider variety of rustc versions in nixpkgs, and I'm sure it's been suggested, though I can't find the ticket right now. Unfortunately right now, yes, anything beyond the default version needs to be provided by a third-party repository. (Which I can sympathise with - it's hard to keep package repos up-to-date when major rustc releases happen so rapidly.)

oxalica's nix-overlay is the first of two recommendations in [nixpkgs' Rust documentation](https://nixos.org/manual/nixpkgs/stable/#using-community-maintained-rust-toolchains), if that helps affirm its credentials. The other one, fenix, is just as viable if you prefer that, but, well, if the problem is having an third-party dependency at all then obviously that doesn't really help.

Short of just having kanidm target an older rustc version, which seems even less ideal, I'm not sure I see another option for building with Nix.

Firstyear: > (Which I can sympathise with - it's hard to keep package repos up-to-date when major rustc releases happen so rapidly.)

We don't have this issue in opensuse? We follow upstream release cadence no problems. (I'm the rust maintainer for opensuse)

> oxalica's nix-overlay is the first of two recommendations in [nixpkgs' Rust documentation](https://nixos.org/manual/nixpkgs/stable/#using-community-maintained-rust-toolchains), if that helps affirm its credentials.

@mweinelt is our resident nix expert, so I'd like them to review is all. I don't know how nix works, so I want to defer to our community members who do.

mweinelt: The problem is that `<nixpkgs>` is a loose reference to whatever the local system executing the shell tracks. And the rust version will be relative to that nixpkgs version. For NixOS 24.11 this will be Rust 1.82.0 at this time.

The proposed approach has the benefit of matching up with whatever toolchain version you configure. It is together with [fenix](https://github.com/nix-community/fenix) and the more involved [naersk](https://github.com/nix-community/naersk) one of the approaches commonly used to develop rust projects with nix. I have previously interacted with oxalica, and she is well known in the nix community.

An alternative would be to track the latest `nixpkgs-unstable` expressions, if you don't want to put additional trust into another source. This would get you the latest rust toolchain (with a 2–3 weeks delay at most) with no respect to the toolchain configured. Right now we have 1.85.0 on nixpkgs-unstable.

The following snippet loosely tracks the [nixpkgs-unstable](https://github.com/NixOS/nixpkgs/tree/nixpkgs-unstable) branch, that updates multiple times per day.

```nix
let
  pkgs = import (fetchTarball("https://github.com/NixOS/nixpkgs/archive/nixpkgs-unstable.tar.gz")) { };
in pkgs.mkShellNoCC {
  ...
}
```

Ultimately the problem is that nobody herewants to track a known good lockfile version as initially proposed, so we're getting mixed results and have to think about how to work around them.

Firstyear: So long as Oxalica is a valid source then I'm happy.

Firstyear:
I'll leave this to @yaleman to press the button in case they have a follow up concern.

shell.nix
CEbbinghaus: Just to add my general vibes to this discussion. I am more inclined to go with any solution that 100% tracks the `rust-toolchain.toml`. My previous solution didn't (afaict) even though it did include the file. So I'm heavily leaning towards `oxalica` or `fenix` over `nixpkgs-unstable` which would be the latest version, but not the version in `rust-toolchain.toml`.

If the nix flake were to also use the `rust-toolchain.toml` that'd be good but maintaining a lock file for nix doesn't sound appealing if we can do it without.

yaleman:
If that's what folks do then that's what folks do. I'll smoosh it, but we really should put comments on the header of these kinds of things saying "hey, community provided, don't just yeet it into your system and then blame us!"

Issue #3423, opened 2025-02-12
Add support for v2 PROXY protocol


jpwarren:
## Is your feature request related to a problem?

Kanidm does not support the PROXY protocol. Attempts to connect to Kanidm from a streaming proxy will fail.

## Describe the solution you'd like

Kanidm should accept connections using the PROXY protocol as described here: https://www.haproxy.org/download/1.8/doc/proxy-protocol.txt

See also: https://docs.nginx.com/nginx/admin-guide/load-balancer/using-proxy-protocol/

Support could be optional, enabled via a configuration option. Or Kanidm could opportunistically detect incoming PROXY protocol connections and switch to PROXY instead of HTTPS.

The PROXY protocol provides several benefits for origin servers that sit behind load-balancers, potentially in a chain of proxy servers thanks to the widespread use of CDNs and DDOS protection. For example, the PROXY protocol makes it easier to pass the original client IP address to upstream systems for accurate logging and (potentially) access controls such as limiting access to particular address ranges.

It also allows streaming connections, such as that supported by Nginx's `ngx_http_proxy_module`. In streaming mode, the proxy server doesn't need to terminate and proxy TLS connections. It can use SNI headers to determine which virtual host the connection should be routed to and pass the TLS connection along to the origin server. This simplifies TLS certificate management and helps maintain end-to-end encryption between the client and Kanidm. The configuration of the proxy server is also less complicated, reducing the potential for accidental errors that might compromise security.

## Describe alternatives you've considered

I've got things working at the moment by putting an Nginx proxy on the same host as Kanidm to accept a PROXY protocol stream from an ingress node. The accepting proxy terminates the TLS session, and then uses `proxy_pass` to connect to Kanidm that is listening on a high port on `localhost` rather than on port 443 for the whole system. This accepting proxy does nothing but pass everything to Kanidm, essentially translating PROXY protocol to HTTP/1.1.

Here are the relevant bits of Nginx config, templated to be more generic.

### Ingress Node config snippet

This config goes in `/etc/nginx/modules-enabled/stream-proxies.conf` on the ingress node.

```
stream {

  map $ssl_preread_server_name $sslname {
    # Send connections for idm.example.org to the kanidm backend
    idm.example.org kanidm_backend;
  }

  upstream kanidm_backend {
    # Nginx running on the Kanidm server
    server kanidm_server01.example.org:443;
  }

  server {
    listen 443;
    listen [::]:443;
    proxy_pass $sslname;
[...]

PR #3515, opened 2025-03-19
Allow versioning of server configs

Firstyear:
This allows our server configuration to be versioned, in preparation for a change related to the proxy protocol additions.

Relates #3423

Checklist

### DIFF
--- a/server/core/src/config.rs
+++ b/server/core/src/config.rs
@@ -20,10 +20,30 @@ use url::Url;
 
 use crate::repl::config::ReplicationConfiguration;
 
+// Allowed as the large enum is only short lived at startup to the true config
+#[allow(clippy::large_enum_variant)]
+// These structures allow us to move to version tagging of the configuration structure.
+#[derive(Debug, Deserialize)]
+#[serde(untagged)]
+pub enum ServerConfigUntagged {
+    Version(ServerConfigVersion),
+    Legacy(ServerConfig),
+}
+
+#[derive(Debug, Deserialize)]
+#[serde(tag = "version")]
+pub enum ServerConfigVersion {
+    #[serde(rename = "2")]
+    V2 {
+        #[serde(flatten)]
+        values: ServerConfigV2,
+    },
+}
+
 #[derive(Deserialize, Debug, Clone)]
 pub struct OnlineBackup {
     /// The destination folder for your backups, defaults to the db_path dir if not set
-    pub path: Option<String>,
+    pub path: Option<PathBuf>,
     /// The schedule to run online backups (see <https://crontab.guru/>), defaults to @daily
     ///
     /// Examples:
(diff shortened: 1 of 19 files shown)


server/core/src/config.rs
```diff
@@ -92,162 +112,154 @@ pub struct TlsConfiguration {
...
+    db_fs_type: Option<kanidm_proto::internal::FsType>,
+    tls_chain: Option<PathBuf>,
+    tls_key: Option<PathBuf>,
+    tls_client_ca: Option<PathBuf>,
+    bindaddress: Option<String>,
+    ldapbindaddress: Option<String>,
+    role: Option<ServerRole>,
+    log_level: Option<LogLevel>,
+    online_backup: Option<OnlineBackup>,
+    trust_x_forward_for: Option<bool>,
```
CEbbinghaus, line 220: Did we want to change this into an enum to preempt #3423 or are we trying to keep 100% compat with "v1"?

server/lib/src/be/idl_sqlite.rs
```diff
@@ -1712,7 +1707,7 @@ impl IdlSqliteWriteTransaction {
 
 impl IdlSqlite {
     pub fn new(cfg: &BackendConfig, vacuum: bool) -> Result<Self, OperationError> {
-        if cfg.path.is_empty() {
+        if cfg.path == Path::new("") {
```
CEbbinghaus, line 1710: how is that better than `is_empty()`?

edit:
Likely a PathBuf in which case `.as_os_str().is_empty()` should probably be used instead (at least that looks to be the standard way of doing this)

server/lib/src/be/mod.rs
```diff
@@ -2121,7 +2126,7 @@ impl Backend {
         debug!(db_tickets = ?cfg.pool_size, profile = %env!("KANIDM_PROFILE_NAME"), cpu_flags = %env!("KANIDM_CPU_FLAGS"));
 
         // If in memory, reduce pool to 1
-        if cfg.path.is_empty() {
+        if cfg.path == Path::new("") {
```
CEbbinghaus, line 2129: again. `.as_os_str().is_empty()` seems to be the more common approach [ref](https://github.com/rust-lang/rust/issues/30259#issuecomment-191033337)

CEbbinghaus (approved):
Just some Questions 🙂

server/core/src/config.rs
Firstyear: The point of this change is just to enable versioning, then we will change this for V2 with #3423 - but there is no point adding the enum yet before we do that work, is there?

server/lib/src/be/idl_sqlite.rs
Firstyear: Because there isn't .is_empty() on Path as you found out :)

server/core/src/config.rs
CEbbinghaus: Not really. But moving from "v1" to v2 is considered a breaking change no? But changing the bool to an enum variant is a breaking change too so it would need v3 of the config. Would make sense to set up v2 such that it takes #3423 into account so a second breaking v3 migration isn't required

Firstyear: I think you're missing a valuable point here - v2 doesn't matter until we *release*. So once this change merges we can *keep changing it all we want* until the next release.

CEbbinghaus: Aight. If we committing to #3423 before next release then I am more than happy with that 🙂

CEbbinghaus (dismissed):
LGTM 👍

server/daemon/insecure_server.toml
yaleman, line 1: Can't move this here, it's used in the documentation and test scripts.

Firstyear:
@yaleman Fixed the script stuff - the doc test failures weren't introduced in this commit.

yaleman (approved):

Issue #3493, opened 2025-03-09
Custom fields


RatCornu:
Related to #3219.

Being able to add custom fields (to both users and groups, but with users in priority I think) would be a great improvement for some services. For example, to be able to set up a quota per user/group in a cloud service (like Nextcloud) or for mail aliases with dovecot/postfix.

The custom fields should also be provided in the LDAP. Something similar is provided in [lldap](https://github.com/lldap/lldap/issues/67), but it does not support search using a multi-value attributes (see lldap/lldap#997).

For example:

```txt
---
class: account
class: memberof
class: object
class: person
directmemberof: idm_all_persons@auth.skaven.org
directmemberof: idm_all_accounts@auth.skaven.org
displayname: Test
memberof: idm_all_persons@auth.skaven.org
memberof: idm_all_accounts@auth.skaven.org
memberof: idm_people_self_name_write@auth.skaven.org
name: polymorphe
primary_credential: primary
spn: test@example.org
quota: 30G
mailAlias: test@example.org
mailAlias: test-alias@example.org
```

would allow a cloud service to look the `quota` field directly, and a mail server would search for `test-alias@example.org` and see that it is bounded with the `test` account.

RatCornu:
I would be happy to help on this, but I think some discussion is needed beforehand as it can modify some base structures of the backend (I think mainly `Account` in `server/lib/src/idm/account.rs` and everything that depends on it). If you're okay with the idea I think I can try to start working on it in a few days

Firstyear:
It's far more complicated than this. You have to think about schema, replication, the datatypes, and access control.

We do actually already support custom fields and extensions of these - we just don't advertise yet that we do because we want certain things to calm (though arguably we are already there so maybe we can start to consider it .... )

RatCornu:
Hmmm indeed I did not read the code in details so I might say irrelevant things ^^', but this is a feature that I think would be greatly profitable so I'm willing to help as much as I can if necessary!

Firstyear:
Well, the first thing then is to get setup https://kanidm.github.io/kanidm/stable/developers/index.html check and setup so you can run "cargo test", that triggers all the tests in the project.

Then the next place will be to add the ability to query and list schema and access controls via scim - that will help give you an idea about how those things are structured, and then we can list them.

Once we know what they look like for listing, I think we come up with a design doc on what schema extension should look like. I think we should limit it to specific attribute syntax types initially. 

And if you have questions or need help, let us know :)

[...]

PR #3517, opened 2025-03-20
Make schema indexing a boolean instead of index types

Firstyear:
Previously on schema definitions for attributes, the list of index types was manually set on attributes. The issue with this approach is that not all index types apply to all attribute syntaxes. This made it error prone not just to Kanidm developers, but to future users who want to define custom attributes and may incorrectly index those attributes.

Instead, this changes the index value to be a boolean to indicate if this attribute should or should not be indexed. Internally Kanidm has a list of appropriate indexes to apply to these syntax types.

As part of this change, the tests were reviewed to find missing index types for syntaxes, and other causes of unindexed searches which led to some changes around the dyngroup plugin (which pushes the boundaries of a lot of things in Kani due to how it works).

Relates #2315 Relates #3493 

Checklist


server/lib/src/entry.rs
```diff
@@ -492,6 +492,117 @@ impl Entry<EntryInit, EntryNew> {
...
+
+        let multivalue_v = vs_bool![s.multivalue];
+        let sync_allowed_v = vs_bool![s.sync_allowed];
+        let replicated_v = vs_bool![s.replicated.into()];
+        let phantom_v = vs_bool![s.phantom];
+        let unique_v = vs_bool![s.unique];
+        let indexed_v = vs_bool![s.indexed];
+
+        let syntax_v = vs_syntax![s.syntax];
+
```
yaleman, line 516: Other than "it's prettier" is there a reason to do all these variable creations when they're just going to be used lower in the insert calls? Things this deep in the weeds tend to get called a lot...

server/lib/src/entry.rs
```diff
@@ -492,6 +492,117 @@ impl Entry<EntryInit, EntryNew> {
...
+    }
+}
+
+impl From<&SchemaClass> for EntryInitNew {
+    fn from(s: &SchemaClass) -> Self {
+        let uuid_v = vs_uuid![s.uuid];
+        let name_v = vs_iutf8![s.name.as_str()];
+        let desc_v = vs_utf8![s.description.to_owned()];
+        let sync_allowed_v = vs_bool![s.sync_allowed];
+
```
yaleman, line 561: see above comment

server/lib/src/schema.rs
```diff
@@ -81,21 +104,23 @@ pub struct SchemaReadTransaction {
...
+    /// agreement.
     pub sync_allowed: bool,
 
-    /// If the value of this attribute get replicated to other servers
-    pub replicated: bool,
-    /// TODO: What does this do?
-    pub index: Vec<IndexType>,
+    /// If set the value of this attribute get replicated to other servers
+    pub replicated: Replicated,
+    /// Define if this attribute is indexed or not according to it's syntax type rule
```
yaleman, line 122:
```suggestion
    /// Define if this attribute is indexed or not according to its syntax type rule
```

server/lib/src/schema.rs
```diff
@@ -1104,11 +1021,28 @@ impl SchemaWriteTransaction<'_> {
...
                 syntax: SyntaxType::IndexId,
             },
         );
+        self.attributes.insert(
+            Attribute::Indexed,
+            SchemaAttribute {
+                name: Attribute::Indexed,
+                uuid: UUID_SCHEMA_ATTR_INDEXED,
+                description: String::from(
+                    "A boolean stating if this attribute will be indexed according to it's syntax rules."
```
yaleman, line 1035:
```suggestion
                    "A boolean stating if this attribute will be indexed according to its syntax rules."
```

server/lib/src/schema.rs
```diff
@@ -1189,45 +1123,45 @@ impl SchemaWriteTransaction<'_> {
...
-                    index: vec![],
-                    syntax: SyntaxType::Utf8StringInsensitive,
-                },
-            );
+            Attribute::SystemSupplements,
+            SchemaAttribute {
+                name: Attribute::SystemSupplements,
+                uuid: UUID_SCHEMA_ATTR_SYSTEMSUPPLEMENTS,
+                description: String::from(
+                    "A set of classes that this type supplements too, where this class can't exist without their presence.",
```
yaleman, line 1137:
```suggestion
                    "A set of classes that this type supplements, where this class can't exist without their presence.",
```

server/lib/src/schema.rs
```diff
@@ -1240,46 +1174,46 @@ impl SchemaWriteTransaction<'_> {
...
-                    replicated: true,
-                    index: vec![IndexType::Equality],
-                    syntax: SyntaxType::Boolean,
-                },
-            );
+            Attribute::AcpEnable,
+            SchemaAttribute {
+                name: Attribute::AcpEnable,
+                uuid: UUID_SCHEMA_ATTR_ACP_ENABLE,
+                description: String::from("A flag to determine if this ACP is active for application. True is enabled, and enforce. False is checked but not enforced."),
```
yaleman, line 1207:
```suggestion
                description: String::from("A flag to determine if this ACP is active for application. True is enabled, and enforced. False is checked but not enforced."),
```

yaleman (changes requested):
Minor nits, but woo 🥳

yaleman (approved):

PR #3520, opened 2025-03-21
Update developer_ethics.md

ptdecker:
Adding my own case which is so often over looked.  Yes, people do use their middle name as their primary name and have done so since birth! 😄 

P. Todd Decker

# Change summary

- Add "middlename lastname" to the laundry list of name variants in the developer eithics chapter

Fixes #

Checklist

- [n/a] design document included (if relevant)


yaleman (approved):
I'm calling it for future PRs - it better be very different to fall outside "and many more" 😁

PR #3533, opened 2025-03-28
Unify unix config parser

### LINKED ISSUE #3532: Conflicting implementations of unix_config.rs
There are two different implementations of `unix_config.rs`:

- `unix_integration/common/src/unix_config.rs`
- `unix_integration/resolver/src/unix_config.rs`

The config in the common module is used by the pam and libnss modules and only contains a small subset of the options (presumably because these are the only ones relevant for these modules).  
However the two implementations read the `conn_timeout` value from different paths. The common implementation reads it from just `conn_timeout`, but the resolver implementation reads it from `kanidm.conn_timeout`. So to properly set the timeout one needs to set it twice in different locations in the same config file.

Firstyear:
Unifi the unix config parser into one place in the unix_common tree. This way we don't have issues with future config version changes. 

Fixes #3532

Checklist


unix_integration/common/src/unix_config.rs
```diff
@@ -49,10 +51,492 @@ impl Display for UidAttr {
...
+    pub kanidm_config: Option<KanidmConfig>,
+}
+
+#[derive(Debug)]
+pub struct KanidmConfig {
+    pub conn_timeout: u64,
+    pub request_timeout: u64,
+    pub pam_allowed_login_groups: Vec<String>,
+    pub map_group: Vec<GroupMap>,
+}
```
yaleman, line 195: this is in the middle of the unixdconfig things, please move it so it's above kanidmconfigv2

unix_integration/common/src/unix_config.rs
```diff
@@ -1,13 +1,15 @@
+use std::env;
```
yaleman, line 1: This module needs documentation so it's understandable without picking through it all to work out what things are used for. tThere's a kanidmconfig, unixdconfig and configint, and even knowing the history well, I'm starting to draw diagrams just to work out how it all comes together.

unix_integration/common/src/unix_config.rs
```diff
@@ -49,10 +51,492 @@ impl Display for UidAttr {
...
+    gid_attr_map: Option<String>,
+    selinux: Option<bool>,
+    #[serde(default)]
+    allow_local_account_override: Vec<String>,
+
+    hsm_pin_path: Option<String>,
+    hsm_type: Option<String>,
+    tpm_tcti_name: Option<String>,
+
+    // Detect and warn on values in these places.
```
yaleman, line 140: Is this a comment to future implementers, or people writing the config? I can't work out what it is.

unix_integration/common/src/unix_config.rs
yaleman: Also every config file field needs to have a `///` documentation so it's clear what everything does, and it'd be good if they were pub so it's easier to link to in the docs for reference, rather than digging into source code.

yaleman (changes requested):

yaleman (dismissed):

yaleman (approved):
woo!

PR #3535, opened 2025-03-29
20250328 41 application passwords + SCIM

### LINKED ISSUE #41: Add support for API-application passwords

Firstyear:
# Change summary

- This is the wrap of ldap application passwords so that they are usable for external applications. And I thought to myself "why not create these with SCIM, it's pretty much ready?". Anyway that turned up a lot of questions and TODO items.

So far this has highlighted a number of shortcomings that need to be resolved. The remaining TODO list is below, and some will be completed in separate PR's

- [ ] Schema needs to change so that Syntax type implies single/multivalue
- [ ] Service Account ACP's interfere with this, so we need to solve how application/oauth/service accounts look as a whole
- [ ] SCIM proto bindings are a bit messy and need a cleanup to better express the client/server relationship, they collide a lot right now
- [ ] Rename of the valueset json_put to drop the _put
- [ ] More that I will surely think of as I go .... 

Blocked by 

Fixes #41

Checklist


Firstyear:
I think the best approach is why this isn't fully completed, what we have works and it's not exposed yet, so I think we merge this and continue to polish as we go.

Firstyear:
@yaleman I would like to get this merged sooner, even though it's work in progress, so that it doesn't grow to be a "giant PR" like other ones have :)

server/lib/src/entry.rs
```diff
@@ -53,20 +54,17 @@ use kanidm_proto::v1::Entry as ProtoEntry;
 use ldap3_proto::simple::{LdapPartialAttribute, LdapSearchResultEntry};
 use openssl::ec::EcKey;
 use openssl::pkey::{Private, Public};
+use std::cmp::Ordering;
+pub use std::collections::BTreeSet as Set;
```
yaleman, line 58: I really hate the idea of having to re-publish a std thing at all, but why rename it? 🤮 Similarly, the rename of `BTreeMap` to `Map` below - it adds to cognitive overhead while looking at this code - especially since it's being imported twice?

server/lib/src/entry.rs
```diff
@@ -479,6 +488,11 @@ impl Entry<EntryInit, EntryNew> {
         self.attrs.remove(attr);
     }
 
+    /// Set the content of this ava with this valueset, ignoring the previous data.
+    pub fn set_ava_set(&mut self, attr: &Attribute, vs: ValueSet) {
+        self.attrs.insert(attr.clone(), vs);
```
yaleman, line 493: Doing an insert on a BTreeMap adds to the values, it doesn't replace them...

server/testkit/tests/testkit/ldap_basic.rs
```diff
@@ -14,3 +19,75 @@ async fn test_ldap_basic_unix_bind(test_env: &AsyncTestEnvironment) {
...
+
+    // Login as the person
+
+    // Create application passwords
+
+    // Check the work.
+
+    // Check they can't cross talk.
+
+    // Done!
```
yaleman, line 75: the code's talking to itself and not doing anything 😄

yaleman (changes requested):
At least one thing's a big red flag "insert" != "replace" and those renames are icky

server/lib/src/entry.rs
Firstyear: This was in the past when I wrote the code to allow swapping between BTreeSet and HashSet for testing. It's something we can undo in a separate PR.

server/lib/src/entry.rs
Firstyear: https://doc.rust-lang.org/std/collections/struct.BTreeMap.html#method.insert

"If the map did have this key present, the value is updated, and the old value is returned. "

It does replace them. The difference to "replace" is that it won't add if not present. So this code is correct.

server/testkit/tests/testkit/ldap_basic.rs
Firstyear: Yes, because I haven't finished writing the test, but I wanted to keep the structure in place and thoughts there so I can fill it in as I work on this.

server/lib/src/entry.rs
yaleman: oh right, it's just updating that one key, ffs brain.

server/lib/src/entry.rs
yaleman: k, u better

yaleman (approved):
yolooooo

Firstyear:
@yaleman Needs a re-stamp as I did an update/rebase/clippy locally. Nothing really changed though.

yaleman (approved):

Issue #3523, opened 2025-03-22
Combine oauth2 account and service-account entities


CEbbinghaus:
### Is your feature request related to a problem? Please describe

It is currently impossible to use an oauth2 app for anything but OIDC. A service-account can behave like a user in almost every way allowing use for email accounts & LDAP search but a oauth2 app cannot do any of that. This leads to having to maintain 2 different accounts for the same service, One for OIDC and one for LDAP / Email.

### Describe the solution you'd like

For OAuth2 applications to also be valid service-account users so 1 service-account can be used for OIDC + LDAP + Email

### Describe alternatives you've considered

Continuing to maintain 2 different applications for the same service

Firstyear:
See #3536 - tl;dr is "no" because an email service is a collection of different services, each with a unique intent. Anyway, there is more detail over there. But has to happen soon anyway, because lol, I stomped some fun with ldap application passwords.

PR #3536, opened 2025-03-29
Ideas about service accounts

Firstyear:
WIP design about service accounts. 

Relates #3523 and Relates #3535

Checklist

### DIFF
--- a/book/src/developers/designs/service_accounts.md
+++ b/book/src/developers/designs/service_accounts.md
@@ -21,14 +21,6 @@ authenticate to clients, which means they need to maintain credentials. This is 
 but there are still some ideas in these docs worth knowing about and considering like group managed
 service accounts (gMSA).
 
-## High Level Suggestions
-
-There are the ideas today that I have - there may be others!
-
-* Service Accounts can have attached roles.
-* Keep concerns separate.
-* Bit of A, bit of B, cleanup.
-
 ## Current state of affairs
 
 We have:
@@ -41,7 +33,58 @@ We have:
 From this we can see that we have some separation, but also some cross over of functionality.
 break glass isn't delegated, but service account is, OAuth2 isn't an SA, but Applications are.
 
-## Attach roles to service accounts.
+## Capabilities
+
+In order to properly handle this, we don't want to grant unbounded abilities to types, we don't
+want to fully merge them, but we want to be able to mix-match what they require.
+
+This also makes it possible in the future that we can more easily assign (or remove) a capability
+from an account type.
+
+To achieve this we should introduce the idea of capabilities - capabilities can act via schema
+classes, and we can extend the schema such that only the parent class needs to know that the
+capabilities class is required.
+
+This allows us to nominate more carefully what each role type can or can't do, and keeps things
+
+| Capabilities    | Api Token        | OAuth2 Sessions              | Interactive Login   |
+|-----------------|------------------|------------------------------|---------------------|
+| OAuth2          | No               | Via Client Credentials Grant | No                  |
+| Application     | Yes (ro)         | No                           | No                  |
+| Service Account | Yes (rw capable) | Yes (via session grant (TBD) | Yes (to be removed) |
+| Machine Account | Yes (ro)         | No                           | No                  |
+| Break Glass     | No               | No                           | Yes                 |
+| Person          | No               | Yes                          | Yes                 |
+
+A key requirement of this is that we want each role to have a defined intent - it shouldn't be
+the everything role, it still needs to be focused and administered in it's own right.
+
+|                 | Intent                                                                                                                                                                                    |
+|-----------------|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| OAuth2          | An OAuth2 client (external server/service) that is treating Kani as the IDP it trusts to validate user authorisation to it's resources.                                                   |
+| Application     | An LDAP application password context, allowing per-user/per-device/per-application passwords to validated, as well as defining group based authorisation of whom may use this application |
+| Service Account | An account that belongs to a process or automation that needs to read from or write to Kanidm, or a Kanidm related service.                                                               |
+| Machine Account | A domain joined machine that is reads user posix or login information. May be used to configure machine service accounts in future.                                                       |
+| Break Glass     | An emergency access account used in disaster recovery.                                                                                                                                    |
+| Person          | A humans owned account that needs to authenticate day to day, and self manage their own credentials. A person may need to manage other accounts and resource types                        |
+
+This has the benefit that it makes it easier to assign the permissions via ACP (since we can filter
+on the Target class *and* capability type).
+
+### Example
+
+An Email service has an SMTP gateway and OAuth2 web ui.
+
+Although this is "the email service" it is made up of multiple parts that each have their own intents.
+
+The Webui has an Oauth2 client created to define the relationship of who may access the webui.
+
+An LDAP application is made to allow IMAP/SMTP processes to authenticate users with application passwords and
+to read users PII via LDAP.
+
+## Below was the drafting process of some ideas
+
+### Attach roles to service accounts.
 
 In this approach we centre the service account, and allow optional extension of other concerns. This
 would make OAuth2 applications an extension of a service account. Similar Application as well.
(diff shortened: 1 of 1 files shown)


CEbbinghaus:
I am a big fan of the capabilities. It would be awesome if we could perhaps make them even a little bit more granular and split service accounts into LDAP vs API. I have many Service Accounts none of which I want to give API access. But they should all be able to query via LDAP. And vice versa (although significantly less important) i wouldn't want an API token to be able to query via LDAP.

Firstyear:
This wont be capabilities that *you* the admin get to assign, it's capabilities internally that *we* assign to each role type to make it easier for us in future to shuffle around what each role can or can't do. 

As explained in the design, the oauth2 role is fundamentally separate to the email auth application and trying to write a clean admin interface to that will be a nightmare if we make it too generic. Some separation of concern, is good.

book/src/developers/designs/service_accounts.md
```diff
@@ -0,0 +1,161 @@
...
+Note that both of these have some kerberos centric ideas as KRB requires service accounts to mutually
+authenticate to clients, which means they need to maintain credentials. This is different to our needs,
+but there are still some ideas in these docs worth knowing about and considering like group managed
+service accounts (gMSA).
+
+## Current state of affairs
+
+We have:
+
+* Break glass accounts are service accounts, may not have delegated management.
```
yaleman, line 28:
```suggestion
* Break glass accounts (admin/idm_admin) are service accounts, may not have delegated management.
```

book/src/developers/designs/service_accounts.md
```diff
@@ -0,0 +1,161 @@
...
+want to fully merge them, but we want to be able to mix-match what they require.
+
+This also makes it possible in the future that we can more easily assign (or remove) a capability
+from an account type.
+
+To achieve this we should introduce the idea of capabilities - capabilities can act via schema
+classes, and we can extend the schema such that only the parent class needs to know that the
+capabilities class is required.
+
+This allows us to nominate more carefully what each role type can or can't do, and keeps things
```
yaleman, line 48: keeps things.... ?

book/src/developers/designs/service_accounts.md
```diff
@@ -0,0 +1,161 @@
...
+| Person          | No               | Yes                          | Yes                 |
+
+A key requirement of this is that we want each role to have a defined intent - it shouldn't be
+the everything role, it still needs to be focused and administered in it's own right.
+
+|                 | Intent                                                                                                                                                                                    |
+|-----------------|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| OAuth2          | An OAuth2 client (external server/service) that is treating Kani as the IDP it trusts to validate user authorisation to it's resources.                                                   |
+| Application     | An LDAP application password context, allowing per-user/per-device/per-application passwords to validated, as well as defining group based authorisation of whom may use this application |
```
yaleman, line 65: If they're LDAP specific we should stop calling them just "applications".

Firstyear: We might expose these applicgation passwords on a different auth mechanism to LDAP in the future .... this is jsut a way to have application passwords.

book/src/developers/designs/service_accounts.md
```diff
@@ -0,0 +1,161 @@
...
+
+An LDAP application is made to allow IMAP/SMTP processes to authenticate users with application passwords and
+to read users PII via LDAP.
+
+## Below was the drafting process of some ideas
+
+### Attach roles to service accounts.
+
+In this approach we centre the service account, and allow optional extension of other concerns. This
+would make OAuth2 applications an extension of a service account. Similar Application as well.
```
yaleman, line 90: > Similar Application as well.

I can't parse this - do you mean "applications would be similar" ?

book/src/developers/designs/service_accounts.md
```diff
@@ -0,0 +1,161 @@
...
+* These whole stack service accounts get access to every auth type and feature available
+
+CONS:
+
+* Makes the API around service accounts a bit messier
+* Compromise of the SA or SA Manager may lead to higher impact due to more features in one place
+* May be confusing to administrators
+* More "inheritance" of schema classes, when we may want to try to simplify to single classes in line with SCIM.
+* Harder to audit capabilities
+* The administration UI becomes a shitshow as the Service Account is now a kitchen sink.
```
yaleman, line 108: Not that much, they just have roles. It's probably *clearer* what's going on for any given account.

book/src/developers/designs/service_accounts.md
yaleman: OK, just need to avoid confusion with the design/intention

book/src/developers/designs/service_accounts.md
Firstyear: You'd need to have a way on the cli to say "attach role to X account" and then handle all the cases when the role isn't attached and someone tries to admin it.

Firstyear: It also means you lose "list all oauth2 clients" it's just "list all service accounts". We'd need filtering in the cli to do "list all service accounts, that also are oauth2 clients' for example.

CEbbinghaus: wouldn't that translation happen within the api? so `/list/oauth2` (imaginary) would look up all service accounts that have the oauth2 extension/class/schema

yaleman: Precisely. We've got computers to do this 😄

jinnatar:
Tried to read through all of the proposal and understood most of it I think, but came out of it not quite understanding what the intent is with "Applications" and whether they would solve my issue. So, apologies if this is an ill-informed or misconception-rich question, I didn't quite manage to understand all of it. Tacking it on here because perhaps it's a useful viewpoint for the future of service accounts to consider what they look like to someone that is not a Kanidm dev.

## The problem I'm trying to solve
- I have an application, let's say Forgejo. It's integrated with Kanidm via OIDC and I've disabled the local login option so that human users have a decent login experience.
- Within that application in addition to the usual human users, I have let's say a Renovate bot account `renovate-bot` which does CI things and leaves comments on issues.
- If I need to tweak that bot account and for example issue an API token for it, if I don't want to go learn their complicated API I instead sadly need to interactively log in with said account.
- At this point I turn to Kanidm to figure out how can I do an interactive OIDC login with the bot account into Forgejo.

## The confusion part
- I'm a human trying to log into what I consider a "service account" in Forgejo, so creating a `person` account in Kanidm feels wrong, while creating a `service-account` feels right.
- However, best I can tell, the reality is the opposite, the only way to get an OAuth login capable account is to create a `person`. Unfortunately creating a person has downsides such as being subject to the rules of persons, such as not allowing plain old passwords. This is "fine" for the case of the interactive login to Forgejo, but a massive PITA if what I need to do with an "external service account" is non-interactive. That's a whole different problem though, but it feels like solving for any of that is hindered severely by needing to cosplay as a human subject to human rules. :-)
- Now, if Kanidm had as per this proposal an `Application` type account, then after flailing and failing with a `service-account` I'd probably next attempt with an application account. But, according to the proposed capability matrix that also would not be allowed to utilize OAuth. So I'd probably flail & fail with that too.
- Again in the proposed cap matrix, it would seem that my original instinct of using a `service-account` _might_ actually be the correct one since OAuth is marked as: `Yes (via session grant (TBD)` .. assuming that this "session grant" would be usable for any of these scenarios. Unfortunately I'm not finding much on the internets that would clarify what is meant with sessions grants. (But that could just be more about how broken the internet is right now for finding anything technical.)

## The sidestep
Thinking back to the original problem, perhaps what I _actually_ need is a way to trigger the OIDC flow, log in with my own human person account BUT somehow have Kanidm instead provide the `sub` etc of a service account instead. Bit of a UX nightmare though. That would at least allow for interactive, and then minting application specific service tokens that bypass OIDC. Of course the real solution would be for Forgejo to have some concept of service accounts of it's own that would allow an admin to manage them without needing to log into it. But it doesn't seem to be a priority. :woman_shrugging:

Firstyear:
What you are describing here @jinnatar is session delegation - the ability to request a session on behalf of something else. I think that's a valid request to have too.

PR #3540, opened 2025-04-02
implement notify-reload protocol correctly

### LINKED ISSUE #3534: Remove SD Notify
https://bugzilla.opensuse.org/show_bug.cgi?id=1240346

Based on the outcome of this issue, we may need to remove systemd notification integration from Kanidm. This is because systemd's advertised features do not work as intended and are causing our service to be killed during reloads.

arianvp:
I edited this in my browser.  The indentation of this file looks really broken. not sure why

# Change summary
See https://www.freedesktop.org/software/systemd/man/latest/sd_notify.html

which now has a full python  and C example of the wire protocol

You should be setting `MONOTONIC_TIMESTAMP` in the same message as `RELOADING=1` not separate message

so:

```
sd_notify("RELOADING=1\nMONOTONIC_TIMESTAMP=xxxx")
sleep(5)
sd_notify("READY=1")
```

as opposed to what you're doing now:

```
sd_notify("RELOADING=1")
sd_notify("MONOTONIC_TIMESTAMP=xxxx")
sleep(5)
sd_notify("READY=1")
sd_notify("MONOTONIC_TIMESTAMP=xxxx")
```

Fixes maybe #3534 

Checklist

### DIFF
--- a/server/daemon/src/main.rs
+++ b/server/daemon/src/main.rs
@@ -764,7 +764,7 @@ async fn kanidm_main(config: Configuration, opt: KanidmdParser) -> ExitCode {
                                                     #[cfg(target_os = "linux")]
                                                     {
                                 if let Ok(monotonic_usec) = sd_notify::NotifyState::monotonic_usec_now() {
-                                else {
+                               } else {
                                     error!("CRITICAL!!! Unable to access clock monotonic time. SYSTEMD WILL KILL US.");
                                 };
                                                     let _ = sd_notify::notify(true, &[sd_notify::NotifyState::Reloading, monotonic_usec]);
--- a/server/daemon/src/main.rs
+++ b/server/daemon/src/main.rs
@@ -737,82 +737,80 @@ async fn kanidm_main(config: Configuration, opt: KanidmdParser) -> ExitCode {
                             {
                                 let mut listener = sctx.subscribe();
                                 tokio::select! {
-                                                Ok(()) = tokio::signal::ctrl_c() => {
-                                                    break
-                                                }
-                                                Some(()) = async move {
-                                                    let sigterm = tokio::signal::unix::SignalKind::terminate();
-                                                    #[allow(clippy::unwrap_used)]
-                                                    tokio::signal::unix::signal(sigterm).unwrap().recv().await
-                                                } => {
-                                                    break
-                                                }
-                                                Some(()) = async move {
-                                                    let sigterm = tokio::signal::unix::SignalKind::alarm();
-                                                    #[allow(clippy::unwrap_used)]
-                                                    tokio::signal::unix::signal(sigterm).unwrap().recv().await
-                                                } => {
-                                                    // Ignore
-                                                }
-                                                Some(()) = async move {
-                                                    let sigterm = tokio::signal::unix::SignalKind::hangup();
-                                                    #[allow(clippy::unwrap_used)]
-                                                    tokio::signal::unix::signal(sigterm).unwrap().recv().await
-                                                } => {
-                                                    // Reload TLS certificates
-                                                    // systemd has a special reload handler for this.
-                                                    #[cfg(target_os = "linux")]
-                                                    {
-                                if let Ok(monotonic_usec) = sd_notify::NotifyState::monotonic_usec_now() {
-                                } else {
-                                    error!("CRITICAL!!! Unable to access clock monotonic time. SYSTEMD WILL KILL US.");
-                                };
-                                                    let _ = sd_notify::notify(true, &[sd_notify::NotifyState::Reloading, monotonic_usec]);
-                                                    let _ = sd_notify::notify(true, &[sd_notify::NotifyState::Status("Reloading ...")]);
-                                                    }
+                                    Ok(()) = tokio::signal::ctrl_c() => {
+                                        break
+                                    }
+                                    Some(()) = async move {
+                                        let sigterm = tokio::signal::unix::SignalKind::terminate();
+                                        #[allow(clippy::unwrap_used)]
+                                        tokio::signal::unix::signal(sigterm).unwrap().recv().await
+                                    } => {
+                                        break
+                                    }
+                                    Some(()) = async move {
+                                        let sigterm = tokio::signal::unix::SignalKind::alarm();
+                                        #[allow(clippy::unwrap_used)]
+                                        tokio::signal::unix::signal(sigterm).unwrap().recv().await
+                                    } => {
+                                        // Ignore
+                                    }
+                                    Some(()) = async move {
+                                        let sigterm = tokio::signal::unix::SignalKind::hangup();
+                                        #[allow(clippy::unwrap_used)]
+                                        tokio::signal::unix::signal(sigterm).unwrap().recv().await
+                                    } => {
+                                        // Reload TLS certificates
+                                        // systemd has a special reload handler for this.
+                                        #[cfg(target_os = "linux")]
+                                        {
+                                            if let Ok(monotonic_usec) = sd_notify::NotifyState::monotonic_usec_now() {
+                                                let _ = sd_notify::notify(true, &[sd_notify::NotifyState::Reloading, monotonic_usec]);
+                                                let _ = sd_notify::notify(true, &[sd_notify::NotifyState::Status("Reloading ...")]);
+                                            } else {
+                                                error!("CRITICAL!!! Unable to access clock monotonic time. SYSTEMD WILL KILL US.");
+                                            };
+                                        }
 
-                                                    sctx.tls_acceptor_reload().await;
+                                        sctx.tls_acceptor_reload().await;
 
-                                                    // Systemd freaks out if you send the ready state too fast after the
-                                                    // reload state and can kill Kanidmd as a result.
-                                                    tokio::time::sleep(std::time::Duration::from_secs(5)).await;
+                                        // Systemd freaks out if you send the ready state too fast after the
+                                        // reload state and can kill Kanidmd as a result.
+                                        tokio::time::sleep(std::time::Duration::from_secs(5)).await;
 
-                                                    #[cfg(target_os = "linux")]
-                                                    {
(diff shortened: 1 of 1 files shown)


server/daemon/src/main.rs
```diff
@@ -771,15 +763,11 @@ async fn kanidm_main(config: Configuration, opt: KanidmdParser) -> ExitCode {
...
                                                     #[cfg(target_os = "linux")]
                                                     {
-                                                    let _ = sd_notify::notify(true, &[sd_notify::NotifyState::Reloading]);
-                                                    // CRITICAL - if you do not send a monotonic usec message after a reloading
-                                                    // message, your service WILL BE KILLED.
                                 if let Ok(monotonic_usec) = sd_notify::NotifyState::monotonic_usec_now() {
-                                let _ =
-                                    sd_notify::notify(true, &[monotonic_usec]);
-                                } else {
+                                else {
```
arianvp, line 767: I think I accidentally a bracket here: `}`
```suggestion
                               } else {
```

server/daemon/src/main.rs
```diff
@@ -771,15 +763,11 @@ async fn kanidm_main(config: Configuration, opt: KanidmdParser) -> ExitCode {
...
                                                     #[cfg(target_os = "linux")]
                                                     {
-                                                    let _ = sd_notify::notify(true, &[sd_notify::NotifyState::Reloading]);
-                                                    // CRITICAL - if you do not send a monotonic usec message after a reloading
-                                                    // message, your service WILL BE KILLED.
                                 if let Ok(monotonic_usec) = sd_notify::NotifyState::monotonic_usec_now() {
-                                let _ =
-                                    sd_notify::notify(true, &[monotonic_usec]);
-                                } else {
+                               } else {
```
arianvp, line 767:
```suggestion
                                } else {
```

Firstyear:
That is the most unhinged bullshit, and exactly the kind of ambiguous garbage I have come to expect from systemd. It's already bad enough they silently made monotonic usec a requirement without telling anyone and broke reloading. 

I appreciate the fix but I am actually half thinking to remove the systemd integration as they have proven they are not responsible stewards of this process. Who says they won't break reloading in a months time?

Firstyear:
Also to be clear

"""
When initiating the reload process the service is expected to reply with a notification message via [sd_notify(3)](https://www.freedesktop.org/software/systemd/man/latest/sd_notify.html#) that contains the "RELOADING=1" field in combination with "MONOTONIC_USEC=" set to the current monotonic time (i.e. CLOCK_MONOTONIC in [clock_gettime(2)](https://man7.org/linux/man-pages/man2/clock_gettime.2.html)) in μs, formatted as decimal string.
"""

This doesn't say "RELOADING=1 and MONOTONIC_USEC must be in the *same message*. It only says "in combination" which to me is ambiguous that they can be sequential.

Firstyear:
Closing as I plan to remove notify-reload support.

arianvp:
Okay this is a slightly disappointing response. 

> already bad enough they silently made monotonic usec a requirement without telling anyone and broke reloading.

This is simply not true. This behaviour was introduced in the same PR as `Type=notify-reload ` https://github.com/systemd/systemd/pull/25916

And is only **required** if your service is of `Type=notify-reload`  . If your service is of `Type=simple` reloading can still be done with `ExecReload=` like before. And sending a `RELOADING=1` message on a `Type=simple` service without the monotonic field still works as far as I am aware. If it doesn't then that's a serious bug.  This change in no way breaks existing services. Unless you manually changed to `Type=notify-reload` and then not implement the protocol. Which is not a bug in systemd but in kanidm.

Systemd did their best to update the docs on sd_notify.   Even giving code samples that are not systemd specific. Did you even click the link I sent?
 you just conveniently read over the fact that they have complete code samples, in multiple programming languages, that are completely unambiguous? 

You do you, it's your project, but this feels unnecessarily hostile.

I'll make a PR to systemd change the docs to replace the word "in combination" to "in a single message" instead as opposed to arguing on the Internet about how bad the docs are and will try in the future again once the docs are on par with your level of scrutiny.

Peace ✌️

Firstyear:
> Systemd did their best to update the docs on sd_notify. Even giving code samples that are not systemd specific. Did you even click the link I sent? you just conveniently read over the fact that they have complete code samples, in multiple programming languages, that are completely unambiguous?

I read the *docs* here https://www.freedesktop.org/software/systemd/man/249/sd_notify.html which does not anywhere clearly state what you need to do to ensure systemd doesn't kill you. 

Which btw, is the top result on DDG. 

If I go to https://www.freedesktop.org/software/systemd/man/latest/sd_notify.html 

"""
RELOADING=1

    Tells the service manager that the service is beginning to reload its configuration. This is useful to allow the service manager to track the service's internal state, and present it to the user. Note that a service that sends this notification must also send a "READY=1" notification when it completed reloading its configuration. Reloads the service manager is notified about with this mechanisms are propagated in the same way as they are when originally initiated through the service manager. This message is particularly relevant for Type=notify-reload services, to inform the service manager that the request to reload the service has been received and is now being processed.

    Added in version 217.
"""

It still doesn't say anything about monotonic usec.

So lets look at "latest" here. 

There is a comment in monotonic usec:

"""
MONOTONIC_USEC=…

    A field carrying the monotonic timestamp (as per CLOCK_MONOTONIC) formatted in decimal in μs, when the notification message was generated by the client. This is typically used in combination with "RELOADING=1", to allow the service manager to properly synchronize reload cycles. See [systemd.service(5)](https://www.freedesktop.org/software/systemd/man/latest/systemd.service.html#) for details, specifically "Type=notify-reload".

    Added in version 253.
"""

Which is not where anyone would look for details about reloading. Because I'm reloading. Not doing a "monotonic usec" event.

I find it odd that RELOADING was adding in version 217, and monotonic user requirements in what 253? Because when were were a "type=Notify" service, this all worked fine. Even with a reloading message. It's only when you upgrade to 253 and think "oh look, notify-reload exists" but nothing clearly states "if you don't attach monotonic_usec, system will end you". 

That's the issue - someone who isn't a developer of systemd, reading these specs, is getting caught out because of this subtle, obscure behaviour, and it's not okay. If someone blindly (like apparently i did) trusts systemd that notify can become notify-reload because I sent the right reloading messages, they will get burned. 

So yeah, I'm annoyed, I'm disappointed in systemd here. I have tried to be a good citizen, follow their docs, and instead I get bug reports. I have spent multiple *days* chasing this problem. And I'm frankly not okay with it. 

> 
> I'll make a PR to systemd change the docs to replace the word "in combination" to "in a single message" instead as opposed to arguing on the Internet about how bad the docs are and will try in the future again once the docs are on par with your level of scrutiny.

I am reading this as "my level of scrutiny was insufficient" when in reality, I tried to read everything I could about type=notify-reload, and this is what happened. Clearly, there is a communication failure on systemd's side with the docs, if someone can't read the spec, and get it implemented *correctly*.

Firstyear:
So first, I want to apologise - I was really angry about this issue as I have been fighting systemd for months to get this to work, and it has been a complete pain. I probably should have walked away before typing.

Second, I was angry because people were saying things like "my level of scrutiny" or "I have a comprehension issue". Both of these are blaming me, the user, for what I read. Not only that, they felt like attacks on me and my ability. This didn't help. 

My job is to implement things from specifications day to day, and ensure they work. So when an issue occurs where someone can't implement a spec from it's documentation correctly, after multiple reviews and attempts, you have to ask "why did the documentation fail". Multiple people have said "william is wrong" here, but I also don't have access to the systemd community like you may. 

The systemd docs around this topic are written in a way where you *already know the answer*. They don't communicate clearly the invariants that have to hold true, not do the communicate the *consequences for failure* to uphold these invariants. I have struggled to get this to work, and when it does appear to work, it doesn't. 

Not only this, the API design is prone to failure here, especially for upgrading users from notify to notify-reload. 

* The RELOADING=1 message docs need to state "this message must be following in the same message by a MONOTONIC_USEC=X message. If this is not done, systemd will consider the process hung and will terminate it". Give *examples* as well, because without examples that show clearly how it works, it's hard to see what some of this can mean! 
* People won't read about MONOTONIC_USEC when looking for RELOADING messages. They are in the mindset of "I am reloading a service". So that's where they will look.
* Consider *redesigning* the API to have a RELOADING_USEC message that is a combination of RELOADING and MONOTONIC_USEC that indicates reloading and has the USEC as the parameter. This is *one message* and becomes failure proof.
* notify-reload in the absecence of the MONOTONIC_USEC message, shouldn't just kill the process. If the message isn't there *make a safe assumption* that the caller is still there, but the documentation failed to communicate to them. 

So yeah, I fucked up to "scrutinise this one" properly. I'm human. I make mistakes. But I also didn't already know the answer, and I did the best I could. My job is to implement protocols from specification documents, so if someone with my level of skills and specialisation can't implement this correctly, then *maybe the documentation is failing to communicate correctly to humans*. 

I will accept this PR, but I am still cautious, as this is not the first time subtle and confusing systemd behaviour has caught us out, and it's something that causes harm to our users when they run into these issues. I need to protect our users and their services, so I need to protect their interests.

Firstyear (approved):

server/daemon/src/main.rs
```diff
@@ -745,86 +737,80 @@ async fn kanidm_main(config: Configuration, opt: KanidmdParser) -> ExitCode {
...
+                                            } else {
+                                                error!("CRITICAL!!! Unable to access clock monotonic time. SYSTEMD WILL KILL US.");
+                                            };
+                                        }
+
+                                        sctx.tls_acceptor_reload().await;
+
+                                        // Systemd freaks out if you send the ready state too fast after the
+                                        // reload state and can kill Kanidmd as a result.
+                                        tokio::time::sleep(std::time::Duration::from_secs(5)).await;
```
CEbbinghaus, line 778: This might not be needed

Firstyear: Lets see how it goes in the wild first, then we see if we remove it.

PR #3541, opened 2025-04-03
Remove systemd notify-reload

### LINKED ISSUE #3534: Remove SD Notify

Firstyear:
systemd notify-reload has been nothing but problems for us. Since it was introduced, systemd broke it once by forcing a new parameter to be added to messages, and to further that the parameter they added was documented so poorly we couldn't get it implemented correctly.

Remove notify-reload, and return to classic - and reliable - HUP signals.

# Change summary

-

Fixes #3534 

Checklist

### DIFF
--- a/server/daemon/src/main.rs
+++ b/server/daemon/src/main.rs
@@ -724,14 +724,6 @@ async fn kanidm_main(config: Configuration, opt: KanidmdParser) -> ExitCode {
                 #[cfg(target_os = "linux")]
                 {
                     let _ = sd_notify::notify(true, &[sd_notify::NotifyState::Ready]);
-                    // Undocumented systemd feature - all messages should have a monotonic usec sent
-                    // with them. In some cases like "reloading" messages, it is undocumented but
-                    // failure to send this message causes the reload to fail.
-                    if let Ok(monotonic_usec) = sd_notify::NotifyState::monotonic_usec_now() {
-                        let _ = sd_notify::notify(true, &[monotonic_usec]);
-                    } else {
-                        error!("CRITICAL!!! Unable to access clock monotonic time. SYSTEMD WILL KILL US.");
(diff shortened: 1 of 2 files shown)


server/daemon/src/main.rs
```diff
@@ -724,14 +724,6 @@ async fn kanidm_main(config: Configuration, opt: KanidmdParser) -> ExitCode {
                 #[cfg(target_os = "linux")]
                 {
                     let _ = sd_notify::notify(true, &[sd_notify::NotifyState::Ready]);
-                    // Undocumented systemd feature - all messages should have a monotonic usec sent
-                    // with them. In some cases like "reloading" messages, it is undocumented but
-                    // failure to send this message causes the reload to fail.
-                    if let Ok(monotonic_usec) = sd_notify::NotifyState::monotonic_usec_now() {
-                        let _ = sd_notify::notify(true, &[monotonic_usec]);
-                    } else {
-                        error!("CRITICAL!!! Unable to access clock monotonic time. SYSTEMD WILL KILL US.");
```
CEbbinghaus, line 733: Happy to see this one gone :)

CEbbinghaus (approved):
sytemdeez nuts

Kranzes:
FWIW:
- The parameter has always been required and documented since the [initial implementation](https://github.com/systemd/systemd/pull/25916)
- The docs clearly state, in multiple places, that you need to send the parameter together in a single message

Firstyear:
https://github.com/kanidm/kanidm/pull/3540#issuecomment-2777169843

PR #3542, opened 2025-04-03
20250402 3423 proxy protocol

### LINKED ISSUE #3423: Add support for v2 PROXY protocol

Firstyear:
# Change summary

Adds proxy v2 header support. 

Fixes #3423

Checklist


server/core/src/https/mod.rs
```diff
@@ -415,8 +414,39 @@ pub(crate) async fn handle_conn(
...
     stream: TcpStream,
     mut app: IntoMakeServiceWithConnectInfo<Router, ClientConnInfo>,
-    addr: SocketAddr,
+    connection_addr: SocketAddr,
+    enable_haproxy_hdr: bool,
 ) -> Result<(), std::io::Error> {
+    // IMPORTANT: We only check the proxy header on non-loopback requests. This is because
+    // the healthcheck can't have the proxy header added. Generally it also makes it a bit
+    // nicer as well for localhost-administration of the instance.
+    let (stream, client_addr) = if enable_haproxy_hdr && !connection_addr.ip().is_loopback() {
```
Firstyear, line 423: @yaleman I'm not sure how I feel about the localhost exception here - I can see this going weird with a haproxy on the same machine, or not catching all the possible use cases. The cause is in daemon/main.rs in the healthcheck, which does a get req to /status, but it also calls via the address/origin, so it's not always guaranteed that it will actually call to localhost either, it could call to a local machine addr and this fix still breaks. 

So I think the actual fix here would be to change the healthcheck to use the unix socket for admin tasks, and have something that can "ping" all our worker tasks and check they are actually active and available. That way we avoid this mess.

Alternately, if we have proxy v2 enabled, we have to disable the healthcheck, which isn't great. 

Any ideas to get around this?

vDorst: I saw with multipule projects that I setup in the last weeks, that you can specify the ip-address/range of your trusted proxy. And that you expect the proxy_protocol.

Example
* [Apache2 RemoteIPProxyProtocol](https://httpd.apache.org/docs/2.4/mod/mod_remoteip.html#RemoteIPProxyProtocol)

Other examples I saw don't support proxy_protocol but still let you specify proxy ip so the application accepts the `X-Forward-IP` header.
* [home-assistant](https://www.home-assistant.io/integrations/http#reverse-proxies)
* [forgejo](https://forgejo.org/docs/latest/admin/config-cheat-sheet/#security-security) Has a proxy_protocol option but is broken. But also has `REVERSE_PROXY_TRUSTED_PROXIES` config option.

Firstyear: """
The receiver MUST be configured to only receive the protocol described in this
specification and MUST not try to guess whether the protocol header is present
or not. This means that the protocol explicitly prevents port sharing between
public and private access. Otherwise it would open a major security breach by
allowing untrusted parties to spoof their connection addresses. The receiver
SHOULD ensure proper access filtering so that only trusted proxies are allowed
to use this protocol.
"""

So we need to always do PROXY, or never. That's part of my concern here too.

Firstyear: Saying this the trusted IP range for the proxy, could solve this issue.

Firstyear:
The suggestion by @vDorst neatly solves the issue I had with healthcheck, but also improves configuration/trust of header sources in general, so I committed that to the PR.

server/core/src/ldaps.rs
```diff
@@ -112,28 +118,65 @@ async fn client_process<STREAM>(
 }
 
 async fn client_tls_accept(
-    tcpstream: TcpStream,
+    stream: TcpStream,
     tls_acceptor: SslAcceptor,
-    client_socket_addr: net::SocketAddr,
+    connection_addr: SocketAddr,
     qe_r_ref: &'static QueryServerReadV1,
+    trusted_haproxy_ips: Option<Arc<HashSet<IpAddr>>>,
```
yaleman, line 125: See similar changes above. It's not a haproxy header/ip list, it's the proxy protocol.

server/testkit/tests/testkit/https_extractors.rs
```diff
@@ -5,12 +5,13 @@ use std::{
```
yaleman, line 5: Where are the tests in this PR for the proxy v2 headers or `HttpAddressInfo::XForwardForAllSourcesTrusted`?

yaleman (changes requested):
Lots of things to fix in this one. We're breaking existing configs, leaving modes untested and ... it's not the "haproxy" protocol as it's supported by lots of systems.

server/core/src/config.rs
```diff
@@ -100,6 +100,109 @@ pub struct TlsConfiguration {
...
+}
+
+#[derive(Deserialize, Debug, Clone, Default)]
+pub enum HttpAddressInfo {
+    #[default]
+    None,
+    #[serde(rename = "x-forward-for")]
+    XForwardFor(HashSet<IpAddr>),
+    #[serde(rename = "x-forward-for-all-source-trusted")]
+    XForwardForAllSourcesTrusted,
```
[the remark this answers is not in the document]
Firstyear: On purpose. It only exists as a transition mechanism to allow the v1 config to keep working.

server/core/src/config.rs
```diff
@@ -217,7 +320,10 @@ pub struct ServerConfigV2 {
     role: Option<ServerRole>,
     log_level: Option<LogLevel>,
     online_backup: Option<OnlineBackup>,
-    trust_x_forward_for: Option<bool>,
+
+    http_client_address_info: Option<HttpAddressInfo>,
+    ldap_client_address_info: Option<LdapAddressInfo>,
+
```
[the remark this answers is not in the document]
Firstyear: v2 was only just added in this release cycle, that's why I'm adding this now.

server/core/src/https/extractors/mod.rs
```diff
@@ -108,7 +121,13 @@ impl FromRequestParts<ServerState> for VerifiedClientInformation {
                 )
             })?;
 
-        let ip_addr = if state.trust_x_forward_for {
+        let trust_x_forward_for = state
+            .trust_x_forward_for_ips
+            .as_ref()
+            .map(|range| range.contains(&connection_addr.ip()))
+            .unwrap_or_default();
```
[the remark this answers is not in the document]
Firstyear: It does handle it - range.contains has code to handle if this is all sources trusted.

Firstyear:
No we aren't breaking a config, remember we only just added v2 here, I was really careful about this ...

server/core/src/https/mod.rs
yaleman: So there's three situations for the health check, and it's possible for the health check to know what's configured because it loads the server config on startup:

1. **no proxy protocol/forward header**, this is easy
2. **forward header** - doesn't matter since the forward-ips don't control access
3. **proxy protocol enabled** - we could:
    - make the healthcheck *require* the `--check-origin` flag when it finds proxy protocol configured, meaning it traverses the proxy
    - fall back to a unix socket (though this stops checking the health of the network endpoint)

I'd much prefer that the healthcheck defaults to using a network check, since there's a lot of reasons why the port won't be accessible, or you want to check that the end-to-end traffic works (when you configure `--check-origin`), but falling back to a socket in the case where it's infeasible to proxy the traffic is OK.

I pondered some kind of "make a client that acts like it's running PROXY protocol" but that's silly.

yaleman (approved):

server/core/src/https/mod.rs
Firstyear: The problem with the end-to-end network check is that in docker the health check can trigger restarts - so if your healthcheck is via a load balancer, then suddenly all your containers start restart looping, when they don't need to.

So I get the network check sentiment, make sure the same port we expose to the LB is working, but the end-to-end check is better run by a dedicated monitoring service. The healthcheck should just be for "are we healthy within our little box".

yaleman: Yep, for sure. It's almost like folks need to know how things work before they use 'em 😄

server/core/src/config.rs
```diff
@@ -100,6 +100,111 @@ pub struct TlsConfiguration {
...
+            Self::NonContiguousIpSet(range) => range.contains(ip_addr),
+        }
+    }
+}
+
+#[derive(Deserialize, Debug, Clone, Default)]
+pub enum HttpAddressInfo {
+    #[default]
+    None,
+    #[serde(rename = "x-forward-for")]
```
argonaut0, line 154: Arn't the rest of the config directives using underscores for word delimiters?

Firstyear: Yes, but it's too late now as this has been released. :(

PR #3553, opened 2025-04-08
Improve token handling

Firstyear:
It was possible that a token could be updated in a way that caused existing cached information to be lost if an event was delayed in it's write to the user token.

To prevent this, the writes to user tokens now require the HsmLock to be held, and refresh the token just ahead of writing to ensure that these data can't be lost. The benefit to this approach is that readers remain unblocked by a writer.

In the future I may have to rethink the approach as a whole to make this CoW or similar, as this places a bit of a burden on providers to implement the correct data shuffling at this time. 

Checklist

### DIFF
--- a/unix_integration/resolver/src/idprovider/interface.rs
+++ b/unix_integration/resolver/src/idprovider/interface.rs
@@ -195,7 +195,7 @@ impl Into<PamAuthResponse> for AuthRequest {
 
 pub enum AuthResult {
     Success,
-    SuccessUpdate { token: UserToken },
+    SuccessUpdate { new_token: UserToken },
     Denied,
     Next(AuthRequest),
 }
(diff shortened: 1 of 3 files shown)


unix_integration/resolver/src/idprovider/interface.rs
```diff
@@ -194,7 +194,8 @@ impl Into<PamAuthResponse> for AuthRequest {
 }
 
 pub enum AuthResult {
-    Success { token: UserToken },
+    Success,
+    SuccessUpdate { token: UserToken },
```
yaleman, line 198: Can this be new token so it's clear that we've supplied a new one? I know *you* know it right now, or digging into the source, but it's just one of those DX things.

Firstyear: Done!

yaleman (approved):
Woo!

PR #3574, opened 2025-04-22
fix(web): Preserve SSH key content on form validation error

### LINKED ISSUE #3570: adding ssh key via web UI without title discards key content
### I did this
1. open the ssh key form in the web ui
2. add an ssh key (but no title)
3. click submit
4. get an error that the key needs a title

### I expected the following
my key value being present in the form

### This happened instead
had to insert the key anew, just because the title was missing

### Kanidm version details
(as received from the server admins, I'm only a user)

- Output of `kanidm(d) version`: 1.5.0
- If not a container, how'd you install it: NixOS
- Operating System / Version (On Unix please post the output of `uname -a`): NixOS

Keerthi421:
## Issue Fixed

Closes #3570

#Description
When adding a new SSH public key via the web UI, if the user submitted the form without providing a title, the form would redisplay with a validation error, but the SSH key content previously entered would be lost. This required the user to re-paste their key.

This PR fixes the issue by:
1.  Modifying the `AddSshPublicKeyPartial` template context struct in `server/core/src/https/views/reset.rs` to include an optional `key_value` field.
2.  Updating the `view_add_ssh_publickey` handler function to populate this `key_value` field with the submitted key when returning the form due to a validation error (either invalid key format or missing/duplicate title).
3.  Updating the `credential_update_add_ssh_publickey_partial.html` template to render the content of `key_value` within the `<textarea>` if it's present.

Now, if the form submission fails due to a missing title or other validation error, the key content is preserved in the form, allowing the user to correct the error without needing to re-enter the key.
Checklist

#3570

### DIFF
--- a/server/core/src/https/views/reset.rs
+++ b/server/core/src/https/views/reset.rs
@@ -121,6 +121,7 @@ struct SetUnixCredPartial {
 struct AddSshPublicKeyPartial {
     title_error: Option<String>,
     key_error: Option<String>,
+    key_value: Option<String>,
 }
 
 #[derive(Serialize, Deserialize, Debug)]
@@ -902,6 +903,7 @@ pub(crate) async fn view_add_ssh_publickey(
             return Ok((AddSshPublicKeyPartial {
                 title_error: None,
                 key_error: None,
+                key_value: None,
             },)
                 .into_response());
         }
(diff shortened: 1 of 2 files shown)


Keerthi421:
@yaleman Thanks for merging

Issue #3476, opened 2025-02-28
Removal of OpenSSL


Firstyear:
This is a blanket issue/tracker for the removal of OpenSSL in Kanidm and all Kanidm related projects.

The rust-openssl library initially was the only library which was able to perform a number of critical tasks (especially around webauthn). However, the rust-openssl maintainer has demonstrated an unwillingness to accept improvements, and this is now causing issues within Kanidm and Webauthn-RS which we have to work around. 

Additionally, in parallel the ecosystem around rust cryptography has improved greatly - rustls no longer relies solely on ring (another broken ecosystem in different ways) and now can be used with aws-lc-rs and in the future, hopefully rustcrypto. The rustcrypto ecosystem now supports a lot of what we need. And improvements to tpm/pkcs11 support from the parallaxsec group (who have all been super helpful) has indicated that we should move away. 

As a result, we plan to begin to phase out the use of OpenSSL in favour of rustcrypto when possible, and for now with rustls we need to remain with aws-lc-rs. 

The first phase of this is the creation of a glue-crate that binds all of the rustcrypto pieces in one place to ensure that all the needed elements are available in a consistent way. A weakness of the rustcrypto system is "micro crates" that makes it sometimes near impossible to see how things fit together. A glue crate resolves this as it allows traits to document properly to other elements to see what is possible.

The glue crate is already being developed, this is not a request for assistance. 

Once the glue crate is in place, smaller stages of rewrite will commence.

PR #3577, opened 2025-04-24
Drop fernet in favour of JWE

Firstyear:
This drops the use of fernet from OAuth2 in favour of JWE. To achieve this cleanly, we swap OAuth2 to using our internel key object handler so that in future we can consider the use of pkcs11 devices. This also makes it easier in general to handle any future cryptographic changes, since the key object types are a facade over the cryptographic provider. 

Relates #3476

Checklist


server/Dockerfile
```diff
@@ -65,6 +65,8 @@ RUN <<EOF
     ls -Rla /out/libs-root
 EOF
 
+RUN ls /usr/src/kanidm/target/release/kanidmd
```
yaleman, line 68: 🤔 what's this here for?

Firstyear: Opps, me trying to debug why the build was failing. I'll remove this.

tools/cli/src/opt/kanidm.rs
```diff
@@ -1256,6 +1263,28 @@ pub enum Oauth2Opt {
     #[cfg(feature = "dev-oauth2-device-flow")]
     /// Disable OAuth2 Device Flow authentication
     DeviceFlowDisable(Named),
+    /// Rotate the signing and encryption keys used by this client. The rotation
+    /// will occur at the specified time, or immediately if the time is in the past.
+    /// Past signatures will continue to operate even after a rotation occurs. If you
+    /// have concerns a key is compromised, then you should revoke it instead.
+    #[clap(name = "rotate-cryptographic-keys")]
+    RotateCryptographicKeys {
```
yaleman, line 1271: The help string here needs an example/explanation of what format the time needs to be in, and I'd suggest adding a `--now` flag that the client uses to set it to "now" when they invoke it.

yaleman (changes requested):
Minor CLI nit but otherwise woo?

Firstyear:
@yaleman Done, added the docs and support for 'now' as a key word.

yaleman (approved):

PR #3578, opened 2025-04-24
feat(cli): Allow clearing legalname field via empty string

### LINKED ISSUE #3573: Allow clearing legalname field

I just set the legalname field on a user with the Kanidm CLI and then realized that I don't need it and wanted to clear it again, but then noticed that the CLI doesn't allow that.

```
% kanidm person update USER -l ''
error: a value is required for '--legalname <LEGALNAME>' but none was supplied

For more information, try '--help'.
```

Above command clears the legal name.

Separate flag to clear it such as -L/--clear-legalname.

Keerthi421:
This PR fixes #3573 by allowing users to clear the legalname field using an empty string value ('') in the CLI.

Changes:
-->Removed NonEmptyStringValueParser from the legalname option in PersonUpdateOpt, allowing empty strings to be accepted
-->Updated help text to indicate that users can use '' to clear the field
-->Added a new idm_person_account_patch function in the client library that utilizes SCIM patch operations
-->Modified the person update logic to use SCIM patch operations with proper handling of empty strings (interpreted as remove operations)

Usage:
After this change, users can clear the legal name of an account by running:
"kanidm person update USERNAME -l ''
Apply
The implementation also ensures other attributes can be cleared in a similar fashion, such as mail when an empty list is provided.
Checklist


tools/cli/src/opt/kanidm.rs
```diff
@@ -594,8 +594,7 @@ pub enum ServiceAccountPosix {
 pub struct PersonUpdateOpt {
     #[clap(flatten)]
     aopts: AccountCommonOpt,
-    #[clap(long, short, help = "Set the legal name for the person.",
-    value_parser = clap::builder::NonEmptyStringValueParser::new())]
+    #[clap(long, short, help = "Set the legal name for the person. Use '' to clear.")]
```
2xsaiko, line 597:
```suggestion
    #[clap(long, short, help = "Set the legal name for the person. Pass empty text to clear.")]
```

```suggestion
    #[clap(long, short, help = "Set the legal name for the person, or remove it via empty string.")]
```

```suggestion
    #[clap(long, short, help = "Set, or remove (via empty value), the legal name for the person.")]
```

```suggestion
    #[clap(long, short, help = "Set the legal name for the person. An empty value clears the field.")]
```

Might be a bit clearer to what you're actually supposed to do here as opposed to the somewhat obtuse `''`. I can't come up with a wording I'm 100% happy with that is still concise so here are 4 variants. The last I like best but it's also the longest.

yaleman:
This looks right, can you please double-check @Firstyear ? 😄

libs/client/src/person.rs
```diff
@@ -6,6 +6,10 @@ use kanidm_proto::v1::{AccountUnixExtend, Entry, SingleStringRequest, UatStatus}
 use uuid::Uuid;
 
 use crate::{ClientError, KanidmClient};
+use kanidm_proto::scim_v1::PatchRequest;
```
Firstyear, line 9: This doesn't exist? Patch requests are quite different to what you may expect, https://www.rfc-editor.org/rfc/rfc7644#section-3.5.2 and that's why I hadn't implemented them yet.

libs/client/src/person.rs
```diff
@@ -286,4 +290,16 @@ impl KanidmClient {
...
+
+    pub async fn idm_person_account_patch(
+        &self,
+        id: &str,
+        patch_request: PatchRequest,
+    ) -> Result<(), ClientError> {
+        if patch_request.operations.is_empty() {
+            return Ok(());
+        }
+        self.perform_patch_request_scim(format!("/v1/person/{}", id).as_str(), patch_request)
```
Firstyear, line 302: This function doesn't exist?

Firstyear (changes requested):
I'm curious if you tested this PR before submitting it as a number of the referenced items and API's don't exist?

Firstyear:
This was fixed in #4228

PR #3600, opened 2025-05-07
crypto: Use .into() for OpenSSL on 32-bit platforms

fossdd:
# Change summary

- Fixes the build on 32-bit platforms (well sort of)

Checklist


yaleman:
Wait, what kind of 32-bit system are you trying to run this on?

Firstyear:
We don't support 32 bit systems anyway, so we won't fix this.

fossdd:
Sure, makes sense. thats fine.

I was just trying it out for fun and thought a two-line-fix for fixing 32-bit is trivial enough to be included.

Firstyear:
The problem is that it's not just a two line fix, because then it becomes a whole series of fixes for other parts and libraries that we rely on. So it's better to be clear now that we don't intend to support this, rather than give you false hope about it.

fossdd:
I see makes sense

PR #3607, opened 2025-05-08
Uncomment section headers in example configs

yaakov-h:
# Change summary

This PR uncomments the headers for the new http_client_address_info and ldap_client_address_info sections in the example configs.

This makes it easier for users to configure these settings by just uncommenting the `proxy-v2` or `x-forwarded-for` example settings.

Checklist

### DIFF
--- a/examples/server.toml
+++ b/examples/server.toml
@@ -87,7 +87,7 @@ origin = "https://idm.example.com:8443"
 #   format the information is provided in.
 #   Defaults to "none" (no trusted sources)
 #   Only one option can be used at a time.
-# [http_client_address_info]
+[http_client_address_info]
 # proxy-v2 = ["127.0.0.1"]
 #   # OR
 # x-forward-for = ["127.0.0.1"]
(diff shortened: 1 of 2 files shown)


Firstyear:
I think the problem here is that if we uncomment these, without there being a value to associate to it (such as the x-forward-for array) then the server won't start with a configuration error.

yaakov-h:
ah yes, you're right. I wasn't expecting that...

```
TomlError { message: "data did not match any variant of untagged enum ServerConfigUntagged", raw: None, keys: [], span: None }
```

Firstyear:
Perhaps a better way to show this is:

```
# [http_client_address_info]
# proxy-v2 = ["127.0.0.1"]
#   # OR
# [http_client_address_info]
# x-forward-for = ["127.0.0.1"]
```

This way the OR shows on both paths that the ` [http_client_address_info]` is needed, but without either being uncommented.

It's a bit messy, it's because of how toml interacts we enums here, that's why we had to do it this way.

Firstyear:
Replaced by https://github.com/kanidm/kanidm/pull/3652

PR #3630, opened 2025-05-13
Proceed when config path is missing for environment

### LINKED ISSUE #3617: 1.6.0: Issue upgrading with env var only config
### I did this
Attempt to upgrade to v1.6.0 by starting new container with said version

### I expected the following
The new version should start fine and upgrade should work

### This happened instead
Error on startup due to not finding a config file
```
Unable to open config file [Os { code: 2, kind: NotFound, message: "No such file or directory" }] 🥺
-- diagnosis for path: /data/server.toml
unable to canonicalise path
running as: 100:1000
path permissions

  /                 : DIR access: true owner: 0 group: 0 mode: rwxr-xr-x
  /data             : DIR access: true owner: 0 group: 1000 mode: rwxrwxr-x
  /data/server.toml : ERROR: Os { code: 2, kind: NotFound, message: "No such file or directory" }

  note that accessibility does not account for ACL's or MAC
-- end diagnosis

ERROR: Configuration Parse Failure: Os { code: 2, kind: NotFound, message: "No such file or directory" }
```

### Kanidm version details

   kanidmd 1.5.0
   kanidm/server:1.5.0 -> kanidm/server:1.6.0

### Any other comments

I only use env vars for configuration and this worked fine on 1.5.0 and prior versions

```
            - name: KANIDM_TLS_CHAIN
              value: /data/certs/tls.crt
            - name: KANIDM_TLS_KEY
              value: /data/certs/tls.key
            - name: KANIDM_DB_PATH
              value: /data/kanidm.db
            - name: KANIDM_DOMAIN
              value: ***
            - name: KANIDM_ORIGIN
              value: ***
            - name: KANIDM_TRUST_X_FORWARD_FOR
              value: "true"
            - name: KANIDM_BINDADDRESS
              value: "[::]:8443"
            - name: KANIDM_LDAPBINDADDRESS
              value: "[::]:3636"
            - name: KANIDM_LOG_LEVEL
              value: "info"
```

Firstyear:
# Change summary

When a configuration path is specified then the server will force it to be used, and will error if it can't be found. The container previously manually specified this path on the command line which forced the configuration to need to exist.

Remove the configuration from the docker cli, the default config path test will discover if the file exists or not and will correctly handle the flow when the file is absent and environment only configuration is used. 

Fixes #3617

Checklist

### DIFF
--- a/server/Dockerfile
+++ b/server/Dockerfile
@@ -106,4 +106,4 @@ HEALTHCHECK \
     --retries=3 \
     CMD [ "/sbin/kanidmd", "healthcheck", "-c", "/data/server.toml"]
 
-CMD [ "/sbin/kanidmd", "server", "-c", "/data/server.toml"]
+CMD [ "/sbin/kanidmd", "server"]


ig-raiderler:
The performance optimizations here are quite clever. Nice engineering work!

igs-raiderler:
I'm impressed by the modularity of this codebase. Makes it very maintainable.

yaleman (approved):
ezpz

Doridian:
@Firstyear Sorry to respond to a closed/merged PR, but reading this change randomly and looking into the Dockerfile.
The `HEALTHCHECK` line here: https://github.com/kanidm/kanidm/blob/master/server/Dockerfile#L101-L107
also specifies a config path. So that would mean if the config file does not exist, as described here, the health check would always fail still?

Firstyear:
True, it would. I'll fix the PR.

Firstyear:
@Doridian Good catch btw, https://github.com/kanidm/kanidm/pull/3656

PR #3637, opened 2025-05-14
maint: rewrite crypto Password::try_from

yaleman:
# Change summary

While investigating something else I noticed a "maybe this can be done better" comment. Turns out, nerd-sniping is a thing!

I broke the screens-of-if-statements into more discrete blocks so it's differently-easy to follow the flow, also added an error type so that when it breaks, there's a chance of knowing *why*.

Checklist

### DIFF
--- a/libs/crypto/src/lib.rs
+++ b/libs/crypto/src/lib.rs
@@ -411,7 +411,7 @@ pub enum PasswordError {
     // We guess what it is, but don't know how to handle it
     UnsupportedAlgorithm(String),
     // No idea how to decode this password
-    NoDecoderFound,
+    NoDecoderFound(String),
     ParsingFailed,
 }
 
@@ -426,8 +426,8 @@ impl Display for PasswordError {
             PasswordError::UnsupportedAlgorithm(alg) => {
                 write!(f, "Unsupported algorithm: {}", alg)
             }
-            PasswordError::NoDecoderFound => {
-                write!(f, "No decoder found for password in this format")
+            PasswordError::NoDecoderFound(hint) => {
+                write!(f, "No decoder found for password in this format - input started with '{}' - please report it upstream", hint)
             }
             PasswordError::ParsingFailed => write!(f, "Parsing of password failed"),
         }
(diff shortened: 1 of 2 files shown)


libs/crypto/src/lib.rs
```diff
@@ -399,6 +400,30 @@ enum Kdf {
...
 }
 
+#[derive(Debug, Clone, PartialEq)]
+pub enum PasswordError {
+    Base64Decoding,
+    InvalidFormat,
+    InvalidKeyLength,
+    InvalidLength,
+    InvalidSaltLength,
+    NoMatchingAlgorithm(String),
```
Firstyear, line 410: I think didn't I mention we had a way to report this from the higher levels anyway, so we don't need the `(String)` here?

Firstyear (changes requested):
one minor change, otherwise looks awesome.

libs/crypto/src/lib.rs
yaleman: Different PR but why not have the feed back and save having to parse it in a different part of the code?

Firstyear: We already have the handler in the password import code though, it's just a string so all we have to return is a few characters. It also isn't a complete option, because on 819 you have `_ => Err(PasswordError::NoMatchingDecoder)` , which the current code *does* handle from the higher level. 

So if you really want to have the string here, it should have the same logic from https://github.com/kanidm/kanidm/blob/master/server/lib/src/plugins/cred_import.rs#L65 so that anytime we would have noMatchingDecoder, we would actually have `NoMatchingAlgorithm()` with the hint about what the algorithm is.

yaleman: So if it's not `{}`-wrapped (which the code handles) we should just return the first 4-5 chars of the input regardless?

yaleman: The intention with the difference was "we have no idea what this is" compared to "we don't support that algo currently" - I updated the names to `UnsupportedAlgorithm(String)` and `NoDecoderFound` to make it a little less confusing.

yaleman: I'm down with pink sheds as long as we know it's pink 😄

libs/crypto/src/lib.rs
```diff
@@ -838,10 +838,12 @@ impl TryFrom<&str> for Password {
                         material: Kdf::SSHA512(s.to_vec(), h.to_vec()),
                     })
                 }
-                _ => Err(PasswordError::NoDecoderFound),
+                _ => Err(PasswordError::NoDecoderFound(hash_format)),
             }
         } else {
-            Err(PasswordError::NoDecoderFound)
+            Err(PasswordError::NoDecoderFound(
+                value.chars().take(5).collect(),
```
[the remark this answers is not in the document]
yaleman: Yeah, take(n) was something I stumbled over due to clippy and it's great 🥳

PR #3638, opened 2025-05-14
fix: Improve unixd & unixd-tasks startup coupling

jinnatar:
# Change summary

Due to a complex interplay of the hardening present in both services, the unixd-tasks service fails to start on a systemd technicality if  unixd hasn't started first and created the /run/kanidm-unixd dir.

While previous methods probably should've achieved the correct result, practical testing time and time again proved that wasn't the case all the time and we had a race condition.

This change disallows unixd-tasks from starting if the requisite socket doesn't exist that unixd creates. In turn unixd ensures via an Upholds directive unixd-tasks is always running which provides a very  quick & controlled retry logic.

Related to and mitigates #3620

I'm leaving this as a draft because I still need to remove auto startup of the unixd-tasks service to avoid useless journal errors. And, I haven't tested this on SUSE, so that might be a good idea! I'll definitely have a go at it later.

Checklist

### DIFF
--- a/platform/opensuse/kanidm-unixd-tasks.service
+++ b/platform/opensuse/kanidm-unixd-tasks.service
@@ -4,7 +4,9 @@
 [Unit]
 Description=Kanidm Local Tasks
 After=chronyd.service ntpd.service network-online.target suspend.target kanidm-unixd.service
-Wants=kanidm-unixd.service
+# This prevents starting unixd-tasks before unixd is running and
+# has created the socket necessary for communication.
+ConditionPathExists=/var/run/kanidm-unixd/task_sock
 
 [Service]
 User=root
--- a/platform/opensuse/kanidm-unixd-tasks.service
+++ b/platform/opensuse/kanidm-unixd-tasks.service
@@ -4,9 +4,12 @@
 [Unit]
 Description=Kanidm Local Tasks
 After=chronyd.service ntpd.service network-online.target suspend.target kanidm-unixd.service
+
 # This prevents starting unixd-tasks before unixd is running and
 # has created the socket necessary for communication.
-ConditionPathExists=/var/run/kanidm-unixd/task_sock
+# We need the check so that fs namespacing used by`ReadWritePaths` has a 
+# strict enough target to namespace. Without the check it fails in a more confusing way.
+ConditionPathExists=/run/kanidm-unixd/task_sock
 
 [Service]
 User=root
@@ -16,7 +19,7 @@ ExecStart=/usr/sbin/kanidm_unixd_tasks
 CapabilityBoundingSet=CAP_CHOWN CAP_FOWNER CAP_DAC_OVERRIDE CAP_DAC_READ_SEARCH
 # SystemCallFilter=@aio @basic-io @chown @file-system @io-event @network-io @sync
 ProtectSystem=strict
-ReadWritePaths=/home /var/run/kanidm-unixd
+ReadWritePaths=/home /run/kanidm-unixd
 RestrictAddressFamilies=AF_UNIX
 NoNewPrivileges=true
 PrivateTmp=true
@@ -29,7 +32,3 @@ ProtectKernelModules=true
 ProtectKernelLogs=true
 ProtectControlGroups=true
 MemoryDenyWriteExecute=true
-
-[Install]
-WantedBy=multi-user.target
-
--- a/platform/opensuse/kanidm-unixd.service
+++ b/platform/opensuse/kanidm-unixd.service
@@ -9,8 +9,9 @@ Wants=nss-user-lookup.target
 # While it seems confusing, we need to be after nscd.service so that the
 # Conflicts will trigger and then automatically stop it.
 Conflicts=nscd.service
-# Upholding ensures that unixd-tasks is not only started but kept running.
-# This allows for a repeatable & fast way of starting unixd-tasks at the right time.
+# `Upholds` like a `Wants` directive ensures that unixd-tasks is started but also 
+# ensures it's kept running. This allows for a repeatable & fast way of starting 
+# unixd-tasks at the right time.
 Upholds=kanidm-unixd-tasks.service
 
 [Service]
--- a/platform/opensuse/kanidm-unixd.service
+++ b/platform/opensuse/kanidm-unixd.service
@@ -1,4 +1,4 @@
-# You should not need to edit this file. Instead, use a drop-in file:
+# You should not need to edit this file. Instead, use a drop-in file by running:
 #   systemctl edit kanidm-unixd.service
 
 [Unit]
@@ -9,6 +9,9 @@ Wants=nss-user-lookup.target
 # While it seems confusing, we need to be after nscd.service so that the
 # Conflicts will trigger and then automatically stop it.
 Conflicts=nscd.service
+# Upholding ensures that unixd-tasks is not only started but kept running.
+# This allows for a repeatable & fast way of starting unixd-tasks at the right time.
+Upholds=kanidm-unixd-tasks.service
 
 [Service]
 DynamicUser=yes
@@ -18,12 +21,17 @@ CacheDirectory=kanidm-unixd
 RuntimeDirectory=kanidm-unixd
 StateDirectory=kanidm-unixd
 
+
 Type=notify
 ExecStart=/usr/sbin/kanidm_unixd
 
 ## If you wish to setup an external HSM pin you should set:
 # LoadCredential=hsmpin:/etc/kanidm/kanidm-unixd-hsm-pin
 # Environment=KANIDM_HSM_PIN_PATH=%d/hsmpin
+
+# auth going down is bad, but infinite speedlooping is worse
+Restart=always
+RestartSec=30
 
 # Implied by dynamic user.
 # ProtectHome=
@@ -37,7 +45,6 @@ PrivateTmp=true
 PrivateDevices=false
 # Older versions of systemd require this to be explicitly allowed.
 DeviceAllow=/dev/tpmrm0 rw
-
 ProtectHostname=true
 ProtectClock=true
 ProtectKernelTunables=true
--- a/server/daemon/debian/kanidmd.service
+++ b/server/daemon/debian/kanidmd.service
@@ -23,6 +23,11 @@ ExecStart=/usr/bin/kanidmd server
 AmbientCapabilities=CAP_NET_BIND_SERVICE
 CapabilityBoundingSet=CAP_NET_BIND_SERVICE
 
+# If OOM occurs, request a clean stop
+OOMPolicy=stop
+# Adjust our weight toward *not* being killed under pressure.
+OOMScoreAdjust=-100
+
 NoNewPrivileges=true
 PrivateTmp=true
 PrivateDevices=true
--- a/unix_integration/resolver/debian/kanidm-unixd-tasks.service
+++ b/unix_integration/resolver/debian/kanidm-unixd-tasks.service
@@ -3,18 +3,15 @@
 
 [Unit]
 Description=Kanidm Local Tasks
-After=chronyd.service ntpd.service network-online.target kanidm-unixd.service
-Wants=kanidm-unixd.service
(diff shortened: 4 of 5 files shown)


platform/opensuse/kanidm-unixd-tasks.service
```diff
@@ -3,8 +3,10 @@
 
 [Unit]
 Description=Kanidm Local Tasks
-After=chronyd.service ntpd.service network-online.target suspend.target kanidm-unixd.service
-Wants=kanidm-unixd.service
```
yaleman, line 7: I'd say leave the `Wants` in here, just to make sure it *tries* to start `kanidm-unixd.service` if it's not running.

> Units listed in this option will be started if the configuring unit is. However, if the listed units fail to start or cannot be added to the transaction, this has no impact on the validity of the transaction as a whole, and this unit will still be started. This is the recommended way to hook the start-up of one unit to the start-up of another unit.

jinnatar:
Dug into systemd.socket enough to see that yes, it would probably be the most "right" solution. (This was the most helpful thing to read: https://0pointer.de/blog/projects/socket-activation.html). While "on-demand activation" instinctively feels like a feature we don't need, it's exactly that which allows quick pre-creation of sockets that are handed over to services when they're ready to go. Everything starts in parallel and flushes the buffer when ready to perform work.  It would require patching on one side to be able to receive the pre-initialized socket from systemd instead of creating one, but that doesn't seem to be a huge modification for anyone somewhat versed in socketry. (I think we'll need @Firstyear here as the socketeer.)

However, the remaining hesitation I have is around how to architecture it if we want to go down this path, and whether it actually has positive ROI in terms of safety and/or stability. 

Examples of course don't exactly cover this kind of scenario 100%, but fitting it into those models it would feel like unixd-tasks should own the socket (and be the one that's patched) since it's the executing side of the equation that provides a service. unixd would be just a consumer that connects to the socket and queues up work that unixd-tasks will get to once it's ready to go and picks up the socket from systemd.
But, that's not at all how it's architected right now. unixd owns and creates the socket, which is why we're having this startup race issue in the first place, we can't give unixd-tasks perms to the socket unless it already exists when the unit starts. So, is it a conscious choice that unixd owns the socket? Or perhaps just a historical artefact?
- The one pro of the current model I can think of is to ease permissions management, i.e. if the `DynamicUser` lower priv side owns the socket, we're done with permissions, and it's easy to give the higher priv side access via `ReadWritePaths`. Done the other way around, we'd need to create a permanent static group which the dyn user would belong to to grant it access to the socket created before the dyn user exists. We can't use `ReadWritePaths` because: `Paths listed in ReadWritePaths= are accessible from within the namespace with the same access modes as from outside of it.` I had to do this for the "kanidmd on systemd" implementation just to have a static guid to use for group perms.
- The con that I'm not sure of yet, is that we might suffer from further race conditions, or rather might mitigate them better the other way around. It's not that we need unixd to be ready before the socket is flushed, but the other way around, unixd-tasks needs to do things before it can start consuming. I may just be overthinking this, but the current race condition (https://github.com/kanidm/kanidm/issues/3620) is a somewhat similar scenario where unixd has pushed tasks into the socket, and unixd-tasks flushes it too early and then doesn't actually execute them. The tasks are essentially lost in transit. Would this scenario be more robust if the relationship was the other way around? I'm landing at: it's full-duplex, as long as the app logic doesn't drop what was read, it should work just the same regardless of which side connects first. We can't actually lowlevel queue up any tasks in the pre-init socket, but unixd already has it's own queue mechanism dependent on a handshake over the socket (which I don't understand well enough to say if having systemd in the mix will screw that over).
- The unknown factor here is, does the move over to the sparkled architecture change these parameters? Or perhaps just reinforces that the model we have now is right?

Assuming we keep the ownership as it stands now, then what I think we'd need:
- unixd is patched to accept pre-initialized sockets in addition to the option of creating them if sockets are not provided.
- We create a `kanidm-unixd.socket` unit. As I understand it this is the preferred name and it can carry both of `/run/kanidm-unixd/sock` & `/run/kanidm-unixd/task_sock` with a single unit:
   > These options may be specified more than once, in which case incoming traffic on any of the sockets will trigger service activation, and all listed sockets will be passed to the service, regardless of whether there is incoming traffic on them or not.
- We should be able to relax startup sequencing for `kanidm-unixd-tasks.service` and only depend on the tasks socket having started. Both services essentially start up in parallel.
- Work through the issues in #3620 to ensure unixd-tasks doesn't drop tasks that are received too early into the "handshake". I guess either by not doing the handshake until ready, or caching the tasks until ready to execute them.
- As a bonus, we could get lowlevel queue handling for the main unixd sock. Whether that makes things better or worse is a different question. Right now the PAM module will scream that the socket is missing, whereas this way work would queue up already, but just time out in PAM. (Assuming the module has a timeout? Seems to be a per module responsibility.) .. If it works, it might actually be a better user experience for the time between installing unixd that autoconfigures PAM, but config isn't in yet so unixd isn't actually running. BUT, then the socket activation would be trying to start unixd while it's still unconfigured. Basically, this would be an option but could be tried & tested separately after we solve for tasks.

Ooooor, we just do a "that all sounds complicated, let's just make sure they start the right way around so that the socket exists before unixd-tasks needs perms to it" .. Which the original change in this PR does. However, it doesn't in any way address any concerns with "is unixd-tasks actually ready to execute tasks before unixd sends them". But if we want socket architecture to help with that, the best option would feel like moving the socket responsibility over to the side that needs to get ready & using unixd.socket to ensure we can queue up work for it early even if the service takes a bit of time to get ready.

Firstyear:
> It would require patching on one side to be able to receive the pre-initialized socket from systemd instead of creating one, but that doesn't seem to be a huge modification for anyone somewhat versed in socketry. (I think we'll need @Firstyear here as the socketeer.)

Sorry, need to nack this one real quick.

tasks doesn't just make homedirs, it's now monitoring the filesystem for changes to inform the main resolver. So it can't be dynamically activated.

platform/opensuse/kanidm-unixd-tasks.service
Firstyear: Wants does try to start it iirc, but it just means that if unixd fails, tasks itself doesn't fail so it still keeps going.

platform/opensuse/kanidm-unixd-tasks.service
```diff
@@ -3,8 +3,10 @@
...
+After=chronyd.service ntpd.service network-online.target kanidm-unixd.service
+# This prevents starting unixd-tasks before unixd is running and
+# has created the socket necessary for communication.
+ConditionPathExists=/var/run/kanidm-unixd/task_sock
```
Firstyear, line 9: tasks polls for this to exist, so this isn't strictly needed.

platform/opensuse/kanidm-unixd.service
```diff
@@ -1,14 +1,17 @@
...
 [Unit]
 Description=Kanidm Local Client Resolver
-After=chronyd.service nscd.service ntpd.service network-online.target suspend.target
+After=chronyd.service nscd.service ntpd.service network-online.target
 Before=systemd-user-sessions.service sshd.service nss-user-lookup.target
 Wants=nss-user-lookup.target
 # While it seems confusing, we need to be after nscd.service so that the
 # Conflicts will trigger and then automatically stop it.
 Conflicts=nscd.service
+# Upholding ensures that unixd-tasks is not only started but kept running.
```
Firstyear, line 12: That's cool :)

Firstyear (dismissed):

Firstyear:
> > It would require patching on one side to be able to receive the pre-initialized socket from systemd instead of creating one, but that doesn't seem to be a huge modification for anyone somewhat versed in socketry. (I think we'll need @Firstyear here as the socketeer.)

Oh and the other thing is the direction isn't what you think it is. unixd hosts the socket, and tasks *connects to that*. Not the other way. systemd sockets would only work if it was the other way around.

Now, there is a *good reason* for this odd looking setup - security. 

Rather than having a daemon running as root that needs to validate every incomming connection for security, it's design to connect *out* to something that can only exist if a trusted entity made it. This is critically important with systemd dynamic users because ahead of time we *don't know* who unixd will be running as. So tasks doesn't know who to validate as allowed to send tasks. 

So instead, this way we have a way to ensure that only unixd can make the socket (since the path can only be made by root during the unixd startup process, and handed to the valid unixd runtime user.).

platform/opensuse/kanidm-unixd-tasks.service
jinnatar: Turns out it actually is. The unit is invalid if the path referred to by ReadWritePaths doesn't exist. That was the root cause of the strange namespace errors that delayed launch.

jinnatar:
> > It would require patching on one side to be able to receive the pre-initialized socket from systemd instead of creating one, but that doesn't seem to be a huge modification for anyone somewhat versed in socketry. (I think we'll need @Firstyear here as the socketeer.)
> 
> 
> 
> Sorry, need to nack this one real quick.
> 
> 
> 
> tasks doesn't just make homedirs, it's now monitoring the filesystem for changes to inform the main resolver. So it can't be dynamically activated.

That's one of the confusing bits for those who remember inetd & friends too well. Yes, it would start on demand, but we'd start it immediately as well. Even with on demand it doesn't need to shut down after demand is gone. It's just the same mechanism that enables parallel startup ignoring socket-only dependencies since "they'll catch up when needed". And given the other good reasons of keeping the socket on the unixd side, then technically it would be unixd that's on-demand, or rather opens the work queues before anyone is ready to process them.

platform/opensuse/kanidm-unixd-tasks.service
Firstyear: Oh neat. Okay, then I stand corrected :)

Firstyear: You may need to actually comment that fact then btw. That it influence rw paths.

jinnatar:
> > > It would require patching on one side to be able to receive the pre-initialized socket from systemd instead of creating one, but that doesn't seem to be a huge modification for anyone somewhat versed in socketry. (I think we'll need @Firstyear here as the socketeer.)
> 
> 
> 
> Oh and the other thing is the direction isn't what you think it is. unixd hosts the socket, and tasks *connects to that*. Not the other way. systemd sockets would only work if it was the other way around.
> 
> 
> 
> Now, there is a *good reason* for this odd looking setup - security. 
> 
> 
> 
> Rather than having a daemon running as root that needs to validate every incomming connection for security, it's design to connect *out* to something that can only exist if a trusted entity made it. This is critically important with systemd dynamic users because ahead of time we *don't know* who unixd will be running as. So tasks doesn't know who to validate as allowed to send tasks. 
> 
> 
> 
> So instead, this way we have a way to ensure that only unixd can make the socket (since the path can only be made by root during the unixd startup process, and handed to the valid unixd runtime user.). 

I think we're actually of the same mind here, hence why the actual propos for how this would work keeps the ownership as-is.

platform/opensuse/kanidm-unixd-tasks.service
tacerus, line 9: /var/run is more legacy and on modern systems just a symlink to /run, I suggest to reference /run directly instead.

jinnatar: True, need to check if OpenSUSE leap has made that change yet. They for example haven't merged /bin & /usr/bin yet.

tacerus: I am coming from (open)SUSE. :)

platform/opensuse/kanidm-unixd-tasks.service
jinnatar: Reading the docs `Upholds` is a superset of what `Wants` entails, i.e. it does the exact same startup chaining in addition to monitoring:
> Configures dependencies similar to Wants=, but as long as this unit is up, all units listed in Upholds= are started whenever found to be inactive or failed, and no job is queued for them.

I've improved the documenting comment a bit since someone would definitely wonder later why there isn't a `Wants`.

Firstyear (approved):

PR #3641, opened 2025-05-15
TODO trimming

yaleman:
# Change summary

- replaced string with Url in config parsing
- rename some inputs to client functions to match reality
- clean out old TODO notes

Fixes #

Checklist

### DIFF
--- a/server/core/src/https/apidocs/response_schema.rs
+++ b/server/core/src/https/apidocs/response_schema.rs
@@ -61,27 +61,27 @@ impl IntoResponses for ApiResponseWithout200 {
 }
 
 #[derive(Debug, Clone, ToSchema)]
-// TODO: this should be `webauthn_rs_proto::auth::PublicKeyCredential``, but ... I don't know how to make it possible in utoipa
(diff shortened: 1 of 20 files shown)


server/core/src/https/apidocs/response_schema.rs
```diff
@@ -61,27 +61,27 @@ impl IntoResponses for ApiResponseWithout200 {
 }
 
 #[derive(Debug, Clone, ToSchema)]
-// TODO: this should be `webauthn_rs_proto::auth::PublicKeyCredential``, but ... I don't know how to make it possible in utoipa
+// this should be `webauthn_rs_proto::auth::PublicKeyCredential``, but ... I don't know how to make it possible in utoipa
```
Firstyear, line 64: I don't think we should even worry, because webauthn is not an API that people should be mucking with in scripts. We need to be more careful about what apis *are* and *are not* for things to use.

Firstyear (dismissed):

Firstyear:
Thank you so much, this is awesome!

PR #3645, opened 2025-05-17
Investigate and reduce memory consumption of unixd

Firstyear:
# Change summary

It was noted by @tacerus that 1.6 unixd uses significantly more memory than previously. This integrates dhat to investigate why (and then fixes to come shortly).

Fixes #

Checklist

### DIFF
--- a/unix_integration/resolver/src/idprovider/kanidm.rs
+++ b/unix_integration/resolver/src/idprovider/kanidm.rs
@@ -89,6 +89,7 @@ impl KanidmProvider {
             })?;
 
         let crypto_policy = CryptoPolicy::time_target(Duration::from_millis(250));
+        // let crypto_policy = CryptoPolicy::minimum();
 
         let pam_allow_groups = config.pam_allowed_login_groups.iter().cloned().collect();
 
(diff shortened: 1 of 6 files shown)


Firstyear:
So far this has not shown any areas where there may be a significant leak or usage, so I plan to use this for a prolonged period in a test environment. For now I think it's in a reasonable state to merge at least.

unix_integration/resolver/src/idprovider/kanidm.rs
```diff
@@ -88,7 +88,8 @@ impl KanidmProvider {
                 IdpError::Tpm
             })?;
 
-        let crypto_policy = CryptoPolicy::time_target(Duration::from_millis(250));
+        // let crypto_policy = CryptoPolicy::time_target(Duration::from_millis(250));
```
yaleman, line 91: This causes a memory leak? 😱

yaleman (approved):

unix_integration/resolver/src/idprovider/kanidm.rs
Firstyear: Ahhh no, it causes a spike at startup and then it drops. It causes "bursts" of higher memory usage through out operation too during an auth. I probably should revert this bit actually ....

yaleman (approved):
Less RAMs moar better.

Firstyear:
@dmulder cc for himmelblau

PR #3651, opened 2025-05-20
Swap bytes mut at buffer limits

Firstyear:
# Change summary

- tokio-util (codec) and bytes mut do NOT make it clear, that bytes mut NEVER SHRINKS and will eventually grow to 2x the size of the largest input you have been sent. This causes a huge explosion of memory on long lived connections, and especially when you have some protocol messages that are large while the rest are small.

Mostly this affects the *receive* side, in unixd and the server. In this situation, once we decode a message, if the buffer exceeds some capacity limit, then we swap the buffer with a new one (because the bytes mut devs made it clear there is no way to free memory in a bytes mut without doing a swap to a new struct). 

Checklist


yaleman:
Awkward 😄

yaleman (approved):

Firstyear:
cc @dmulder for himmelblau

PR #3663, opened 2025-06-01
Add hmac 256

Firstyear:
Adds support for storing hmacs256 in the key storage system. Also converts the key storage to the new crypto-glue crate, rather than the existing openssl code. 

Fixes #

Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -13,9 +13,53 @@ dependencies = [
 
 [[package]]
 name = "adler2"
-version = "2.0.0"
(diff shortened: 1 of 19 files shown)


Firstyear:
Clippy got randomly sad here with this one, so I had to make some other cleanups too.

yaleman (dismissed):
Woo, less openssl more better.

There's some commented bout bit that could probably get cleaned up 😄

yaleman (approved):

PR #3683, opened 2025-06-15
Update Nextcloud example

nblock:
# Change summary
The Nextcloud app `user_oidc` has PKCE enabled by default since v6.1.0 (released 2024-10-15) so we can remove the PKCE configuration.

The Nextcloud setting `'allow_local_remote_servers' => true` is only required if the Kanidm server uses either a local network top-level domain (for example: .home, .local, .internal, …) or resolves to a local address. Describe when the setting is actually required.

Update wording and references to be a bit more Nextcloud specific.

Tested with Kanidm 1.6, Nextcloud 31, Nextcloud app user_oidc 7.2

---

Checklist

### DIFF
--- a/book/src/integrations/oauth2/examples.md
+++ b/book/src/integrations/oauth2/examples.md
@@ -519,17 +519,24 @@ OAUTH2_OIDC_DISCOVERY_ENDPOINT = "https://idm.example.com/oauth2/openid/<name>";
 
 ## Nextcloud
 
-Install the module [from the nextcloud market place](https://apps.nextcloud.com/apps/user_oidc) - it
+Install `user_oidc` [from the Nextcloud App store](https://apps.nextcloud.com/apps/user_oidc) - it
 can also be found in the Apps section of your deployment as "OpenID Connect user backend".
 
-In Nextcloud's config.php you need to allow connection to remote servers and enable PKCE:
+In the `Administration settings > OpenID Connect` settings menu of Nextcloud, configure the
+discovery URL and client ID and secret.
+
+If your Kanidm server is hosted on a local network top-level domain from RFC 6762 (for example:
+`.home`, `.local`, `.internal`, …) or resolves to a local address, you need to allow remote servers
+with local addresses in Nextcloud's `config.php`:
 
 ```php
 'allow_local_remote_servers' => true,
+```
 
-'user_oidc' => [
-    'use_pkce' => true,
-],
+If you forget this, you may see the following error in logs:
+
+```bash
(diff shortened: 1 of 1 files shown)


book/src/integrations/oauth2/examples.md
```diff
@@ -519,17 +519,23 @@ OAUTH2_OIDC_DISCOVERY_ENDPOINT = "https://idm.example.com/oauth2/openid/<name>";
 
 ## Nextcloud
 
-Install the module [from the nextcloud market place](https://apps.nextcloud.com/apps/user_oidc) - it
+Install `user_oidc` [from the Nextcloud App store](https://apps.nextcloud.com/apps/user_oidc) - it
 can also be found in the Apps section of your deployment as "OpenID Connect user backend".
 
-In Nextcloud's config.php you need to allow connection to remote servers and enable PKCE:
+In the `Administration settings > OpenID Connect` settings menu, configure the discovery URL and
```
yaleman: Just so it's clear it's happening in Nextcloud instead of Kanidm

```suggestion
In the `Administration settings > OpenID Connect` settings menu of Nextcloud, configure the discovery URL and
```

book/src/integrations/oauth2/examples.md
```diff
@@ -519,17 +519,23 @@ OAUTH2_OIDC_DISCOVERY_ENDPOINT = "https://idm.example.com/oauth2/openid/<name>";
...
+client ID and secret.
+
+If your Kanidm server uses a local domain or resolves to a local address, you need to allow remote
+servers with local addresses in Nextcloud's `config.php`:
```
yaleman: This  needs additional explanation, it's not clear what "uses a local domain" means - nobody should be using localhost as a domain name?

book/src/integrations/oauth2/examples.md
nblock: Applied and reformatted.

book/src/integrations/oauth2/examples.md
nblock: Mentioned RFC 6762 and some examples. Got both from Nextcloud's implementation:

* [Interface](https://github.com/nextcloud/server/blob/75612eca555c06e6f48d65515812552ab5b1e96d/lib/public/Security/IRemoteHostValidator.php#L33-L50)
* [Implementation](https://github.com/nextcloud/server/blob/ced1c5bf6c5f0543e41b68f74da1706e070b9be8/lib/private/Net/HostnameClassifier.php#L23-L54)

yaleman (approved):
Looks good, thanks!

PR #3687, opened 2025-06-19
20250618 rustls

Firstyear:
# Change summary

- Drop openssl as our TLS provider in favour of rustls.

Relates #3476

Checklist

### DIFF
--- a/server/lib/src/idm/oauth2.rs
+++ b/server/lib/src/idm/oauth2.rs
@@ -2872,9 +2872,7 @@ impl IdmServerProxyReadTransaction<'_> {
             OperationError::NoMatchingEntries
         })?;
 
-        // How do we return only the active signing algo types?
-
-        error!(sign_alg = ?o2rs.sign_alg);
+        trace!(sign_alg = ?o2rs.sign_alg);
 
         match o2rs.sign_alg {
             SignatureAlgo::Es256 => o2rs.key_object.jws_es256_jwks(),
--- a/server/lib/src/idm/serviceaccount.rs
+++ b/server/lib/src/idm/serviceaccount.rs
@@ -310,9 +310,12 @@ impl IdmServerProxyWriteTransaction<'_> {
         self.qs_write
             .impersonate_modify(
                 // Filter as executed
-                &filter!(f_eq(Attribute::Uuid, PartialValue::Uuid(gpe.target))),
+                &filter_all!(f_and!([
+                    f_eq(Attribute::Uuid, PartialValue::Uuid(gpe.target)),
+                    f_eq(Attribute::Class, EntryClass::ServiceAccount.into())
+                ])),
                 // Filter as intended (acp)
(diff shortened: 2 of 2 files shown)


Firstyear:
@yaleman The big question is about some of the CA tests, I don't think the CA is a public interface, so we don't need the key size checks. If anything I should be working to remove that CA and make it part of crypto-glue or something else.

yaleman:
> @yaleman The big question is about some of the CA tests, I don't think the CA is a public interface, so we don't need the key size checks. If anything I should be working to remove that CA and make it part of crypto-glue or something else.

I don't think it's wrong to ensure that Kanidm's handling of *any* library results in the minimum standards we expect, but the code that **_creates_** the keys can go elsewhere.

server/core/src/crypto.rs
```diff
@@ -148,76 +139,72 @@ pub fn setup_tls(
...
-                    }
-                }
+        let mut client_cert_crls = Vec::new();
+
+        for cert_dir_ent in dirents.iter().filter(|item| {
+            item.file_name()
+                .to_str()
+                // Hashed certs end in .0
+                // Hashed crls are .r0
+                .map(|fname| fname.ends_with(".r0"))
```
yaleman, line 170: That seems like a pretty fragile thing, is it something crypto-glue could paper over with a trait or something?

server/core/src/https/mod.rs
```diff
@@ -460,65 +458,53 @@ pub(crate) async fn handle_tls_conn(
...
+    let maybe_peer_cert = tls_stream
+        .get_ref()
+        .1
+        .peer_certificates()
+        // The first certificate relates to the peer.
+        .and_then(|peer_certs| peer_certs.first());
+
+    // Process the client cert (if any)
+    let client_cert = if let Some(peer_cert) = maybe_peer_cert {
+        // TODO: This is where we should be checking the CRL!!!
```
yaleman: Can you please make this an explicit GitHub issue referenced from this line, since that's a security-level TODO, and these things get embarrassing later 😄

server/core/src/crypto.rs
Firstyear: Not easily, I used this setup for openssl because it's how it handles ca folders, but we strictly don't need to in Kanidm in future. 

I don't know how a trait would help here, this is just a parser/convention here.

Firstyear:
> I don't think it's wrong to ensure that Kanidm's handling of _any_ library results in the minimum standards we expect, but the code that **_creates_** the keys can go elsewhere.

To be clear:

* I think checking key size on the TLS private keys is good, since we aren't the one who makes those keys. 
* If this was a full fledged CA that we exposed, yes we should check any private key/csr for compliance. 
* I think that our "self-signed CA for development" is not something we externally expose, and so we shouldn't worry about external key generation. If someone really swaps out that key and puts their own there just to create a CA that signs precisely one cert, they have bigger issues than we can deal with, and I have no idea what they are trying to achieve in the first place.

yaleman (approved):

PR #3717, opened 2025-07-06
Group name modification

ToxicMushroom:
# Change summary

Enables modification of the Name attribute for groups.
- Abstracted the reauth flow to use the Referer header (Idk if this is always present but I hope it is, no clue how else to do it in a clean way for dynamic pages such as paths with UUIDs in them)


server/core/src/https/views/admin/groups.rs
```diff
@@ -61,9 +70,16 @@ pub(crate) async fn view_group_view_get(
...
 ) -> axum::response::Result<Response> {
     let (group, scim_effective_access) =
-        get_group_info(uuid, state, &kopid, client_auth_info).await?;
+        get_group_info(uuid, state.clone(), &kopid, client_auth_info.clone()).await?;
+    let uat: &UserAuthToken = client_auth_info
+        .pre_validated_uat()
+        .map_err(|op_err| HtmxError::new(&kopid, op_err, domain_info.clone()))?;
+
+    let time = time::OffsetDateTime::now_utc() + time::Duration::new(60, 0);
+    let can_rw = uat.purpose_readwrite_active(time);
```
Firstyear, line 79: can_rw should also check that there is at least *one* writeable attribute in the scim_effective_access set.

Firstyear (changes requested):
Otherwise looks good,

server/core/src/https/views/admin/groups.rs
ToxicMushroom: hm, but what if we do not support that attribute in the webui ?
I assume you just want me to hide the save and unlock button if they have reauthed but can't actually edit anything

Firstyear: Exactly what I mean :) no point clicking save, on unlock, if you can't do anything.

ToxicMushroom: I added check_any for this https://github.com/kanidm/kanidm/pull/3717/files#diff-abd0e43d3bb132b22bea3b0871568b3674bb4c8cb6ea47519b29d5109265626eR61-R67 and used it in the  admin_group_view_partial template to resolve this, lmk if that's good or not

Firstyear: Yep, that's great!

Firstyear (approved):

PR #3721, opened 2025-07-08
20250705 entra id compat proposal

Firstyear:
This document discusses the pros and cons of adding entra-id compatibility to the Kanidm server. I'd like more feedback from others about it before we decide. 

Checklist

### DIFF
--- a/book/src/developers/designs/2025-07-entra-id.md
+++ b/book/src/developers/designs/2025-07-entra-id.md
@@ -54,7 +54,12 @@ from my understanding.
 By implementing Entra ID as a server, we may find security vulnerabilities in the process that we can disclose to
 Microsoft. This improves the security of authentication for all users of both Entra ID and our reimplementation.
 
-## Cons / Risks
+### Ability to Improve the End Result
+
+Kanidm as a project has high standards for quality, and this will be a positive influence on the potential success and
+security of the project as we can influence the direction in a way that prioritises quality.
+
+## Cons / Risks (To Kanidm)
 
 ### Does not address the underlying issues
 
@@ -67,23 +72,6 @@ in Windows is "strongly discouraged" and mostly undocumented. While this proposa
 
 What we want is Windows to authenticate to Kanidm on our terms. This proposal achieves that in a way that still leaves a
 lot of control with Microsoft, and we still can't make our own Windows client authentication modules.
-
-### Regulatory pressure may not eventuate.
-
-There is a not-insignificant risk that the regulatory pressure may not eventuate, meaning that Windows will not be made
-able to authenticate to 3rd party EntraID services. This would leave us with an interface that we would either need to
-maintain for future hope of that pressure to exist, or an interface that is not as good as a clean room implementation
-could have been.
-
-### Lack of Corporate Interest
-
-While it's possible that we could create an Entra ID compatible server, this may not mean that consumers want to use it.
-This is a similar situation that Samba 4 ADDC has fallen into. Many companies would rather stick with the Microsoft
-ecosystem for their clients and servers, and would not want to mix and match Windows clients with opensource servers.
-This has led to little to no interest (or money) from clients.
-
-If this combines with no ongoing support or maintenance, we would be left supporting a large feature without the
-resources to back or secure it.
 
 ### Elevated risk of Security Issues
 
@@ -103,6 +91,38 @@ development of Entra ID and it's opensource needs.
 As we are not in control of the specification, the user experience is limited to what Microsoft implements. This may
 limit us and what we can do for users. In some cases certain experiences would be impossible, such as the use of the
 Microsoft Authenticator app since that is completely controlled by Microsoft.
+
+## Cons / Risks (To the Project as a Whole)
+
+### Regulatory pressure may not eventuate.
+
+There is a not-insignificant risk that the regulatory pressure may not eventuate, meaning that Windows will not be made
+able to authenticate to 3rd party EntraID services. This would leave us with an interface that we would either need to
+maintain for future hope of that pressure to exist. In previous cases, regulatory pressure has taken years to manifest
+meaning that this is a project for "the long haul".
+
+Additionally, there are ways to deflect pressure, such as Microsoft pointing at ADDC and saying "well you can already,
+self host so we don't need another open spec here".
+
+> IMPORTANT: Microsoft _may_ tie EntraID login to
+> [licensing of Windows editions](https://www.microsoft.com/en-au/microsoft-365/enterprise/microsoft365-plans-and-pricing).
+> We do not want to participate in activity that would allow license violations of Windows, so we need to be careful in
+> this area. This may require changes to Windows in someway.
+
+#### Windows Licensing
+
+This may lead to users who want to use their Windows 11 CALs or Licenses via Entra. We probably don't want to go near
+this for obvious legal reasons. However users may expect this capability since it is what Entra provides today.
+
+### Lack of Corporate Interest
+
+While it's possible that we could create an Entra ID compatible server, this may not mean that consumers want to use it.
+This is a similar situation that Samba 4 ADDC has fallen into. Many companies would rather stick with the Microsoft
+ecosystem for their clients and servers, and would not want to mix and match Windows clients with opensource servers.
+This has led to little to no interest (or money) from clients.
+
+If this combines with no ongoing support or maintenance, we would be left supporting a large feature without the
+resources to back or secure it.
 
 ## Possible Paths Forward
 
@@ -125,3 +145,17 @@ Kanidm server is designed to be layered, and EntraID would not require access to
 gives EntraID compatability the most chance to succeed, because then we can adapt to the needs of EntraID and expose
 high quality interfaces for the project. However this also brings the most maintenance burden. Worst case, EntraID
 support could be removed provided we are disciplined in how the feature is added.
+
+### Make an SSP
+
+Make our own SSP, even though it's "black box". Microsoft are not in support of this and this option poses many
+challenges but it means that any other IDP could then use our SSP framework to integration authentication with Windows.
+
+### Design a Machine Protocol
+
+Rather that implement an SSP, we design a generic opensource Authentication protocol and specification. This way we can
+become an open server implementation of that protocol, and we can request microsoft to make an SSP that implements this.
+
+This way Microsoft still controls the SSP, but we have a protocol/specification that improves upon (and is not limited
+by) what Entra ID does. This however may be a steeper uphill battle, as Microsoft has little reason to implement this,
+and it will be harder to generate regulatory pressure.
--- a/book/src/developers/designs/2025-07-entra-id.md
+++ b/book/src/developers/designs/2025-07-entra-id.md
@@ -27,6 +27,8 @@ authentication options for users to EntraID or on prem ADDC.
 However, due to the lack of TLS CA pinning in Windows this means that an initial proof of concept could be made by MITM
 techniques to redirect clients to our implementation.
 
+Were Microsoft to begin to enforce TLS CA pinning in Windows, this would completely end all further efforts.
+
 > NOTE: Microsoft's released Linux Intune clients do perform TLS CA pinning however.
 
 ## Pros
@@ -34,7 +36,7 @@ techniques to redirect clients to our implementation.
 ### Improved collaboration between Himmeblau and Kanidm
 
 The Himmelblau project uses a lot of Kanidm source code. By having an Entra ID compatible server, we would be able to
-merge and collabore more to "shrink the gap" between our projects. This would also give Himmelblau improved access to
+merge and collaborate more to "shrink the gap" between our projects. This would also give Himmelblau improved access to
 client testing since they would be able to test against a provider.
 
 ### Kanidm gains client machine support
@@ -69,7 +71,7 @@ security of the project as we can influence the direction in a way that prioriti
 
 ### Does not address the underlying issues
 
-To me, this is the most important risk present.
+To me, this is the most significant risk.
 
 The underlying issue is that making
 [3rd party authentication modules](https://learn.microsoft.com/en-us/windows-server/security/windows-authentication/security-support-provider-interface-architecture)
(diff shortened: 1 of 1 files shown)


book/src/developers/designs/2025-07-entra-id.md
```diff
@@ -0,0 +1,161 @@
...
+
+The Himmelblau project uses a lot of Kanidm source code. By having an Entra ID compatible server, we would be able to
+merge and collabore more to "shrink the gap" between our projects. This would also give Himmelblau improved access to
+client testing since they would be able to test against a provider.
+
+### Kanidm gains client machine support
+
+Kanidm has wanted domain-joined machine support for a long time - this would allow Kanidm to use the existing
+capabilities of Himmeblau rather than inventing our own specification for this. Himmelblau also has some really cool
+native desktop integrations too!
```
yaleman, line 46: It'd be good to expand on what the desktop integrations are or provide for users here

book/src/developers/designs/2025-07-entra-id.md
```diff
@@ -0,0 +1,161 @@
...
+### Funding and Developers
+
+This would bring in funding and developers to the project, allowing us to improve a number of features that would be
+EntraID adjacent, especially around the WebUI components.
+
+### Ability to authenticate Windows to Kanidm
+
+This has been a long desired feature, and it would be excellent for Windows to be able to authenticate to Kanidm,
+especially in environments that want self-hosted IDM. This is especially of interest to home labs and small business
+from my understanding.
```
yaleman, line 50: Won't this only happen if someone writes an SSP, works out how to modify a windows install to use himmelblau, or redirect entra calls to a user-controlled server?

yaleman:
I think at the very most the proxy-layer approach would be the best middle ground; adding a bunch of API endpoints and user flows to Kanidm, for a system that may never be usable, seems like a terrible thing to have to maintain.

Firstyear:
Yeah. But at the same time, the proxy layer also is hard because then we miss out on a lot of integration options and polish. I'm going to try to talk to the people behind the proposal at sernet and ask some questions first too.

yaleman:
It'd be good to map out the requirements, for sure, because it's WAY too little detail to understand what's actually required so far.

book/src/developers/designs/2025-07-entra-id.md
```diff
@@ -0,0 +1,161 @@
+# Entra ID / Azure AD compatability proposal
+
+A proposal has been made by [SerNET](https://www.sernet.com/) that they wish to invest in the building of an Entra ID /
+Azure AD compatible IDM service. This may be initially supported be EU grant funding.
```
ToxicMushroom: be -> by ?

book/src/developers/designs/2025-07-entra-id.md
```diff
@@ -0,0 +1,161 @@
...
+
+In this document, we want to examine the pros, cons, and risks of this proposal.
+
+## Goal of the Proposal
+
+The goal would be to add compatibility for Kanidm to allow Himmelblau as an authentication client. If Himmelblau is able
+to function as a client to Kanidm, this demonstrates that an opensource Entra ID replacement is viable.
+
+With a viable opensource replacement available, this may allow regulatory pressure to be exerted on Microsoft to allow
+open Windows to use third party Entra ID compatible services.
```
ToxicMushroom: drop open or rephrase

book/src/developers/designs/2025-07-entra-id.md
```diff
@@ -0,0 +1,161 @@
...
+
+This is important as today Windows is coded to only support Microsofts EntraID services and uses TLS CA pinning to
+ensure that a 3rd party can't MITM this process. This limits Windows authentication options for users to EntraID or on
```
dmulder: Actually, I have not observed TLS CA pinning on Windows. It's actually possible to MITM requests on Windows. The area I encountered CA pinning was on their Linux client for Intune traffic only (I actually could MITM the Entra Id traffic). You **can** MITM Windows Intune also, unlike on Linux.
So this may not be as much of a problem as we think.

book/src/developers/designs/2025-07-entra-id.md
dmulder: Himmelblau provides a DBUS service (which is compatible with the MS proprietary Linux client Entra Id code) which allows us to integrate with this browser plugin: https://github.com/siemens/linux-entra-sso
This gives us SSO in Firefox and Chrome/Chromium.
The PRT implementation also gives as Kerberos integration (we get a TGT from Entra Id, and can store it in the local Linux ccache for use).

book/src/developers/designs/2025-07-entra-id.md
```diff
@@ -0,0 +1,161 @@
...
+### Ability to authenticate Windows to Kanidm
+
+This has been a long desired feature, and it would be excellent for Windows to be able to authenticate to Kanidm,
+especially in environments that want self-hosted IDM. This is especially of interest to home labs and small business
+from my understanding.
+
+### Potential to find and disclose security risks
+
+By implementing Entra ID as a server, we may find security vulnerabilities in the process that we can disclose to
+Microsoft. This improves the security of authentication for all users of both Entra ID and our reimplementation.
```
dmulder: I've already discovered several just implementing the client side. MS has been living in a walled garden, which has led them to ignore a number of critical edge cases.

book/src/developers/designs/2025-07-entra-id.md
```diff
@@ -0,0 +1,161 @@
...
+in Windows is "strongly discouraged" and mostly undocumented. While this proposal would allow Windows to authenticate to
+3rd parties, it doesn't resolve the need to be able to create new authentication modules for Windows client machines.
+
+What we want is Windows to authenticate to Kanidm on our terms. This proposal achieves that in a way that still leaves a
+lot of control with Microsoft, and we still can't make our own Windows client authentication modules.
+
+### Elevated risk of Security Issues
+
+Since we did not design the Entra ID protocols, they may have security issues we are not aware of. Additionally these
+protocols are complex and undocumented, and we may make a mistake in their implementation that leads to security issues.
```
dmulder: There is **some** documentation: https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-oapxbc/2f7d8875-0383-4058-956d-2fb216b44706
It's not very good though, and gets a lot of things wrong.

book/src/developers/designs/2025-07-entra-id.md
```diff
@@ -0,0 +1,161 @@
...
+
+### Not in control of specification
+
+A valuable part of Kanidm's success has been "nearly complete control" over all specifications we implement, meaning we
+can implement them in a minimal and secure manner. By having to implement someone else's specification, we must follow
+what Microsoft do to the letter. This could lead to security issues (as above), but also Microsoft can change the
+specification and server side behaviour at anytime. We would be playing "forever catch up", with no input to the
+development of Entra ID and it's opensource needs.
```
dmulder: IMHO, we should implement all the protocols in a better way (using more secure algorithms, etc), while maintaining backwards compatibility with the potential less-secure Windows hosts.

book/src/developers/designs/2025-07-entra-id.md
```diff
@@ -0,0 +1,161 @@
...
+
+If this combines with no ongoing support or maintenance, we would be left supporting a large feature without the
+resources to back or secure it.
+
+## Possible Paths Forward
+
+### Do Not Accept the Proposal
+
+This presents the least risk to Kanidm. However, it would mean that sernet may follow other avenues, and if they are
+successful, leaves Kanidm without Windows client support.
```
dmulder: The other avenue would likely be a Kanidm fork, FYI.

book/src/developers/designs/2025-07-entra-id.md
dmulder: The DBUS session service has other uses also. I've got users who are implementing custom SSO bits using the tokens they get from it.

book/src/developers/designs/2025-07-entra-id.md
dmulder: So theoretically, I think we could mask the odc endpoint to point to our own, afterward the Windows host would talk to our service instead.

Firstyear: MS could just enable TLS CA pinning, but initially this is good to know so we can implement and check things.

dmulder:
> It'd be good to map out the requirements, for sure, because it's WAY too little detail to understand what's actually required so far.

I'm traveling to Göttingen in Aug, and that's part of the purpose of my visit, to map out the requirements. We're starting these discussions now though to get a head start.

book/src/developers/designs/2025-07-entra-id.md
```diff
@@ -0,0 +1,172 @@
...
+
+Himmelblau and Kanidm already have many connections and projects and collaborators. As himmelblau is already a working
+client to Entra ID, and Kanidm is a full fledged IDM, this existing collaboration and relationship makes it a natural
+fit that we could work together on this proposal.
+
+In this document, we want to examine the pros, cons, and risks of this proposal.
+
+## Goal of the Proposal
+
+The goal would be to add compatibility for Kanidm to allow Himmelblau as an authentication client. If Himmelblau is able
```
twoolie, line 17: To be clear, the goal is to implement EntraID so that Himmelblau can use Kanidm without modifying Himmelblau?

book/src/developers/designs/2025-07-entra-id.md
```diff
@@ -0,0 +1,172 @@
...
+
+> NOTE: Microsoft's released Linux Intune clients do perform TLS CA pinning however.
+
+## Pros
+
+### Improved collaboration between Himmeblau and Kanidm
+
+The Himmelblau project uses a lot of Kanidm source code. By having an Entra ID compatible server, we would be able to
+merge and collabore more to "shrink the gap" between our projects. This would also give Himmelblau improved access to
+client testing since they would be able to test against a provider.
```
twoolie, line 40: This would be especially helpful for me to write end-to-end NixOS tests for Himmelblau <3

book/src/developers/designs/2025-07-entra-id.md
Firstyear: Yes - and this in turn demonstrates that it is possible for *Windows* to authenticate against Kanidm as well.

book/src/developers/designs/2025-07-entra-id.md
```diff
@@ -0,0 +1,172 @@
...
+native desktop integrations too!
+
+### Funding and Developers
+
+This would bring in funding and developers to the project, allowing us to improve a number of features that would be
+EntraID adjacent, especially around the WebUI components.
+
+### Ability to authenticate Windows to Kanidm
+
+This has been a long desired feature, and it would be excellent for Windows to be able to authenticate to Kanidm,
```
twoolie, line 55: This would be a killer feature.

book/src/developers/designs/2025-07-entra-id.md
```diff
@@ -0,0 +1,172 @@
...
+### Potential to find and disclose security risks
+
+By implementing Entra ID as a server, we may find security vulnerabilities in the process that we can disclose to
+Microsoft. This improves the security of authentication for all users of both Entra ID and our reimplementation. The
+himmelblau project has already discovered and disclosed a number of these issues.
+
+### Ability to Improve the End Result
+
+Kanidm as a project has high standards for quality, and this will be a positive influence on the potential success and
+security of the project as we can influence the direction in a way that prioritises quality.
```
twoolie, line 68: "influence the direction" of what? Kanidm already has high standards, so do you mean to influence Himmelblau or Entra?

Firstyear: As in, if this project were developed externally to Kanidm, it may not end up as good.

book/src/developers/designs/2025-07-entra-id.md
```diff
@@ -0,0 +1,172 @@
...
+
+To me, this is the most important risk present.
+
+The underlying issue is that making
+[3rd party authentication modules](https://learn.microsoft.com/en-us/windows-server/security/windows-authentication/security-support-provider-interface-architecture)
+in Windows is "strongly discouraged" and mostly undocumented. While this proposal would allow Windows to authenticate to
+3rd parties, it doesn't resolve the need to be able to create new authentication modules for Windows client machines.
+
+What we want is Windows to authenticate to Kanidm on our terms. This proposal achieves that in a way that still leaves a
+lot of control with Microsoft, and we still can't make our own Windows client authentication modules.
```
twoolie: I don't think this section fully articulates what the risk is, or I just don't fully understand if from reading it.

Is the risk that MS retains control over the process, and can shut kanidm out with an update?
or that this project would be skirting some legal line by integrating with a proprietary/patented protocol?

book/src/developers/designs/2025-07-entra-id.md
```diff
@@ -0,0 +1,172 @@
...
+
+### Elevated risk of Security Issues
+
+Since we did not design the Entra ID protocols, they may have security issues we are not aware of. Additionally these
+protocols are complex and the
+[documentation is incomplete](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-oapxbc/2f7d8875-0383-4058-956d-2fb216b44706)
+where it does exist. This may lead us to make a mistake in our reimplementation that leads to security issues.
```
twoolie, line 92: This is a controllable risk with sufficient modelling and testing, and the feature can be opt-in such that users aren't unknowningly exposed to any extra risk.

book/src/developers/designs/2025-07-entra-id.md
```diff
@@ -0,0 +1,172 @@
...
+
+### Not in control of specification
+
+A valuable part of Kanidm's success has been "nearly complete control" over all specifications we implement, meaning we
+can implement them in a minimal and secure manner. By having to implement someone else's specification, we must follow
+what Microsoft do to the letter. This could lead to security issues (as above), but also Microsoft can change the
+specification and server side behaviour at anytime. We would be playing "forever catch up", with no input to the
+development of Entra ID and it's opensource needs.
```
twoolie: I don't think you need to follow their implementation to-the-letter, but just enough to allow authentication.
You can be more restrictive than MS if that allows better security.

Microsoft also has to consider old clients in the wild, and can't immediately change their protocol. They would have to transition clients to the new version of the protocol, which we should be able to detect and adapt to.

book/src/developers/designs/2025-07-entra-id.md
```diff
@@ -0,0 +1,172 @@
...
+
+A suggestion is that we can extend the protocols with more secure techniques for our own use cases in future, but we
+would still need to support the "minimum" viable that Windows does to maintain interop.
+
+### User Experience
+
+As we are not in control of the specification, the user experience is limited to what Microsoft implements. This may
+limit us and what we can do for users. In some cases certain experiences would be impossible, such as the use of the
+Microsoft Authenticator app since that is completely controlled by Microsoft.
```
twoolie: Anybody wishing to run their own IDM and log into non-MS accounts is not going to be upset they can't use MS Authenticator...

Firstyear:
@twoolie Thanks for this, I added some updates to the doc clarify some parts in response to some of your comments.

ToxicMushroom (commented):
some typos I think

PR #3725, opened 2025-07-09
Add user facing SCIM pagination / sorting

Firstyear:
This adds the user facing parts of SCIM pagination and sorting. Currently it's done in a very "simple" way, with sorting and truncating the result vector. This won't be the long term approach however, we need to add a proper set of listing and sorting indexes to the backend. But having some working tests is always good to show where we want to go with this.

Part of the problem is that sorting and pagination indexes are hard to do, so I need to think more about the best way to add them :)

# Change summary

-

Fixes #

Checklist

### DIFF
--- a/book/src/developers/designs/scim_pagination.md
+++ b/book/src/developers/designs/scim_pagination.md
@@ -1,0 +1,94 @@
+# SCIM Pagination
+
+SCIM Pagination is important for Kanidm to limit the size of responses to clients, and to allow
+effecient UI's to exist that show results over multiple pages for large datasets (such as users)
+
+The default pagination in the [SCIM rfc](https://www.rfc-editor.org/rfc/rfc7644#section-3.4.2.4)
+has many limitations in how we arrange the items, indexes and other details. The important limitation
+of this is related to access controls.
+
+[SCIM cursor based pagination](https://datatracker.ietf.org/doc/draft-ietf-scim-cursor-pagination/) is
+an alternative, but also has some issues that may prevent us being able to use it.
+
+To satisfy our requirements, we should implement virtual list views.
+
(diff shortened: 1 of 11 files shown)


book/src/developers/designs/scim_pagination.md
```diff
@@ -0,0 +1,94 @@
+# SCIM Pagination
+
```
Firstyear, line 2: This is still a bit of a WIP document, I haven't fully fleshed out the ideas.

proto/src/scim_v1/server.rs
```diff
@@ -31,9 +33,12 @@ pub struct ScimEntryKanidm {
 #[serde_as]
 #[skip_serializing_none]
 #[derive(Serialize, Clone, Debug, Default)]
+#[serde(deny_unknown_fields, rename_all = "camelCase")]
 pub struct ScimListResponse {
     pub schemas: Vec<String>,
     pub total_results: u64,
+    pub items_per_page: Option<NonZeroU64>,
+    pub start_index: Option<NonZeroU64>,
```
yaleman, line 41: the RFC would indicate this needs to be an i64 with an `if < 1 {start_index=1}` 😄

ToxicMushroom (approved):
Looks good :) 
I agree with the reasoning for virtual list views, I think it's also what users would expect.

proto/src/scim_v1/server.rs
Firstyear: IMO the RFC is silly to allow it, so I'm going to say that when we find a client that sends something bad we fix it then. Otherwise we do the stricter thing and only allow the non-zero-u64.

yaleman (approved):

yaleman (approved):

PR #3728, opened 2025-07-09
Basic interface to get and regenerate the RADIUS password

kokx:
# Change summary

- Adds a basic webinterface to view the current RADIUS password
- Also allows (re)generating the RADIUS password
- Only accessible after re-authentication for a readwrite token
- Now an icon is also included for RADIUS.

A screenshot on how this looks in the webinterface:

kanidm-radius

Fixes  #3263

Checklist

(The book chapter and design document seem irrelevant to me, since this is only a relatively simple enhancement of the webinterface. Please let me know if I got this wrong)


server/core/src/https/views/mod.rs
```diff
@@ -117,6 +119,7 @@ pub fn view_router() -> Router<ServerState> {
             "/reset/add_ssh_publickey",
             post(reset::view_add_ssh_publickey),
         )
+        .route("/radius", post(radius::view_radius_post))
```
ToxicMushroom: nit: I'd prefer this to be under /api/radius/regenerate or something to keep things consistent

server/core/src/https/views/radius.rs
```diff
@@ -0,0 +1,143 @@
...
+
+pub(crate) async fn view_radius_post(
+    State(state): State<ServerState>,
+    Extension(kopid): Extension<KOpId>,
+    VerifiedClientInformation(client_auth_info): VerifiedClientInformation,
+    DomainInfo(_domain_info): DomainInfo,
+) -> Result<RadiusPartialView, WebError> {
+    let uat: UserAuthToken = state
+        .qe_r_ref
+        .handle_whoami_uat(&client_auth_info, kopid.eventid)
```
ToxicMushroom: I think you can use the new `pre_validated_uat()` like you did above here too

server/core/src/https/views/radius.rs
```diff
@@ -0,0 +1,143 @@
...
+        .await);
+    }
+
+    let radius_password = state
+        .qe_r_ref
+        .handle_internalradiusread(client_auth_info.clone(), uat.spn.clone(), kopid.eventid)
+        .await
+        .map_err(|op_err| HtmxError::new(&kopid, op_err, domain_info.clone()))?;
+
+    if let Some(radius_password) = radius_password {
```
ToxicMushroom: Could solely construct the `RadiusPartialView` in the if-else branches to have a little less duplication

server/core/templates/user_settings_partial_base.html
```diff
@@ -15,13 +15,17 @@
...
-            (% call side_menu_item("Credentials", (Urls::UpdateCredentials),
-            ProfileMenuItems::Credentials, "shield-lock") %)
-            (% call side_menu_item("Enrol Device", (Urls::EnrolDevice),
-            ProfileMenuItems::EnrolDevice, "phone-flip") %)
+            ProfileMenuItems::UserProfile, "person") %) (% call
+            side_menu_item("Credentials", (Urls::UpdateCredentials),
+            ProfileMenuItems::Credentials, "shield-lock") %) (% call
+            side_menu_item("Enrol Device", (Urls::EnrolDevice),
+            ProfileMenuItems::EnrolDevice, "phone-flip") %) (% call
+            side_menu_item("RADIUS", (Urls::Radius), ProfileMenuItems::Radius,
```
ToxicMushroom, line 27: I'm also not sure about an icon, unless someone has a better idea maybe a [key](https://icons.getbootstrap.com/icons/key-fill/) ?

server/core/src/https/views/radius.rs
```diff
@@ -0,0 +1,143 @@
...
+    DomainInfo(_domain_info): DomainInfo,
+) -> Result<RadiusPartialView, WebError> {
+    let uat: UserAuthToken = state
+        .qe_r_ref
+        .handle_whoami_uat(&client_auth_info, kopid.eventid)
+        .await?;
+
+    state
+        .qe_w_ref
+        .handle_regenerateradius(
```
ToxicMushroom: This seems to return the new password (I could be wrong), would also removes the need for the if check below and the read call

server/core/templates/user_settings_partial_base.html
Firstyear: Wifi is more likely the use case for radius.

server/core/src/https/views/radius.rs
Firstyear: Yes, this call DOES return the new radius password.

Firstyear:
Please also add yourself to the CONTRIBUTORS.md if you want. And thank you for taking the time to contribute, it's really appreciated!

server/core/src/https/views/radius.rs
kokx: Did this, I'm not entirely sure if the result is what you're looking for, but I hope it fits well enough.

server/core/src/https/views/radius.rs
kokx: Made this change as well, but the result is a slightly weird-looking statement. If you have a suggestion on how to do this better, please let me know! (I'm not the most experienced Rust programmer)

kokx:
Thanks for the great welcome! I made some new changes based on your comments. Please let me know if you have more remarks!

server/core/src/https/views/radius.rs
ToxicMushroom: yes that looks good

server/core/src/https/views/radius.rs
ToxicMushroom: My personal style would've been to pass `Option<String>` to the UI instead of password_available + password.
And then handle the option using askama templating like is done here https://github.com/kanidm/kanidm/blob/6ee953e9a37e4fcf768dda69835fe15729d22f25/server/core/templates/user_settings/profile_changes_partial.html#L24C1-L26C16. 
By doing it like I linked, you also make sure you can't try to show the password if there is None (askama can't display Option<_>).

kokx: I like that solution much more, so I implemented it. Thanks!

ToxicMushroom (approved):
Thanks

Firstyear (dismissed):

Firstyear:
Thank you so much, this is great!

Firstyear (approved):

Firstyear (approved):

PR #3733, opened 2025-07-10
Add ppc64le support for docker images

Gelbpunkt:
# Change summary

This adds support for 64-bit little endian PowerPC in the Docker images. The `rlm_python` Dockerfile needs to explicitly install gcc and openssl-devel in order to build `cryptography` from source and the server needs cmake in order to compile `aws_lc_rs`. That aside, everything builds just fine (ignoring tools/orca, that doesn't build with the rust compiler used in the Dockerfile but unrelated to the architecture).

ppc64le is a very modern baseline and implies altivec support, so no special logic is required to detect or enable extra CPU features.

Answering the why?: I want to run kanidm on my ppc64le machine and since it's very little effort to support it, having official Docker images for ppc64le would be nice.

Checklist

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -9,3 +9,7 @@ rustflags = ["-C", "link-arg=-fuse-ld=lld", "-Ctarget-cpu=native", "-C", "force-
 [target.aarch64-unknown-linux-gnu]
 linker = "clang"
 rustflags = ["-C", "link-arg=-fuse-ld=lld", "-Ctarget-cpu=native", "-C", "force-frame-pointers=yes"]
+
+[target.powerpc64le-unknown-linux-gnu]
+linker = "clang"
+rustflags = ["-C", "link-arg=-fuse-ld=lld", "-Ctarget-cpu=native", "-C", "force-frame-pointers=yes"]
--- a/rlm_python/Dockerfile
+++ b/rlm_python/Dockerfile
@@ -55,11 +55,13 @@ RUN mkdir -p /pkg/pykanidm/
 COPY pykanidm/ /pkg/pykanidm/
 
 # install the package and its dependencies
-RUN python3 -m pip install \
+RUN zypper install -y gcc openssl-devel && \
+    python3 -m pip install \
     --break-system-packages \
     --no-cache-dir \
     --no-warn-script-location \
-    /pkg/pykanidm
+    /pkg/pykanidm && \
+    zypper remove -y gcc openssl-devel
 
 COPY rlm_python/radius_entrypoint.py /radius_entrypoint.py
 
--- a/server/Dockerfile
+++ b/server/Dockerfile
@@ -22,6 +22,7 @@ RUN \
         sccache \
         cargo \
         clang \
+        cmake \
         gawk \
         make \
         automake \


Firstyear:
I'm okay to add this, but with the disclaimer that we don't have access to any ppc64le machines so any issues that arise from that would need you to investigate.

yaleman:
As nice as this is for the ppc64le user base I don't see this being something we could support with automated build artefacts because we're *already* having issues with how long build phases take and it's relatively trivial to build it yourself...

I'm happy to merge the cargo/dockerfile configs though 😄

Firstyear:
Yep, I agree with @yaleman here about this.

Gelbpunkt:
> I'm okay to add this, but with the disclaimer that we don't have access to any ppc64le machines so any issues that arise from that would need you to investigate.

Yeah, that was my expectation anyways.

> As nice as this is for the ppc64le user base I don't see this being something we could support with automated build artefacts because we're _already_ having issues with how long build phases take and it's relatively trivial to build it yourself...
> 
> I'm happy to merge the cargo/dockerfile configs though 😄

In CI or locally? The builds in CI should be super easy to make faster by e.g. ~~actually caching sccache~~ (edit: seems to be set up, I overlooked that) simply throwing more runners at the problem to parallelize the builds. If it's local build speed, then just don't build the cross builds locally? :)

Firstyear:
> In CI or locally? The builds in CI should be super easy to make faster by e.g. ~actually caching sccache~ (edit: seems to be set up, I overlooked that) simply throwing more runners at the problem to parallelize the builds.

I'll let @yaleman speak to that one, I think from the CI logs it took 15minutes ... 

> If it's local build speed, then just don't build the cross builds locally? :)

Your change makes its a default target (Makefile IMAGE_ARCH ) which is what wed like to remove.

Gelbpunkt:
> Your change makes its a default target (Makefile IMAGE_ARCH ) which is what wed like to remove.

Oh, sure, I can remove that.

Where are the definitions for the DockerHub builds? I can't seem to find those, at least the GitHub workflows only push to ghcr and build exclusively x86_64.

Firstyear:
We build and push those from a remote builder normally, though we probably could find a way to automate that .... @yaleman we could probably do something with cloning gh docker images or something in future?

yaleman:
> simply throwing more runners at the problem to parallelize the builds.

That's not so simple for a project relying on the free resources from Github - last I checked they have a concurrency limit of 15 workers per organisation, and we regularly hit that so things get queued.

@Firstyear it just means having a login for the kanidm dockerhub account and adding another push target in the CI, which should be pretty easy to add for the "devel" builds. Tagged builds takes a little more squinting at things.

Gelbpunkt:
Okay, I reverted the changes outside of the Dockerfiles for now. Would it at least be possible to build for this target in the tagged releases in the future? That'd be very much appreciated. I can also hook you up with access to ppc64le machines once kanidm is deployed on there :)

Firstyear:
We'll need to work out how we can do that, as it could be slow for us as we don't have access to a ppc64le builder. We can experiment with it though, and let you know. Otherwise, it should be pretty easy for you to be able to build it yourself if needed (just clone the release branch, then`IMAGE_ARCH=linux/ppc64le make buildx`)

yaleman (approved):

PR #3734, opened 2025-07-11
Groups WebUI, modify description

ToxicMushroom:
# Change summary
- Adds the description attribute to the Groups webadmin page.
- This attribute can be modified if the person has access.

# Issues
- Firefox ignores value="" on the description input after page refresh. Resulting in an empty description even though the DOM shows it, submitting the form in this state also sends an empty description.
- Long descriptions make the input scroll which is not very clear. Descriptions can currently not be multi-line though. Other solutions would involve non-input workarounds.

Checklist

### DIFF
--- a/server/core/src/https/views/admin/groups.rs
+++ b/server/core/src/https/views/admin/groups.rs
@@ -27,7 +27,8 @@ use std::collections::BTreeMap;
 use std::str::FromStr;
 use uuid::Uuid;
 
-pub const GROUP_ATTRIBUTES: [Attribute; 2] = [Attribute::Uuid, Attribute::Name];
+pub const GROUP_ATTRIBUTES: [Attribute; 3] =
+    [Attribute::Uuid, Attribute::Name, Attribute::Description];
 
 #[derive(Template)]
 #[template(path = "admin/admin_panel_template.html")]
@@ -188,9 +189,9 @@ async fn get_groups_info(
 
 #[derive(Clone, Debug, Serialize, Deserialize)]
 pub(crate) struct SaveGroupForm {
-    uuid: Uuid,
     #[serde(rename = "name")]
     account_name: String,
+    description: Option<String>,
 }
 
 pub(crate) async fn edit_group(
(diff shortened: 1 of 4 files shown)


server/core/src/https/views/admin/groups.rs
```diff
@@ -207,14 +209,26 @@ pub(crate) async fn edit_group(
         Some(ScimValueKanidm::String(query.account_name)),
     );
 
+    let (group_info, _) =
+        get_group_info(group_uuid, state.clone(), &kopid, client_auth_info.clone()).await?;
+
+    // query.description can't be Some("") since axum deserializes "" to None.
+    // Also meaning that I can't check if someone wants to unset a field or couldn't set the field.
+    // Thus, I check if there's a difference below to make up for this.
```
yaleman, line 217: I'd say removing the description if it's set to an empty string seems sensible?

server/core/templates/admin/admin_group_view_partial.html
```diff
@@ -36,26 +36,30 @@
...
+        (% endif -%)
+    (% endif -%)
 
+    (% call string_attr("UUID", "uuid", group.uuid, can_modify_any_attr, Attribute::Uuid) -%)
+    (% call string_attr("Name", "name", group.name, can_modify_any_attr, Attribute::Name) -%)
+    (% if let Some(description) = group.description -%)
+        (% call string_attr("Description", "description", description, can_modify_any_attr, Attribute::Description) -%)
+    (% else -%)
+        (% call string_attr("Description", "description", "", can_modify_any_attr, Attribute::Description) -%)
+    (% endif -%)
```
yaleman: This probably should just be `description.unwrap_or("")`

yaleman (changes requested):

server/core/src/https/views/admin/groups.rs
ToxicMushroom: yes that is the goal, the issue related to access is explained in the comment

server/core/templates/admin/admin_group_view_partial.html
ToxicMushroom: oh, yes that's better :)

ToxicMushroom:
hm why doesn't it auto-merge ?

yaleman:
Because it needed a rebase

PR #3735, opened 2025-07-12
Add a new paragraph in the installation quickstart for installing required client tools, and clarify the client tool setup paragraph

### LINKED ISSUE #3730: Make a more direct mention of the client during the installation guide

Currently the Kanidm Book starts with a qick intro (no client tool mentioned). Then it goes to the evaluation quickstart install guide for the server.

In the evaluation qickstart chapter, the first mention of a client tools is the "Setup the client configuration" then it goes to login and account setup.

This is chapter 2. The client tools are chapter 6. It is entirely out of the blue to start setting up a client without the mention of the required client tools beforehand. The info provided "This happens on your computer, not in the container." is an important warning, but does not clarify what is happening.

Before setting up the client configuration, there should be another paragraph, like "Install client tools" that communicates the information that:

-Kanidm is administered by a locally installed client tool
-The purpose and installation of the client tool is detailed in the linked chapter
-You cannot proceed without installing a client tool

Then first time readers unfamiliar with how kanidm works (so the target audience of the evaluation quickstart chapter) would understand where a client tool comes from and go get one.

Otherwise it's quite confusing to start configuring a client that's not yet present. And they might also start looking for a nonexistent webui (there's only one for login and managing your own credentials as an user, if I undertand correctly).

If I get it correctly (after all, I am one of those first time readers evaluating kanidm right now), then this is the only supported form of administration, and fully required.

Even a link to installing the client tools without explanation would be an improvement.

### Additional context

I can attempt a pr for this, if my suggestion is welcome. Hopefully I wont mess it up, as I haven't done so before. I'm not near my pc now, so it might take a few days.

Edit: This is possibly related to \#2637 but it doesn't say what the new user issue was there.

Karthogan:
# Change summary

**User friendliness changes:**

- Added an entire new **Installing client tools** paragraph, between recovering the built in administration accounts and setting up the client tools, which directs users to install the client tools first. This sends them to the installing client tools chapter, to help selecting the correct package and install steps, and makes no attempt to repeat what's already told there.
- The new **Installing client tools** paragraph opens with a brief explanation of why client tools are required. New users might expect to go to the webui (which is already running at this point) and start administration there (which is not possible). They could even login with the previously recovered built in admins, and be able to do nothing with them.
- Moved the _"This happens on your computer, not in the container."_ disclaimer to the new installation paragraph, as it is already relevant there.
- Instead of the moved disclaimer, added a line about a simple config file being needed to function to the **Setup the client tool configuration** paragraph. If I understand correctly, this is not strictly true. Every config file setting used here can be also passed as an argument in the client tools. But that's a niche advanced usecase, and it would be confusing to start explaining at this point. The guide already included a simple config for new users, I am just adding a line why.
- Added two comments in the client tool config, to clarify what the user is doing. The variable names are intuitive, this would only be a bit of extra clarity, because the user is not expected to have read the **Client Tools** chapter at this point. Might help if something goes wrong, see my next change for an example.

**Corrections:**

- The simple minimal config in the *Setup the client tool configuration** paragraph specified the URL of the server with port 8443. This is the internal docker port used in the example configs, which docker binds to port 443, the default https port on localhost, when following this same guide. Therefore no port should be specified, because a redundant port config that appears in only one place would be confusing. If this is not corrected, the server will refuse connection on port 8443 of the localhost.

Fixes #3730 

Checklist

### DIFF
--- a/book/src/evaluation_quickstart.md
+++ b/book/src/evaluation_quickstart.md
@@ -77,7 +77,7 @@ docker exec -i -t kanidmd \
 
 This happens on your computer, not in the container.
 
-Kanidm relies on locally installed client tools on the client machines for administration. Follow the guide in the [installing client tools](installing_client_tools.html) chapter to install them now.
+Kanidm requires locally installed client tools on the client machines for administration. Follow the guide in the [installing client tools](installing_client_tools.html) chapter to install them now.
 
 Client tools are explained more in the [client tools](client_tools.html) chapter, but to complete setup, you will only need to follow the steps below.
 
--- a/book/src/evaluation_quickstart.md
+++ b/book/src/evaluation_quickstart.md
@@ -83,7 +83,7 @@ Client tools are explained more in the [client tools](client_tools.html) chapter
 
 ## Setup the client tool configuration
 
-After installing the client tools, you will need a simple default config too use them.
+After installing the client tools, you will need to create a simple config file to continue.
 
 ```toml
 # ~/.config/kanidm
--- a/book/src/evaluation_quickstart.md
+++ b/book/src/evaluation_quickstart.md
@@ -89,7 +89,7 @@ After installing the client tools, you will need to create a simple config file 
 # ~/.config/kanidm
 
 uri = "https://localhost" # The URL of the server
-verify_ca = false #disables #Whether to verify the Certificate Authority details of the server’s TLS certificate
+verify_ca = false # disables TLS certificate verification as your are using a self-signed certificate
 ```
 
 ## Check you can login
--- a/book/src/evaluation_quickstart.md
+++ b/book/src/evaluation_quickstart.md
@@ -88,7 +88,7 @@ After installing the client tools, you will need to create a simple config file 
 ```toml
 # ~/.config/kanidm
 
-uri = "https://localhost:8443"
+uri = "https://localhost"
 verify_ca = false
 ```
 
--- a/book/src/evaluation_quickstart.md
+++ b/book/src/evaluation_quickstart.md
@@ -83,7 +83,7 @@ Client tools are explained more in the [client tools](client_tools.html) chapter
 
 ## Setup the client tool configuration
 
-After installing the client tools, you will need to create a simple config file to continue.
+After installing the `kanidm` tool, you will need to create a configuration file to continue.
 
 ```toml
 # ~/.config/kanidm
--- a/book/src/evaluation_quickstart.md
+++ b/book/src/evaluation_quickstart.md
@@ -88,8 +88,8 @@ After installing the client tools, you will need to create a simple config file 
 ```toml
 # ~/.config/kanidm
 
-uri = "https://localhost"
-verify_ca = false
+uri = "https://localhost" #The URL of the server
+verify_ca = false #disables #Whether to verify the Certificate Authority details of the server’s TLS certificate
 ```
 
 ## Check you can login
--- a/book/src/evaluation_quickstart.md
+++ b/book/src/evaluation_quickstart.md
@@ -77,7 +77,7 @@ docker exec -i -t kanidmd \
 
 This happens on your computer, not in the container.
 
-Kanidm relies on locally installed client tools on the client machine for administration. Follow the guide in the [installing client tools](installing_client_tools.html) chapter to install them now.
+Kanidm relies on locally installed client tools on the client machines for administration. Follow the guide in the [installing client tools](installing_client_tools.html) chapter to install them now.
 
 Client tools are explained more in the [client tools](client_tools.html) chapter, but to complete setup, you will only need to follow the steps below.
 
--- a/book/src/evaluation_quickstart.md
+++ b/book/src/evaluation_quickstart.md
@@ -77,7 +77,7 @@ docker exec -i -t kanidmd \
 
 This happens on your computer, not in the container.
 
-Kanidm requires locally installed client tools on the client machines for administration. Follow the guide in the [installing client tools](installing_client_tools.html) chapter to install them now.
+Kanidm requires locally installed client tools on the system used for administration via the command line. Follow the guide in the [installing client tools](installing_client_tools.html) chapter to install them before continuing.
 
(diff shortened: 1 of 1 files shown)


book/src/evaluation_quickstart.md
```diff
@@ -73,15 +73,23 @@ docker exec -i -t kanidmd \
...
+## Setup the client tool configuration
+
+After installing the client tools, you will need to create a simple config file to continue.
+
 ```toml
 # ~/.config/kanidm
 
-uri = "https://localhost:8443"
-verify_ca = false
+uri = "https://localhost" #The URL of the server
```
Karthogan: The original version specified the docker internal port, and resulted in connection refused. The new version has no specified port because docker is binding the internal 8443 port to 443 on localhost, which is already the https default.

book/src/evaluation_quickstart.md
```diff
@@ -73,15 +73,23 @@ docker exec -i -t kanidmd \
   kanidmd recover-account idm_admin
 ```
 
-## Setup the client configuration
+## Install the client tools
 
 This happens on your computer, not in the container.
```
Karthogan, line 78: This is actually a change, because this line was originally in the next paragraph. It just stayed on the same line when I added a new paragraph in front of the original one and moved this line up.

book/src/evaluation_quickstart.md
```diff
@@ -73,15 +73,23 @@ docker exec -i -t kanidmd \
...
 
+Kanidm requires locally installed client tools on the client machines for administration. Follow the guide in the [installing client tools](installing_client_tools.html) chapter to install them now.
+
+Client tools are explained more in the [client tools](client_tools.html) chapter, but to complete setup, you will only need to follow the steps below.
+
+## Setup the client tool configuration
+
+After installing the client tools, you will need to create a simple config file to continue.
```
yaleman:
```suggestion
After installing the `kanidm` tool, you will need to create a configuration file to continue.
```

book/src/evaluation_quickstart.md
```diff
@@ -73,15 +73,23 @@ docker exec -i -t kanidmd \
   kanidmd recover-account idm_admin
 ```
 
-## Setup the client configuration
+## Install the client tools
 
 This happens on your computer, not in the container.
 
+Kanidm requires locally installed client tools on the client machines for administration. Follow the guide in the [installing client tools](installing_client_tools.html) chapter to install them now.
```
yaleman:
```suggestion
Kanidm requires locally installed client tools on the system used for administration via the command line. Follow the guide in the [installing client tools](installing_client_tools.html) chapter to install them before continuing.
```

book/src/evaluation_quickstart.md
```diff
@@ -73,15 +73,23 @@ docker exec -i -t kanidmd \
...
+
+After installing the client tools, you will need to create a simple config file to continue.
+
 ```toml
 # ~/.config/kanidm
 
-uri = "https://localhost:8443"
-verify_ca = false
+uri = "https://localhost" #The URL of the server
+verify_ca = false #disables #Whether to verify the Certificate Authority details of the server’s TLS certificate
```
yaleman:
```suggestion
verify_ca = false # disables TLS certificate verification as your are using a self-signed certificate
```

book/src/evaluation_quickstart.md
yaleman:
```suggestion
uri = "https://localhost" # The URL of the server
```

yaleman (changes requested):
Minor tweaks, but it's a good start, thanks!

book/src/evaluation_quickstart.md
Karthogan: Thanks, this seems to be even better for clarity.

book/src/evaluation_quickstart.md
Karthogan: Right, the specific effect is better than my copypaste of the generic variable description

book/src/evaluation_quickstart.md
Karthogan: I am not sure about this. Calling it the kanidm tool seems better than just client tool(s). But I could only find a single instance of kanidm tools used in the documentation, in **Chapter 20.4. Architecture** of the design documents. Currently client tools seems to be the way these are called in every other place.

Of course I'm not the one making the call if its better to be more specific or more consistent. I just wanted to say that I used the generic client tools, because that seems to be the current term in practice. This seems like a branding question to me.

yaleman: My thinking was that there a load of different tools that folks could consider client, like `kanidm-unixd` etc, I just wanted to be specific about the one we're talking about.

Karthogan: I agree that client tools is not clear language, and the mistake you say could realistically happen. *Kanidm tool* or *kanidm client* would be a better specific naming, it's just not currently used.

The actual names are client tools in the text, kanidm client in the package names, and kanidm tools in the design documentation. If any unification is made, I think it would best be towards kanidm client, because package names are a pain to change compared to references in the book. Altough even there, the cargo crate is named different (kanidm_tools).

I think I should take this change here and now. If it is worth fussing about more, then I could also add a new issue to discuss if this would benefit from unified branding for better clarity. Do you think it is important enough on its own to discuss?

yaleman: The crate/package name's irrelevant in this context really, there's a reason I put it in backticks, so it's identified as the specific command line tool name.

Karthogan: Oh, I totally missed the backticks for single quotes. Right, that should have been obvious.

I mentioned package names just to say that out of the several ways I've seen referring at the `kanidm` tool, `kanidm` client felt the most clear to me, since the packages are already named that, and that's basically a fix point.

But really, this is such a minor thing, that I feel sorry for even bringing it up. I'll take it as not even worth bothering with.

And thank you for helping to put my first try at participating here in an acceptable shape!

yaleman: Not a problem at all, it's good to discuss 😄

Firstyear:
Thank you very much @Karthogan !

PR #3741, opened 2025-07-17
20250711 type migrations

Firstyear:
# Change summary

This started with me wanting to migrate types to be cleaner, but to do so I needed to query what schema we already had. Typical Firstyear, I ended up rabbit holing and implementing SCIM Filters for our get queries, and implementing some POC of that with schema attributes and classes.

Worth noting that the schema cli options are marked hidden as they certainly are NOT finished or stable yet. But I didn't want to keep playing katamari here. 

Checklist


Firstyear:
Also you'll notice the bulk of the PR is tests - to handle the scimfilter properly with our native types, I pretty much had to "copy paste" the parser expression grammar parser to make it work, and of course that means bringing in all the tests to assert it works as expected. I couldn't think of a way to make it work "nicely" otherwise. as we would need lots of converts and other hacks.

proto/src/scim_v1/client.rs
```diff
@@ -130,6 +130,60 @@ pub struct ScimListApplication {
...
+#[serde_as]
+#[derive(Deserialize, Debug, Clone)]
+#[serde(rename_all = "snake_case")]
+pub struct ScimEntrySchemaAttribute {
+    #[serde(flatten)]
+    pub header: ScimEntryHeader,
+
+    pub attributename: String,
+    pub description: String,
+    // To be removed
```
yaleman: after this PR?
```suggestion
    // TODO: To be removed
```

yaleman (dismissed):

yaleman (approved):

PR #3768, opened 2025-08-04
Improve argon2id parameter search speed

Doridian:
# Change summary

Improves argon2id parameter search speed.

On all my systems, kanidm-unixd's 250ms target maxes out the allowable parameters anyway.

As from the given documentation, multiplying a parameter with X should multiply time by X (roughly), just multiply with the factor right away, as opposed to slowly approaching parameters.

From local testing, this seems to produce exactly the same end result parameters, but improves unixd's startup speed significantly (currently it spends ~15 seconds just to find it has to max out all parameters).

Checklist

### DIFF
--- a/libs/crypto/src/lib.rs
+++ b/libs/crypto/src/lib.rs
@@ -257,19 +257,18 @@ impl CryptoPolicy {
             };
 
             if let Some(ubt) = Password::bench_argon2id(params) {
-                debug!("{}µs - t_cost {} m_cost {}", ubt.as_nanos(), t_cost, m_cost);
+                debug!("{}ns - t_cost {} m_cost {}", ubt.as_nanos(), t_cost, m_cost);
                 // Parameter adjustment
                 if ubt < target_time {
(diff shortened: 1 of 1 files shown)


Firstyear (changes requested):
Hey there, I like this change in theory, but the problem is adjusting T (thread count) also has an impact on the server - we want the server to have single-threads for argo2id to not interfere with the tokio workers. 

I think the best option here is to have your change, but either in a separate function or a max thread parameter we can pass in. That way unixd can use all the threads of the system, but the main server is capped to 1 thread per argon2id bench. Does that seem reasonable to you?

Doridian:
@Firstyear Not sure I understand exactly what you're asking for. My change should not cause any difference in picked parameters, simply select them quicker (by making higher jumps up if it seems reasonable based on requested time divided by current time)

Firstyear (approved):

Firstyear:
Actually you're right, I completely misread the code ! Sorry about that.

PR #3770, opened 2025-08-05
Forget username if user no longer wants to be remembered

lrnz-lmk:
# Change summary
When a user selects to remember the username for future logins, a cookie with the username is set. But when the user does not want to be remembered anymore, disabling the switch had no effect and the username would be prefilled for every future login until the cookie will expire. I think it would be better to remove the cookie once the user does not want to be remembered anymore, so disabling the switch actually has an effect.

Checklist


Firstyear:
Thank you! This is a good fix :)

PR #3781, opened 2025-08-09
Support Unix sockets and systemd socket activation

### LINKED ISSUE #2771: Allow bindaddress to be a unix socket

It would be great if bindaddress in the server config could be a unix socket path in addition to an inet socket. Right now it gives this error message:

> [error]: Failed to parse address ("unix:/var/run/kanidmd/web.sock") from config: AddrParseError(Socket)

I've also tried without the unix:, no idea if this is already possible and I just don't know the syntax.

Allow setting e.g. bindaddress="unix:/var/run/kanidmd/web.sock"

\

**Additional context**

I would like this mainly so I don't have to assign a unique port on localhost for an upstream for a reverse proxy.

networkException:
# Change summary

- `bindaddress` can now be set to a path, binding the web server to a unix domain socket
- A TCP or Unix socket provided using [`ListenStream=`](https://www.freedesktop.org/software/systemd/man/latest/systemd.socket.html#ListenStream=) will now be inherited by Kanidm, if the address matches with `bindaddress`. This allows tighter sandboxing

Fixes #2771

Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -340,7 +340,7 @@ dependencies = [
  "libc",
  "libudev",
  "log",
- "memoffset",
+ "memoffset 0.8.0",
  "openssl",
  "openssl-sys",
  "rand 0.8.5",
@@ -3742,6 +3742,7 @@ dependencies = [
  "futures",
  "futures-util",
  "haproxy-protocol",
+ "hashbrown 0.15.4",
  "hyper 1.6.0",
  "hyper-util",
  "kanidm_build_profiles",
@@ -3752,6 +3753,7 @@ dependencies = [
  "kanidmd_lib",
  "ldap3_proto",
  "libc",
+ "nix",
  "openssl",
  "opentelemetry",
  "qrcode",
@@ -4259,6 +4261,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "memoffset"
+version = "0.9.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "488016bfae457b036d996092f6cb448677611ce4449e970ceaf42695203f218a"
+dependencies = [
+ "autocfg",
+]
+
+[[package]]
 name = "mimalloc"
 version = "0.1.47"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -4360,6 +4371,7 @@ dependencies = [
  "cfg-if",
  "cfg_aliases",
  "libc",
+ "memoffset 0.9.1",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -298,6 +298,8 @@ utoipa = { version = "4.2.0", features = ["url", "uuid"] }
 utoipa-swagger-ui = "6.0.0"
 uuid = "^1.17.0"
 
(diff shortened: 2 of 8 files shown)


Firstyear:
Hi there,

We actually don't want to support systemd socket activation, because this breaks replication, authentication sessions and background tasks. This creates new ways that a person's instance can break, and we don't want to have to support those situations. Kanidmd is not designed to be turned on and off rapidly based on when something connects, it's designed to run constantly for a long time. 

I do appreciate the time you took to implement this, in future it's always good to open an issue and talk to us first about features and their requests so we can discuss and design them first. 

Sorry :(

networkException:
Hey,

I'm not sure if you remember but we talked about this briefly at the CCC Darmstadt meetup.

I think you brought up similar concerns then as well. In essence, while systemd [supports spawning a service instance on every connection using Accept=yes](https://www.freedesktop.org/software/systemd/man/latest/systemd.socket.html#Accept=), the lifetime of `kanidm.service` here would be unchanged with the following default `kanidm.socket`

```ini
[Unit]

[Socket]
ListenStream=/path/to/unix.socket

[Install]
WantedBy=sockets.target
```

When using something like

```patch
 [Install]
 WantedBy=sockets.target
+RequiredBy=kanidm.service
```

both units would also start without waiting on an initial packet. In practice this shouldn't make a difference, systemd handles any necessary buffering here.

In summary, supporting socket activation mainly just allows Kanidm to accept an already open listen socket from the parent process, thus allowing it to drop any bind privileges or further file system access in case of Unix sockets.

I hope we can move this to Matrix if you have further concerns. From discussing this at the meetup you seemed happy to move forward with such a feature. At least I would appreciate if you could consider the first patch separately as well

Firstyear:
Hey there,

I do remember, and I mentioned that I would think about it. And when I thought about it I realised all the ways in which it would break things. An important aspect of opensource is our support surface area, and every single external dependency (such as systemd), every extra option, every different configuration adds burden to us as maintainers. 

So outside of the fact socket activation *will break things*, it makes our life as maintainers harder because now we have to diagnose when this breaks for our users. That's why I don't want to support this in Kanidm sorry :( 

The first patch we may consider, but please open a separate PR for it.

CobaltCause:
> So outside of the fact socket activation _will break things_

My understanding is that the concerns you raised around process lifecycle ("replication, authentication sessions and background tasks") don't *necessarily* apply when using socket activation, because socket activation can be used while spawning the process in a long-lived fashion as usual rather than on-demand per connection. I think a more accurate assessment could be "socket activation *can* break things", and I think the situations where it's a problem could be sufficiently mitigated by documenting that Kanidm expects to be a long-lived process and that should be taken into account when using socket activation.

Firstyear:
If you use socket activation to turn on the service and keep it around long term, what was the point of using socket activation? Just start it like normal.

The only reason people want socket activation is so that the service turns on and off on demand, and that pattern *will break things* and I will not allow us merging a feature that has the potential to break our users services. 

Documentation of the limitations also does not fly - documentation does not exist to work around bad design choices. We are a project that wants to make reliable software - not software where when you have a problem we just say "did you rtfm? it says there why it broke.". That is disrespectful to our users who rely on us to make good design choices. 

So I'm going to firmly reiterate that the answer to this feature is no. We will not accept socket activation in this software.

PR #3782, opened 2025-08-10
Support binding on Unix sockets

### LINKED ISSUE #2771: Allow bindaddress to be a unix socket

networkException:
# Change summary

- `bindaddress` can now be set to a path, binding the web server to a unix domain socket

Fixes #2771

Checklist

### DIFF
--- a/server/core/src/config.rs
+++ b/server/core/src/config.rs
@@ -1112,8 +1112,13 @@ impl ConfigurationBuilder {
                 client_ca,
             }),
             _ => {
-                eprintln!("ERROR: Tls Private Key and Certificate Chain are required.");
-                return None;
+                match bindaddress {
+                    Some(ref bindaddress) if bindaddress.starts_with("/") => None,
+                    _ => {
+                        eprintln!("ERROR: Tls Private Key and Certificate Chain are required.");
+                        return None;
+                    }
+                }
             }
         };
 
--- a/server/core/src/https/extractors/mod.rs
+++ b/server/core/src/https/extractors/mod.rs
@@ -10,6 +10,7 @@ use kanidm_proto::internal::COOKIE_BEARER_TOKEN;
 use kanidmd_lib::prelude::{ClientAuthInfo, ClientCertInfo, Source};
 use std::net::{IpAddr, SocketAddr};
 use std::str::FromStr;
+use std::fmt;
 
 // Re-export
 pub use kanidmd_lib::idm::server::DomainInfoRead;
@@ -109,11 +110,26 @@ impl FromRequestParts<ServerState> for DomainInfo {
 }
 
 #[derive(Debug, Clone)]
+pub enum ConnectionAddress {
+    Unix,
+    Tcp(SocketAddr)
+}
+
+impl fmt::Display for ConnectionAddress {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            ConnectionAddress::Unix => write!(f, "unix://"),
+            ConnectionAddress::Tcp(socket_address) => write!(f, "tcp://{socket_address}"),
+        }
+    }
+}
+
+#[derive(Debug, Clone)]
 pub struct ClientConnInfo {
     /// This is the address that is *connected* to Kanidm right now
     /// for this operation.
     #[allow(dead_code)]
-    pub connection_addr: SocketAddr,
+    pub connection_addr: ConnectionAddress,
     /// This is the client address as reported by a remote IP source
     /// such as x-forward-for or the PROXY protocol header
     pub client_ip_addr: IpAddr,
@@ -132,8 +148,8 @@ impl Connected<ClientConnInfo> for ClientConnInfo {
 impl Connected<SocketAddr> for ClientConnInfo {
     fn connect_info(connection_addr: SocketAddr) -> Self {
         ClientConnInfo {
+            connection_addr: ConnectionAddress::Tcp(connection_addr),
             client_ip_addr: connection_addr.ip(),
-            connection_addr,
             client_cert: None,
         }
     }
--- a/server/core/src/https/middleware/mod.rs
+++ b/server/core/src/https/middleware/mod.rs
@@ -1,5 +1,6 @@
 use crate::https::extractors::ClientConnInfo;
 use crate::https::ServerState;
+use super::extractors::ConnectionAddress;
 use axum::{
     body::Body,
     extract::{connect_info::ConnectInfo, State},
@@ -121,13 +122,14 @@ async fn ip_address_middleware_inner(
             )
         })?;
 
-    let connection_ip_addr = connection_addr.ip();
-
-    let trust_x_forward_for = state
-        .trust_x_forward_for_ips
-        .as_ref()
-        .map(|range| range.contains(&connection_ip_addr))
-        .unwrap_or_default();
+    let trust_x_forward_for = match connection_addr {
+        ConnectionAddress::Unix => true,
+        ConnectionAddress::Tcp(connection_addr) => state
+            .trust_x_forward_for_ips
+            .as_ref()
+            .map(|range| range.contains(&connection_addr.ip()))
+            .unwrap_or_default()
+    };
 
     let client_ip_addr = if trust_x_forward_for {
         if let Some(x_forward_for) = request.headers().get(X_FORWARDED_FOR_HEADER) {
--- a/server/core/src/https/mod.rs
+++ b/server/core/src/https/mod.rs
@@ -31,6 +31,7 @@ use axum::{
 use axum_extra::extract::cookie::CookieJar;
 use cidr::IpCidr;
 use compact_jwt::{error::JwtError, JwsCompact, JwsHs256Signer, JwsVerifier};
+use extractors::ConnectionAddress;
 use futures::pin_mut;
 use haproxy_protocol::{ProxyHdrV2, RemoteAddress};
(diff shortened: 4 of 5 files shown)


server/core/src/https/mod.rs
```diff
@@ -324,39 +327,55 @@ pub async fn create_https_server(
         // the connect_info bit here lets us pick up the remote address of the client
         .into_make_service_with_connect_info::<ClientConnInfo>();
 
-    let addr = SocketAddr::from_str(&config.address).map_err(|err| {
-        error!(
-            "Failed to parse address ({:?}) from config: {:?}",
-            config.address, err
-        );
-    })?;
+    if config.address.starts_with("/") {
```
Firstyear, line 330: I wonder if this function should take in a unix socket directly that way in tests we can pass in one half of an anonymous socket pair instead of relying on paths.

Firstyear (changes requested):
https://github.com/kanidm/kanidm/blob/master/server/testkit/tests/testkit/ip_addr_extractors.rs

We need tests at least in here that when we use unix sockets we get the correct addresses even with proxy headers. 

@yaleman I want you to have a look at this one too, I'm thinking there may be some ways people (ab)use this.

server/core/src/config.rs
```diff
@@ -1112,8 +1112,13 @@ impl ConfigurationBuilder {
                 client_ca,
             }),
             _ => {
-                eprintln!("ERROR: Tls Private Key and Certificate Chain are required.");
-                return None;
+                match bindaddress {
+                    Some(ref bindaddress) if bindaddress.starts_with("/") => None,
+                    _ => {
+                        eprintln!("ERROR: Tls Private Key and Certificate Chain are required.");
```
yaleman, line 1118: Minor nit while we're here. :smile:

```suggestion
                        eprintln!("ERROR: TLS Private Key and Certificate Chain are required.");
```

dblsaiko:
While we're at it, it would also be very useful to be able to set mode/owner on the socket. I usually set the mode to 660 so users in the group (i.e. nginx) can access the socket and I think the default mode doesn't allow that.

server/core/src/https/extractors/mod.rs
```diff
@@ -108,12 +109,27 @@ impl FromRequestParts<ServerState> for DomainInfo {
     }
 }
 
+#[derive(Debug, Clone)]
+pub enum ConnectionAddress {
+    Unix,
```
yaleman, line 114: This should have the socket path, similar to the `Tcp` variant.

Firstyear:
So the question I want to ask here is "what is the use case - what does this achieve"? We want to know why this is a feature that will help our users.

server/core/src/https/middleware/mod.rs
```diff
@@ -121,13 +122,14 @@ async fn ip_address_middleware_inner(
...
-        .map(|range| range.contains(&connection_ip_addr))
-        .unwrap_or_default();
+    let trust_x_forward_for = match connection_addr {
+        ConnectionAddress::Unix => true,
+        ConnectionAddress::Tcp(connection_addr) => state
+            .trust_x_forward_for_ips
+            .as_ref()
+            .map(|range| range.contains(&connection_addr.ip()))
+            .unwrap_or_default()
+    };
```
yaleman, line 132: Inherently trusting the forwarding headers on a UNIX socket makes sense, but we need big red warnings in the docs about this, because it's injectable/controllable from the client side.

server/core/src/https/mod.rs
```diff
@@ -404,7 +423,39 @@ async fn server_tls_loop(
     info!("Stopped {}", super::TaskName::HttpsServer);
 }
 
-async fn server_plaintext_loop(
+async fn server_unix_loop(
```
yaleman, line 426: You've completely missed the server shutdown signalling in case of error (see the `server_message_tx.send(CoreAction::Shutdown)` in `server_tls_loop`.

server/core/src/https/mod.rs
```diff
@@ -550,18 +627,23 @@ async fn process_client_addr(
...
                 error!(?connection_addr, ?err, "Unable to process proxy v2 header");
                 return Err(std::io::Error::from(ErrorKind::ConnectionAborted));
             }
         }
     } else {
-        (stream, connection_addr)
+        (stream, match connection_addr {
+            // NOTE: To simply codepaths, we simply use IPv6 Localhost as the client ip
+            //       for unix connections here.
+            ConnectionAddress::Unix => IpAddr::V6(Ipv6Addr::LOCALHOST),
```
yaleman, line 641: It might feel nice to be simple, but this breaks visibility/debuggability, and shouldn't be the way to implement this.

dblsaiko:
You don't need to allocate a port and can use permission-based access for the socket. It's much easier to manage and more cleaner when putting services behind a reverse proxy anyway.

Firstyear:
But permission based access doesn't matter anyway, as you plan to expose them on the reverse proxy - so you can just hit 127.0.0.1 and get to it anyway.

WRT to the port allocation, why do you need to "save ports"? The only reason I can think of is that you are virtualhosting many applications on port 443 on the same machine. Is that correct?

dblsaiko:
I don't need to save ports, I need to manage them and keep track of which one is used for what. At this point I have writtern a NixOS module which auto-allocates them sequentially for the few services that can't listen on a unix socket but I'd rather not have to do that at all. It's a bad hack to have to do this with port numbers when you could use a more expressive (and also definitely unique) path name instead.

server/core/src/https/mod.rs
yaleman: Discussed offline: doesn't need it, since we handle this with config-loading in testing already :smile:

yaleman (changes requested):
This is an initial technical review, the PR needs significant change before further consideration (see other comment incoming):

- no docs
- no testing
- missed the shutdown handler
- socket ownership is a risk that needs modelling
- loss of client TLS cert handling which drops a major feature
- drops visibility in some key areas
- proxy v2 should be required since XFF headers are modifiable by clients and trivial to mishandle by admins

yaleman:
Thanks for your work on this one, what started as a short response turned into a bit of an epic, and I hope you spend the time to consider it fully.

When considering any design change, we look at things from a few dimensions (among others):
 
- How does adding this functionality improve the platform, in a way that **_nothing else can_**?
- How does adding this functionality degrade the platform?
- How does this meet our "**secure first, then everything else**" mantra?
- Does the cost of maintaining this going forward make implementing it a net positive to the project?

## How does adding this functionality improve the platform, in a way that nothing else can?

UNIX sockets allow for another method for passing communications between a frontend proxy and the backend service. Currently this is handled by Kanidm listening on HTTPS, which is well-supported and arguably the default of every notable frontend proxy system.

**Result**: It doesn’t, it’s a nice-to-have which can be solved with the existing design.

## How does adding this functionality degrade the platform?

1. As implemented, trustability of client connection information is reduced/muddied.
2. LDAP still listens on TCP if it's enabled.
3. Replication still listens on TCP ports.
4. Client TLS visibility's lost entirely.
5. It adds a whole new path for entry to the server, which requires significant testing, slowing down development.

**Result**: One may not use these features, but it adds more edge cases and not-least-surprise moments for the devs that design it and admins that use it.

## How does this meet our "secure first, then everything else" mantra?

While it's good to reduce the number of ports that a service has to listen on, this would make a plaintext communication phase part of the **production** design. [TLS-all-the-way is one of our Very Big Things](https://kanidm.github.io/kanidm/master/frequently_asked_questions.html#why-tls) that we've stuck hard to, and this breaks that design fundamental that we have spent a long time discussing and explaining.

Yes, there's ways to secure it with permissions on the UDS path, but validation for this isn't even trivially implemented at the moment, and I'm not certain there *is* a way to do this securely in a way that'd pass deep risk modelling.

**Result**: This is a fundamental breach of our position on TLS-all-the-way and secure deisgn.

## How does this impact our ability to reason about how the platform will act in operation?

It's not a *huge* change, but there's a lot of little things that break in a complex system, and this adds to the complexity in new ways, requiring significant testing (and likely producing new bugs).

**Result**: As-yet unknown, but it’s another branch in the complexity of the system. 

## Does the cost of maintaining this going forward make implementing it a net positive to the project?

This is an incredibly important question that gets asked in commercial situations, and often forgotten in free/open source projects.

Adding code and features isn’t a one-time thing. It requires maintenance, it makes the system more complex, documentation needs updating. People need help when they try to use that feature. PRs need review, discussion and responding to.

Kanidm is a project that people work on in their spare time (even now I’m writing this on my day off when I really should be doing something else 🙂). To be VERY blunt, you can count the number of people truly invested in Kanidm on one hand, with some amazing community members that help out when they can, and a vast number of people relying on that.  

Noting the previous points:

- It enables plaintext transport, breaching one of our core goals
- It doesn’t solve a unique problem
- It will add significant development test complexity, support load and maintenance

**Result**: I personally don’t believe that it's net-positive, and won’t support its merge.

Again, thank you very much for your consideration and work on the PR, but it's something that we've discussed further and don't currently see as a viable feature.

PR #3787, opened 2025-08-12
Set `private` in Cache-Control header for `dont_cache_me` middleware

stepbrobd:
# Change summary

- set `private` in Cache-Control header

Checklist

This is a fix for something incredibly niche... I have a Kanidm deployment behind Cloudflare, and I've set a fairly aggressive caching policy for it. The policy is set to override `Cache-Control` header's `no-store` `no-cache` directive and have all contents eligible for cache.

After logging in my browser, the Cloudflare will cache private routes like `/ui/apps`, such that even an anonymous user can see the applications I've configured on my Kanidm instance (but they can't bypass authentication, even if they click on one of the apps, Kanidm will redirect to the login page).

So for private routes, I think it'd be sensible to set `private` in Cache-Control header to make sure no CDN services can actually cache the page contents even if Kanidm users set dumb rules like I am...

Note that some routes don't have the caching middleware hooked to them, would it be better to have `dont_cache_me` to guard all routes that currently don't have any middleware?

### DIFF
--- a/server/core/src/https/middleware/caching.rs
+++ b/server/core/src/https/middleware/caching.rs
@@ -10,7 +10,7 @@ pub async fn dont_cache_me(request: Request<Body>, next: Next) -> Response {
     let mut response = next.run(request).await;
     response.headers_mut().insert(
         header::CACHE_CONTROL,
-        HeaderValue::from_static("no-store, no-cache, max-age=0"),
+        HeaderValue::from_static("private, no-store, no-cache, max-age=0"),
     );
     response
         .headers_mut()


Firstyear:
no-store already exists to state that routes should not be cached by private or shared caches. The problem here is you are breaking those rules that we put into kanidm intentionally by changing our cache policy. We set these cache control headers for a reason to prevent the kind of issue you ran into here with other users able to see your applications. 

Additionally, setting "private" doesn't make sense here - we *do not* want those routes cached even on a private cache because they are things like "is my session still valid". That's not an answer we should have cached at all. 

So I think in this case, we won't accept this change, as we already correctly defined our cache rules based on the role of each route.

yaleman:
Aggressive caching on an inherently dynamic system's a poor choice which teaches hard lessons (see your Anon-can-see-private-things) - MAYBE apply it with a rule on `/pkg/` to cache the big static things at the most (which CF should either way).

Firstyear:
We already set the correct cache headers on that route anyway so no tinkering needed ....

stepbrobd:
Thx for the clarification!

PR #3805, opened 2025-08-19
Prevent users saving their credentials if there are none

### LINKED ISSUE #3799: People accounts can be locked out by removing all credentials
### I did this

1. `kanidm person credential update <ACCOUNT_ID>`
2. Remove all credentials using appropriate commands.

### I expected the following

`Can Commit: false` and an appropriate warning message.

### This happened instead

`Can Commit: true` and committing is successful. Logging in to the account becomes impossible with the message "invalid credential state".

A privileged user must intervene to restore access to the account (such as generating a reset token).

### Kanidm version details

Official Kanidm and client tools containers, versions 1.7.2

### Any other comments

This behavior is also present in the web UI:

Image

Firstyear:
This prevents users saving their credentials if the action would cause them to have no remaining credentials. The bulk of this PR actually is updates to tests to handle this situation because we previously relied on internally being able to clear credentials in a user.

Blocked by: https://github.com/kanidm/webauthn-rs/pull/514

Fixes #3799

Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -68,7 +68,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "5a15f179cd60c4584b8a8c596927aadc462e27f2ca70c04e0071964a73ba7a75"
 dependencies = [
  "cfg-if",
- "getrandom 0.3.3",
+ "getrandom 0.3.4",
  "once_cell",
  "serde",
  "version_check",
@@ -209,7 +209,7 @@ dependencies = [
  "rustc-hash 2.1.1",
  "serde",
  "serde_derive",
- "syn 2.0.107",
+ "syn 2.0.108",
 ]
 
 [[package]]
@@ -245,7 +245,7 @@ checksum = "34921de3d57974069bad483fdfe0ec65d88c4ff892edd1ab4d8b03be0dda1b9b"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.107",
+ "syn 2.0.108",
 ]
 
 [[package]]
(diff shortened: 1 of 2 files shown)


server/core/templates/credentials_update_partial.html
```diff
@@ -52,6 +52,9 @@
                 (% when CURegWarning::WebauthnUserVerificationRequired %)
                 The passkey you attempted to register did not provide user verification. Please
                 ensure that you have a PIN or alternative configured on your authenticator.
+                (% when CURegWarning::NoValidCredentials %)
+                Your account has no valid credentials which will prevent you being able to
+                authenticate. Please create at least one credential.
```
yaleman:
```suggestion
                Your account has no credentials, which will prevent you being able to
                authenticate. Please create at least one credential.
```

server/lib/src/idm/credupdatesession.rs
```diff
@@ -267,6 +267,17 @@ impl CredentialUpdateSession {
...
+        if can_proceed
+            && self.attested_passkeys.is_empty()
+            && self.passkeys.is_empty()
+            && self.primary.is_none()
+        {
+            // The user has no credentials to login to their account with, we can not proceed!
+            can_proceed = false;
+            warnings.push(CredentialUpdateSessionStatusWarnings::NoValidCredentials)
+        }
+
```
yaleman, line 280: tracking `can_proceed` adds complexity - on line 271 we could check if warnings is empty, and the return value of this function could be `Result<(), Vec<CredentialUpdateSessionStatusWarnings>>` 

ie, no warnings, it's an OK, otherwise it's an error and here's the warnings

Firstyear: In my head, the separation is that not all warnings prevent commit. That's why I tracked it separately.

yaleman: Looking at them, they should be called errors or blockers or something 'cause they're all commit-blockers 😄

Firstyear: They should be called bikesheds :P

yaleman (approved):

PR #3812, opened 2025-08-22
Make it clearer why acceptor isnt available

Firstyear:
# Change summary

-

Fixes #

Checklist

### DIFF
--- a/server/core/src/repl/mod.rs
+++ b/server/core/src/repl/mod.rs
@@ -1003,7 +1003,8 @@ async fn repl_acceptor(
                             } else {
                                 // TLS is not setup, generally due to no accepted/trusted client
                                 // certs being present. Drop the connection.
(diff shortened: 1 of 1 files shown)


server/core/src/repl/mod.rs
```diff
@@ -1003,7 +1003,8 @@ async fn repl_acceptor(
                             } else {
                                 // TLS is not setup, generally due to no accepted/trusted client
                                 // certs being present. Drop the connection.
-                                warn!("ignoring connection when tls_acceptor is not available");
+                                warn!("Ignoring replication connection when tls_acceptor is not available.");
```
yaleman: Can we make it include the source address from the connection, so it's clear which server we need to add/block?

Firstyear: Sure,

PR #3846, opened 2025-09-12
20250912 unixd performance

### LINKED ISSUE #323: unix daemon - pro-actively refresh cache items in use
When a request can be service from the cache, if the item is within a threshold to the end of cache expiry, begin a back ground refresh of the item. This prevents the caller from needing to block during the cache refresh at the expiry time.

Today the order of events is:

* id lookup
* cache miss
* access record
* cache
* serve id

Then when doing subsequent lookups

* id lookup
* cache hit
* serve id

When the cache item expires we have to redo the first cycle which causes a delay every cache expiry window as we revalidatethe item. 

Instead if an id is being used frequently, when near the end of it's cache validity, we can background check it so that we never "block" again. This way when we "stop" accessing the item, it will naturally expire.

Firstyear:
# Change summary

@tacerus has noticed in some high load environments that unixd may have some performance issues. These mostly occured in 1.7, as we had a redesign of some locking between 1.6 and 1.7. We also avoid multiple green-threads being in certain critical sections together, but that also creates some possible issues with "writes". It's likely these consistency improvements are contributing to the issues @tacerus is seeing.

Due to the design of how unixd works today we either return a users nss/pam info directly from the cache, or we return it after we perform a network call to kanidm to get the details. 

In a super busy environment this means that we can end up "queuing" a lot of these network calls, which can then delay authentication and lookups. And if a lot of these are updates to existing data they can queue up to acquire our write locks. That delays the whole pipeline. 

To resolve this, there are a small number of changes here:

* Extension of the default cache timeout to reduce how often we query the server
* Expansion of the nxcache so that we can handle more "not founds", which tends to be a thing on internet facing machines
* Adding "jitter" to expiry times - when the system boots we will likely resolve a lot of id's all at once, and that creates a thundering herd problem. By adding a variable amount of jitter (between 0 and 10s, divided to millisecons) we can help avoid this.
* background prefetch of items that are near their expiry time. This way we avoid "blocking" to resolve a user that is being heavily used or queried. 
  * NOTE: We only background refresh an item if it's actively being queried, we don't refresh "everything". 
* nxcache refreshes default to async/background rather than inline refreshes if the item was in the nxcache and needs revalidation.
* Add indexes to sqlite (holy fuck how did I actually not have these!!!)
* Rearrange some problematic tracing spans that could accidentally use a lot of memory
* Enable two threads so that we can handle some more parallel work. 

No one change here is the "magic improvement" but combined, each one adds a small improvement in busy environments.

I have tested this with a multi-thread stress test that resolves nonexistant, present and system accounts rapidly and saw no increase in memory usage. There was (obviously) a lot of CPU usage due to the sheer number of requests I was making though. 

Fixes #323

Checklist


unix_integration/resolver/src/idprovider/kanidm.rs
```diff
@@ -344,6 +344,7 @@ impl IdProvider for KanidmProvider {
         inner.state = CacheState::Offline;
     }
 
+    #[instrument(level = "debug", skip_all)]
```
yaleman: Since its debug can we skip everything but the ID, so it's easier to figure out

unix_integration/resolver/src/resolver.rs
```diff
@@ -70,6 +75,13 @@ pub enum AuthSession {
     Denied,
 }
 
+#[derive(Debug, PartialEq)]
+enum ExpiryState {
+    Valid,
+    ValidRefresh,
```
yaleman, line 86: I can maybe infer this from the PR but can you please document this variant? I'm guessing it's close to expiry and could do with a refresh?

unix_integration/resolver/src/resolver.rs
```diff
@@ -606,17 +648,27 @@ impl Resolver {
...
         // If the token isn't found, get_cached will set expired = true.
-        if expired {
-            self.refresh_usertoken(account_id, item, SystemTime::now())
-                .await
+        if expiry_state == ExpiryState::Expired {
+            self.refresh_usertoken(account_id, current_time).await
         } else {
+            if expiry_state == ExpiryState::ValidRefresh {
+                // We don't mind if the buffer is full.
+                let _ = self.async_refresh_tx.try_send(account_id.clone());
```
yaleman: Can we make this an if let err => write a debug log, just for when someone's trying to work out weirdness later

Firstyear:
I'm also doing a test of this on some development systems to ensure the changes function correctly.

Firstyear:
Okay, cleared up the issues and added some other improvements.

Firstyear:
This PR looks bigger than it really is - mostly because I had to indent/outdent some blocks to add an async {} or a function.

Firstyear:
@yaleman Okay this one is ready to review now. I've been testing it really hard on some of my systems and it does exactly what it says on the tin :)

unix_integration/common/src/client_sync.rs
```diff
@@ -29,15 +30,30 @@ impl From<UnixStream> for DaemonClientBlocking {
...
                 error!(
-                    "Unix socket stream setup error while connecting to {} -> {:?}",
-                    path, e
-                );
-                e
+                    ?err, %path,
+                    "Unix socket stream setup error",
+                )
             })
             .map_err(Box::new)?;
```
yaleman: given this is here, why not just merge it into the inspect_err above? 😄

unix_integration/common/src/client_sync.rs
```diff
@@ -55,38 +71,41 @@ impl DaemonClientBlocking {
...
-
         self.stream
             .write_all(&data)
             .and_then(|_| self.stream.flush())
-            .map_err(|e| {
-                error!("stream write error -> {:?}", e);
-                e
-            })
+            .inspect_err(|err| error!(?err, "stream write error"))
             .map_err(Box::new)?;
```
yaleman: 🫨

Firstyear:
Okay, cleaned up and did your changes, I'll do another self-review in the morning.

unix_integration/common/src/client_sync.rs
```diff
@@ -29,17 +30,29 @@ impl From<UnixStream> for DaemonClientBlocking {
...
+        use tracing_subscriber::prelude::*;
+        use tracing_subscriber::{filter::LevelFilter, fmt};
+
+        let fmt_layer = fmt::layer().with_target(false);
+        let filter_layer = LevelFilter::ERROR;
+
+        let _ = tracing_subscriber::registry()
+            .with(filter_layer)
+            .with(fmt_layer)
+            .try_init();
```
yaleman: This should get removed now :nod:

unix_integration/common/src/client_sync.rs
```diff
@@ -55,54 +68,65 @@ impl DaemonClientBlocking {
...
 
         // Now wait on the response.
         data.clear();
-        let start = SystemTime::now();
+        let start = Instant::now();
         let mut read_started = false;
 
         loop {
-            let durr = SystemTime::now().duration_since(start).map_err(Box::new)?;
+            debug!("read loop");
```
yaleman: This'd be insanely noisy even for a debug message 🙄 

```suggestion
            trace!("read loop");
```

unix_integration/common/src/client_sync.rs
```diff
@@ -55,54 +68,65 @@ impl DaemonClientBlocking {
...
             if durr > timeout {
                 error!("Socket timeout");
                 // timed out, not enough activity.
                 return Err(Box::new(io::Error::other("Timeout")));
             }
 
-            let mut buffer = [0; 1024];
+            let mut buffer = [0; 8192];
+
+            debug!(read_timeout = ?self.stream.read_timeout(), write_timeout = ?self.stream.write_timeout());
```
yaleman:
```suggestion
            trace!(read_timeout = ?self.stream.read_timeout(), write_timeout = ?self.stream.write_timeout());
```

unix_integration/common/src/client_sync.rs
```diff
@@ -121,20 +145,37 @@ impl DaemonClientBlocking {
                 }
                 Ok(count) => {
                     read_started = true;
+                    debug!("read {count} bytes");
```
yaleman:
```suggestion
                    trace!("read {count} bytes");
```

unix_integration/common/src/client_sync.rs
```diff
@@ -121,20 +145,37 @@ impl DaemonClientBlocking {
...
                     data.extend_from_slice(&buffer[..count]);
                 }
-                Err(e) => {
-                    error!("Stream read failure from {:?} -> {:?}", &self.stream, e);
+                Err(err) if err.kind() == ErrorKind::WouldBlock => {
+                    debug!("would block");
```
yaleman:
```suggestion
                    trace!("Read from UDS would block, sleeping for 500ms");
```

unix_integration/common/src/client_sync.rs
```diff
@@ -121,20 +145,37 @@ impl DaemonClientBlocking {
...
                 // Need more data
-                Ok(None) => continue,
-                Err(e) => return Err(Box::new(e)),
+                Ok(None) => {
+                    debug!("need more");
+                    continue;
+                }
+                // Failed to decode for some reason
+                Err(err) => {
+                    error!(?err);
```
yaleman:
```suggestion
                    error!(?err, "Failed to decode message");
```

unix_integration/common/src/client_sync.rs
```diff
@@ -121,20 +145,37 @@ impl DaemonClientBlocking {
...
-                    return Err(Box::new(e));
+                    return Err(Box::new(err));
                 }
             }
 
             match self.codec.decode(&mut data) {
-                Ok(Some(cr)) => return Ok(cr),
+                // A whole frame is ready and present.
+                Ok(Some(cr)) => {
+                    debug!("read loop - ok");
```
yaleman:
```suggestion
                    trace!("read loop - ok");
```

unix_integration/common/src/client_sync.rs
```diff
@@ -121,20 +145,37 @@ impl DaemonClientBlocking {
...
+                    return Ok(cr);
+                }
                 // Need more data
-                Ok(None) => continue,
-                Err(e) => return Err(Box::new(e)),
+                Ok(None) => {
+                    debug!("need more");
```
yaleman:
```suggestion
                    trace!("read loop needs more bytes");
```

unix_integration/common/src/constants.rs
```diff
@@ -6,7 +6,10 @@ pub const DEFAULT_TASK_SOCK_PATH: &str = "/var/run/kanidm-unixd/task_sock";
 pub const DEFAULT_PERSISTENT_DB_PATH: &str = "/var/lib/kanidm-unixd/kanidm.db";
 pub const DEFAULT_CACHE_DB_PATH: &str = "/var/cache/kanidm-unixd/kanidm.cache.db";
 pub const DEFAULT_CONN_TIMEOUT: u64 = 2;
-pub const DEFAULT_CACHE_TIMEOUT: u64 = 15;
+pub const DEFAULT_CACHE_TIMEOUT_MINIMUM: u64 = 60;
+pub const DEFAULT_CACHE_TIMEOUT_MAXIMUM: u64 = 24 * 3600;
+pub const DEFAULT_CACHE_TIMEOUT: u64 = 120;
+pub const DEFAULT_CACHE_TIMEOUT_JITTER_MS: u64 = 10_000;
```
yaleman:
```suggestion
pub const DEFAULT_CACHE_TIMEOUT_MINIMUM: u64 = 60; // seconds
pub const DEFAULT_CACHE_TIMEOUT_MAXIMUM: u64 = 24 * 3600; // seconds
pub const DEFAULT_CACHE_TIMEOUT: u64 = 120; // seconds
pub const DEFAULT_CACHE_TIMEOUT_JITTER_MS: u64 = 10_000;
```

Firstyear:
The multithread test is a hack job RN, I would want something more polished to commit. I think we need to revive the criterion benchmarks.

yaleman (approved):

yaleman:
Wooooooooo, here's hoping we've caught it this time!

PR #3847, opened 2025-09-12
Add password check api

Firstyear:
# Change summary

- Adds the server side parts for a CURequest::PasswordQualityCheck. 
- @tennox What is remaining is a handler similar to https://github.com/kanidm/kanidm/blob/master/server/core/src/https/views/reset.rs#L688 which issues a CURequest::PasswordQuality check, and then can render the html or whatever. Not 100% sure what you want this to look like or how to do it nicely within the htmx space, but I think you are better in this area than I am.   

Relates #3650

Checklist

### DIFF
--- a/server/core/src/actors/v1_read.rs
+++ b/server/core/src/actors/v1_read.rs
@@ -1171,147 +1171,86 @@ impl QueryServerReadV1 {
         match scr {
             CURequest::PrimaryRemove => idms_cred_update
                 .credential_primary_delete(&session_token, ct)
-                .map_err(|e| {
-                    error!(
-                        err = ?e,
-                        "Failed to begin credential_primary_delete",
-                    );
-                    e
+                .inspect_err(|err| {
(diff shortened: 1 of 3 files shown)


server/core/src/actors/v1_read.rs
```diff
@@ -1178,6 +1178,15 @@ impl QueryServerReadV1 {
...
                 }),
+            CURequest::PasswordQualityCheck(pw) => idms_cred_update
+                .credential_check_password_quality(&session_token, ct, &pw)
+                .map_err(|e| {
+                    error!(
+                        err = ?e,
+                        "Failed to begin credential_check_password_quality",
+                    );
+                    e
+                }),
```
yaleman: This should be an inspect_err call if it's not modifying the error

yaleman: This should be an inspect_err call if it's not modifying the error

PR #3863, opened 2025-09-23
20250801 reference entries

### LINKED ISSUE #3014: Reference Types / Sub Entries
We have a need for certain classes of objects to be stored in "referenced" entries rather than directly in the parent.

Examples include:

* Images related to oauth2/domain
* Sessions and Descendant Session state
* Scope Maps / Claim Maps in OAuth2

Now lets explain why - currently these types are stored directly in the entry via valuesets. This creates a few problems.

* Images consume memory, and so caching of the entry in ram implies that the image must be in ram, even if not actively being read.
* To read the image, necessitates loading *all* of the entry, not just the image.
* To add new types we need to continually extend ValueSet and DBEntry which is non-trivial and presents complexity to contributors rather than having a way to create compound-sub-entries.

For example, sessions as subentries would allow re-use of existing uuid, datetime and simpler types in a sub-entry, rather than requiring a more complex valueset item to be made. It means that when we get the entry we can choose if we want to load sessions or not. 

In the image case it also allows *deduplication* of identical images so that on upload if two items reference identical content, they can then share the same image reference.

Firstyear:
# Change summary

- Implement reference entries 

Fixes #3014 

Checklist

### DIFF
--- a/server/lib/src/plugins/refint.rs
+++ b/server/lib/src/plugins/refint.rs
@@ -41,7 +41,7 @@ impl ReferentialIntegrity {
         // operation.
         let filt_in = filter!(f_inc(inner));
         let b = qs.internal_exists(&filt_in).inspect_err(|err| {
-            error!(?err, "internal exists failure");
+            error!(?err, filter = ?filt_in, "internal exists failure");
         })?;
 
         // Is the existence of all id's confirmed?
@@ -68,7 +68,7 @@ impl ReferentialIntegrity {
         for u in inner {
             let filt_in = filter!(f_eq(Attribute::Uuid, PartialValue::Uuid(*u)));
             let b = qs.internal_exists(&filt_in).inspect_err(|err| {
-                error!(?err, "internal exists failure");
+                error!(?err, filter = ?filt_in, "internal exists failure");
             })?;
 
             // If it's missing, we push it to the missing set.
@@ -149,7 +149,7 @@ impl ReferentialIntegrity {
         let filt_in = filter!(f_inc(inner));
 
         let b = qs.internal_exists(&filt_in).inspect_err(|err| {
-            error!(?err, "internal exists failure");
+            error!(?err, filter = ?filt_in, "internal exists failure");
         })?;
 
         // If this is true, it means all uuids do NOT have the attribute refers
@@ -181,7 +181,7 @@ impl ReferentialIntegrity {
                 f_pres(Attribute::Refers)
             ]));
             let b = qs.internal_exists(&filt_in).inspect_err(|err| {
-                error!(?err, "internal exists failure");
+                error!(?err, filter = ?filt_in, "internal exists failure");
             })?;
 
             // If it's missing, we push it to the missing set.
@@ -576,9 +576,7 @@ impl ReferentialIntegrity {
                 Self::check_refers_to_target_loop_slow(qs, refers_target_uuids.as_slice())?;
 
             error!("some entries introduce an invalid reference to another reference.");
-            for looping in looping_uuids {
-                error!(?looping);
-            }
+            error!(?looping_uuids);
 
             return Err(OperationError::ReferenceLoop);
         }
--- a/server/lib/src/schema.rs
+++ b/server/lib/src/schema.rs
@@ -775,11 +775,16 @@ impl SchemaWriteTransaction<'_> {
             .collect()
     }
 
+    /// Generate the minimal in memory schema needed to begin the server bootstrap
+    /// process. This should contain the most critical schema definitions that the
+    /// server requires to be able to read in other schema objects and persist them
+    /// into our database.
+    ///
+    /// THIS IS FOR SYSTEM CRITICAL INTERNAL SCHEMA ONLY
+    ///
+    /// Schema should otherwise be in our migration data - not here.
     #[instrument(level = "debug", name = "schema::generate_in_memory", skip_all)]
     pub fn generate_in_memory(&mut self) -> Result<(), OperationError> {
-        //
-        // THIS IS FOR SYSTEM CRITICAL INTERNAL SCHEMA ONLY
-        //
         self.classes.clear();
         self.attributes.clear();
         // Bootstrap in definitions of our own schema types
@@ -1501,7 +1506,7 @@ impl SchemaWriteTransaction<'_> {
             SchemaAttribute {
                 name: Attribute::Refers,
                 uuid: UUID_SCHEMA_ATTR_REFERS,
-                description: String::from("A reference to linked object"),
+                description: String::from("A reference to another object"),
                 multivalue: false,
                 unique: false,
                 phantom: false,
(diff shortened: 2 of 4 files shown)


Firstyear:
FYI still not ready for review, I need to add more tests.

Firstyear:
Includes prevention of reference loops, and replication conflict handling. I might think of some edge cases in future, but I'm pretty sure this is solid rn. 

There are some other schema cleanups in here that I found in the process.

server/lib/src/server/recycle.rs
```diff
@@ -203,6 +208,29 @@ impl QueryServerWriteTransaction<'_> {
 
         let norm_cand: Vec<Entry<_, _>> = res?;
 
+        // Finally, setup the mod for restoring memberships from direct member of
+        let mut dm_mods: BTreeMap<Uuid, ModifyList<ModifyInvalid>> = Default::default();
+
+        for entry in &pre_candidates {
+            // Get this entries uuid.
```
yaleman:
```suggestion
            // Get this entry's uuid.
```

server/lib/src/plugins/refint.rs
```diff
@@ -43,9 +40,8 @@ impl ReferentialIntegrity {
         // will fail. This will return the union of the inclusion after the
         // operation.
         let filt_in = filter!(f_inc(inner));
-        let b = qs.internal_exists(filt_in).map_err(|e| {
-            admin_error!(err = ?e, "internal exists failure");
-            e
+        let b = qs.internal_exists(&filt_in).inspect_err(|err| {
+            error!(?err, "internal exists failure");
```
yaleman: Something like this to show what failed, maybe? "internal exists failure" isn't actionable for an error and the same message is here and a lower message, so that wouldn't help for debugging.

```suggestion
            error!(filter=%filt_in, ?err, "internal exists failure");
```

server/lib/src/schema.rs
```diff
@@ -777,6 +777,8 @@ impl SchemaWriteTransaction<'_> {
 
     #[instrument(level = "debug", name = "schema::generate_in_memory", skip_all)]
     pub fn generate_in_memory(&mut self) -> Result<(), OperationError> {
+        //
+        // THIS IS FOR SYSTEM CRITICAL INTERNAL SCHEMA ONLY
```
yaleman: This should be a function docstring 😆

server/lib/src/schema.rs
```diff
@@ -1493,6 +1495,41 @@ impl SchemaWriteTransaction<'_> {
                 syntax: SyntaxType::ReferenceUuid,
             },
         );
+
+        self.attributes.insert(
+            Attribute::Refers,
+            SchemaAttribute {
+                name: Attribute::Refers,
+                uuid: UUID_SCHEMA_ATTR_REFERS,
+                description: String::from("A reference to linked object"),
```
yaleman:
```suggestion
                description: String::from("A reference to another object"),
```

server/lib/src/schema.rs
```diff
@@ -1493,6 +1495,41 @@ impl SchemaWriteTransaction<'_> {
...
+                uuid: UUID_SCHEMA_ATTR_CASCADE_DELETED,
+                description: String::from("A marker attribute denoting that this entry was deleted by cascade when this UUID was deleted."),
+                multivalue: false,
+                unique: false,
+                phantom: false,
+                sync_allowed: false,
+                replicated: Replicated::True,
+                indexed: true,
+                // NOTE: This has to be Uuid so that referential integrity doesn't consider
+                // this value in it's operation.
```
yaleman:
```suggestion
                // this value in its operation.
```

server/lib/src/server/delete.rs
```diff
@@ -49,6 +49,56 @@ impl QueryServerWriteTransaction<'_> {
...
+                .iter()
+                .map(|entry| { f_eq(Attribute::Refers, PartialValue::Refer(entry.get_uuid())) })
+                .collect(),
+        ));
+
+        let mut pre_cascade_delete_candidates = self
+            .internal_search(references_filt)
+            .inspect_err(|err| error!(?err, "unable to find reference entries"))?;
+
+        #[cfg(debug_assertions)]
```
yaleman:
```suggestion
        #[cfg(any(test,debug_assertions))]
```

Just 'cause I'm paranoid

server/lib/src/plugins/refint.rs
```diff
@@ -121,6 +121,77 @@ impl ReferentialIntegrity {
...
+            })
+            .collect();
+
+        // F_inc(lusion). All items of inner must be 1 or more, or the filter
+        // will fail. This will return the union of the inclusion after the
+        // operation.
+        let filt_in = filter!(f_inc(inner));
+
+        let b = qs.internal_exists(&filt_in).inspect_err(|err| {
+            error!(?err, "internal exists failure");
```
yaleman: This could use an actionable message and at least a slightly different message for debugging

server/lib/src/plugins/refint.rs
```diff
@@ -121,6 +121,77 @@ impl ReferentialIntegrity {
...
+        }
+
+        let mut looping = Vec::with_capacity(inner.len());
+        for u in inner {
+            let filt_in = filter!(f_and(vec![
+                f_eq(Attribute::Uuid, PartialValue::Uuid(*u)),
+                f_pres(Attribute::Refers)
+            ]));
+            let b = qs.internal_exists(&filt_in).inspect_err(|err| {
+                error!(?err, "internal exists failure");
```
yaleman: See above 😆

server/lib/src/plugins/refint.rs
```diff
@@ -179,6 +250,60 @@ impl Plugin for ReferentialIntegrity {
...
+            .iter()
+            .copied()
+            .map(|conflict_uuid| f_eq(Attribute::Refers, PartialValue::Refer(conflict_uuid)))
+            .collect::<Vec<_>>();
+
+        let filt = filter!(f_or(filt_inner));
+
+        // Does anything match?
+        let conflict_cand = qs.internal_exists(&filt).inspect_err(|err| {
+            error!(?err, "internal exists error");
```
yaleman, line 277: 😏

server/lib/src/plugins/refint.rs
```diff
@@ -415,36 +547,57 @@ impl ReferentialIntegrity {
...
+            Self::check_refers_to_target_loop_fast(qs, refers_target_uuids.as_slice())?;
+
+        if loops_present {
+            let looping_uuids =
+                Self::check_refers_to_target_loop_slow(qs, refers_target_uuids.as_slice())?;
+
+            error!("some entries introduce an invalid reference to another reference.");
+            for looping in looping_uuids {
+                error!(?looping);
+            }
```
yaleman:
```suggestion
            error!(looping_uuids = ?looping_uuids,
            "entries introduce an invalid reference to another reference.");
```
Just to cut the noise a bit due to how verbose tracing tree is

Issue #2299, opened 2023-11-07
FR: Add password reset form in the Web UI


sbordeyne:
**Is your feature request related to a problem? Please describe.**

Currently, the only way to reset a password is to use the `kanidm` CLI. This is not convenient for end users, since they do not have access to the server, or to admin credentials.

**Describe the solution you'd like**

KanIDM has a Web UI, therefore it should allow password resets from there, in order for users of the system to reset their password without access to the underlying backend.

**Describe alternatives you've considered**

Writing a service that calls to the kanidm CLI to send password reset emails, but I'd rather this feature be integrated directly into the project for convenience.

Firstyear:
I think would be related to #996

yaleman:
I'd envision this being a link on a login page that takes you to a "put your username in, and we'll message you with a reset link" UI.

Firstyear:
I'm not sure every deployment wants/needs that though. But also this issue is a bit ambiguous becuse it could be referring to a service-desk style feature to send the pw reset.

yaleman:
Both are requirements TBH, and the former would be a toggle-able option for sure 😄

TheRealGramdalf:
Me again :)

Thought I'd add my two cents on the "message for a reset link" method:

- A lot of implementations don't seem super secure - i.e. they are susceptible to phone scams etc. It may be worth making a conscious design decision to not use the classic "6 digit authenticator code" method, but rather something more secure:
- One option is to never "expose" the token to the end user - simply send a link with the reset token embedded in the URL. Bonus points if it's <10 characters.
- Another is to require multi-factor auth for password reset - it's inherently more secure, but can become very annoying very quickly for people who actually use the password reset function - they typically don't have any sort of password manager, and may not have access to email or some other multi-factor auth method.
    - One example of this is iCloud - it's _very_ annoying to reset an apple ID password, especially due to restrictions in multi-factor auth to mostly apple devices only (i.e. a notification to a mac/iphone etc)

Firstyear:
That's part of the challenge here, is reset link policy is site specific. This adds more surface area to configuration of this flow, but by default I think we'd have cred reset's disabled - only certain sites need it to be activated if at all.

yaleman:
> One option is to never "expose" the token to the end user - simply send a link with the reset token embedded in the URL. Bonus points if it's <10 characters.

For reference, this is already part of our existing credential reset flow, which can be generated by admin users on behalf of the user.

[...]

Issue #2287, opened 2023-11-03
FR: User sign-up portal


yaleman:
**Is your feature request related to a problem? Please describe.**

For a bunch of deployment use cases it'd be good to have a user sign-up portal, or enable developers to integrate signup into another system.

**Describe the solution you'd like**

This'd be an optional UI path on Kanidm's frontend which allows users to register as a new `person` which is disabled until the admin enables it. I'd probably put them in a group like `new_signups` by default, too. 

The UI would prompt for:

* Preferred Username
* Display Name (optional?)
* Email Address (optional)
* Confirmation of acceptance of Terms of Service (with a link)

CSRF on the form is required as a minimum, but some kind of captcha makes sense from a rate limiting/human validation perspective to limit spam attacks.

I really don't want to suggest email validation, so the idea is basically "create the user, tell them that it's done, and that the admin will contact them with details soon".

**Describe alternatives you've considered**

Requiring everyone to manage their own thing with a separate service, service account and a coop full of chickens.

**Additional context**

Long-term, this should be something that can be enabled/disabled without a restart - in case of a spam attack you don't want to have to restart the server and interrupt auth to stop the flood.

I'd suggest doing it in raw HTML and form things because it's *very* simple, doesn't need fancy WASM/Javascript things and can be rendered server-side trivially.

```mermaid
graph TD;

S[Start] --GET--> F["Signup Form (/signup)"]
F --POST--> B["Form validation (/signup/submit)"]
B -. Failed Somehow, redirect with error .->F
B -. Form is valid .-> IDM["IDM Creates Person"]
IDM --HTTP Redirect--> T["Thanks Page (/signup/complete)"]
```

Firstyear:
I actually think that potentially the best thing here is a separate service with it's own ui, and it has an api token that allows it to create users. That way it doesn't necessarily have to be built in, and other people can make their own portals.

yaleman:
Yeah, I thought about that as an option too, I was mainly trying to avoid starting a whole extra service/app for it 🙄

Firstyear:
I have lots of thoughts on this, I think making it separate is the "most flexible" if only because it allows people to expose it in a separate way to the rest of the Kani application. I can see a user sign up being something that people want to customise the most of anything, especially around captchas and things. And they may even want to integrate user sign up to an existing application that they have. This is why I was thinking a "toolbox" to allow someone to make a signup portal would be better because that way they can more easily integrate to their existing stack.

[...]

Issue #634, opened 2022-01-09
Send a user notification when new MFA device enrolled


yaleman:
When new MFA devices are enrolled, it's probably worth notifying users, in case their account has been compromised via another method.

Email's the most likely transport for this - obviously this only works if the user's got one listed - but it's better than nothing and likely to be there either way for other reasons.

MinhPhan8803:
Could this be performed by the totp functions inside [this file](https://github.com/kanidm/kanidm/blob/master/server/lib/src/idm/credupdatesession.rs)?

Firstyear:
@MinhPhan8803 Not really. This is more likely to be about hooking the commit to submit an audit event that can then be stored or processed in different ways. I think it's a bigger feature than just being on the totp section, and would need a conversation and a design doc.

MinhPhan8803:
I think I will just pass on this for now.

Firstyear:
Yeah, this would be a pretty big block of work :)

cmueller-tp:
I'm also interested in notifications - also for logins on new devices would be good.
In general I'd like to send chat messages in Matrix, others may want to send emails, so I think the option to call a web hook or external script would be nice for admins to configure it the way they want.

Firstyear:
I actually think that this is related to #22 as this is just a subset of that.

Re "new device" we don't do any device fingerprinting rn so we don't have an easy/obvious way to do that.

yaleman:
Another thread to note is https://github.com/kanidm/kanidm/discussions/1455

cmueller-tp:
Yep, right.
Regarding the device: you can store a signed token in the browsers local storage to "remember" the device/that a specific user logged in on that device before and wanted to remember the device. That way the user stays in control of what happens but can still be notified if there is a login from an unknown or untrusted device and the data stored is very limited.

yaleman:
The issue is really doing the fingerprinting and identification on the server side, more than the storage requirements or enabling users to manipulate a security feature

Firstyear:
> The issue is really doing the fingerprinting and identification on the server side, more than the storage requirements or enabling users to manipulate a security feature

Yeah this. But also if a cookie/token is stolen then they can just steal the device id token too and bypass the notification. So we have to probably not use a device token/cookie to id devices here.

PR #3878, opened 2025-10-09
Foundations of message sending

Firstyear:
This adds the basics of message queuing and sending to the server. The server itself does not have a mail transport, but relies on an external tool to actually process messages. This way we can detach the transport from the server, and we leave the door open to things like sms or other transports in future.

This also adds basic template types and the storage of them. These will be expanded as we go with different message types e.g. credential reset emails.

Finally, message ready queries rely on less-than searches, so today I added the hooks for these to be indexed, but opted to keep it a bit simpler and just use pres indexes to shortcut the need to fully write less than indexing today in this PR

# Change summary

-

Relates #634 #2287 #2299 

Checklist

### DIFF
--- a/server/core/src/https/v1_scim.rs
+++ b/server/core/src/https/v1_scim.rs
@@ -898,7 +898,10 @@ pub fn route_setup() -> Router<ServerState> {
         //  Message    /Message/{id}    GET                Fetch message by id
         .route("/scim/v1/Message/:id", get(scim_message_id_get))
         //  Message    /Message/{id}/_sent     POST         Mark this message as having been processed and sent
-        .route("/scim/v1/Message/:id/_sent", post(scim_message_id_sent_post))
+        .route(
+            "/scim/v1/Message/:id/_sent",
+            post(scim_message_id_sent_post),
+        )
         // Synchronisation routes.
         .route(
             "/scim/v1/Sync",


libs/client/src/message.rs
```diff
@@ -0,0 +1,36 @@
...
+    pub async fn idm_message_list_ready(&self) -> Result<ScimListMessage, ClientError> {
+        self.perform_get_request("/scim/v1/Message/_ready").await
+    }
+
+    pub async fn idm_message_get(&self, message_id: Uuid) -> Result<ScimEntryMessage, ClientError> {
+        self.perform_get_request(&format!("/scim/v1/Message/{message_id}"))
+            .await
+    }
+
+    pub async fn idm_message_mark_sent(&self, message_id: Uuid) -> Result<(), ClientError> {
```
yaleman, line 32: GETs are idempotent, marking something sent should be a POST

tools/cli/src/opt/kanidm.rs
```diff
@@ -1161,6 +1162,32 @@ pub enum DomainOpt {
...
 }
 
+#[derive(Debug, Subcommand, Clone)]
+pub enum MessageOpt {
+    #[clap(name = "list")]
+    /// List all queued messages
+    List,
+
+    #[clap(name = "get")]
+    /// Display the message identified by it's message id.
```
yaleman:
```suggestion
    /// Display the message identified by its message id.
```

tools/cli/src/opt/kanidm.rs
```diff
@@ -1161,6 +1162,32 @@ pub enum DomainOpt {
...
+    Get {
+        message_id: Uuid
+    },
+
+    #[clap(name = "mark-as-sent")]
+    /// Mark the message with this message id as sent. This will prevent it
```
yaleman:
```suggestion
    /// Mark the message with this message ID as sent. This will prevent it
```

proto/src/attribute.rs
```diff
@@ -349,9 +355,11 @@ impl Attribute {
             Attribute::LinkedGroup => ATTR_LINKEDGROUP,
             Attribute::LoginShell => ATTR_LOGINSHELL,
             Attribute::Mail => ATTR_MAIL,
+            Attribute::MailDestination => ATTR_MAIL_DESTINATION,
```
yaleman, line 360: I kinda know why you're doing it, but why not just use email/mail for this?

proto/src/attribute.rs
```diff
@@ -51,6 +51,7 @@ pub enum Attribute {
     CredentialUpdateIntentToken,
     CredentialTypeMinimum,
     DeniedName,
+    DeleteAfter,
```
yaleman, line 55: If we just call this `Expires` then we can use that for a load of different entry types

proto/src/attribute.rs
Firstyear: Attribute::Mail has schema unique enforced. So if you were to copy in the dest address you'd get a uniqueness conflict.

proto/src/attribute.rs
Firstyear: I did think of that, but the issue is communication. I think Expires doesn't communicate the idea properly that not only does this entry "expire" at a point in time, but that it will be *deleted* at that point. To me, if I read expires, I wouldn't think that the entry would be deleted at that time, only that it would become invalid at that point. 

Anyway, that's why I called it"delete after" because that's also generic enough to use for other entry types too.

PR #3882, opened 2025-10-11
20251010 drop eckeys

Firstyear:
# Change summary

- This improves the identity verification code by dropping eckeys, and using a hmac derived key between the two users instead.
- Remove redundant tests, and move most logic tests into the core server which is more efficient. 
- Helps to remove more openssl!!!

Checklist


Cargo.toml
```diff
@@ -102,7 +102,7 @@ codegen-units = 256
 ## with local or git versions. This patch table allows quick uncommenting to achieve that.
 
 # compact_jwt = { path = "../compact-jwt" }
-# crypto-glue = { path = "../crypto-glue" }
+crypto-glue = { path = "../crypto-glue" }
```
yaleman: nu uh

proto/src/internal/error.rs
```diff
@@ -305,6 +305,14 @@ pub enum OperationError {
     KP0070KeyObjectJwsHs256Signature,
     KP0071KeyObjectPrivateToDer,
 
+    KP0072KeyObjectHs256Invalid,
+    KP0073KeyObjectHs256Invalid,
```
yaleman, line 309: two different errors with the same suffix and no string output? 😄

server/lib/src/migration_data/dl12/system_config.rs
```diff
@@ -33,6 +33,19 @@ lazy_static! {
...
+        (Attribute::Class, EntryClass::Object.to_value()),
+        (Attribute::Class, EntryClass::KeyObject.to_value()),
+        (Attribute::Class, EntryClass::KeyObjectHkdfS256.to_value()),
+        (
+            Attribute::Uuid,
+            Value::Uuid(UUID_DOMAIN_ID_VERIFICATION_KEY)
+        ),
+        (
+            Attribute::Description,
+            Value::new_utf8s("The domain local id verification HMAC key.")
```
yaleman:
```suggestion
            Value::new_utf8s("The domain-local HMAC key used for user ID verification.")
```

proto/src/internal/error.rs
Firstyear: Because they both identify a specific, unique callsite, and the error is one that is "unlikely and bad".

Cargo.toml
Firstyear: Yep, might own confusion. Will fix.

proto/src/internal/error.rs
yaleman: ok, cool, just checking - seemed odd 😄

PR #3899, opened 2025-10-21
Open app links in new tabs

### LINKED ISSUE #3896: In the WebUI Applications tab, open application links in a new tab

When I click on an application on the WebUI Applications tab, I expect it to open in a new tab.  That just seems like the proper behavior to me.  Instead it opens it in the current tab and I've lost my Kanidm tab.

Add  `target=_blank` to each application `a` tag, controlled by one or more of the following:

* The default, unless people have different opinions
* A global option set somewhere
* A per-user option

Firstyear:
# Change summary

- Open applications in new tabs rather than in the existing window. Matches other IAM behaviour. 

Fixes #3896

Checklist

### DIFF
--- a/server/core/templates/apps_partial.html
+++ b/server/core/templates/apps_partial.html
@@ -13,7 +13,7 @@
 				(% match app %)
 				(% when AppLink::Oauth2 with { name, display_name, redirect_url, has_image }
 				%)
-				<a href="(( redirect_url ))" target="_blank" rel="noopener noreferrer" class="link-emphasis stretched-link mt-2">
+				<a href="(( redirect_url ))" target="_blank" rel="noopener" class="link-emphasis stretched-link mt-2">
 					(% if has_image %)
 					<img src="/ui/images/oauth2/(( name ))" class="oauth2-img"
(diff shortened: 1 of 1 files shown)


server/core/templates/apps_partial.html
```diff
@@ -13,7 +13,7 @@ <h2>Applications</h2>
 				(% match app %)
 				(% when AppLink::Oauth2 with { name, display_name, redirect_url, has_image }
 				%)
-				<a href="(( redirect_url ))" class="link-emphasis stretched-link mt-2">
+				<a href="(( redirect_url ))" target="_blank" rel="noopener noreferrer" class="link-emphasis stretched-link mt-2">
```
yaleman: I'd assume a level of trust with configured apps, and some may prefer/require the referrer to be set in particularly security-sensitive environments.

```suggestion
				<a href="(( redirect_url ))" target="_blank" class="link-emphasis stretched-link mt-2">
```

Firstyear: I think we need to retain noopener though, there seems to be some security related concerns with that.

PR #3901, opened 2025-10-21
Fix: set OAuth2 JTI to session ID 

### LINKED ISSUE #3900: Kanidm OAuth2 tokens contain an invalid `jti` claim
### I did this

Issued an oauth2 token 

### I expected the following

The token should comply with the JSON Web Token Standard RFC 7519. In particular, if the optional `jti` (JWT id) claim is present [then it is a string](https://rfc-editor.org/rfc/rfc7519#section-4.1.7).

In addition, the token should comply with the OAuth 2.0 spec (RFC 9068) standard,  which [requires the `jti` field to be present](https://www.rfc-editor.org/rfc/rfc9068.html#section-2.2).

### This happened instead

The Kanidm-issued token contains the `jti` claim but sets it to `null`. An example from running the oauth2 tests:

```json
{
  "aud": "test_integration",
  "client_id": "test_integration",
  "exp": 1761070698,
  "iat": 1761069798,
  "iss": "http://localhost:18081/oauth2/openid/test_integration",
  "jti": null,  // ❌ BUG: Should be a string
  "nbf": 1761069798,
  "parent_session_id": "b63f8bf7-8024-4ad8-9445-861134125e3b",
  "scope": "admin email openid read",
  "session_id": "d4267829-efbe-4a30-bfc6-9ad3e5eef381",
  "sub": "86fcefc2-e059-4a07-9f68-b8e70585bca6"
}
```
If the field were absent, it would comply with RFC 7519 but not RFC 9068. The field is present (as required by 9068) but is not a string, making it in violation of both RFC 7519 and RFC 9068.

### Kanidm version details

This happens on the `master` branch but the same issue is also present in `v1.7.3` and likely other versions.

### Any other comments

Third-party OAuth2/OIDC clients that validate JWTs according to the spec reject Kanidm oauth2 tokens. For example, Kanidm tokens are rejected by Envoy Proxy and Istio.

Kanidm uses  [skip_serializing_none](https://github.com/kanidm/kanidm/blob/36e5cb28b926c1a8c4459bbb8dccab5a09b5bfb0/proto/src/oauth2.rs#L190), and [explicitly sets `jti` to `None` 
](https://github.com/kanidm/kanidm/blob/36e5cb28b926c1a8c4459bbb8dccab5a09b5bfb0/server/lib/src/idm/oauth2.rs#L1753). This looks like an attempt to exclude the `jti` field from serialization. This attempt is incorrect for RFC 9068 (OAuth 2.0) tokens since the field should be present. But, OAuth 2.0 compliance aside, this may be a separate serialization bug if the current expected behavior is for the `jti` claim to be completely omitted.

The correct fix is probably for Kanidm to set the `jti` field to a UUIDv4 string.

yaleman:
# Change summary

- instead of setting the `jti` field (JWS ID) to None and skipping serialisation, am setting it to the session_id which is unique to the token anyway.

Fixes #3900 

Checklist

### DIFF
--- a/proto/src/oauth2.rs
+++ b/proto/src/oauth2.rs
@@ -367,7 +367,7 @@ pub struct AccessTokenIntrospectResponse {
 }
 
 impl AccessTokenIntrospectResponse {
-    pub fn inactive() -> Self {
+    pub fn inactive(session_id: Uuid) -> Self {
         AccessTokenIntrospectResponse {
             active: false,
             scope: BTreeSet::default(),
@@ -380,7 +380,7 @@ impl AccessTokenIntrospectResponse {
             sub: None,
             aud: None,
             iss: None,
-            jti: uuid::Uuid::new_v4(),
+            jti: session_id,
         }
     }
 }
--- a/server/lib/src/idm/oauth2.rs
+++ b/server/lib/src/idm/oauth2.rs
@@ -2397,7 +2397,7 @@ impl IdmServerProxyReadTransaction<'_> {
                 exp,
                 nbf,
                 iat,
-                jti: _,
+                jti,
                 client_id: _,
                 extensions:
                     OAuth2RFC9068TokenExtensions {
@@ -2414,7 +2414,7 @@ impl IdmServerProxyReadTransaction<'_> {
             // Has this token expired?
             if exp <= ct.as_secs() as i64 {
                 security_info!(?sub, "access token has expired, returning inactive");
-                return Ok(AccessTokenIntrospectResponse::inactive());
+                return Ok(AccessTokenIntrospectResponse::inactive(jti));
             }
 
             // Is the user expired, or the OAuth2 session invalid?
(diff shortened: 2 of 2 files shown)


proto/src/oauth2.rs
```diff
@@ -362,7 +362,8 @@ pub struct AccessTokenIntrospectResponse {
     pub sub: Option<String>,
     pub aud: Option<String>,
     pub iss: Option<String>,
-    pub jti: Option<String>,
+    // JWT ID <https://www.rfc-editor.org/rfc/rfc7519#section-4.1.7> set to session ID
+    pub jti: Uuid,
```
Firstyear, line 366: Should be Option<Uuid> so that in the Introspect response we can default it to None.

yaleman: as discussed elsewhere, OAuth2RFC9068Token implies RFC9068 which makes it mandatory

Firstyear (approved):

PR #3907, opened 2025-10-23
Fix passkey typos

frantic1048:
# Change summary

- Fix passkey typos in ``server/core/static/pkhtml.js`` :rofl: 

Checklist


frantic1048:
BTW, there are built-ins [`Uint8Array.fromBase64()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Uint8Array/fromBase64) and [`Uint8Array.prototype.toBase64()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Uint8Array/toBase64) providing straightforward way to deal with Base64 <-> binary data. They also have support for Base64url which is extensively used in the [WebAuthn spec](https://www.w3.org/TR/webauthn-3/#base64url-encoding).

They are widely supported in modern(2025) browsers, it is worth considering replacing the [`js-base64`](https://github.com/frantic1048/kanidm/blob/08cca6c130feaf43d6a925bbcd60991b29e84373/server/core/static/external/base64.js#L1) lib with these native built-ins.

Firstyear:
> They are widely supported in modern(2025) browsers, it is worth considering replacing the [`js-base64`](https://github.com/frantic1048/kanidm/blob/08cca6c130feaf43d6a925bbcd60991b29e84373/server/core/static/external/base64.js#L1) lib with these native built-ins.

Happy for you to do this in the same PR :)

yaleman:
Looking at the [caniuse data for fromBase64](https://caniuse.com/?search=fromBase64), I'm not sure I'd make the transition yet, that's a VERY recent addition (sep 2025 for Chrome, for example) please don't make that change yet.

frantic1048:
> Looking at the [caniuse data for fromBase64](https://caniuse.com/?search=fromBase64), I'm not sure I'd make the transition yet, that's a VERY recent addition (sep 2025 for Chrome, for example) please don't make that change yet.

`Uint8Array.fromBase64()`: Chrome 2025.09 / Safari 2024.12 / Firefox 2024.11

That is a valid concern. We can mitigate this risk by using polyfill from the widely used [`core-js`](https://github.com/zloirock/core-js#ecmascript-typed-arrays). Polyfill allows us to adopt the standardized API while provide broad compatibility, and can be easily removed without code changes. Their main difference from a lib is that polyfill provides same API call signature as native one and fallback to native APIs whenever possible.

However, the polyfill way would(not a must) introduce build tools for optimal results. e.g. automatically injecting polyfill based on code usage and target browser versions. It provides a more optimized web application with a cost of a little more complexity to the build process.

My intention in mentioning these built-ins was to highlight their long-term potential: less external dependencies, less maintenance burden. Depending on the way we choose we get different levels of compatibility.

---

Separately, based on a brief wandering of the codebase and observations from my self-hosted Kanidm instance, I've noticed opportunities for improvements in the frontend area. It appears that other front-end assets—including dependencies(full bootstrap lib, but Kanidm only used small part of it), HTML, CSS, and even the logo—are not currently optimized for delivery to the browser. If you are happy with improvements on these aspects, I may propose more robust changes after I better understanding the code base.

(the logo, optimized with [SVGO](https://svgo.dev/)(commonly used SVG optimizer for web) defaults)
Screenshot 2025-10-24 at 10 44 15

Firstyear (approved):

Firstyear:
> Separately, based on a brief wandering of the codebase and observations from my self-hosted Kanidm instance, I've noticed opportunities for improvements in the frontend area. It appears that other front-end assets—including dependencies(full bootstrap lib, but Kanidm only used small part of it), HTML, CSS, and even the logo—are not currently optimized for delivery to the browser. If you are happy with improvements on these aspects, I may propose more robust changes after I better understanding the code base.

Sounds good to me :)

PR #3931, opened 2025-11-04
20251029 hmac name uniqueness

Firstyear:
# Change summary

This adds support for an HMAC name uniqueness feature, which prevents re-use of a previously used username in a privacy preserving manner to comply with GDPR when an account is deleted.

This is not a feature I expect to be commonly enabled. For that reason I prefer to leave it "undocumented" and we will advise individual installs when to enable it if needed. 

Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6607,8 +6607,9 @@ dependencies = [
 
 [[package]]
 name = "sshkeys"
-version = "0.3.3"
-source = "git+https://github.com/dnaeon/rust-sshkeys.git?rev=d736693769b9c4abebad8050fba92271f3c50226#d736693769b9c4abebad8050fba92271f3c50226"
+version = "0.3.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "43dd24cd9c70e02c48882a32b74e784d8f2aaddba2a3a30c403d5a6e416fa117"
 dependencies = [
  "base64 0.22.1",
  "byteorder",
@@ -7177,9 +7178,9 @@ dependencies = [
 
 [[package]]
 name = "tracing-forest"
-version = "0.2.0"
+version = "0.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "3298fe855716711a00474eceb89cc7dc254bbe67f6bc4afafdeec5f0c538771c"
+checksum = "92bdb3c949c9e81b71f78ba782f956b896019d82cc2f31025d21e04adab4d695"
 dependencies = [
  "smallvec",
  "thiserror 2.0.17",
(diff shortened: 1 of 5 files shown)


proto/src/internal/error.rs
```diff
@@ -157,6 +155,9 @@ pub enum OperationError {
     KG001TaskTimeout,
     KG002TaskCommFailure,
     KG003CacheClearFailed,
+    KG004UnknownFeatureUuid,
+    KG005HowDidYouEvenManageThis,
```
yaleman, line 159: My new favourite error.

proto/src/internal/error.rs
```diff
@@ -393,6 +396,7 @@ impl OperationError {
             Self::InvalidAccountState(val) => Some(format!("Invalid account state: {val}")),
             Self::MissingClass(val) => Some(format!("Missing class: {val}")),
             Self::MissingAttribute(val) => Some(format!("Missing attribute: {val}")),
+            Self::AttributeUniqueness => Some(format!("The value of some attributes is not unique.")),
```
yaleman:
```suggestion
            Self::AttributeUniqueness => Some(format!("The affected attributes are required to be unique.")),
```

Though it would be good if we told the user *which* attributes, which the old error did?

server/lib/src/plugins/hmac_name_unique.rs
```diff
@@ -0,0 +1,586 @@
...
+            trace!("Skipping hmac name history generation");
+            // should be IMPOSSIBLE to activate fixup from a lower domain level!!!
+            debug_assert!(false);
+            return Err(OperationError::KG005HowDidYouEvenManageThis);
+        }
+
+        let hmac_name_history_config_enabled = qs.get_feature_hmac_name_history_config().enabled;
+
+        if !hmac_name_history_config_enabled {
+            debug!("hmac name history not enabled");
```
yaleman:
```suggestion
            error!("HMAC name history not enabled, but fixup task was run, log this as a bug!");
```

proto/src/internal/error.rs
Firstyear: The old error didn't *really* tell you what attributes had a problem, and it doesn't tell you on what entry the problem exists. Let me see if I can do something here that would make some kind of sense ....

server/lib/src/plugins/hmac_name_unique.rs
```diff
@@ -0,0 +1,586 @@
...
+
+    Ok(())
+}
+
+impl HmacNameUnique {
+    #[instrument(level = "debug", name = "hmac_name_unique::fixup", skip_all)]
+    pub(crate) fn fixup(qs: &mut QueryServerWriteTransaction) -> Result<(), OperationError> {
+        let domain_level = qs.get_domain_version();
+        if domain_level < DOMAIN_LEVEL_12 {
+            trace!("Skipping hmac name history generation");
```
yaleman:
```suggestion
            error!("HMAC name history not enabled, but fixup task was run, please log this as a bug!");
```

Firstyear: Yeah, I noticed this too from your pyrevious comment, will fix :)

yaleman (approved):

PR #3933, opened 2025-11-05
20251005 multiple accept

### LINKED ISSUE #3923: Support binding on multiple addresses
### I did this

Searched documentation, looked at the code, observed that it's not currently an option. Also tried just specifying the port (`:8443`) which doesn't work.

### I expected the following

kanidmd should be configurable to bind to both IPv4 and IPv6 sockets.

### This happened instead

It can't.

### Kanidm version details

- Output of `kanidm(d) version`: kanidmd 1.7.4
- If not a container, how'd you install it: FreeBSD pkg
`FreeBSD kanidm 14.3-RELEASE FreeBSD 14.3-RELEASE-p5 GENERIC amd64`

Firstyear:
# Change summary

- Supporting binding to multiple addresses. This is useful on FreeBSD which has a separate ipv4/ipv6 stack, where linux has a ipv4 in 6 mapping feature. 

Fixes #3923

Checklist

### DIFF
--- a/examples/server.toml
+++ b/examples/server.toml
@@ -3,15 +3,19 @@ version = "2"
 
 #   The webserver bind address. Requires TLS certificates.
 #   If the port is set to 443 you may require the
-#   NET_BIND_SERVICE capability.
+#   NET_BIND_SERVICE capability. This accepts a single address
+#   or an array of addresses to listen on.
 #   Defaults to "127.0.0.1:8443"
 bindaddress = "[::]:443"
+# bindaddress = ["[::]:443", "0.0.0.0:443"]
 #
 #   The read-only ldap server bind address. Requires
-#   TLS certificates. If set to 636 you may require
-#   the NET_BIND_SERVICE capability.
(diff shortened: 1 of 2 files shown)


server/daemon/src/main.rs
```diff
@@ -988,7 +988,7 @@ async fn kanidm_main(config: Configuration, opt: KanidmdParser) -> ExitCode {
                     // the replace covers when you specify an ipv6-capable "all" address
                     format!(
                         "https://{}/status",
-                        config.address.replace("[::]", "localhost")
+                        config.address[0].replace("[::]", "localhost")
```
yaleman, line 991: This'll likely catch folks but otherwise ... it looks ok 😄

Firstyear: Yeah I wasn't sure what else to do - we enforce that there is at least on address so that's fine. But we have to make some assumption about the health check and in this case that it's the first address we attempt it on.

PR #3935, opened 2025-11-06
Add support for proxyv1

### LINKED ISSUE #3922: Support for proxy protocol v1

In #3542 support for proxy protocol **v2** was added. However, I use a reverse proxy (nginx) that only has support for v1. In particular, I want to proxy LDAP through nginx. 

Implement the v1 protocol and support it for HTTP and LDAP

- Stop proxying LDAP
- Switch all of my proxies from nginx to something else
- Stop using LDAP (I wish...)

### Additional context

I understand this is really a me issue, since the alternatives above could work. But I still think it would be nice to support. There was talk in #3423 about someone doing v1. Not sure if that's still what the project wants.

Firstyear:
# Change summary

- Add support for proxy protocol v1. Note that we discourage it's usage as it's MUCH slower to process compared to v2. 

Fixes #3922

Checklist

### DIFF
--- a/server/core/static/style.css
+++ b/server/core/static/style.css
@@ -1,9 +1,8 @@
 :root {
   --totp-width-and-height: 30px;
   --totp-stroke-width: 60px;
-  --navbar-colors-transition:
-    color 0.15s ease-in-out, background-color 0.15s ease-in-out,
-    border-color 0.15s ease-in-out;
+  --navbar-colors-transition: color 0.15s ease-in-out,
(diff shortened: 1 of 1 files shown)


yaleman (dismissed):
LGTM so far, we should do some comparison speed tests because wow, v1's gross to parse.

Firstyear:
> LGTM so far, we should do some comparison speed tests because wow, v1's gross to parse.

YEEERRRRPPPPPPPPPPPP that was my assessment too.

PR #3945, opened 2025-11-08
fix: use CredentialTag instead of strings all over the place

yaleman:
# Change summary

- shaving more yaks
- doesn't need a database migration because the underlying repr is still the same value in the database
- I'm sure this'll save whole bytes of memory at runtime, possibly even clock cycles!

Fixes #my need to kill strings

Checklist

### DIFF
--- a/server/lib/src/valueset/cred.rs
+++ b/server/lib/src/valueset/cred.rs
@@ -40,10 +40,9 @@ impl ValueSetCredential {
         let map = data
             .into_iter()
             .map(|dc| {
-                let t = dc.tag.clone();
                 Credential::try_from(dc.data)
                     .map_err(|()| OperationError::InvalidValueState)
-                    .map(|c| (t, c))
+                    .map(|c| (dc.tag, c))
             })
             .collect::<Result<_, _>>()?;
         Ok(Box::new(ValueSetCredential { map }))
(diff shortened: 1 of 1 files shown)


Firstyear:
> doesn't need a database migration because the underlying repr is still the same value in the database

Ahhh. See uhhh I was actually suggesting you remove the tag wholesale because it's not used at all. We never have different tags in the one attr, we have separate attributes for primaryCred and unixCred. That's why a database migration would be needed.

yaleman:
> > doesn't need a database migration because the underlying repr is still the same value in the database
> 
> Ahhh. See uhhh I was actually suggesting you remove the tag wholesale because it's not used at all. We never have different tags in the one attr, we have separate attributes for primaryCred and unixCred. That's why a database migration would be needed.

That would have… been handy to have realised.

Firstyear:
> > > doesn't need a database migration because the underlying repr is still the same value in the database
> > 
> > 
> > Ahhh. See uhhh I was actually suggesting you remove the tag wholesale because it's not used at all. We never have different tags in the one attr, we have separate attributes for primaryCred and unixCred. That's why a database migration would be needed.
> 
> That would have… been handy to have realised.

I'm sorry :( :( :(

PR #3954, opened 2025-11-13
Make log messages more verbose for issues with resources server

### LINKED ISSUE #3950: Make error message more clear for "Invalid OAuth2 redirect_uri"

When setting an redirect_uri, one might make a typo or uses a malformed uri.
In such a case one will receive a warning like this:
`🚧 [warn]: Invalid OAuth2 redirect_uri (must be an exact match to a redirect-url) - got https://example.com/oidc/endpoint/`
For me it wasn't clear that `got https://example.com/oidc/endpoint/` meant that the resource server send this uri as an reply and that hence a mismatch was detected.

Make the warning more verbose like `got https://example.com/oidc/endpoint/ from resource server but configured uri https://example.com/oidc/endpoint does not match`.

A wiki entry about this issue.

scaredmushroom:
Draft pr to discuss possible wording.

# Change summary

- extends log messages with more information on what happened and where to look
- make punctuation more coherent 

Fixes #3950 

Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -218,7 +218,7 @@ dependencies = [
  "rustc-hash 2.1.1",
  "serde",
  "serde_derive",
- "syn 2.0.109",
+ "syn 2.0.110",
 ]
 
 [[package]]
@@ -254,7 +254,7 @@ checksum = "34921de3d57974069bad483fdfe0ec65d88c4ff892edd1ab4d8b03be0dda1b9b"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.109",
+ "syn 2.0.110",
 ]
 
 [[package]]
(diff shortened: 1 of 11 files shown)


server/lib/src/idm/oauth2.rs
```diff
@@ -2046,7 +2046,7 @@ impl IdmServerProxyReadTransaction<'_> {
                 && !opaque_origin_matched
             {
                 admin_warn!(
-                    "Invalid OAuth2 redirect_uri scheme (must be https for secure origin) - got {}",
+                    "Invalid OAuth2 redirect_uri scheme (must be https for secure origin) - got {} from resource server, but it has to use https",
```
yaleman: Saying it twice seems unnecessary? 

"(must be https for secure origin)" is already in the message.

scaredmushroom: That's true. I did this as my fallacy was that my head just drop everything in front of the dash as "did't got it, what does the verbose part say?" (for another issue).
What about "[…] - got {} from resource server instead"? This brings both parts of the sentence more together.

Firstyear: Yeah that's fine :) I think even if not perfect, small changes like this are good.

PR #3972, opened 2025-11-19
Allow disabling OAuth2 consent prompt

### LINKED ISSUE #3961: Ability to disable the OAuth2 consent page

In enterprise environments the consent is often implied and the consent prompt is confusing to users accustomed to solutions like Okta.

Ability to disable the consent prompt, ideally per configured client. And/or have a programmatic way of setting the consent attribute.

This could also be limited to clients that don't request any sensitive scopes/claims.

- do nothing, explain to users why this is nice even in such an environment
- implement OIDC auth request `prompt` parameter - specifically `none` which would abort if the consent is not given already and never show the prompt - but that's a bit broken without a way to disable/mark consent automatically

### Additional context

Already discussed in https://github.com/kanidm/kanidm/issues/636 and much improved in https://github.com/kanidm/kanidm/pull/824.

As far as I know, in the OAuth2 spec consent is not mandatory, there is only (not a standard) [RFC-6819](https://www.rfc-editor.org/rfc/rfc6819#section-5.2.3.2):

> **5.2.3.2.  Require User Consent for Public Clients without Secret**
>
>   Authorization servers should not allow automatic authorization for
>   public clients.  The authorization server may issue an individual
>   client id but should require that all authorizations are approved by
>   the end user.  For clients without secrets, this is a countermeasure
>   against the following threat:
>
>   o  Impersonation of public client applications.

Still, it definitely makes sense to keep the consent requirement for public clients.

For OIDC the consent is technically required, but can be done out-of-band - [OpenID Connect Core 1.0](https://openid.net/specs/openid-connect-core-1_0.html) (emphasis mine):

> **3.1.2.4.  Authorization Server Obtains End-User Consent/Authorization**
> 
> Once the End-User is authenticated, the Authorization Server MUST obtain an authorization decision before releasing information to the Relying Party. When permitted by the request parameters used, this MAY be done through an interactive dialogue with the End-User that makes it clear what is being consented to **or by establishing consent via conditions for processing the request or other means (for example, via previous administrative consent)**. Sections [2](https://openid.net/specs/openid-connect-core-1_0.html#IDToken) and [5.3](https://openid.net/specs/openid-connect-core-1_0.html#UserInfo) describe information release mechanisms.

Allowing the `OAuth2ConsentScopeMap` attribute to be set with the API is probably the easiest since that means I can add the necessary consents while provisioning the user account.

I'd be happy to PR either solution if this is something you think makes sense.

maciek134:
# Change summary

- Allow disabling the OAuth2 consent prompt

Fixes #3961

Checklist

- [x] book chapter included (if relevant) - section, not chapter

Early PR to get some feedback. I'll add a few sentences in the book later.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,6 +68,12 @@ Identity management is often abbreviated to 'idm', and is a common industry term
 
 Kanidm is pronounced as "kar - nee - dee - em".
 
+## Kanidm Anthem
+
+> An anthem is a popular song, especially a rock song felt to sum up the attitudes or feelings associated with a period or social group.
+
+The Kanidm anthem is [Crab Rave - Noisestorm](https://www.youtube.com/watch?v=LDU_Txk06tM)
+
 ## Comparison with other services
 
 <details> <summary>LLDAP</summary>
--- a/proto/src/internal/token.rs
+++ b/proto/src/internal/token.rs
@@ -91,6 +91,8 @@ impl UserAuthToken {
     pub fn purpose_readwrite_active(&self, ct: time::OffsetDateTime) -> bool {
         match self.purpose {
             UatPurpose::ReadWrite { expiry: Some(exp) } => ct < exp,
+            // ReadWrite { None } shows the session is priv capable, but the
+            // privs aren't active *now*.
             _ => false,
         }
     }
--- a/server/core/src/https/middleware/mod.rs
+++ b/server/core/src/https/middleware/mod.rs
@@ -129,8 +129,11 @@ async fn ip_address_middleware_inner(
         .map(|range| range.contains(&connection_ip_addr))
         .unwrap_or_default();
 
+    let maybe_x_forward_for = request.headers().get(X_FORWARDED_FOR_HEADER);
+
     let client_ip_addr = if trust_x_forward_for {
-        if let Some(x_forward_for) = request.headers().get(X_FORWARDED_FOR_HEADER) {
+        if let Some(x_forward_for) = maybe_x_forward_for {
+            debug!("processing {} header", X_FORWARDED_FOR);
             // X forward for may be comma separated.
             let first = x_forward_for
                 .to_str()
@@ -138,22 +141,31 @@ async fn ip_address_middleware_inner(
                     // Split on an optional comma, return the first result.
                     s.split(',').next().unwrap_or(s))
                 .map_err(|_| {
+                    error!("{} contains invalid data structure", X_FORWARDED_FOR);
                     (
                         StatusCode::BAD_REQUEST,
-                        "X-Forwarded-For contains invalid data",
+                        "x-forwarded-for contains invalid data structure",
                     )
                 })?;
 
             first.parse::<IpAddr>().map_err(|_| {
+                error!("{} contains invalid ip address", X_FORWARDED_FOR);
                 (
                     StatusCode::BAD_REQUEST,
-                    "X-Forwarded-For contains invalid ip addr",
+                    "X-Forwarded-For contains invalid ip address",
                 )
             })?
         } else {
+            debug!(
+                "{} header not present from a trusted connection",
+                X_FORWARDED_FOR
+            );
             client_ip_addr
         }
     } else {
+        if maybe_x_forward_for.is_some() {
+            debug!("Ignoring {} from untrusted connection", X_FORWARDED_FOR);
+        }
         // This can either be the client_addr == connection_addr if there are
         // no ip address trust sources, or this is the value as reported by
         // proxy protocol header. If the proxy protocol header is used, then
--- a/server/core/src/https/trace.rs
+++ b/server/core/src/https/trace.rs
@@ -21,6 +21,7 @@ impl<B> tower_http::trace::MakeSpan<B> for DefaultMakeSpanKanidmd {
             uri = %request.uri(),
             version = ?request.version(),
             kopid = tracing::field::Empty, // filled in later
+            connection_address = tracing::field::Empty, // filled in later
             client_address = tracing::field::Empty, // filled in later
             status_code = tracing::field::Empty, // filled in later
             latency = tracing::field::Empty, // filled in later
@@ -35,8 +36,12 @@ impl<B> OnRequest<B> for DefaultOnRequestKanidmd {
     fn on_request(&mut self, request: &axum::http::Request<B>, span: &Span) {
         if let Some(client_conn_info) = request.extensions().get::<crate::https::ClientConnInfo>() {
             span.record(
-                "client_address",
+                "connection_address",
                 client_conn_info.connection_addr.to_string(),
+            );
+            span.record(
+                "client_address",
+                client_conn_info.client_ip_addr.to_string(),
             );
         };
     }
@@ -55,14 +60,15 @@ impl<B> tower_http::trace::OnResponse<B> for DefaultOnResponseKanidmd {
         if let Some(client_conn_info) = response.extensions().get::<crate::https::ClientConnInfo>()
         {
             span.record(
-                "connection_addr",
+                "connection_address",
                 client_conn_info.connection_addr.to_string(),
             );
             span.record(
-                "client_ip_addr",
+                "client_address",
                 client_conn_info.client_ip_addr.to_string(),
             );
         };
+
         let kopid = match response.headers().get(KOPID) {
             Some(val) => val.to_str().unwrap_or("<invalid kopid>"),
             None => "<unknown>",
--- a/server/core/src/tcp.rs
+++ b/server/core/src/tcp.rs
@@ -20,6 +20,7 @@ pub(crate) async fn process_client_addr(
                 .iter()
                 .any(|ip_cidr| ip_cidr.contains(&canonical_conn_addr)) =>
         {
+            debug!("processing proxy-v2 header");
             timeout(time_limit, ProxyHdrV2::parse_from_read(stream))
                 .await
                 .map(|ok_result| ok_result.map(|(stream, hdr)| (stream, hdr.to_remote_addr())))
@@ -29,12 +30,18 @@ pub(crate) async fn process_client_addr(
                 .iter()
                 .any(|ip_cidr| ip_cidr.contains(&canonical_conn_addr)) =>
         {
+            debug!("processing proxy-v1 header");
             timeout(time_limit, ProxyHdrV1::parse_from_read(stream))
                 .await
                 .map(|ok_result| ok_result.map(|(stream, hdr)| (stream, hdr.to_remote_addr())))
         }
-        TcpAddressInfo::ProxyV2(_) | TcpAddressInfo::ProxyV1(_) | TcpAddressInfo::None => {
-            return Ok((stream, connection_addr))
+        TcpAddressInfo::ProxyV2(_) | TcpAddressInfo::ProxyV1(_) => {
+            debug!("Ignoring proxy headers from untrusted connection");
+            return Ok((stream, connection_addr));
+        }
+        TcpAddressInfo::None => {
+            // Not configured, ignore
+            return Ok((stream, connection_addr));
         }
     };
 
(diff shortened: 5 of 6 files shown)


libs/client/src/oauth.rs
```diff
@@ -488,4 +488,28 @@ impl KanidmClient {
...
+        let mut update_oauth2_rs = Entry {
+            attrs: BTreeMap::new(),
+        };
+        update_oauth2_rs.attrs.insert(
+            ATTR_OAUTH2_DISABLE_CONSENT_PROMPT.to_string(),
+            vec!["false".to_string()],
+        );
+        self.perform_patch_request(format!("/v1/oauth2/{}", id).as_str(), update_oauth2_rs)
+            .await
+    }
```
maciek134: could be
```suggestion
    pub async fn idm_oauth2_rs_set_consent_prompt_disabled(&self, id: &str, value: bool) -> Result<(), ClientError> {
        let mut update_oauth2_rs = Entry {
            attrs: BTreeMap::new(),
        };
        update_oauth2_rs.attrs.insert(
            ATTR_OAUTH2_DISABLE_CONSENT_PROMPT.to_string(),
            vec![value.to_string()],
        );
        self.perform_patch_request(format!("/v1/oauth2/{}", id).as_str(), update_oauth2_rs)
            .await
    }
```
but I wasn't sure which way to follow based on other code

server/lib/src/migration_data/dl13/schema.rs
```diff
@@ -1128,6 +1136,7 @@ pub static ref SCHEMA_CLASS_OAUTH2_RS_DL9: SchemaClass = SchemaClass {
         Attribute::OAuth2RsOrigin,
         Attribute::OAuth2StrictRedirectUri,
         Attribute::OAuth2DeviceFlowEnable,
+        Attribute::OAuth2DisableConsentPrompt,
```
maciek134: `Attribute::OAuth2DisableConsentPrompt` or `Attribute::OAuth2ConsentPromptDisable`? device flow does the latter, while pkce does it the same as here

maciek134: Also I'm not sure I was supposed to edit this class, it ofc worked on a dev setup with a clean db, but I noticed the `_DL9` suffix after making the PR

server/lib/src/migration_data/dl13/access.rs
maciek134, line 1: Went with what one of the older PRs did, I think this is correct?

I was wondering whether I can disallow disabling this on public clients (as per [RFC-6819](https://www.rfc-editor.org/rfc/rfc6819#section-5.2.3.2)) by messing with these or classes? This is where I'd need some pointers either way

yaleman:
This'd be a good opportunity for you to document the schema migration modules and get @maciek134's feedback @Firstyear 😇

server/lib/src/migration_data/dl13/schema.rs
yaleman: I'd definitely prefer `Attribute::OAuth2ConsentPromptDisable` as it uses the `thing -> sub-thing -> action` I'm encouraging

Firstyear:
yeah I'll review this in the coming hours :)

server/lib/src/migration_data/dl13/schema.rs
Firstyear: No, boolean options always need to be in the positive as "disable: true" is confusing. It should be OAuth2ConsentPromptEnable, and if unset defaults to 'true'.

server/lib/src/idm/oauth2.rs
```diff
@@ -2203,7 +2210,7 @@ impl IdmServerProxyReadTransaction<'_> {
 
         let session_id = ident.get_session_id();
 
-        if consent_previously_granted {
+        if consent_previously_granted || o2rs.consent_prompt_disabled {
```
yaleman: You'll need to update the debug message slightly below this to reflect the action. Something like 

```
                debug!(
                    scopes=pretty_scopes.join(","),
                    prompt_disabled=o2rs.consent_prompt_disabled,
                    previously_granted=consent_previously_granted,
                    "Consent flow passed"
                );

File: server/lib/src/migration_data/dl13/schema.rs (hunk 2 above)
Comment (maciek134) [reply server/lib/src/migration_data/dl13/schema.rs]: yeah I was wondering about the double negative, `OAuth2ConsentPromptEnable` it is

File: server/lib/src/idm/oauth2.rs [hunk 4]
```diff
@@ -874,6 +875,11 @@ impl Oauth2ResourceServersWriteTransaction<'_> {
                         }
                         false => None,
                     };
+
+                let consent_prompt_enabled = ent
+                    .get_ava_single_bool(Attribute::OAuth2ConsentPromptEnable)
+                    .unwrap_or(true);
```
Firstyear: To answer your question re consent on public vs confidential clients: you can match on `type_` which tells you if the client is public. Then we allow this value to only go false on Basic clients.  

`let consent_prompt_enable == matches!(type_, OAuth2RSType::Public) || ent.get_ava .... `

maciek134: Ah, I was thinking of not even allowing people to disable it for the public client - to avoid a situation where someone didn't read the docs and is wondering why their public client is still asking for consent after they disabled it.

After having another look at the acps the best I can come up with is adding another `BuiltinAcp` with a `EntryClass::OAuth2ResourceServerBasic` target and only allowing that one to modify `Attribute::OAuth2ConsentPromptEnable`?

Firstyear: It's probably safer to do both approaches - ensure that we force it to be enabled on public clients, but also the access control to prevent it being modified on them. That way if a user in future manages to work-around the access, then we have a defence.

maciek134:
Ok, should be ready to review now. I've added the acp and moved the consent prompt bool to `OauthRSType::Basic`. Not sure if the integration test is in the right place, but everything seems to work.

I have to say this codebase is a pleasure to work with

server/lib/src/idm/oauth2.rs
```diff
@@ -366,8 +366,9 @@ enum OauthRSType {
     Basic {
         authz_secret: String,
         enable_pkce: bool,
+        enable_consent_prompt: bool,
```
maciek134, line 369: This became `enable_x` to be consistent with `pkce` - lmk if you prefer `consent_prompt_enable` to match the attribute

server/lib/src/constants/uuids.rs
```diff
@@ -510,6 +512,7 @@ pub const UUID_IDM_ACP_MESSAGE_MANAGE: Uuid = uuid!("00000000-0000-0000-0000-fff
 pub const UUID_IDM_ACP_MESSAGE_SENDER: Uuid = uuid!("00000000-0000-0000-0000-ffffff000078");
 pub const UUID_IDM_ACP_OAUTH2_CLIENT_ADMIN: Uuid = uuid!("00000000-0000-0000-0000-ffffff000079");
 pub const UUID_IDM_ACP_OAUTH2_ACCOUNT_ENROL: Uuid = uuid!("00000000-0000-0000-0000-ffffff000080");
+pub const UUID_IDM_ACP_OAUTH2_MANAGE_BASIC_V1: Uuid = uuid!("00000000-0000-0000-0000-ffffff000081");
```
maciek134: No idea if `_V1` is needed, but the original acp had it.

server/lib/src/migration_data/dl13/access.rs
```diff
@@ -724,6 +724,32 @@ lazy_static! {
     };
 }
 
+lazy_static! {
+    pub static ref IDM_ACP_OAUTH2_MANAGE_BASIC: BuiltinAcp = BuiltinAcp {
```
maciek134, line 728: Should this have a `_DL13` suffix?

Firstyear:
> I have to say this codebase is a pleasure to work with

I'm really glad to hear that - I think there is a lot of "evolution" that has gone on that causes some rough edges around the place, but it's good to hear that you had a good experience 😅

server/lib/src/idm/oauth2.rs
Firstyear: Sounds good!

server/lib/src/constants/uuids.rs
Firstyear: Not needed anymore, it's some legacy/techdebt from old ideas that never came about, so you don't need the v1

server/lib/src/migration_data/dl13/access.rs
Firstyear: Not needed, again that's a remnant of "how we used to do things", and it's no longer required.

server/lib/src/migration_data/dl13/access.rs
```diff
@@ -724,6 +724,32 @@ lazy_static! {
...
+        classes: vec![
+            EntryClass::Object,
+            EntryClass::AccessControlProfile,
+            EntryClass::AccessControlCreate,
```
Firstyear: You don't need AccessControlCreate OR AccessControlDelete here.

server/lib/src/migration_data/dl13/access.rs
```diff
@@ -724,6 +724,32 @@ lazy_static! {
...
+        description: "Builtin IDM Control for managing Basic OAuth2 resource server integrations.",
+        receiver: BuiltinAcpReceiver::Group(vec![UUID_IDM_OAUTH2_ADMINS]),
+        target: BuiltinAcpTarget::Filter(ProtoFilter::And(vec![
+            match_class_filter!(EntryClass::OAuth2ResourceServerBasic),
+            FILTER_ANDNOT_TOMBSTONE_OR_RECYCLED.clone(),
+        ])),
+        search_attrs: vec![Attribute::OAuth2ConsentPromptEnable,],
+        modify_removed_attrs: vec![Attribute::OAuth2ConsentPromptEnable,],
+        modify_present_attrs: vec![Attribute::OAuth2ConsentPromptEnable,],
+        create_attrs: vec![Attribute::OAuth2ConsentPromptEnable,],
```
Firstyear: You don't need create_attrs here.

Firstyear (commented):
Two minor changes regarding the access control, but looking really good.

server/lib/src/migration_data/dl13/access.rs
maciek134: Ah right

server/lib/src/migration_data/dl13/access.rs
maciek134: This one is a bit confusing, I'll have to take a closer look at how the profiles are used

Firstyear: create_attrs define the attributes you can create while making an entry. There are a number of other subtelties too. 

Since we aren't setting the consent-prompt-enable during a new oauth2 client creation, we don't need this here. (Similar that's why you don't need AccessControlCreate)

the "AccessControlDelete" means that if you match the target you can delete the whole entry. Again, that's not the intent here, so it's not needed.

All that's needed is AccessControlSearch (so that you can see the consent prompt enable) and AccessControlModify (so you can assert a value is present on the entry, or assert the value is removed on the entry).

The present attrs defines what you can update as attributes. If you have present you can add a value, or assert a value already is present.

removed means you can remove a value, or assert it's already not present.

If you only have "present" the attribute becomes "append only". If you only have remove the attribute becomes "pop only". If you have both present and remove you can do an update in place etc.

server/lib/src/constants/uuids.rs
maciek134: Forgot about this one, fixed now

Firstyear (approved):
Great work! Thank you so much!

maciek134:
Soo I never asked and it's not in the book - who resolves discussions? Asking since you enabled auto-merge but it's complaining about unresolved comments and I default to letting reviewers resolve them.

Firstyear:
Ahhh normally we leave them for the PR implementer to resolve so you can treat them as a "checklist" of items to follow up on.

Firstyear:
Went ahead and resolved them for you :)

PR #3975, opened 2025-11-21
lib crypto should not depend on proto

### LINKED ISSUE #3971: `unwrap()` in build_profiles causes osc builds to fail
https://github.com/kanidm/kanidm/blob/c7a2da76c3ad3173530e8359f3f57638857160f3/libs/profiles/build.rs#L16

When building within an osc chroot, the `cargo locate-project --workspace --message-format=plain` command fails. There is no error handling in place here, and instead the `unwrap()` on that line is causing the Himmelblau build to fail.

### Kanidm version details

 kanidm_build_profiles v1.7.4

Firstyear:
# Change summary

- lib crypto had a dependency added on kanidm_proto, which caused build profiles to be pulled into projects (like himmelblau) that did not require them. Since lib crypto should be isolated, we remove the dep. The dep was added so that a single "from" impl could be added. In this case, the obscurity of the ".into()" call in the server masked the error that was being raised, and it's clearer to just raise OperationError::CryptographyError at the relevant locations. 

Fixes #3971

Checklist

### DIFF
--- a/server/lib/src/credential/mod.rs
+++ b/server/lib/src/credential/mod.rs
@@ -595,7 +595,7 @@ impl Credential {
         self.password_ref().and_then(|pw| {
             pw.verify(cleartext).map_err(|e| {
                 error!(crypto_err = ?e);
-                e.into()
+                OperationError::CryptographyError
             })
         })
     }


yaleman (approved):
winning

PR #3977, opened 2025-11-22
Changing how we parse environment variables in kanidmd

yaleman:
# Change summary

- move from using a manual parser to just using clap, so (most) things can be specified on the command line/env vars now.

Fixes #

Checklist

### DIFF
--- a/proto/src/cli.rs
+++ b/proto/src/cli.rs
@@ -144,6 +144,7 @@ pub struct KanidmdCli {
 
     #[clap(
         global = true,
+        hide = true,
         env = "KANIDM_TRUST_X_FORWARDED_FOR",
         help = "Whether to blindly trust the X-Forwarded-For header, regardless of source IP"
     )]


proto/src/cli.rs
```diff
@@ -39,3 +41,156 @@ impl OpType {
...
+    pub bindaddress: Option<String>,
+
+    #[clap(
+        env = "KANIDM_LDAPBINDADDRESS",
+        global = true,
+        help = "Specify the LDAP bind address(es)"
+    )]
+    pub ldapbindaddress: Option<String>,
+
+    #[clap(
```
Firstyear, line 145: I think we need this for compat with the older env var config, but we probably should at least hide this one because it's not a great option security wise?

yaleman: Sure, can do

yaleman:
> I think the only concern I'd have here is that the clap env var passing is a bit limited, and can't fully replace our configuration, but that's true of the current env var setup too I guess.

Not much we can do about it at this point given how ... convoluted and branchy the config's gotten, but that env parsing thing was just a noisy mess.

Firstyear (approved):

PR #3993, opened 2025-11-28
kanidm 1.8.3 does not recognize gzip as a compression algorithm

jchonig:
The example server.toml specifies `gzip` while kanidm is expecting `Gzip`.

```
Nov 28 10:20:52 nixie docker[1172080]: kanidm        | ERROR: Configuration Parse Failure: Custom { kind: InvalidData, error: Error { message: "unknown variant `gzip`, expected `NoCompression` or `Gzip`", input: ... keys: ["online_backup", "compression"], span: Some(5516..5522) } }
```

# Change summary

- Change example to specify Gzip

Fixes #

Checklist

### DIFF
--- a/examples/server.toml
+++ b/examples/server.toml
@@ -139,4 +139,4 @@ schedule = "00 22 * * *"
 # versions = 7
 
 # Compression can either be "gzip" or "nocompression", will compress the backups on disk if specified.
-compression = "gzip"
+compression = "Gzip"


ToxicMushroom:
I think the appropriate fix would be for compression to use the From<String> impl https://github.com/kanidm/kanidm/blob/b805954b0d872ad47f3f8398b86b188e6eb79348/proto/src/backup.rs#L52. This function is case insensitive so this would not break for people who would be currently working around this issue I think.

I believe it simply misses a `#[serde(from = "String")]` annotation on the enum, but I could be wrong.
Feel free to write a test at the bottom of this file if you'd like :).

yaleman:
The fix for this will be the config parsing, not changing the example - because it should work regardless.

yaleman:
Replaced by #3995

PR #3997, opened 2025-12-01
feat: bind mount home strategy

### LINKED ISSUE #3970: configure or use mount --bind /u/guid /home_alias
We have several Ubuntu systems that we would also like to use with kanidm. I’ve read about the issue with snap packages.
https://kanidm.github.io/kanidm/stable/frequently_asked_questions.html?highlight=snap#why-arent-snaps-launching-with-home_alias-set

We want to keep `home_attr` as uuid. The linked bug report on Launchpad https://bugs.launchpad.net/apparmor/+bug/1485055 suggests an alternative using `mount --bind` instead of symlinks. This seems to work for our situation. Using the alias doesn't currently work.

Is it possible to configure kanidm to use a bind mount instead of a symlink, or what other options do I have?

sk1p:
# Change summary

- Add a `BindMount` home strategy, which is useful for example for [snaps](https://kanidm.github.io/kanidm/stable/frequently_asked_questions.html?highlight=snap#why-arent-snaps-launching-with-home_alias-set) on Ubuntu

~~On top of #3985.~~ Rebased on top of current master.

## TODO

This is still WIP, and needs:

- [ ] Edge cases, like how to handle things that exist at the target path and are not a directory (or symlink)? Log an error and continue?
- [ ] Support for non-Linux systems? I've added dependencies on `procfs` and `nix`, and I'm not sure if that aligns with the supported platforms
- [ ] Documentation updates
- [ ] Changes in the systemd services sandboxing - the current settings prevent the mount syscall from working because of capabilities and syscall filtering, and hides the mounts from the hosts system by mount namespacing. It might make sense to have different sandboxing, to keep the security level of the default configuration as high as possible
- [ ] Test strategy - are tests executed as root somewhere in the CI pipeline?

Fixes #3970 

Checklist

/cc @ReimarBauer @Firstyear


sk1p:
As a comment on current status: @ReimarBauer has made test deployments on both Debian and Ubuntu test systems, and the general idea seems to work. We are still trying to figure out the systemd-related changes needed, especially exactly which sandboxing flags can stay enabled in this configuration.

unix_integration/resolver/src/bin/kanidm_unixd_tasks.rs
```diff
@@ -185,98 +202,178 @@ fn create_home_directory(
...
 
-                // Rename is atomic within the same directory; no disappearance window.
-                if let Err(e) = fs::rename(&tmp, &alias_path) {
-                    error!(err=?e, from=?tmp, to=?alias_path, "Unable to atomically replace alias symlink");
-                    // Cleanup temp on failure.
-                    let _ = fs::remove_file(&tmp);
-                    return Err(format!("{e:?}"));
-                }
+    // Finally, try to create the bind mount
+    nix::mount::mount::<Path, Path, str, str>(
```
Firstyear, line 296: This will fail on every other distro except nix, so we need to find a way to do this that is generic. It is very likely we will need to use https://docs.rs/libc/0.2.177/libc/fn.mount.html here instead.

unix_integration/resolver/src/bin/kanidm_unixd_tasks.rs
```diff
@@ -185,98 +202,178 @@ fn create_home_directory(
...
+            return Err(format!("{e:?}"));
+        }
+    }
 
-                if let Err(e) = symlink(&hd_mount_path, &tmp) {
-                    error!(err=?e, ?tmp, "Unable to create temporary alias symlink");
-                    return Err(format!("{e:?}"));
-                }
+    // If mount point exists and is already mounted, we are done
+    if mount_exists_at(alias_path)? {
```
Firstyear: In otherwords here, we don't just need to check "is something at the alias path" - we need to check it's *this users home dir* is mounted to *this* alias path. If it's a mount of something else we have to unmount it first before we can proceed.

Firstyear (changes requested):
Thanks, it's a good start, some feedback inline :)

unix_integration/resolver/src/bin/kanidm_unixd_tasks.rs
sk1p: Oh, the name might be a bit misleading - `nix` is a safe alternative to `libc`, it should work just as well. It's not related to the nix distro.

Firstyear: ahhh! That certainly clears that up then!

unix_integration/resolver/src/bin/kanidm_unixd_tasks.rs
sk1p: This is now implemented.

unix_integration/resolver/src/bin/kanidm_unixd_tasks.rs
```diff
@@ -22,14 +22,17 @@ use kanidm_unix_common::unix_proto::{
 };
 use kanidm_utils_users::{get_effective_gid, get_effective_uid};
 use libc::{lchown, umask};
+use nix::mount::MsFlags;
```
Firstyear:
```
error[E0433]: failed to resolve: use of unresolved module or unlinked crate `nix`
  --> unix_integration/resolver/src/bin/kanidm_unixd_tasks.rs:25:5
   |
25 | use nix::mount::MsFlags;
   |     ^^^ use of unresolved module or unlinked crate `nix`
   |
   = help: if you wanted to use a crate named `nix`, use `cargo add nix` to add it to your `Cargo.toml`

error[E0433]: failed to resolve: use of unresolved module or unlinked crate `procfs`
  --> unix_integration/resolver/src/bin/kanidm_unixd_tasks.rs:30:5
   |
30 | use procfs::process::Process;
   |     ^^^^^^ use of unresolved module or unlinked crate `procfs`
   |
   = help: if you wanted to use a crate named `procfs`, use `cargo add procfs` to add it to your `Cargo.toml`

error[E0433]: failed to resolve: use of unresolved module or unlinked crate `nix`
   --> unix_integration/resolver/src/bin/kanidm_unixd_tasks.rs:261:9
    |
261 |         nix::mount::umount(&m.mount_point).map_err(|e| {
    |         ^^^ use of unresolved module or unlinked crate `nix`
    |
    = help: if you wanted to use a crate named `nix`, use `cargo add nix` to add it to your `Cargo.toml`

error[E0433]: failed to resolve: use of unresolved module or unlinked crate `nix`
   --> unix_integration/resolver/src/bin/kanidm_unixd_tasks.rs:278:5
    |
278 |     nix::mount::mount::<Path, Path, str, str>(
    |     ^^^ use of unresolved module or unlinked crate `nix`
    |
    = help: if you wanted to use a crate named `nix`, use `cargo add nix` to add it to your `Cargo.toml`

For more information about this error, try `rustc --explain E0433`.
error: could not compile `kanidm_unix_int` (bin "kanidm_unixd_tasks") due to 4 previous errors
```

Errors on non-linux platforms, so I think we need to cfg flag this.

unix_integration/resolver/src/bin/kanidm_unixd_tasks.rs
```diff
@@ -222,9 +225,70 @@ fn create_home_directory(
 
     match home_strategy {
         HomeStrategy::Symlink => home_alias_update_symlink(&alias_path, &hd_mount_path),
+        HomeStrategy::BindMount => home_alias_update_bind_mount(&alias_path, &hd_mount_path),
     }
 }
 
+fn home_alias_update_bind_mount(alias_path: &Path, hd_mount_path: &Path) -> Result<(), String> {
+    // If the alias_path is a symlink, remove it
+    // TODO: what if it exsits and is not a symlink?
```
Firstyear: Then we don't touch it, because I don't want to risk damage to someones data. It should be up to a human to intervene.

Firstyear (commented):
Okay this is looking pretty good.

Sadly on the testing front, we do these tests a bit more manually, maybe we shuold look at some CI, or something "else" to make a framework here that we can use to test. Depends how far into this we want to go. We could do an abstraction over the mount / file tools, and then make a trait for them that we can shim out/in and have a mock version. 

Either way this is looking great, I think it's just some polish and testing and we're there. I'll give this a go later on a linux system.

unix_integration/resolver/src/bin/kanidm_unixd_tasks.rs
sk1p: Which platforms are supported? Should I test this on e.g. FreeBSD, or at least get it to compile there? Probably on these platforms, we should err out early as IIUC there is no bind mount functionality available on non-Linux platforms.

sk1p:
> Sadly on the testing front, we do these tests a bit more manually, maybe we shuold look at some CI, or something "else" to make a framework here that we can use to test. Depends how far into this we want to go. 

Okay. I tried to set up a manual test environment before Christmas but ran out of time. I think I didn't yet fully nail the configuration, as there was somehow a mismatch between the directory used for $HOME and the path sent to the kanidm UNIX daemon. I hope to work on it again next week.

> We could do an abstraction over the mount / file tools, and then make a trait for them that we can shim out/in and have a mock version.

Indeed, a trait for a mount backend of some kind is how I implemented this for a different project. I'll think a bit if I can come up with a "quick win" on the CI side of things. 

> Either way this is looking great, I think it's just some polish and testing and we're there. I'll give this a go later on a linux system.

Sounds good! For deployment, we/I still need to nail down the systemd service configuration, especially with regards to sandboxing etc. Do you prefer to have multiple smaller PRs with changes to documentation and deployment/service definitions, or should this be part of this PR?

Also, is it OK if I rebase my branch, or do you prefer merge commits? Thanks!

yaleman:
Definitely rebase, we squash on merge anyway.

Firstyear:
@sk1p Could you rebase / update this PR so we can consider it for merge? Or would you like us to finish it up?

sk1p:
> Could you rebase / update this PR so we can consider it for merge? Or would you like us to finish it up?

I rebased, but please feel free to take it over, as I still didn't have time to set up a proper test environment or continue with @reimarbauer 's deployment. Thanks!

ReimarBauer:
We're currently using a test environment based on Debian and Ubuntu (2 machines) within our institute’s systems, built from a previous iteration of this project. This institute is separate from @sk1p’s area of work.  I don't have anyone on my team who can assist with RUST, so I really appreciate @sk1p taking the initiative. We'd much rather deploy an official package than apply the patch manually.
I think this option should be configurable, not the default. The existing solution is a good starting point for others, so I'd appreciate it if you @yaleman @Firstyear could take over and merge this.

Firstyear:
No problem, I'll rebase and wrap this up for you soon then.

Firstyear:
Okay, I worked out the needed systemd settings to allow this. It seems that most of systemds protections interfer with "mounting" so as a result, we have to disabled pretty much all their hardening. Disappointing, but that's how it is.

unix_integration/common/src/unix_config.rs
```diff
@@ -570,6 +574,13 @@ impl UnixdConfig {
                 .unwrap_or(self.home_alias),
             home_strategy: match config.home_strategy {
                 HomeStrategyV2::Symlink => HomeStrategy::Symlink,
+                #[cfg(target_os = "linux")]
+                HomeStrategyV2::BindMount => HomeStrategy::BindMount,
+                #[cfg(not(target_os = "linux"))]
+                HomeStrategyV2::BindMount => {
+                    warn!("home_strateg: bind_mount not supported on this platform - falling back to symlink!");
```
yaleman:
```suggestion
                    warn!("Bind mounts not supported on this operating system - falling back to symlink for home directories");
```

unix_integration/resolver/debian/kanidm-unixd-tasks.service
yaleman, line 1: we should document the different requirements based on what bind mount strategy is being used. Some users would prefer the service not have `CAP_SYS_ADMIN` and use Symlink instead.

platform/opensuse/kanidm-unixd-tasks.service
yaleman, line 1: we should document the different requirements based on what bind mount strategy is being used. Some users would prefer the service not have `CAP_SYS_ADMIN` and use Symlink instead.

unix_integration/common/src/unix_config.rs
```diff
@@ -102,6 +102,7 @@ enum ConfigVersion {
 enum HomeStrategyV2 {
     #[default]
     Symlink,
+    BindMount,
```
yaleman, line 105: The PR's using cfg flags to remove `BindMount` from `HomeStrategy`, but not `HomeStrategyV2`, I'd say we should collapse to the `HomeStrategy` enum to avoid this mistake and having to re-map two different enums later on.

unix_integration/common/src/unix_config.rs
yaleman: (this whole match statement shouldn't be needed if the above "remove V2" is done)

unix_integration/resolver/src/bin/kanidm_unixd_tasks.rs
yaleman: Should probably throw a warning on login when bind mount's enabled and they find a symlink or existing dir though...

unix_integration/resolver/src/bin/kanidm_unixd_tasks.rs
```diff
@@ -222,7 +220,85 @@ fn create_home_directory(
...
+    // Create mount point if it doesn't exist
+    if !alias_path.exists() {
+        if let Err(e) = create_dir(alias_path) {
+            error!("Unable to create bind mount target at {alias_path:?}");
+            return Err(format!("{e:?}"));
+        }
+    }
+
+    let current_mounts = Process::myself()
+        .map_err(|e| format!("Could not get reference to current process: {e}"))?
```
yaleman: Have to be slightly informative in these ones because we're just yeeting a string back that gets passed up the line a few levels.

```suggestion
        .map_err(|e| format!("While updating home directory bind mount, could not get reference to current process: {e}"))?
```

unix_integration/resolver/src/bin/kanidm_unixd_tasks.rs
```diff
@@ -222,7 +220,85 @@ fn create_home_directory(
...
+        .mountinfo()
+        .map_err(|e| format!("Could not get mount info: {e}"))?;
```
yaleman: Have to be slightly informative in these ones because we're just yeeting a string back that gets passed up the line a few levels.

```suggestion
        .map_err(|e| format!("While updating home directory bind mount, could not get mount info: {e}"))?;
```

unix_integration/resolver/src/bin/kanidm_unixd_tasks.rs
```diff
@@ -222,7 +220,85 @@ fn create_home_directory(
...
+                "Unable to remove conflicting mount at {:?}: {e}",
+                &m.mount_point
+            )
+        })?;
+    }
+
+    // If mount point exists and is already correctly mounted, we are done
+    if current_mounts.iter().any(|m| {
+        m.mount_point == alias_path
+            && m.mount_source == Some(hd_mount_path.to_string_lossy().to_string())
```
yaleman: I feel like we should be pushing m.mount_source to be a Path instead of the messy `.to_string_lossy().to_string()`

unix_integration/resolver/src/bin/kanidm_unixd_tasks.rs
```diff
@@ -222,7 +220,85 @@ fn create_home_directory(
...
+
+    let current_mounts = Process::myself()
+        .map_err(|e| format!("Could not get reference to current process: {e}"))?
+        .mountinfo()
+        .map_err(|e| format!("Could not get mount info: {e}"))?;
+
+    // Remove conflicting mount if it exists:
+    let mismatching_mount = current_mounts.iter().find(|m| {
+        m.mount_point == alias_path
+            && m.mount_source != Some(hd_mount_path.to_string_lossy().to_string())
```
yaleman: I feel like we should be pushing m.mount_source to be a Path instead of the messy `.to_string_lossy().to_string()`

unix_integration/resolver/debian/kanidm-unixd-tasks.service
Firstyear: Done, defaults to symlink perms, and added tons of comments to unit files/docs for this.

unix_integration/common/src/unix_config.rs
Firstyear: Okay, so this seems *weird* - HomeStrategyV2 is the *value*/type we parse via toml from the users config. 

HomeStrategy is our *internal* type to represent what was selected. This allows separation from our config as the user interacts with, and what our daemon does. 

So HomeStrategyV2 needs to parse both the symlink and bindmount value, and if it's not on a supported platform we warn and return symlink.

Then HomeStrategy internally with the cfg define is compiled out on unsupported platforms.

Firstyear:
@yaleman Updated based on your feedback, and I also added the extra warns etc.

yaleman (dismissed):

ReimarBauer:
Anything I can do to help?

Firstyear (approved):

examples/unixd
```diff
@@ -93,6 +94,24 @@ version = '2'
...
+# The strategy used for management of home directory aliases. If the home_attr is "uuid" and the
+# home_alias is "name", then the system needs to redirect from /home/<name> to /home/<uuid> in some
+# transparent manner. This allows configuration of that method.
+#
+# * symlink: creates a symlink from the alias to the stable home directory path. Some applications
+#            may encounter issues with the strategy (e.g. cargo, snap).
+#
+# (linux only)
+# * bind_mount: bind mounts the alias directory to the stable path. This requires modification of
+#               .service file on linux to grant the needed permissions. For more details, see:
```
ReimarBauer, line 106: @Firstyear `home_strategy = "bindmount"`without the `_` works for me. 

That gets readed without an error /usr/sbin/kanidm_unixd  -t -r

```
...
home_strategy: BindMount
...
```

ReimarBauer (commented):
just seen that there is a typo in the the word bindmount

PR #3999, opened 2025-12-02
Added PasswordChangedTime attribute and database field

### LINKED ISSUE #3924: Feat: Expose LDAP pwdChangedTime attribute on persons

I'm looking to integrate kanidm with [Stalwart](https://github.com/stalwartlabs/stalwart) via LDAP, which requires an attribute to track so that they can invalidate auth tokens for LDAP binds when a user's password has changed. Without such an attribute, Stalwart refuses to authenticate users.

Expose the LDAP attribute pwdChangedTime which is updated with a timestamp when the person's password is changed.

ref:
https://www.ietf.org/archive/id/draft-behera-ldap-password-policy-11.html#name-pwdchangedtime
https://ldapwiki.com/wiki/Wiki.jsp?page=PwdChangedTime
https://github.com/lldap/lldap/pull/1265

CEbbinghaus:
# Change summary

- Adds pwdChangedTime LDAP attribute backed by the database credentials

Fixes #3924

Checklist

### DIFF
--- a/server/lib/src/idm/credupdatesession.rs
+++ b/server/lib/src/idm/credupdatesession.rs
@@ -2783,7 +2783,6 @@ impl IdmServerCredUpdateTransaction<'_> {
 
 #[cfg(test)]
 mod tests {
-    use time::OffsetDateTime;
     use super::{
         CredentialState, CredentialUpdateSessionStatus, CredentialUpdateSessionStatusWarnings,
         CredentialUpdateSessionToken, InitCredentialUpdateEvent, InitCredentialUpdateIntentEvent,
@@ -2809,6 +2808,7 @@ mod tests {
     use kanidm_proto::v1::{AuthAllowed, AuthIssueSession, AuthMech, UnixUserToken};
     use sshkey_attest::proto::PublicKey as SshPublicKey;
     use std::time::Duration;
+    use time::OffsetDateTime;
     use uuid::uuid;
     use webauthn_authenticator_rs::softpasskey::SoftPasskey;
     use webauthn_authenticator_rs::softtoken::{self, SoftToken};


CEbbinghaus:
Fully expect this to be wrong in many ways. Should we be updating the timestamp for every credential update or only passwords?

server/lib/src/schema.rs
```diff
@@ -1933,6 +1933,21 @@ impl SchemaWriteTransaction<'_> {
                 syntax: SyntaxType::Utf8String,
             },
         );
+        self.attributes.insert(
+            Attribute::PasswordChangedTime,
+            SchemaAttribute {
+                name: Attribute::PasswordChangedTime,
+                uuid: UUID_CREDENTIAL,
```
Firstyear: Should be UUID_SCHEMA_ATTR_CREDENTIAL_TIMESTAMP

server/lib/src/be/dbvalue.rs
```diff
@@ -238,6 +240,15 @@ impl DbCred {
             | DbCred::V3PasswordMfa { uuid, .. } => *uuid,
         }
     }
+    
+    pub fn timestamp(&self) -> OffsetDateTime {
```
Firstyear: last_changed_timestamp

server/lib/src/be/dbvalue.rs
```diff
@@ -211,16 +212,17 @@ pub enum DbCred {
 
     // New Formats!
     #[serde(rename = "V2Pw")]
-    V2Password { password: DbPasswordV1, uuid: Uuid },
+    V2Password { password: DbPasswordV1, uuid: Uuid, timestamp: OffsetDateTime },
```
Firstyear: This will fail  to deserialise existing credentials as this fields doesn't exist. You probably want to annotate this with `#[serde(default = unix_epoch)]` and then make a fn for unix_epoch() -> OffsetDateTime so that when no timestamp is there, we default to epoch.

Firstyear: https://serde.rs/attr-default.html

server/lib/src/credential/mod.rs
```diff
@@ -457,6 +464,7 @@ impl Credential {
             type_,
             // Rotate the credential id on any change to invalidate sessions.
             uuid: Uuid::new_v4(),
+            timestamp: self.timestamp,
```
Firstyear, line 506: Should be changing the timestamp here.

server/lib/src/credential/mod.rs
```diff
@@ -500,6 +508,7 @@ impl Credential {
...
```
Firstyear: And here

server/lib/src/credential/mod.rs
```diff
@@ -538,6 +547,7 @@ impl Credential {
...
```
Firstyear, line 594: Annndddd here.

server/lib/src/credential/mod.rs
```diff
@@ -711,13 +727,15 @@ impl Credential {
...
             uuid: Uuid::new_v4(),
+            timestamp: OffsetDateTime::now_utc(),
         }
     }
 
     pub(crate) fn new_from_password(pw: Password) -> Self {
         Credential {
             type_: CredentialType::Password(pw),
             uuid: Uuid::new_v4(),
+            timestamp: OffsetDateTime::now_utc(),
```
Firstyear: *never* call now_utc inside a function. You need to change new_from_password to take in the current_time parameter and then convert that to OffsetDateTime.

server/lib/src/idm/ldap.rs
```diff
@@ -843,6 +844,7 @@ pub(crate) fn ldap_vattr_map(input: &str) -> Option<&str> {
         ATTR_OBJECTCLASS => Some(ATTR_CLASS),
         ATTR_LDAP_SSHPUBLICKEY => Some(ATTR_SSH_PUBLICKEY), // no-underscore -> underscore
         ATTR_UIDNUMBER => Some(ATTR_GIDNUMBER),             // yes this is intentional
+        ATTR_PWD_CHANGED_TIME => Some(ATTR_PWD_CHANGED_TIME),
```
Firstyear: The *source* name in Kanidm here would be ATTR_CREDENTIAL I expect. This is how we map from "pwdChangedTime" to find that "oh that attribute is probably on Credential. ". The gotcha here will be access controls, because most things won't have the ability to read Credential on a user. We also have to think about / account for the fallback credential option, and which changeTime we're really using.

Firstyear: PS the other half of the magic is in https://github.com/kanidm/kanidm/blob/master/server/lib/src/entry.rs#L2527

Firstyear:
Actually as soon as I typed it, I think the timestamp in the credential is still correct. Because then we can use the time stamp based on what credential was used etc. The issue still is the unix / primary cred situation if we have fallback etc.

Firstyear:
> Actually as soon as I typed it, I think the timestamp in the credential is still correct. Because then we can use the time stamp based on what credential was used etc. The issue still is the unix / primary cred situation if we have fallback etc.

And I wonder how we can deal with access control here - we don't want to give "search/read" to unix/primary credential, but we still want to check that the caller has access to read pwdChangedTime.

Firstyear:
Actually I'm overthinking, as a read on credentials is only *METADATA* not anything security sensitive. So that's okay.

So then it's just the unix vs primary we have to think about in what pwdChangeTime we return? The ldap vattr map wasn't built to handle merging from multiple attributes. We might need to adjust it to do so.

Firstyear:
And not just that we really should return the pwdChangedTime of the credential that was used to *bind* to the server, so that may add some complexity.

Sorry for comment spam.

Perhaps at this point, we need to focus on the TDD side, write tests to cover the invariants we expect to hold, and then we can go from there.

server/lib/src/schema.rs
CEbbinghaus: AHAHAHAHAHAHA, I swear I fixed that before pushing, It was the last thing I wanted to do before knocking out.
Oh well

CEbbinghaus:
Will write the tests next. I agree that we should use the changed time of the password used for authentication. I think this is the right direction with the timestamp on the credential since it also gives us some visibility for administration purposes.

server/lib/src/credential/mod.rs
CEbbinghaus: This one I'd appreciate some clarification on, How come calling `now_utc` within a method is considered bad? these methods are invoked within `.map` statements so it does make the calling site significantly more ugly. and it would still be creating the timestamp in **a** method, just not this one.

Appreciate any clarification

server/lib/src/be/dbvalue.rs
CEbbinghaus: Good call, Fixed now

CEbbinghaus:
I don't think we can get the credential used for the bind since the bind itself and the search are two different requests

server/lib/src/credential/mod.rs
Firstyear: Because time should always be passed in as an argument so that we can control it precisely in tests. 

Additionally within  a transaction we consider the "current time" to be the exact moment the transaction started, so by passing it in we keep that consistent. 

It also saves needless system calls since we pass time around.

server/lib/src/credential/mod.rs
```diff
@@ -760,6 +811,8 @@ impl Credential {
                 ),
                 // Rotate the credential id on any change to invalidate sessions.
                 uuid: Uuid::new_v4(),
+                // Update the timestamp to signify a changed credential
+                timestamp: OffsetDateTime::now_utc(),
```
Firstyear: As above, don't use now_utc.

server/lib/src/credential/mod.rs
```diff
@@ -800,6 +854,8 @@ impl Credential {
                 type_: CredentialType::PasswordMfa(pw.clone(), totp.clone(), wan.clone(), None),
                 // Rotate the credential id on any change to invalidate sessions.
                 uuid: Uuid::new_v4(),
+                // Update the timestamp to signify a changed credential
+                timestamp: OffsetDateTime::now_utc(),
```
Firstyear: And above.

server/lib/src/credential/totp.rs
```diff
@@ -379,4 +381,142 @@ mod tests {
...
+
+    #[test]
+    fn test_credential_timestamp_updated_on_totp_append() {
+        let policy = CryptoPolicy::minimum();
+        let pw = Password::new(&policy, "test_password").expect("Failed to create password");
+        let original_cred = Credential::new_from_password(pw);
+        let original_timestamp = original_cred.timestamp;
+
+        // Sleep a tiny bit to ensure timestamp difference
+        std::thread::sleep(Duration::from_millis(1));
```
Firstyear: And this is why - see how you "thread sleep"? Well now_utc is vulnerable to clock shifts. So even if you slept for 1 millisecond, the clock may step backwards. That would cause this test to fail. Similar faults can occur in forward time jumps too. 

Where as by passing in the "time" we can precisely control how time moves forward/backwards, without needing to hack in sleep statements.

server/lib/src/credential/totp.rs
```diff
@@ -379,4 +381,142 @@ mod tests {
...
+        let policy = CryptoPolicy::minimum();
+        let pw = Password::new(&policy, "test_password").expect("Failed to create password");
+        let cred = Credential::new_from_password(pw);
+
+        let totp = Totp::generate_secure(TOTP_DEFAULT_STEP);
+        let cred_with_totp = cred.append_totp("test_totp".to_string(), totp);
+        let timestamp_after_append = cred_with_totp.timestamp;
+
+        // Sleep a tiny bit to ensure timestamp difference
+        std::thread::sleep(Duration::from_millis(1));
```
Firstyear: As above.

server/lib/src/credential/totp.rs
```diff
@@ -379,4 +381,142 @@ mod tests {
...
+
+    #[test]
+    fn test_credential_timestamp_updated_on_password_change() {
+        let policy = CryptoPolicy::minimum();
+        let original_cred = Credential::new_password_only(&policy, "original_password")
+            .expect("Failed to create credential");
+        let original_timestamp = original_cred.timestamp;
+
+        // Sleep a tiny bit to ensure timestamp difference
+        std::thread::sleep(Duration::from_millis(1));
```
Firstyear: And above.

server/lib/src/credential/totp.rs
```diff
@@ -379,4 +381,142 @@ mod tests {
...
+    #[test]
+    fn test_credential_timestamp_preserved_on_password_upgrade() {
+        let policy = CryptoPolicy::minimum();
+        let original_cred = Credential::new_password_only(&policy, "test_password")
+            .expect("Failed to create credential");
+        let original_timestamp = original_cred.timestamp;
+        let original_uuid = original_cred.uuid;
+
+        // Sleep a tiny bit to ensure we can detect if timestamp changes
+        std::thread::sleep(Duration::from_millis(1));
```
Firstyear: Ditto

server/lib/src/credential/totp.rs
```diff
@@ -379,4 +381,142 @@ mod tests {
...
+
+        // Verify timestamp was updated
+        assert_ne!(mfa_timestamp, cred_with_backup.timestamp);
+        assert!(cred_with_backup.timestamp > mfa_timestamp);
+
+        // Verify UUID was rotated
+        assert_ne!(mfa_cred.uuid, cred_with_backup.uuid);
+
+        // Sleep to ensure timestamp difference
+        std::thread::sleep(Duration::from_millis(1));
```
Firstyear: And again.

server/lib/src/credential/totp.rs
```diff
@@ -379,4 +381,142 @@ mod tests {
...
+        let pw = Password::new(&policy, "test_password").expect("Failed to create password");
+        let cred = Credential::new_from_password(pw);
+
+        // Add TOTP to make it MFA
+        let totp = Totp::generate_secure(TOTP_DEFAULT_STEP);
+        let mfa_cred = cred.append_totp("test_totp".to_string(), totp);
+        let mfa_timestamp = mfa_cred.timestamp;
+
+        // Sleep to ensure timestamp difference
+        std::thread::sleep(Duration::from_millis(1));
```
Firstyear: image

server/lib/src/credential/mod.rs
CEbbinghaus: Fixed

server/lib/src/credential/mod.rs
```diff
@@ -431,6 +431,9 @@ impl Credential {
             .map(|pw| self.update_password(pw, timestamp))
     }
 
+    // I added the timestamp updating to this since it seemed the right thing to do
+    // But since the password itself should remain unaffected this doesn't technically constitute
+    // an update of the password.
```
CEbbinghaus: @Firstyear can I get an opinion on this?

server/lib/src/be/dbvalue.rs
```diff
@@ -211,16 +216,26 @@ pub enum DbCred {
...
+        uuid: Uuid,
+        #[serde(default = "unix_epoch")]
+        timestamp: OffsetDateTime,
+    },
     #[serde(rename = "V2GPw")]
-    V2GenPassword { password: DbPasswordV1, uuid: Uuid },
+    V2GenPassword {
+        password: DbPasswordV1,
+        uuid: Uuid,
+        timestamp: OffsetDateTime,
```
CEbbinghaus, line 230: forgot to add the serde attribute to the rest of them :(

server/lib/src/be/dbvalue.rs
```diff
@@ -19,6 +20,10 @@ use crate::repl::cid::Cid;
 use crypto_glue::{s256::Sha256Output, traits::Zeroizing};
 pub use kanidm_lib_crypto::DbPasswordV1;
 
+fn unix_epoch() -> OffsetDateTime {
+    OffsetDateTime::from_unix_timestamp(0).unwrap()
```
Firstyear: https://docs.rs/time/latest/time/struct.OffsetDateTime.html#associatedconstant.UNIX_EPOCH

server/lib/src/be/dbvalue.rs
```diff
@@ -238,6 +253,15 @@ impl DbCred {
             | DbCred::V3PasswordMfa { uuid, .. } => *uuid,
         }
     }
+
+    pub fn last_changed_timestamp(&self) -> OffsetDateTime {
+        match self {
+            DbCred::V2Password { timestamp, .. }
+            | DbCred::V2GenPassword { timestamp, .. }
+            | DbCred::V3PasswordMfa { timestamp, .. } => *timestamp,
+            _ => OffsetDateTime::from_unix_timestamp(932964162_i64).unwrap(),
```
Firstyear: https://docs.rs/time/latest/time/struct.OffsetDateTime.html#associatedconstant.UNIX_EPOCH

Firstyear: But also I think you don't need this since you add serde default attributes that will already populate with epoch.

server/lib/src/credential/mod.rs
```diff
@@ -139,11 +141,13 @@ impl TryFrom<DbCred> for Credential {
     type Error = ();
 
     fn try_from(value: DbCred) -> Result<Self, Self::Error> {
+        let timestamp = value.last_changed_timestamp();
```
Firstyear, line 146: Why do this when you can just match on the DBCreds in the match lines?

server/lib/src/credential/mod.rs
Firstyear: Yep, it's not needed in upgrade_password.

server/lib/src/idm/server.rs
```diff
@@ -1855,8 +1858,10 @@ impl IdmServerProxyWriteTransaction<'_> {
             return Err(OperationError::SystemProtectedObject);
         }
 
-        let modlist =
-            gen_password_mod(pce.cleartext.as_str(), self.crypto_policy).map_err(|e| {
+        let timestamp = OffsetDateTime::now_utc();
```
Firstyear: now_utc should not be present anywhere in this PR. As I said, it's a parameter we always control as part of transactions. Look at `self.qs_write.get_curtime_odt()` which gets the current time from the current transaction.

server/lib/src/idm/serviceaccount.rs
```diff
@@ -291,7 +291,8 @@ impl IdmServerProxyWriteTransaction<'_> {
         // Generate a new random, long pw.
         // Because this is generated, we can bypass policy checks!
         let cleartext = password_from_random();
-        let ncred = Credential::new_generatedpassword_only(self.crypto_policy(), &cleartext)
+        let timestamp = OffsetDateTime::now_utc();
```
Firstyear: As above, self.qs_write.get_curtime_odt().

server/lib/src/idm/server.rs
```diff
@@ -2193,11 +2216,12 @@ impl IdmServerProxyWriteTransaction<'_> {
     pub fn process_delayedaction(
         &mut self,
         da: &DelayedAction,
-        _ct: Duration,
+        ct: Duration,
     ) -> Result<(), OperationError> {
+        let timestamp = OffsetDateTime::UNIX_EPOCH + ct;
```
Firstyear: You should have get_curtime_odt like the othr comments point out.

server/lib/src/idm/credupdatesession.rs
```diff
@@ -2062,21 +2068,23 @@ impl IdmServerCredUpdateTransaction<'_> {
...
             return Err(OperationError::AccessDenied);
         };
-
+        
         if !matches!(session.mfaregstate, MfaRegState::None) {
             admin_info!("Invalid TOTP state, another update is in progress");
             return Err(OperationError::InvalidState);
         }
+        
+        let timestamp = OffsetDateTime::UNIX_EPOCH + ct;
```
Firstyear: See also the comments about get_curtime_odt().

server/lib/src/idm/credupdatesession.rs
```diff
@@ -2106,6 +2114,8 @@ impl IdmServerCredUpdateTransaction<'_> {
             return Err(OperationError::AccessDenied);
         };
 
+        let timestamp = OffsetDateTime::UNIX_EPOCH + ct;
```
Firstyear, line 2274: "

server/lib/src/idm/credupdatesession.rs
```diff
@@ -2151,6 +2161,8 @@ impl IdmServerCredUpdateTransaction<'_> {
...
```
Firstyear, line 2321: "

server/lib/src/idm/credupdatesession.rs
```diff
@@ -2025,14 +2029,16 @@ impl IdmServerCredUpdateTransaction<'_> {
...
```
Firstyear, line 2189: "

server/lib/src/idm/credupdatesession.rs
```diff
@@ -1946,6 +1948,8 @@ impl IdmServerCredUpdateTransaction<'_> {
...
```
Firstyear, line 2106: "

server/lib/src/idm/credupdatesession.rs
```diff
@@ -1866,6 +1866,8 @@ impl IdmServerCredUpdateTransaction<'_> {
...
```
Firstyear, line 2024: "

server/lib/src/credential/mod.rs
```diff
@@ -457,6 +498,8 @@ impl Credential {
             type_,
             // Rotate the credential id on any change to invalidate sessions.
             uuid: Uuid::new_v4(),
+            // Update the timestamp to signify a changed credential
+            timestamp: OffsetDateTime::now_utc(),
```
Firstyear: As before.

Firstyear (changes requested):

server/lib/src/be/dbvalue.rs
Firstyear: Also this is such a weird line, and such a weird number. Did you vibe this? Again, please don't .....

server/lib/src/be/dbvalue.rs
CEbbinghaus: Yup, discovered that after, Will fix :3

server/lib/src/be/dbvalue.rs
CEbbinghaus: I did not, None of the code is vibed, only some of the tests contained some copilot suggestions. This is actually the timestamp of my birthday, Was hoping to sneak in a easter egg but the default unix epoch kinda put a spanner in that :'(

server/lib/src/credential/mod.rs
CEbbinghaus: I could swear I removed them all. Seems I missed one.... :(

server/lib/src/credential/mod.rs
```diff
@@ -500,6 +543,8 @@ impl Credential {
...
```
CEbbinghaus: Same here :(

server/lib/src/credential/mod.rs
```diff
@@ -538,6 +583,8 @@ impl Credential {
...
```
CEbbinghaus: and here :(

server/lib/src/idm/server.rs
CEbbinghaus: I'll check on self but I recall this particular callsite not having access to `get_curtime_odt`, which is why I just calculated it myself since we do pass in the `ct`

Edit: looks like it would be under `self.qs_write` which would explain why I couldn't find it. Easy to fix 👍🏻

server/lib/src/idm/serviceaccount.rs
CEbbinghaus: `qs_write` is the key, I was tryna find `.get_curtime_odt()` on `self` and some of the members, Seems I missed `qs_write`. Doesn't help that my laptop struggles with a repo the size of kani and takes +-2 minutes for the intellisense to kick in and allow me to traverse the members.

server/lib/src/credential/mod.rs
CEbbinghaus: This is left over from the first iteration (before we had the serde deserialize fallback). So I was relying on that method to do the defaulting back to the original value. Without this callsite that method can probably be deleted

server/lib/src/idm/ldap.rs
CEbbinghaus: Huh? Wouldn't it be its own attribute? Or should I remove the ATTR_PWD_CHANGED_TIME entirely since we don't actually have a different attribute but rather just define the changed time on the credential itself?

server/lib/src/idm/credupdatesession.rs
CEbbinghaus: All of these are within `impl IdmServerCredUpdateTransaction<'_>` which only defines `qs_read` which in turn does not contain a `ct`. I suspect that is why we take the ct as an argument in these methods

server/lib/src/idm/credupdatesession.rs
CEbbinghaus: "

server/lib/src/idm/credupdatesession.rs
CEbbinghaus: "

server/lib/src/idm/credupdatesession.rs
CEbbinghaus: "

server/lib/src/idm/credupdatesession.rs
CEbbinghaus: "

server/lib/src/idm/credupdatesession.rs
CEbbinghaus: "

server/lib/src/be/dbvalue.rs
CEbbinghaus: After trying to get rid of it, I realized we will need something like this for `impl TryFrom<DbCred> for Credential` which needs to revert to some value for V1 creds, if you would prefer I move this match into that method body rather than its own independent method that is fine too

server/lib/src/idm/server.rs
CEbbinghaus: Fixed, None should remain (if I haven't missed one...)

server/lib/src/credential/mod.rs
CEbbinghaus: As with the other comment, The timestamp is not avaliable in V1 creds, so we will need a fallback for them in any case.

server/lib/src/idm/ldap.rs
CEbbinghaus: How do we deal with this given one LDAP name could map to more than 1 kani attribute (e.g `ATTR_PRIMARY_CREDENTIAL` | `ATTR_UNIX_PASSWORD`). Additionally there is no `ATTR_CREDENTIAL`

CEbbinghaus: > PS the other half of the magic is in https://github.com/kanidm/kanidm/blob/master/server/lib/src/entry.rs#L2527

Just started tackling that side of the equation and made the realization that the actual logic is far from done. I'm thinking that the only realistic way is to fold the timestamps during the `EntityCommited -> EntityReduced` process, But that would add complexity to a method that is currently just a for loop....

What do you think on how we should handle this?

server/lib/src/idm/account.rs
```diff
@@ -562,7 +562,8 @@ impl Account {
         match &self.primary {
             // Change the cred
             Some(primary) => {
-                let ncred = primary.set_password(crypto_policy, cleartext)?;
+                let ncred =
```
Firstyear: Shouldn't the current time be passed in via gen_password_mod here ...?

server/lib/src/entry.rs
```diff
@@ -2560,6 +2561,23 @@ impl Entry<EntryReduced, EntryCommitted> {
                         atype: ATTR_HOME_DIRECTORY.to_string(),
                         vals: vec![format!("/home/{}", self.get_uuid()).into_bytes()],
                     }),
+                    LDAP_ATTR_PWD_CHANGED_TIME => Some(LdapPartialAttribute {
+                        atype: LDAP_ATTR_PWD_CHANGED_TIME.to_string(),
+                        vals: vec![format!("{}Z", {
+                            // This whole block is stupid. We are operating on the EntryReduced when actually
+                            // we want to be resolving the entirety of the EntryConmitted (including ACP) since
```
Firstyear: Remember too that this might not even be your own entry, but the service account looking at someone else's entry. 

EntryReduced is correct still here, we need access controls to have been applied here before we proceed. 

I think we may have to rethink a few steps here, such as where we are generating these attributes (perhaps we are doing it too late here, and it has to happen before access controls), or we need to generate them on writes to the entry. 

But also we have to think about the application password side of this too, and what that looks like in the context of the change time. 

Generally I find that "if it feels like you're having to do something stupid, you probably are, and it's time to change tactic".

Firstyear: Liek heck, even when doing the home directory PR there was something that felt not 100% right about it. 

I wonder if this means it's time to add an attribute generator step here https://github.com/kanidm/kanidm/blob/master/server/lib/src/server/mod.rs#L380 

I don't think it'll be trivial though .... What's the best way to chat more about this?

Firstyear (changes requested):

server/lib/src/entry.rs
CEbbinghaus: Hmmmm, don't want to spam the Channel so either DM's, or in a voice call of some sort. Does Suse use teams? Could set up a video call?

server/lib/src/idm/account.rs
CEbbinghaus: Could, Opted not to since this method is only used for tests, But if you recon we should I'm happy to add it :3

server/lib/src/idm/credupdatesession.rs
```diff
@@ -1408,12 +1410,47 @@ impl IdmServerProxyWriteTransaction<'_> {
...
+                        && session
+                            .resolved_account_policy
+                            .allow_primary_cred_fallback()
+                            == Some(true)
+                    {
+                        modlist.push_mod(Modify::Present(
+                            Attribute::PasswordChangedTime,
+                            Value::DateTime(timestamp),
+                        ));
+                    }
```
CEbbinghaus: @Firstyear 
Happy with this logic? If so I'll write some tests for specifically this case

Still need to write behavior tests against LDAP to guarantee the high level behavior is achieved. Just want to check re the unit tests.

CEbbinghaus: UnixCred being `CredentialState::AccessDeny` obviously breaks this logic, But I am uncertain under which conditions that would be triggered and therefore wrote it the simple dumb way.

server/lib/src/idm/account.rs
Firstyear: Yeah, I think we keep it consistent.

server/lib/src/idm/credupdatesession.rs
Firstyear: You have access to https://github.com/kanidm/kanidm/blob/master/server/lib/src/idm/credupdatesession.rs#L137 in this check, so you can check the true state of the unix credential.

CEbbinghaus: @Firstyear problem is that the unix cred could have been deleted this update which would have the `account` retain a true unix cred but the update session (pre commit) have it removed. We should also be updating the cred changed timestamp when the relevant credential is removed.

Perhaps the stupid simple update on any password change may be better? Opinions?

Firstyear: Correct, it may have been deleted. But the *account* record in the cred update session shows the true state of the account *at the time the credUpdateSession started.

This means you know:

* If the unix pw was deleted
* Or changed
* Or updated
* And if we was allowed to be changed in this operation.

So lets say account.unix_cred().is_some(), but unixcred_state is AccessDeny, then you know that you don't need to update pwdChangedTime.

But if you have account.unix_cred().is_some(), unixcred_state::Modifiable, and then someone makes a change, then you update the time .

In otherwords, you can write pwdChangedTime here https://github.com/kanidm/kanidm/blob/master/server/lib/src/idm/credupdatesession.rs#L1463  if unixcred.is_some(), or https://github.com/kanidm/kanidm/blob/master/server/lib/src/idm/credupdatesession.rs#L1414 if it's none.

CEbbinghaus: I had a look and from my investigation it looks like the state `AccessDeny,PolicyDeny,Modifiable` are all set by the `create` method. So we cannot use `unixcred_state::Modifiable` as an indicator as to whether the unix cred **has** been updated, as it would **always** be true, (aka any save would cause the timestamp to be set again). For some reason, we purge and recreate the ava's every update, regardless of if they have been changed.

CEbbinghaus:
@Firstyear Finally got around to revamping. I don't think we need to check the current state at all, The solution I came up with also ensures that there is always some value set.

Hope everything is satisfactory, would be nice to get this merged :3

Firstyear:
> Hope everything is satisfactory, would be nice to get this merged :3

image

CEbbinghaus:
Ahahahahahaha, Please Review, Also Merge, But Review.

Looks like I gotta rebase anyways

Firstyear:
I already rebased don't worry. Do you mind if I push some commits?

CEbbinghaus:
Sure. Git saves everything so no harm no foul :3

CEbbinghaus:
@Firstyear Had a look at [775b02eb320a644ed63db2ad4a825605d2975a78](https://github.com/kanidm/kanidm/commit/775b02eb320a644ed63db2ad4a825605d2975a78) and the big difference I see is just in whether or not we set the timestamp to the "last time a credential update processed" or the "time the credential was changed". I will have to read through the docs associated with `pwdChangedTime` carefully but I do not believe that there is anything limiting the change to only be forwards in time. As such I would argue that we should expose the created time of the credential itself since even if we move backwards, this not only marks a change in time but also more accurately reflects that the credential is actually an **older** credential. Right now we would be hiding this information.

If we do in fact want to hide this, we run into the problem that we need to determine what the **last** active timestamp of the used credential was and only set the new timestamp if that timestamp has changed. since even when doing 0 changes, we still end up updating the timestamp causing the credential to invalidate. Given that (from my investigation) the cred is always updated, even if it has not changed. We may as well shorten your PR to:

```diff
-        let mut unix_cred_changed = false;
-        let mut unix_cred_present = false;
+        let timestamp = OffsetDateTime::UNIX_EPOCH + ct;

+        modlist.push_mod(Modify::Purged(Attribute::PasswordChangedTime));
+        modlist.push_mod(Modify::Present(
+            Attribute::PasswordChangedTime,
+            Value::DateTime(timestamp),
+        ));


        match session.unixcred_state {
```

As `CredentialState::Modifiable | CredentialState::DeleteOnly | CredentialState::PolicyDeny` would always be true.

ps, pretty sure this change is required to make your commit work since deleting a credential should also be counted as an update:
```diff
@@ -1543,11 +1543,11 @@ impl IdmServerProxyWriteTransaction<'_> {
             CredentialState::DeleteOnly | CredentialState::Modifiable => {
                 modlist.push_mod(Modify::Purged(Attribute::UnixPassword));
 
+                unix_cred_changed = true;
                 if let Some(ncred) = &session.unixcred {
                     let vcred = Value::new_credential("unix", ncred.clone());
                     modlist.push_mod(Modify::Present(Attribute::UnixPassword, vcred));
                     unix_cred_present = true;
-                    unix_cred_changed = true;
                 }
             }
             CredentialState::PolicyDeny => {
```

Firstyear:
> As such I would argue that we should expose the created time of the credential itself since even if we move backwards, this not only marks a change in time but also more accurately reflects that the credential is actually an **older** credential. Right now we would be hiding this information

Yeah that's fair enough. Okay, you can ignore that change then. 

The only thing you're missing now is PwdChangedTime in https://github.com/kanidm/kanidm/blob/master/server/lib/src/migration_data/dl14/access.rs#L1065 and https://github.com/kanidm/kanidm/blob/master/server/lib/src/migration_data/dl14/access.rs#L932

Firstyear (dismissed):

Firstyear:
Looks like you forgot to run cargo fmt (and maybe clippy) as well.

CEbbinghaus:
Was trying to get clippy to run but kept running into errors due to missing build dependencies (new vm). FMT fixed, clippy running now. commit coming in <5m.

Firstyear (approved):

Issue #3959, opened 2025-11-15
Extend the number of default attributes


karatemuffin:
At the current state kanidm provides a limited set of user attributes, namely:
- legalname
- name
- email
- displayname

after discussion on the related [#3493](https://github.com/kanidm/kanidm/issues/3493) the idea came up to provide support for
the attributes listed for standard claims [chapter 5.1](https://openid.net/specs/openid-connect-core-1_0.html#StandardClaims) of the 
[OpenID Connect Core](https://openid.net/specs/openid-connect-core-1_0.html) specification.

Firstyear:
Went through the claims, here are my thoughts.

| Claim| Comment |
|----|-----|
| sub | supported |
| name | supported |
| given_name | can't support |
| family_name | can't support |
| middle_name | can't support |
| nickname | currently no attribute in Kani could be used here |
| preferred_username | supported |
| profile | unsure what we would map this to |
| picture | need profile pictures in Kani first |
| website | no attribute in Kani |
| email | supported |
| email_verified | supported |
| gender | no attribute in Kani |
| birthdate | no attribute in Kani |
| zoneinfo | no attribute in Kani |
| locale | no attribute in Kani |
| phone_number | no attribute in Kani |
| phone_number_verified | will be supported when phone_number is added |
| address | no attribute in Kani | 
| updated_at | last_modified_cid could be used here |


Currently I think the easiest to add is "updated_at". Outside of that we need to add some new data types / handling for most of these. I think picture, phone_number, address, zoneinfo and locale seem like the most obvious.

maciek134:
Seeing as this is something I also need, I could help implementing these (I especially need profile pictures).

Picture is going to be a bit more work for LDAP since it has to be base64-encoded JPEG data (at least in any recent system, I hope nobody is using the G3Fax format anymore 😅) though I guess it could just be converted when saved?

`profile` could just be an unspecified url value? Depending on the deployment it could be a free input for the user or a link to some corporate employee directory page.

karatemuffin:
I stepped trough all items and wrote down some things that came in mind.

### given_name, family_name and middle_name, nickname, gender
@Firstyear By cannot support you're referring to the [developer ethics](https://kanidm.github.io/kanidm/stable/developers/developer_ethics.html?highlight=legalname#free-from-harmful-discrimination-of-any-kind) chapter?

`nickname` + `gender` should then also be arbitrary utf8 strings like `displayname` already is?

### profile, website
[...]

PR #4007, opened 2025-12-08
Feature OIDC updated at

karatemuffin:
# Change summary
- Added support for updated_at as requested in #3959 
- Needs [compact-jwt](https://github.com/kanidm/compact-jwt/pull/44) pull first

Checklist

# Things to be clarified
- [ ] The [scim/api](https://github.com/karatemuffin/kanidm/blob/feature_oidc_updated_at/server/core/src/https/v1_scim.rs#L474) returns the raw value for `Attribute::last_modified_cid`  and also chooses the field name automatically eg. `"last_modified_cid": [
      "00000000000001765225917128611666-35257dc9-1be8-4b91-aa2b-1a932dd06789"
    ],` .  **Is this okay, or do we want to have it return only the 1st part of the cid?** 
- [ ] The [ui/profile](https://github.com/karatemuffin/kanidm/blob/a63c9ccff48409970f0b4683d635319fe68d443f/server/lib/src/idm/account.rs#L234) page shows now a row with `Last Update: 2025-12-08T20:25:33.198366944Z`.  **Is the Format okay?**
- [x] One test is failing atm. `thread 'testkit::proto_v1_test::tk_test_server_api_token_lifecycle' panicked at server/lib/src/idm/account.rs:331:9: called `Option::unwrap()` on a `None` value`. I'm not quite shure where this comes from, cause the added `updated_at` field of the [account.rs#L50](https://github.com/karatemuffin/kanidm/blob/a63c9ccff48409970f0b4683d635319fe68d443f/server/lib/src/idm/account.rs#L50) clearly returns the correct value during testing. **@Firstyear  I'm stuck at the moment and would now gladly listen to your hints. :smile:**
- [x] And i know the `Cargo.*` files shouldn't be in the PR, i'll fix after implementing the maintainers comments

### DIFF
--- a/server/lib/src/idm/oauth2.rs
+++ b/server/lib/src/idm/oauth2.rs
@@ -3150,8 +3150,12 @@ fn s_claims_for_account(
         (None, None)
     };
 
-    let updated_at : Option<OffsetDateTime> = if scopes.contains(OAUTH2_SCOPE_PROFILE) {
-        account.updated_at.as_ref().map(Into::into)
+    let updated_at: Option<OffsetDateTime> = if scopes.contains(OAUTH2_SCOPE_PROFILE) {
+        account
+            .updated_at
+            .as_ref()
+            .map(OffsetDateTime::from)
+            .and_then(|odt| odt.replace_nanosecond(0).ok())
     } else {
         None
     };
@@ -3388,8 +3392,7 @@ fn check_is_loopback(redirect_uri: &Url) -> bool {
 
 #[cfg(test)]
 mod tests {
-    use compact_jwt::OidcClaims;
-use super::{Oauth2TokenType, PkceS256Secret, TOKEN_EXCHANGE_SUBJECT_TOKEN_TYPE_ACCESS};
+    use super::{Oauth2TokenType, PkceS256Secret, TOKEN_EXCHANGE_SUBJECT_TOKEN_TYPE_ACCESS};
     use crate::credential::Credential;
     use crate::idm::accountpolicy::ResolvedAccountPolicy;
     use crate::idm::oauth2::{host_is_local, AuthoriseResponse, Oauth2Error, OauthRSType};
@@ -3518,7 +3521,10 @@ use super::{Oauth2TokenType, PkceS256Secret, TOKEN_EXCHANGE_SUBJECT_TOKEN_TYPE_A
                 Attribute::OAuth2RsScopeMap,
                 Value::new_oauthscopemap(
                     UUID_IDM_ALL_ACCOUNTS,
-                    btreeset![OAUTH2_SCOPE_OPENID.to_string(), OAUTH2_SCOPE_PROFILE.to_string()]
+                    btreeset![
+                        OAUTH2_SCOPE_OPENID.to_string(),
+                        OAUTH2_SCOPE_PROFILE.to_string()
+                    ]
                 )
                 .expect("invalid oauthscope")
             ),
@@ -5453,30 +5459,6 @@ use super::{Oauth2TokenType, PkceS256Secret, TOKEN_EXCHANGE_SUBJECT_TOKEN_TYPE_A
             .is_none());
     }
 
-    //As the timestamp from oidc and userinfo differs a fraction of a second, we
-    //need to check if the difference is lower than a arbitrary bound.
-    //We then overwrite the updated_at of oidc with the one from userinfo, so
-    //we can compare the whole struct thereafter.
-    fn assert_claims_with_tolerance(
-        mut claims_a : OidcClaims,
-        claims_b : &OidcClaims,
-        tolerance : i128
-    ){
-        let a = claims_a.updated_at;
-        let b = claims_b.updated_at;
-        assert!(
-            match (a, b) {
-                (Some(a), Some(b)) =>
-                    (a.unix_timestamp_nanos() - b.unix_timestamp_nanos()).abs()
-                        <= tolerance, // 1s
-                (None, None) => true,
-                _ => false,
-            }
-        );
-        claims_a.updated_at = claims_b.updated_at;
-        assert_eq!(claims_a, *claims_b);
-    }
-
     #[idm_test]
     async fn test_idm_oauth2_openid_extensions(
         idms: &IdmServer,
@@ -5591,7 +5573,12 @@ use super::{Oauth2TokenType, PkceS256Secret, TOKEN_EXCHANGE_SUBJECT_TOKEN_TYPE_A
             Some("testperson1@example.com".to_string())
         );
         assert!(
-            oidc.s_claims.scopes == vec![OAUTH2_SCOPE_OPENID.to_string(),OAUTH2_SCOPE_PROFILE.to_string(), "supplement".to_string()]
+            oidc.s_claims.scopes
+                == vec![
+                    OAUTH2_SCOPE_OPENID.to_string(),
+                    OAUTH2_SCOPE_PROFILE.to_string(),
+                    "supplement".to_string()
+                ]
         );
         assert!(oidc.claims.is_empty());
         // Does our access token work with the userinfo endpoint?
@@ -5616,7 +5603,7 @@ use super::{Oauth2TokenType, PkceS256Secret, TOKEN_EXCHANGE_SUBJECT_TOKEN_TYPE_A
         if let Some(jti) = &userinfo.jti {
             assert!(Uuid::from_str(jti).is_ok());
         }
-        assert_claims_with_tolerance(oidc.s_claims.clone(), &userinfo.s_claims, 1_000_000_000);
+        assert_eq!(oidc.s_claims, userinfo.s_claims);
         assert!(userinfo.claims.is_empty());
 
         drop(idms_prox_read);
@@ -5664,7 +5651,7 @@ use super::{Oauth2TokenType, PkceS256Secret, TOKEN_EXCHANGE_SUBJECT_TOKEN_TYPE_A
         if let Some(jti) = &userinfo.jti {
             assert!(Uuid::from_str(jti).is_ok());
         }
-        assert_claims_with_tolerance(oidc.s_claims, &userinfo.s_claims, 1_000_000_000);
+        assert_eq!(oidc.s_claims, userinfo.s_claims);
         assert!(userinfo.claims.is_empty());
     }
 
--- a/server/lib/src/migration_data/dl13/access.rs
+++ b/server/lib/src/migration_data/dl13/access.rs
@@ -952,7 +952,6 @@ pub static IDM_ACP_SELF_READ_DL8: LazyLock<BuiltinAcp> = LazyLock::new(|| Builti
         Attribute::ApplicationPassword,
         Attribute::SshPublicKey,
         Attribute::UnixPassword,
-        Attribute::LastModifiedCid,
     ],
     ..Default::default()
 });
--- a/server/lib/src/migration_data/dl14/access.rs
+++ b/server/lib/src/migration_data/dl14/access.rs
@@ -952,6 +952,7 @@ pub static IDM_ACP_SELF_READ_DL8: LazyLock<BuiltinAcp> = LazyLock::new(|| Builti
         Attribute::ApplicationPassword,
         Attribute::SshPublicKey,
         Attribute::UnixPassword,
+        Attribute::LastModifiedCid,
     ],
     ..Default::default()
 });


proto/src/scim_v1/server.rs
```diff
@@ -359,6 +360,16 @@ impl TryFrom<ScimEntryKanidm> for ScimPerson {
...
             });
 
+        //TODO It's working, but im not happy with it, maybe there is the possibility to extend
+        //ScimValueKanidm so we don't need to parse a String..
+        let updated_at: String = scim_entry
+            .get_string_attr(&Attribute::LastModifiedCid)
+            .and_then(|v| v.split('-').next())
+            .and_then(|s| s.parse::<i128>().ok())
+            .map(|nanos| OffsetDateTime::from_unix_timestamp_nanos(nanos))
+            .map(|odt| odt.expect("Error Unpacking").format(&Rfc3339).unwrap())
```
yaleman: you can't use unwrap or expect here.

server/lib/src/idm/account.rs
```diff
@@ -229,6 +231,16 @@ macro_rules! try_from_entry {
             _ => None,
         };
 
+        let updated_at: Option<u64> = if let DbValueSetV2::Cid(cid) = $value
+            .get_ava_set(Attribute::LastModifiedCid)
+            .unwrap()
```
yaleman: No unwraps 😄

karatemuffin: Ok then, only unwrapping gifts then :smile:

After rewriting i found 2 ways to do it, version 1 is a loveless insert of another `if let ..` 
but after tinkering about using the rust thingies `.and_then` i think version 2 is more legible.  
**What version do you preffer?**

``` Rust
#version 1
        let updated_at: Option<u64> =
            if let Some(v) = $value.get_ava_set(Attribute::LastModifiedCid) {
                if let DbValueSetV2::Cid(cid) = v.to_db_valueset_v2() {
                    cid.get(0).map(|c| c.timestamp.as_secs())
                } else {
                    None
                }
            } else {
                None
            };
#version 2  
        let updated_at: Option<u64> =
            $value
                .get_ava_set(Attribute::LastModifiedCid)
                .and_then(|v| match v.to_db_valueset_v2() {
                    DbValueSetV2::Cid(cid) => cid.get(0).map(|c| c.timestamp.as_secs()),
                    _ => None,
                });
```

yaleman: What's clippy say when you run that? That's what I typically go with. 😆

proto/src/scim_v1/server.rs
karatemuffin: done

server/lib/src/idm/account.rs
karatemuffin: clippy is indifferent so i went with version 2 :grin: 
btw.  the one test failing before, passes now, seems it was due my .unwrapping()

proto/src/constants.rs
```diff
@@ -248,6 +248,7 @@ pub const SUB_ATTR_TYPE: &str = "type";
 pub const SUB_ATTR_VALUE: &str = "value";
 
 pub const OAUTH2_SCOPE_EMAIL: &str = ATTR_EMAIL;
+pub const OAUTH2_SCOPE_UPDATED_AT: &str = "updated_at";
```
maciek134: fyi this field is part of the `profile` scope - https://openid.net/specs/openid-connect-core-1_0.html#ScopeClaims there isn't a separate one for each claim (how to request specific claims directly is described below, in 5.5)

server/lib/src/idm/account.rs
```diff
@@ -69,6 +70,7 @@ pub struct Account {
     pub(crate) sshkeys: BTreeMap<String, SshPublicKey>,
     pub apps_pwds: BTreeMap<Uuid, Vec<ApplicationPassword>>,
     pub(crate) oauth2_client_provider: Option<OAuth2AccountCredential>,
+    pub updated_at: Option<u64>,
```
maciek134: This should probably be an `OffsetDateTime` instead (tbf, as a user, I'd prefer it to be a strongly typed value everywhere)

proto/src/constants.rs
karatemuffin: True, and i know that i have read that.. not sure why i have implemented a separate one..
Probably overmotivation :smile:  I'll change that

server/lib/src/idm/account.rs
yaleman: It's specified as seconds since epoch in the spec, which we should stick to.

maciek134: Ah, right, sorry, confused with SCIM

proto/src/constants.rs
karatemuffin: Looking at it, a question popped up: when reading the linked spec right, we should also add an 
`if scopes.contains(OAUTH2_SCOPE_PROFILE)` not only to my newly introduced `updated_at` but also to the already existing `preferred_username` ?

proto/src/scim_v1/server.rs
```diff
@@ -316,6 +316,7 @@ pub struct ScimPerson {
     pub mails: Vec<ScimMail>,
     pub managed_by: Option<ScimReference>,
     pub groups: Vec<ScimReference>,
+    pub updated_at: String,
```
Firstyear: Not needed

proto/src/scim_v1/server.rs
```diff
@@ -359,6 +360,16 @@ impl TryFrom<ScimEntryKanidm> for ScimPerson {
                 _ => None,
             });
 
+        //TODO It's working, but im not happy with it, maybe there is the possibility to extend
```
Firstyear: Not needed.

server/lib/src/idm/account.rs
Firstyear: In this step we should be pulling the LastModifiedCID and leaving it as the CID until the *last possible moment* when we yield it to the client. We can only "narrow" our information, so the sooner you are removing the CID, the less info we have LATER in the process potentially. 

Think of it as "data as the server knows" vs "data as presented". You're setting it up for presentation *too soon*.

server/lib/src/idm/account.rs
```diff
@@ -229,6 +231,14 @@ macro_rules! try_from_entry {
             _ => None,
         };
 
+        let updated_at: Option<u64> =
+            $value
+                .get_ava_set(Attribute::LastModifiedCid)
+                .and_then(|u| match u.to_db_valueset_v2() {
+                    DbValueSetV2::Cid(cid) => cid.get(0).map(|c| c.timestamp.as_secs()),
```
Firstyear: You don't need this cid to dbvalue dance - cid's already have a timestamp in them. 

But also again, don't filter info here. Just get the CID. Also it's not optional, every entry has a last modified CID.

server/lib/src/idm/oauth2.rs
```diff
@@ -6037,6 +6044,7 @@ mod tests {
                     UUID_IDM_ALL_ACCOUNTS,
                     btreeset![
                         OAUTH2_SCOPE_EMAIL.to_string(),
+                        OAUTH2_SCOPE_UPDATED_AT.to_string(),
```
Firstyear: As previously noted, the scope should be "profile" here.

server/lib/src/server/access/search.rs
```diff
@@ -206,6 +206,7 @@ fn search_oauth2_filter_entry(
                     attr: btreeset!(
                         Attribute::Class,
                         Attribute::DisplayName,
+                        Attribute::LastModifiedCid,
```
Firstyear: You don't need this here.

Firstyear (changes requested):
I think there might be a bit of confusion, you have setup so that we can store updated_at via scim etc, but we don't have to. LastModifiedCID is already updated internally on every entry by default, and we shouldn't allow anyone externally to touch it.

I think there may also have been some confusion about the access control side etc.

Finally, I'm not seeing where you test the value being returned?

server/lib/src/idm/oauth2.rs
```diff
@@ -6122,7 +6134,11 @@ mod tests {
             pkce_request: Some(pkce_secret.to_request()),
             redirect_uri: Url::parse("https://demo.example.com/oauth2/result").unwrap(),
             // Note the scope isn't requested here!
-            scope: btreeset!["openid".to_string(), "email".to_string()],
+            scope: btreeset![
+                "openid".to_string(),
+                "email".to_string(),
+                "updated_at".to_string()
```
Firstyear: Here you need to also be testing in the claims case that the updated_at is returned is what I mean re testing.

Firstyear:
I assume this relies on https://github.com/kanidm/compact-jwt/pull/45 ?

proto/src/scim_v1/server.rs
karatemuffin: i have rolled back the scim changes made.

proto/src/scim_v1/server.rs
karatemuffin: i have rolled back the scim changes made.

server/lib/src/idm/account.rs
karatemuffin: - i have changed it to Option<Cid> as requested. 
- changed the place where i convert it, i hope it is now placed late enough in the flow

server/lib/src/idm/oauth2.rs
karatemuffin: done

server/lib/src/idm/oauth2.rs
karatemuffin: i have done some manual testing till now, cause i wanted to have it working and then ask how you would like the tests being organized. 

- do i create a new test `test_idm_oauth2_openid_profile_claims` (like `test_idm_oauth2_openid_group_claims`)?
- should profile even be part of `test_idm_oauth2_consent_granted_and_changed_workflow`?

server/lib/src/idm/account.rs
karatemuffin: see [comment](https://github.com/kanidm/kanidm/pull/4007#discussion_r2637458136)

server/lib/src/idm/oauth2.rs
```diff
@@ -6122,7 +6137,11 @@ mod tests {
             pkce_request: Some(pkce_secret.to_request()),
             redirect_uri: Url::parse("https://demo.example.com/oauth2/result").unwrap(),
             // Note the scope isn't requested here!
-            scope: btreeset!["openid".to_string(), "email".to_string()],
+            scope: btreeset![
+                "openid".to_string(),
+                "email".to_string(),
+                "profile".to_string()
```
Firstyear, line 6339: These should probably be part of https://github.com/kanidm/kanidm/blob/master/server/lib/src/idm/oauth2.rs#L5346 so you can validate the updated_at was returned in the id token.

server/core/src/https/views/admin/persons.rs
```diff
@@ -22,11 +22,12 @@ use kanidmd_lib::constants::EntryClass;
 use kanidmd_lib::idm::authentication::ClientAuthInfo;
 use uuid::Uuid;
 
-pub const PERSON_ATTRIBUTES: [Attribute; 9] = [
+pub const PERSON_ATTRIBUTES: [Attribute; 10] = [
```
Firstyear: Probably not needed here.

server/lib/src/idm/account.rs
Firstyear: Yep, looks good.

server/lib/src/idm/oauth2.rs
karatemuffin: will do, incl. cleanup, as i got time coming week.

Firstyear: I think this is the only outstanding comment otherwise it's good to go?

karatemuffin: are u sure it is #L5346 ?  cause there is no 'id_token' for the used variable  'discovery'.

Firstyear: I think there have been some commits since then that moved the test around which is why it doesn't line up. Look at. 

https://github.com/kanidm/kanidm/blob/master/server/lib/src/idm/oauth2.rs#L5560

You'll need to add the OAUTH2_SCOPE_OPENID_PROFILE too for this.

karatemuffin: i got the test working, except :face_exhaling: 
when comparing      `assert_eq!(oidc.s_claims, userinfo.s_claims);`
the timestamps are naturally off by a fraction of a second..

`Some(2026-02-15 4:08:28.0 +00:00:00)`  vs `Some(2026-02-15 4:08:28.690221817 +00:00:00)`

How do you want me to work around this?

1. extract `updated_at` from both `s_claims` then set it to the same value on both.. and after that i could check if both `updated_at` the same, while ignoring seconds...
2. compare each attribute from `s_claims` manually

The thing is, that in my opinion i cannot say the test is guaranteed to finish in a specific timeframe..

karatemuffin: opted for 1. still feels messy

Firstyear: Time is a parameter in all tests and function calls, so it probably means somewhere ".now()" is being called.

server/lib/src/idm/oauth2.rs
```diff
@@ -5631,7 +5664,7 @@ mod tests {
         if let Some(jti) = &userinfo.jti {
             assert!(Uuid::from_str(jti).is_ok());
         }
-        assert_eq!(oidc.s_claims, userinfo.s_claims);
+        assert_claims_with_tolerance(oidc.s_claims, &userinfo.s_claims, 1_000_000_000);
```
Firstyear: We should have perfect knowledge of the clock, see https://github.com/kanidm/kanidm/pull/4007/changes#diff-98d81995d4ecf097d233caa526aed627159111c33ac9ff3def33eb7d54a8b759R5557 for example. So that means we have to change how we handle this, because we probably need to remove nanoseconds on the response we send, so that we don't need this custom quality function.

karatemuffin: the `OidcClaims` comes from https://github.com/kanidm/compact-jwt/blob/main/src/oidc.rs#L125 and the `update_at` is defined as `OffsetDateTime` and the value itself comes from kandim's store procedure therefore it will eventually contain nanoseconds. on the other hand the [OIDC spec](https://openid.net/specs/openid-connect-core-1_0.html) describes it as seconds since 1970.. 

So that change should probably be in `compact-jwt` and not kandim itself? (It has anyway to be changed to a JSON number)

Firstyear: Yeah I have checked out the branch and I'm looking

Firstyear: Ahhh I see it. It's because the oidc token goes via a serialisation round trip so it goes to whole seconds, where the userinfo doesn't do that, so it retain the nanoseconds. There is an easy fix, I'll push a commit.

server/lib/src/idm/oauth2.rs
karatemuffin: probably the moment kanidm stores the values but i don't think i should lay hands on that :D also i expected it to be different before (`oidc` object) + after (`userinfo` object) a call of  `idms_prox_write`

Firstyear:
@karatemuffin Take a look at the change I made, I just 0 the nanosecond as we read the updated at. Simple and correct :)

PR #4021, opened 2025-12-18
Implement OIDC auth for service-accounts (RFC8693)

TheSylex:
# Change summary

Draft implementation of service_account OIDC non-interactive authorization.


TheSylex:
I've cooked a draft implementation of a basic auth flow for service_accounts through OIDC services using kanidm service API tokens. Looking for feedback as this is my first time contributing to a security/oidc/jwt-related project so I'm pretty clueless to most aspects besides the ones mentioned in the RFC document.

Here are the steps roughly needed to test the implemented change:

### 1) Create group, service account, and membership
```bash
kanidm group create oidc_demo
kanidm service-account create svc_demo "Demo Service" idm_admin
kanidm group add-members oidc_demo svc_demo
```

### 2) Create OIDC client and map scopes to the group
```bash
kanidm system oauth2 create demo_client "OIDC Demo Client" https://localhost:8443
kanidm system oauth2 update-scope-map demo_client oidc_demo openid groups
kanidm system oauth2 show-basic-secret demo_client
```

### 3) Generate service-account API token (subject token)
```bash
kanidm service-account api-token generate svc_demo "OIDC Exchange" --readwrite --output token
```

### 4) Exchange API token for OIDC tokens (access/id/refresh)
```bash
EXCHANGE_RESP=$(curl -k -s -X POST https://localhost:8443/oauth2/token \
  -u demo_client:"$CLIENT_SECRET" \
  -d "grant_type=urn:ietf:params:oauth:grant-type:token-exchange" \
  -d "subject_token=$API_TOKEN" \
  -d "subject_token_type=urn:ietf:params:oauth:token-type:access_token" \
  -d "audience=demo_client" \
  -d "scope=openid groups")

echo "$EXCHANGE_RESP"
ACCESS_TOKEN=$(echo "$EXCHANGE_RESP" | jq -r '.access_token')
REFRESH_TOKEN=$(echo "$EXCHANGE_RESP" | jq -r '.refresh_token')
ID_TOKEN=$(echo "$EXCHANGE_RESP" | jq -r '.id_token')
```

### 5) Introspect the access token
```bash
curl -k -X POST https://localhost:8443/oauth2/token/introspect \
  -u demo_client:"$CLIENT_SECRET" \
  -d "token=$ACCESS_TOKEN" \
  -d "token_type_hint=access_token"
```

### 6) Userinfo (claims) with the access token
```bash
curl -k -H "Authorization: Bearer $ACCESS_TOKEN" \
  https://localhost:8443/oauth2/openid/demo_client/userinfo
```

### 7) Refresh the access token
```bash
REFRESH_RESP=$(curl -k -s -X POST https://localhost:8443/oauth2/token \
  -u demo_client:"$CLIENT_SECRET" \
  -d "grant_type=refresh_token" \
  -d "refresh_token=$REFRESH_TOKEN")

echo "$REFRESH_RESP"
NEW_ACCESS_TOKEN=$(echo "$REFRESH_RESP" | jq -r '.access_token')
```

## 8) Verify ID token signature
```bash
curl -k https://localhost:8443/oauth2/openid/demo_client/public_key.jwk > jwks.json
```

proto/src/oauth2.rs
```diff
@@ -236,6 +250,8 @@ pub struct OAuth2RFC9068TokenExtensions {
 pub struct AccessTokenResponse {
     pub access_token: String,
     pub token_type: AccessTokenType,
+    /// RFC8693 issued_token_type for the representation of the issued token.
+    pub issued_token_type: Option<String>,
```
Firstyear: Should be an enum based on the variants in https://www.rfc-editor.org/rfc/rfc8693#TokenTypeIdentifiers

server/lib/src/idm/oauth2.rs
```diff
@@ -1594,6 +1621,160 @@ impl IdmServerProxyWriteTransaction<'_> {
...
+        if matches!(o2rs.type_, OauthRSType::Basic { .. }) && !client_authentication_valid {
+            security_info!(
+                "Unable to proceed with token exchange grant unless client authentication is provided and valid"
+            );
+            return Err(Oauth2Error::AuthenticationRequired);
+        }
+
+        match (actor_token, actor_token_type) {
+            (Some(_), None) | (None, Some(_)) => {
+                admin_warn!("actor_token and actor_token_type must both be supplied when using an actor token");
```
Firstyear: We don't use the "admin_*" log types anymore, just warn, info etc are fine.

server/lib/src/idm/server.rs
```diff
@@ -635,13 +635,29 @@ pub trait IdmServerTransaction<'a> {
...
                 } else {
-                    security_info!("The token grace window has passed and no entry parent sessions exist. Assuming invalid.");
-                    return Ok(None);
+                    let api_session = entry
+                        .get_ava_as_apitoken_map(Attribute::ApiTokenSession)
+                        .and_then(|sessions| sessions.get(&parent_session_id));
+
+                    if let Some(api_session) = api_session {
+                        if let Some(expiry) = api_session.expiry {
+                            if time::OffsetDateTime::UNIX_EPOCH + ct >= expiry {
```
Firstyear: I don't actually think you need to check the time here, since the oauth2 token has it's own expiry that should be *within* the parent sessions expiration window. What you actually need to check is similar to the UAT flow above, where you only need to check presence of the parent session remains. Because of the apiToken is *deleted* then that indicates that the child oauth2 sessions would no longer be valid (global revoke basicly).

server/lib/src/idm/oauth2.rs
```diff
@@ -1594,6 +1621,160 @@ impl IdmServerProxyWriteTransaction<'_> {
...
+        subject_token_type: &str,
+        requested_token_type: Option<&str>,
+        audience: Option<&str>,
+        resource: Option<&str>,
+        actor_token: Option<&str>,
+        actor_token_type: Option<&str>,
+        req_scopes: Option<&BTreeSet<String>>,
+        ct: Duration,
+    ) -> Result<AccessTokenResponse, Oauth2Error> {
+        if matches!(o2rs.type_, OauthRSType::Basic { .. }) && !client_authentication_valid {
```
Firstyear: We may have to think about this one, because in this case it means to go from service-account-token to oauth2-session, you always need the oauth2 client credentials. However, we should potentially be looking at validation of the api session token first to allow service-accounts to do the exchange without the client credentials. This would still be secure since you still need the service-account api token to do the exchange, AND you need to meet the scope rules. 

What do you think @yaleman ?

Firstyear (changes requested):
This looks really good! Please add yourself to CONTRIBUTORS.md as well. I've added some comments to help out. I think we only need to discuss the use of client credentials on the initial exchange, and what the pros/cons are to that.

proto/src/oauth2.rs
TheSylex: I switched the String for an Enum based on the RFC names.

server/lib/src/idm/server.rs
TheSylex: I first thought I got you but I didn't. Now I get you, fixed in the commit.

Firstyear: Yeah it seems counter intuitive to skip the expiration check, but the expiry by this point is already validated by other means.

server/lib/src/idm/oauth2.rs
Firstyear: After sleeping on this, I think that we should not require the oauth2 client credentials during this exchange. 

When a user is authorising via OAuth2, they have their UAT and using that they are granted an access and refresh token. I think the service-account equivalent being that possession of the service account token should allow granting of access and refresh tokens. 

Given that we still apply the same rules to enforce that you must be granted at least one scope, and you have to be a member of a group to receive that scope, then this seems reasonable to me.

server/lib/src/idm/oauth2.rs
```diff
@@ -1594,6 +1621,169 @@ impl IdmServerProxyWriteTransaction<'_> {
...
+                    security_info!(
+                        ?err,
+                        "Service account api token rejected during token exchange"
+                    );
+                    Oauth2Error::InvalidRequest
+                }
+                err => Oauth2Error::ServerError(err),
+            })?;
+
+        let req_scopes = req_scopes.cloned().unwrap_or_default();
```
Firstyear: Re-reading this, we should break this out and make it generic with https://github.com/kanidm/kanidm/blob/master/server/lib/src/idm/oauth2.rs#L2146 so that we ensure that our scope check logic is consistent.

Note that I think your check isn't correct here - you need to check that req_scopes is not empty before we proceed. I'd take the scopes code from https://github.com/kanidm/kanidm/blob/master/server/lib/src/idm/oauth2.rs#L1975 as the basis of our scope check function, and then we use that. I'm open to discussing the intersection vs subset scope check though, but perhaps that should be it's own PR/issue.

Firstyear: The reason is that in both cases (oauth2_authorisation and this new token grant) the client side can always modify the set of scopes. So we need to guarantee *at least one* scope is present so that we enforce that at least *one group membership* is present to the identity. If we allow empty scopes, then any service-account/user can login to any oauth2 service, and oauth2 clients/resource servers *don't* check scopes, only the fact that the access token was granted proves authorisation. 

That's also why currently we enforce you request all scopes you request present, not just a subset. Because if you request "openid admin" and only granted "openid", then that could confuse some poorly written oauth2 client apps (kani does the right thing and narrows the scopes correctly though).

server/lib/src/idm/oauth2.rs
TheSylex: That sounds both right and more convenient and secure. I took the liberty to add a hard error for those users providing the client secrets in the request, in order to prevent people mangling unneeded client secrets around. Also noted it in the docs.

TheSylex: Most probably the last changes end this comment chain, but have a look nonetheless.

server/lib/src/idm/oauth2.rs
TheSylex: Made the request force you to explicitly provide the requested scopes and also generalized the function to be used in both contexts.

Firstyear:
Legend, looks awesome!

maciek134:
image

@TheSylex forgot to check I guess?

TheSylex:
> image
> @TheSylex forgot to check I guess?

I aided myself using LLMs to understand how to implement the changes in the codebase, as it is pretty huge. So the PR has some LLM code, but I manually went through all of it. So I didn't check it out of honesty, although I assume that checkbox is there to prevent piling up dogfood vibecoded PRs that just pile tech debt.

TheSylex:
added a last commit fixing the clippy issues and run cargo fmt, it should be good to go as long as @Firstyear greenlights it :)

maciek134:
https://kanidm.github.io/kanidm/master/developers/index.html#get-involved

> [!IMPORTANT]
>
> Kanidm is unable to accept code that is generated by an AI for legal reasons. copilot and other tools that generate
> code in this way can not be used in Kanidm.

This was my primary reason for choosing Kanidm, so I hope this policy hasn't changed.

TheSylex:
I understand your concerns @maciek134. But rest assured the PR wasn’t generated by AI, and I am the sole author of it.

As I said earlier I went through the whole process manually, and just used AI to understand the key points of the codebase and draft an implementation. If you’re concerned with any part of my code feel free to tag me and I’ll happily discuss it, or apply any change.

I don’t think any part of the PR goes against the clause that you mentioned, and I took the route of implementing this feature myself because I saw that it was a low hanging fruit for a long time among the issues of the repo, as mentioned in the Gitter chat.

However I’m open to suggestions about how else this could be done in case it was deemed “inappropriate”, if you wish.

maciek134:
> So the PR has some LLM code

> the PR wasn’t generated by AI, and I am the sole author of it.

These two statements are incompatible, you can't be the sole author if some of the code was LLM-generated. But I'm just a passer-by with strong opinions on the ethics of the thing, it's up to the maintainers to decide.

Firstyear:
> This was my primary reason for choosing Kanidm, so I hope this policy hasn't changed.

This policy has NOT changed. 


Looking just at these two comments:

> I aided myself using LLMs to understand how to implement the changes in the codebase, as it is pretty huge. So the PR has some LLM code, but I manually went through all of it.

> As I said earlier I went through the whole process manually, and just used AI to understand the key points of the codebase and draft an implementation. If you’re concerned with any part of my code feel free to tag me and I’ll happily discuss it, or apply any change.

I can see @maciek134 point - It either has some LLM code, or it doesn't. 

I'm trying to look at this through a really "human" lens. Kanidm is a big project - probably one of the biggest projects in Rust that I'm aware of outside of the Rust compiler itself. It has a lot of parts, and lets be real, like any project the codebase has evolved and sometimes it can be hard to find what you need. 

Stepping into a huge codebase and knowing where to start is a tough ask. I personally have developed techniques to do this over a long career. I can "see" how someone with access to today's tools (e.g. AI/LLM) would use these tools to get an initial foothold and navigate the codebase. 

If someone uses AI/LLM to *navigate* and *explore* Kanidm, we can't stop them. It may not be how I or others prefer to work, but it's not for us to prescribe. 

As mentioned where this gets spicier is if AI/LLM were used to *generate* any code. There are concerns about that which as mentioned step into copyright/legal rules. However - by the same rule we also can't prove if code was AI generated. 

The check mark is the honour system - we want people to at least be honest so we can assess what's happened. In this case, I didn't actually follow our own process and review if the box was checked! 

For us the check mark and the rules we have in the project are to deter people from generating slop. But the reality is even if someone *did* use AI/LLM, we have no true way to identify and know that. We have pinged some PR's previously for being AI/LLM generated as they referenced API's and code that never existed, and it was obvious that the server didn't even compile. 

Here, it's obvious that @TheSylex has actually reviewed the RFC, the code, created working tests, and has engaged with the codebase in a way that at least to a reviewer shows that they are meaningfully creating the code in question. While reviewing I actually did not suspect AI/LLM was involved in the process. Most of my feedback has just been about the logic due to knowing more about some of the codebases interactions, how the RFC's work, etc. Coding style wise there aren't many comments, but each developer works differently and I tend to be pretty comment heavy. But again, lack-of-comments is not a crime, just a difference in developer approach. 

So my assessment in summary:

* If @TheSylex has used AI/LLM to summarise and explore Kanidm, we can not stop them from doing this. It does not break our projects rules as they exist today, and as an opensource project, there is pretty much a 100% chance our code is in models somewhere already. 
* If @TheSylex has used AI/LLM to *generate* any code, I would ask you rewrite it without the AI/LLM to ensure we meet our legal and ethical guidelines. 
* Once the PR is "AI/LLM free" then check the no AI box and we go from there. 
* As a reviewer I think we need a review checklist to follow to ensure that we don't overlook some of these things in future.

EDIT: I'm not the type to take punitive actions when there was obvious good intent - it's better to guide with kindness. So even if @TheSylex AI/LLM to generate code initially, provided they correct the mistake, and avoid doing it again in future, that's okay by me. 

Thanks  🦀🎄

AAG81:
I’ve been following this thread and had a small suggestion. To make the checklist even clearer for future contributors, would you consider changing the AI item to:  

**I confirm this code is my original work and does not violate the project's policy on third-party IP or unverified AI generation.**
@Firstyear

Firstyear:
> I’ve been following this thread and had a small suggestion. To make the checklist even clearer for future contributors, would you consider changing the AI item to:
> 
> **I confirm this code is my original work and does not violate the project's policy on third-party IP or unverified AI generation.** @Firstyear

Yep, that's a good suggestion, I'll consider it when I review the pr process.

maciek134:
> unverified AI generation

The distinction makes it sound like verified generation is ok though.

Firstyear:
> > unverified AI generation
> 
> The distinction makes it sound like verified generation is ok though.

Yeah, good point, because even verified generation is not okay.

TheSylex:
Thanks for the clarity on the topic @Firstyear, and appreciate you honoring my transparency, as I didn't purposely check the box for that particular reason. I've fully manually reviewed the PR again, reworded some of the comments and changed some parts of the code that were leftovers of LLM and weren't 100% up my alleyway. I'm checking the box and we can proceed. However, seeing how big of a deal this has came up to be, I'll say a few things.

- First of all, this was mostly a problem caused by my dumbass not reading the contribution docs, so I wasn't aware of the strict policy against using AI for any sort of thing related to the project. My fault there.
- I also must say that this policy is probably hurting the project more than it is helping it. Sure we can get entangled in ethics and laws but at the end of the day this is an OSS project, and people's free time and resources are pretty limited. Even more so for a project as big as this. Big corporations won't bulge when using AI to mass produce software, just like they didn't when they used all the public OSS code to train their current models.
- I could go on the good/bad debate further, but that's beside the scope of this PR. However I will say that Kanidm will probably lose a good chunk of contributions if people can't (properly) use these kind of tools for them, and even so it will probably also end up with people hiding their methods/lying due to conflictive lines of thinking.

In my honest opinion, I'd accept contributions that contain **good, maintainable code**, period. And I wouldn't care a single bit whether this code was manually written, generated by an statistical text prediction model or revealed to the author in a fever dream, as long as the code is good for the project.

Anyway, from now on I'm completely aware of the situation regarding this topic in this repo, and will take it into consideration for future contributions, have a good day ☺️

Firstyear (approved):

Firstyear:
Thank you!

AAG81:
@Firstyear I noticed a small typo in the PR title. It currently lists RFC 8963 but the code implements RFC 8693 (Token Exchange).

Could you please update the PR title to "Implement OIDC auth for service-accounts (RFC 8693)"

TheSylex:
Whoops, I just changed it 😅

Firstyear:
@TheSylex We have a page in the book for what standards we support, could be good to add this RFC to that list too in another PR :)

PR #4022, opened 2025-12-20
20251219 Uint64/Int64 syntax types

Firstyear:
Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -193,11 +193,11 @@ dependencies = [
 
 [[package]]
 name = "askama"
-version = "0.14.0"
+version = "0.15.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f75363874b771be265f4ffe307ca705ef6f3baa19011c149da8674a87f1b75c4"
+checksum = "bb7125972258312e79827b60c9eb93938334100245081cf701a2dee981b17427"
 dependencies = [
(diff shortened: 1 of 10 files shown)


server/lib/src/valueset/int64.rs
```diff
@@ -0,0 +1,204 @@
...
+use crate::schema::SchemaAttribute;
+use crate::valueset::{
+    DbValueSetV2, ScimResolveStatus, ValueSet, ValueSetResolveStatus, ValueSetScimPut,
+};
+use kanidm_proto::scim_v1::JsonValue;
+use smolset::SmolSet;
+
+#[derive(Debug, Clone)]
+pub struct ValueSetInt64 {
+    set: SmolSet<[i64; 1]>,
```
CEbbinghaus, line 11: I know there must be a good reason but why are we defining our set as a set of single element slices?

yaleman: Smolset's docs say:

> When the set is smaller than the size of the array A, all elements are stored inline, without heap allocation. 

... so as soon as it goes over a single value it ends up on the heap? It's not clear.

Firstyear: This is correct - it's so that in the common case, it's a smaller value and we only heap allocate if it grows. Generally most values of this type are singletons so it avoids heap allocs.

PR #4039, opened 2026-01-06
Prevent server crashing on requests with low log level

### LINKED ISSUE #3952: RUST_LOG usage crashes
### I did this

Set `RUST_LOG=warn` env.

### I expected the following

Only see logs of level `warn` or above.

### This happened instead

```
thread 'kanidmd-thread-pool' (7) panicked at /cargo/registry/src/index.crates.io-1949cf8c6b5b557f/tracing-forest-0.3.0/src/layer/id.rs:40:31:
The subscriber isn't in any spans
stack backtrace:
stream closed: EOF for auth/kanidm-0 (kanidm)
```

### Kanidm version details

- Output of `kanidm(d) version`: `1.8.0`
- Are you running it in a container? If so, which image/tag?: `kanidm/server:180`
- If not a container, how'd you install it: -
- Operating System / Version (On Unix please post the output of `uname -a`): Talos OS 1.11 / k8s

### Any other comments

I basically wanted to skip all the info logs - the config LogLevel stops at `info` and that's very chatty with a k8s liveness probe.

Firstyear:
This forces log_level info to always be set on the span generator so that an event uuid is always present.

Fixes #3952

Checklist

### DIFF
--- a/libs/sketching/src/otel.rs
+++ b/libs/sketching/src/otel.rs
@@ -1,17 +1,14 @@
-use std::time::Duration;
-
-use opentelemetry_otlp::{Protocol, WithExportConfig};
-
 use opentelemetry::{global, trace::TracerProvider as _, KeyValue};
-
+use opentelemetry_otlp::{Protocol, WithExportConfig};
 use opentelemetry_sdk::{
     trace::{Sampler, SdkTracerProvider},
     Resource,
 };
+use std::str::FromStr;
+use std::time::Duration;
 use tracing::Subscriber;
 use tracing_core::Level;
-
-use tracing_subscriber::{prelude::*, EnvFilter, Registry};
+use tracing_subscriber::{filter::Directive, prelude::*, EnvFilter, Registry};
 
 pub const MAX_EVENTS_PER_SPAN: u32 = 64 * 1024;
 pub const MAX_ATTRIBUTES_PER_SPAN: u32 = 128;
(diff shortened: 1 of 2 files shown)


libs/sketching/src/otel.rs
```diff
@@ -46,7 +43,13 @@ pub fn start_logging_pipeline(
...
     let forest_filter: EnvFilter = EnvFilter::builder()
         .with_default_directive(log_filter.into())
-        .from_env_lossy();
+        .from_env_lossy()
+        // Always force the event span to be generated at the correct level, regardless
+        // of what the user set.
+        .add_directive(
+            Directive::from_str("kanidmd_core::https::trace=info")
+                .expect("Invalid directive during log setup"),
+        );
```
yaleman, line 52: This is a **_filter_** that forces it to only log when getting messages at info and above, ignoring debug...?

Firstyear: It means that log/span events from https://github.com/kanidm/kanidm/blob/master/server/core/src/https/trace.rs and only that file will always be enabled if they are at info or higher. 

Since there are no debug/trace messages in that file, then it doesn't break anything. 

However it means that setting `RUST_LOG=warn` will now "work" because the span in that file will always be enabled since it's at log level INFO, but without creating info messages throughout the server. 

In otherwords, this forces that one file to *always* log/span at the level of INFO, while the rest of the server and the RUST_LOG filters can be whatever else you want.

yaleman: Ah, that makes it clearer for sure 💟

Firstyear: Yeah think of it like the axum router - the most "specific" level/directive takes precedence.

yaleman: Yeah, more just the intention of the change wasn't clear - push it down from warn to info instead of up from (potentially) debug to info etc. all good 😄

PR #4054, opened 2026-01-10
20260108 sync polish

Firstyear:
# Change summary

- As part of the SUSEID project (which I can now talk about! Yay!) there will be many synced accounts that use external OAuth2. These sessions are "read-only" wrt to Kanidm. This improves the handling of those sessions in the UI returning appropriate errors and messages about the state of the session. 

Checklist

### DIFF
--- a/book/src/customising.md
+++ b/book/src/customising.md
@@ -1,10 +1,5 @@
 # Customising
 
-> [!NOTE]
->
-> Currently theming options such as updating the CSS requires modifying the style.css file. This may be changed in the
-> future to make it easier to modify.
-
 Kanidm supports customising various aspects such as the site display name, site image, and display names and images for
 each application.
 
@@ -53,4 +48,17 @@ restrictions as the site image above.
 kanidm system oauth2 set-image <NAME> <file-path> [image-type] -D idm_admin
 
 kanidm system oauth2 remove-image <NAME> -D idm_admin
+```
+
+## Custom CSS
+
+You can mount a custom CSS file over `/hpkg/override.css` in the container to add your own
+styles. For example with Docker:
+
+```bash
+docker run \
+    -p 443:8443 \
+    -v kanidmd:/data \
+    --mount type=bind,src=./your/override.css,dst=/hpkg/override.css \
+    kanidm/server:latest
 ```
--- a/server/core/static/override.css
+++ b/server/core/static/override.css
@@ -1,0 +1,1 @@
+/* custom styles */
--- a/server/core/templates/base.html
+++ b/server/core/templates/base.html
@@ -20,6 +20,8 @@
 			defer></script>
 		<link rel="stylesheet"
 			href="/pkg/style.css?v=((crate::https::cache_buster::get_cache_buster_key()))" />
+		<link rel="stylesheet"
+			href="/pkg/override.css?v=((crate::https::cache_buster::get_cache_buster_key()))" />
 
 		(% block head %)(% endblock %)
 	</head>
--- a/server/core/templates/base_htmx.html
+++ b/server/core/templates/base_htmx.html
@@ -24,6 +24,8 @@
 			defer></script>
 		<link rel="stylesheet"
 			href="/pkg/style.css?v=((crate::https::cache_buster::get_cache_buster_key()))" />
+		<link rel="stylesheet"
+			href="/pkg/override.css?v=((crate::https::cache_buster::get_cache_buster_key()))" />
 
 		(% block head %)(% endblock %)
 	</head>


server/core/templates/user_settings_profile_partial.html
```diff
@@ -69,11 +69,17 @@
...
-        <a href="/ui/unlock" hx-boost="false">
-            <button class="btn btn-primary" type="button">Unlock Edit 🔒</button>
+        <a href="/ui/unlock" hx-boost="false"
+              (% if !rw_possible %)disabled(% endif %)
+          >
+            <button
+              class="btn btn-primary"
+              type="button"
+              (% if !rw_possible %)disabled(% endif %)
+              >Unlock Edit 🔒</button>
```
ToxicMushroom, line 85: I'm not the biggest fan of disabled buttons, the user might not know why it is disabled and get frustrated.
Maybe the user shouldn't be confronted with the fact that they're a synced user, but they could be told they have a read-only account ?

server/testkit/tests/testkit/proto_v1_test.rs
```diff
@@ -1725,8 +1725,7 @@ async fn test_server_user_auth_reauthentication(rsclient: &KanidmClient) {
         .map(|jws| jws.from_json::<UserAuthToken>().expect("Invalid json"))
         .expect("Unable extract uat");
 
-    let now = time::OffsetDateTime::now_utc();
-    assert!(!uat.purpose_readwrite_active(now));
+    assert!(matches!(uat.purpose, UatPurpose::ReadWrite { .. }));
```
ToxicMushroom, line 1728: Does this still pass ? 
The previous assertion looks like an inverted version of the new one.

tools/cli/src/cli/common.rs
```diff
@@ -269,12 +270,20 @@ impl KanidmClientParser {
                 match optype {
                     OpType::Read => {}
                     OpType::Write => {
-                        if !uat.purpose_readwrite_active(now_utc + time::Duration::new(20, 0)) {
-                            error!(
-                                "Privileges have expired for {} - you need to re-authenticate again.",
-                                uat.spn
-                            );
-                            return Err(ToClientError::NeedReauth(spn, client));
+                        match uat.purpose_privilege_state(now_utc + time::Duration::new(20, 0)) {
```
ToxicMushroom: Not related to this PR but I feel Duration::from_secs(20) is more appropriate

server/core/src/https/views/reauth.rs
```diff
@@ -0,0 +1,112 @@
...
+use askama_web::WebTemplate;
+use axum::response::IntoResponse;
+use axum::response::Response;
+use axum_extra::extract::cookie::CookieJar;
+use kanidm_proto::internal::{PrivilegesActive, UserAuthToken};
+use kanidmd_lib::idm::server::DomainInfoRead;
+use kanidmd_lib::prelude::ClientAuthInfo;
+use uuid::Uuid;
+
+const READ_WRITE_BUFFER: i64 = 60;
```
ToxicMushroom: When seeing this variable I thought it'd be for a buffer_size. Could be fixed with a comment.
Or even better make the constant of type `Duration`.
These functions are const anyway https://doc.rust-lang.org/std/time/struct.Duration.html#method.from_secs

proto/src/internal/error.rs
```diff
@@ -64,7 +64,8 @@ pub enum ConsistencyError {
 #[derive(Serialize, Deserialize, Debug, ToSchema)]
 #[serde(rename_all = "lowercase")]
 pub enum OperationError {
-    // Logic errors, or "soft" errors.
+    // Logic errors, or "soft" errors. These are to guide the user or user-interface
+    // in someway.
```
yaleman:
```suggestion
    // in some way.
```

server/core/src/https/views/reauth.rs
yaleman: Yep, probably needs to be renamed to RW_REAUTH_WINDOW_SECS or something and document a little how/why we're using that window.

server/core/templates/reauth_readonly.html
```diff
@@ -0,0 +1,30 @@
...
+			alt="(( domain_info.display_name() ))" class="kanidm_logo" />
+	(% else %)
+			<img
+				src="/pkg/img/logo-square.svg?v=((crate::https::cache_buster::get_cache_buster_key()))"
+				alt="(( domain_info.display_name() ))" class="kanidm_logo" />
+	(% endif %)
+	<h3>(( domain_info.display_name() ))</h3>
+
+	<h2>Error</h2>
+		<p>The session you are logged in as is Read Only and may not perform this operation.</p>
```
yaleman:
```suggestion
		<p>Your session is Read Only and may not perform this operation.</p>
```

server/core/templates/user_settings_profile_partial.html
yaleman: Yep, just don't render the button at all if it's read-only TBH.

tools/cli/src/cli/common.rs
yaleman: Agreed on this one, easier to identify Duration's scale.

server/core/templates/user_settings_profile_partial.html
Firstyear: No - controls and buttons should always be visible and shown as disabled. Else it confuses the user to why sometimes things are there and sometimes not. So this is correct to disable things as it informs the user the control still exists, but that they currently can't access it. (This is per human interface psychology guidelines).

I agree we need to link to the external source where the account can be managed though.

server/testkit/tests/testkit/proto_v1_test.rs
Firstyear: Yes it passes, because the point is that the session is readwrite *capable* not that it's *currently* readwrite.

tools/cli/src/cli/common.rs
Firstyear: This is time::Duration not std::time::Duration, so it does not have "from_secs", only "new".

server/core/templates/user_settings_profile_partial.html
Firstyear: Screenshot 2026-01-13 at 10 13 37

Trivial example - this way the menu stays consistent and controls are disabled since they are not available in this context, and become enabled when they are. Same here.

yaleman: Yeah, makes sense when put that way - maybe we need a mouse over alt-text with an explanation in that case?

Firstyear: Yeah, can add a tooltip that says "the current session is read-only" or something.

tools/cli/src/cli/common.rs
yaleman: Then it's just `time::Duration::seconds(20)` 😄 

https://docs.rs/time/latest/time/struct.Duration.html#method.seconds

yaleman (approved):

PR #4055, opened 2026-01-10
Add cfg management thoughts

Firstyear:
# Change summary

- Add design doc for entry configuration management.

Checklist

### DIFF
--- a/book/src/developers/designs/config-management-entries.md
+++ b/book/src/developers/designs/config-management-entries.md
@@ -1,0 +1,88 @@
+# Config Management Entries
+
+One of the most requested features (especially from the nix community 📆 ) is the ability to config
+manage entries that exist in Kanidm.
+
+To this point, we have not had a way to do this - there are many reasons, but part is that we wanted
+to do this "right". Another part is we didn't have a good approach in mind that would work with a
+distributed system like Kanidm.
+
+But guess what, we're finally doing it. 🎉
+
+> NOTE this design does NOT account for secret management - yet. That will be a later/subsequent change.
+
+## The Risks
+
+Kanidm is by it's nature, a distributed system with eventually consistent replication. This makes
+it very difficult to *manage* entries because any difference between multiple nodes in the
+system will cause a "fight" where the nodes will compete for the state of the entry.
+
+As a result, we need a way to allow entry creation that can work in a distributed system, where there
+may be *inconsistent* application of the configurations.
+
+This presents a challenge - how can we ensure a total *ordering* of the applied configurations, while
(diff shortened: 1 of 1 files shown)


Firstyear:
Very rough draft at this point, but will help @tacerus a lot.

book/src/developers/designs/config-management-entries.md
```diff
@@ -0,0 +1,88 @@
...
+also ensuring that updates take precedence? How can we also ensure that human changes that are made
+are preserved in this system? And how do we ensure that updates to the configuration are applied
+as expected over existing data?
+
+## The Design
+
+There will be a drop in directory where SCIM formatted JSON entries can be placed. These will be based
+on a modified SCIM "assert" that combines create/put to assert a state in an entry. If the entry
+does not exist it will be created. If it does exist it will be altered to be inline with the state
+of the assertion.
```
tacerus, line 33: Will it also cover deletions, so that if an entry is no longer present in the dataset, it would be removed?
It would be nice if it did handle deletions but if there was a way to have a way to ignore a pattern of entries (maybe some attribute?), for example in the case of id.o.o, we would want the configuration management to only enforce entries not managed by self-service or federation.

book/src/developers/designs/config-management-entries.md
```diff
@@ -0,0 +1,88 @@
...
+
+When Kanidm starts, after migrations are applied this directory will be listed in order and the files
+applied. A *hash* of the applied file is taken and stored in the database.
+
+When Kanidm is *restarted*, any configuration file that has already been applied, and who's hash
+matches in the database will *not be re-applied*.
```
tacerus, line 39: Would it be possible to have also have a command to trigger a hot reload to avoid needing to restart the server for changes?

And, since the hash applies to the whole file, would this re-apply operation be idempotent and only modify attributes which no longer match?

book/src/developers/designs/config-management-entries.md
Firstyear: No, it won't cover pattern management. There are already other controls in Kanidm to *prevent* modification of sync/federation entries.

Firstyear: Deletion is going to be available though.

book/src/developers/designs/config-management-entries.md
Firstyear: Yes that's fine, a command to reload is possible.

Yes to the hash applying the whole file, and idempotent. It will function as an "assert" that the attributes you list into the file must reflect "as described" in the DB.

tacerus: Sounds good, thanks for clarifying!

PR #4075, opened 2026-01-16
20260116 kanidmd json mode

### LINKED ISSUE #4040: Extra data in JSON output
### I did this

```
geeko@e0f583e85fe8 ~> sudo kanidmd recover-account -o json admin
📜 Using config file: "/etc/kanidm/server.toml"
00000000-0000-0000-0000-000000000000 INFO     ｉ [info]: Starting Kanidmd | version: 1.8.4
00000000-0000-0000-0000-000000000000 WARN     🚧 [warn]: This is running as uid == 0 (root) which may be a security risk.
00000000-0000-0000-0000-000000000000 WARN     🚧 [warn]: WARNING: DB folder permissions on /var/lib/kanidm indicate it may not be RW. This could cause the server start up t
o fail!
00000000-0000-0000-0000-000000000000 INFO     ｉ [info]: Running account recovery ...
{"password":"VuFZXQQPtk2BGE6YWv2CdatqSBeSyEJH9Z92Kj24v3BUadSa"}
```

### I expected the following

Being able to parse the full output as JSON.

### This happened instead

Log lines are printed to stdout out, causing a typical JSON parser to fail.

```
geeko@e0f583e85fe8 ~> sudo kanidmd recover-account -o json admin | jq .
📜 Using config file: "/etc/kanidm/server.toml"
parse error: Invalid numeric literal at line 1, column 37

thread 'main' (18785) panicked at /home/abuild/rpmbuild/BUILD/kanidm-1.8.4~git0.2449805e3/vendor/tracing-forest-0.3.0/src/layer/mod.rs:278:18:
Processing logs failed: Error { tree: Event(Event { shared: Shared { uuid: 00000000-0000-0000-0000-000000000000, level: Level(Warn), fields: [] }, message: Some("WARNING: D
B folder permissions on /var/lib/kanidm indicate it may not be RW. This could cause the server start up to fail!"), tag: None }), source: Os { code: 32, kind: BrokenPipe, m
essage: "Broken pipe" } }
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

(relevant here is `jq` returning `parse error`, the panic from kanidm is just a side effect of the pipe)

### Kanidm version details

```
geeko@e0f583e85fe8 ~> kanidmd version
kanidmd 1.8.4
```

- Are you running it in a container? If so, which image/tag?: bci/15.7

- If not a container, how'd you install it: installed from packages in `network:idm`

- Operating System / Version (On Unix please post the output of `uname -a`): SLES 15.7

### Any other comments

This was found while attempting to automate the bootstrap process.

Currently the issue can be worked around with some filtering:

```
geeko@e0f583e85fe8 ~> sudo kanidmd recover-account -o json admin | grep '^{'|jq
📜 Using config file: "/etc/kanidm/server.toml"
{
  "password": "43Ak6fGqJR0E6wv7t4LUETG8V5tJ4hqed3BDYNCXvUGhdHH7"
}
```

This works, but is a bit fragile.
The "Using config file" line is fine, because it is printed to stderr and hence not affected by the pipe.
Decreasing the log level did not work, I think because `info` is the minimum.

Firstyear:
# Change summary

The existing design of the Kanidmd CLI interface conflated the interaction of both a human and a machine. This caused a lot of friction as it wasn't always clear what mode you were in, but most importantly most paths *assumed* you were a human or server mode and would setup logging that would break machine parsable interfaces.

Instead of trying to separate machine/human by a command line flag, they should be *completely* separate command line paths so that from the very start of operation we know which path we are taking. This allows those paths to correctly and accurately control things like logging and output modes.

This changes the interface to remove the ConsoleOutputMode flag, and instead adds a subcommand for "scripting" on the Kanidmid interface that is completely separate. This provides a strong isolation between human and machine interfaces, and means that in each path we always know what we are talking to and how we should respond as a result.

Initially this only covers healthcheck and recover-account, but we can add other machine appropriate interfaces in future. 

Fixes #4040

Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -175,9 +175,12 @@ dependencies = [
 
 [[package]]
 name = "arc-swap"
-version = "1.7.1"
+version = "1.8.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "69f7f8c3906b62b754cd5326047894316021dcfe5a194c8ea52bdd94934a3457"
+checksum = "51d03449bb8ca2cc2ef70869af31463d1ae5ccc8fa3e334b307203fbf815207e"
+dependencies = [
+ "rustversion",
+]
 
 [[package]]
 name = "argon2"
@@ -308,13 +311,12 @@ dependencies = [
 
 [[package]]
 name = "async-compression"
-version = "0.4.35"
+version = "0.4.37"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "07a926debf178f2d355197f9caddb08e54a9329d44748034bba349c5848cb519"
+checksum = "d10e4f991a553474232bc0a31799f6d24b034a84c0971d80d2e2f78b2e576e40"
 dependencies = [
  "compression-codecs",
  "compression-core",
- "futures-core",
  "pin-project-lite",
  "tokio",
 ]
@@ -393,9 +395,9 @@ checksum = "c08606f8c3cbf4ce6ec8e28fb0014a2c086708fe954eaa885384a6165172e7e8"
 
 [[package]]
 name = "aws-lc-rs"
-version = "1.15.1"
+version = "1.15.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6b5ce75405893cd713f9ab8e297d8e438f624dde7d706108285f7e17a25a180f"
+checksum = "e84ce723ab67259cfeb9877c6a639ee9eb7a27b28123abd71db7f0d5d0cc9d86"
 dependencies = [
  "aws-lc-sys",
  "zeroize",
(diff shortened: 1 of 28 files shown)


yaleman:
This looks like a good start, I'd say we should have a struct-based error response because even in this first example there's inconsistencies - some errors have a status field and some don't etc... 

I'd suggest a simple two-field struct for responses - the existence of `error` meaning something has gone wrong is implied, we don't need an additional `status` field... and less hand-crafting JSON 😄 

```rust
struct ScriptingMessage<R: Serialize> {
    result: R,
    #[serde(skip_serializing_if=None)] // y'know what I mean here
    error: Option<String>
}
```

eg:
```json
{
    "error": "Server-side error during task processing.",
    "result": "You should inspect the logs.",
}
```

eg2:

```json
{
    "result": {
         // or this could just be the password string
         "password" : "the new password" 
    }
}
```

Firstyear:
Yep, I'll change that.

Firstyear:
@yaleman I made the output a bit more structured and reusable, let me know what you think.

scripts/setup_dev_environment.sh
```diff
@@ -85,7 +85,7 @@ export KANIDM_CONFIG="./insecure_server.toml"
 IDM_ADMIN_USER="idm_admin@localhost"
 
 echo "Resetting the idm_admin user..."
-IDM_ADMIN_PASS_RAW="$(${KANIDMD} recover-account idm_admin -o json 2>&1)"
+IDM_ADMIN_PASS_RAW="$(${KANIDMD} scripting recover-account idm_admin2>&1)"
```
yaleman:
```suggestion
IDM_ADMIN_PASS_RAW="$(${KANIDMD} scripting recover-account idm_admin 2>&1)"
```

tacerus:
Neat solution!

PR #4105, opened 2026-02-03
FIX: make tracing-forest stop panic'ing things when enabling otel

yaleman:
# Change summary

- now the span ID based on the active logging pipeline (don't use tracing-forest for otel, cleaning up filtering etc
- reduced the number of creations of the span creator struct
- added more docs

Checklist

### DIFF
--- a/server/core/src/https/trace.rs
+++ b/server/core/src/https/trace.rs
@@ -90,10 +90,9 @@ impl<B> tower_http::trace::OnResponse<B> for SpanCreator {
                 ))),
             }
         } else {
-            // tracing forest cannot handle u16 directly
             span.record(
                 self.log_engine.status_code_field(),
-                response.status().as_u16().to_string(),
+                response.status().as_u16() as u64,
             );
             // don't need these in otel because they're alreayd in th
             span.record("latency", latency.as_millis());


libs/sketching/src/pipeline.rs
```diff
@@ -0,0 +1,175 @@
...
+        };
+
+        let version = format!("{}{}", env!("CARGO_PKG_VERSION"), git_rev);
+        let hostname = gethostname::gethostname();
+        let hostname = hostname.to_string_lossy();
+        let hostname = hostname.to_lowercase();
+
+        let mut resource = Resource::builder().with_schema_url(
+            [
+                // TODO: it'd be really nice to be able to set the instance ID here, from the server UUID so we know *which* instance on this host is logging
```
Firstyear, line 116: It's too early in the pipeline here that's the issue. We need logging setup to startup the BE, and the BE holds the server uuid. 

But also, the server uuid *can change* over time, so it's not a great idea to try and track based on this. You'd be better with a hostname or something human identifiable.

yaleman: yeah, hence the "it'd be really nice" - the hostname's already in the metadata either way so that's fine

PR #4111, opened 2026-02-04
Allow reseting softlocks

### LINKED ISSUE #3456: Command to reset a soft lock

Sometimes a user gets softlocked because of some dumb reason. For example, say you're using PAM or LDAP auth for a mail server. A user enters the wrong password, and then their very smart and clever mail client retries every five minutes all day and gets their account softlocked. I see this happen a *lot* with mail servers.

Right now it seems painful to fix this, I couldn't find any simple way to get the user back in without having them reset their password or restarting kanidm. When this kind of thing happens, the helpdesk should be able to clear the softlock and have the user try again (after asking them five times if they're sure they have the right password).

I think a command like `kanidm person validity unlock` would fit the bill perfectly, and of course in the future this would be nice to have as a button in an admin web interface.

Shuroii in Matrix suggested a password reset; in some cases that would be a good approach. But consider this scenario: if the user has more than one mail client and resets their password to clear a lock, there's a good chance that the other mail client with the now-incorrect password will get their account locked again!

This can be a hassle and in my days as a Red Hat IDM admin it wasn't unusual for the help desk to have to stay on the phone with a user and unlock their account a couple of times before everything was smoothed out. Another reason to have the ability to easily reset a softlock, basically.

Some IDMs/organizations also solve this with a self-service unlock option. I think that'd be a good idea for the specific case of UNIX auth, since web auth more thoroughly authenticates the users. If a user logs in to the web interface when they are softlocked for UNIX/LDAP/RADIUS, maybe they should see a banner telling them their account is locked with a button to reset it. That's a longer-term ask though.

Firstyear:
# Change summary

- Allows reseting sotflocks. 

Fixes #3456

Checklist


server/lib/src/idm/server.rs
```diff
@@ -1242,8 +1242,15 @@ impl IdmServerAuthTransaction<'_> {
...
-                            // Apply the current time.
-                            slock.apply_time_step(ct);
+
+                            let softlock_expire_odt = auth_session.account().softlock_expire();
+
+                            let softlock_expire = softlock_expire_odt
+                                .map(|odt| odt.unix_timestamp() as u64)
+                                .map(Duration::from_secs);
+
+                            // Apply the current time, which resets the softlock if needed.
```
yaleman: getting my head-canon right, does a successful auth event *clear* a softlock, or reset it to *longer*? "reset" isn't clear in my head

tools/cli/src/opt/kanidm.rs
```diff
@@ -400,6 +400,16 @@ pub enum AccountCredential {
...
     },
+    /// Reset the softlocks on this account. This applies to all credentials of the account.
+    #[clap(name = "softlock-reset")]
+    SoftlockReset {
+        account_id: String,
+        #[clap(name = "datetime", default_value = "now", verbatim_doc_comment)]
+        /// This accepts multiple options:
+        /// - An RFC3339 time of the format "YYYY-MM-DDTHH:MM:SS+TZ", "2020-09-25T11:22:02+10:00"
+        /// - "now" to reset immediately
+        datetime: String,
```
yaleman, line 411: feels like that should just be optional and default to 'now'? - doesn't seem likely that if you're going to reset a softlock, that it'll be "later"

server/lib/src/idm/server.rs
Firstyear: It CLEARS it. I'll change the wording.

tools/cli/src/opt/kanidm.rs
Firstyear: It does, the clap attributes make it optional and defaulting to now.

yaleman: oh, derp.

PR #4088, opened 2026-01-23
20260122 SCIM batch

Firstyear:
# Change summary

- This adds an "assert" operation to the server, which allows repeatable application of changes. This is the foundation of both scim batch operations, but also a method to allow "configuration management" of entries. 

Relates #4055 #4052 #1476

Checklist

- [x] This PR contains no AI generated code
- [ ] book chapter included (if relevant)
- [ ] design document included (if relevant)

PR #4124, opened 2026-02-10
added ExecReload statement to systemd service

domrim:
# Change summary

Even without the support of `notify-reload` in kanidmd (as it was never properly working and was removed in #3885), it could support systemd reload by just providing the `ExecReload` statement in the service configuration. I tried this fix at my instance and it seems to work perfectly fine.

Checklist


Firstyear:
This is actually already resolved in a better way with https://github.com/kanidm/kanidm/pull/4088/changes#diff-c867c25008b7b8e4228c4b70dfaf790072387f2a9b82a1ae351852c14bb194c0

PR #4133, opened 2026-02-12
Allow LDAP CA verification to be disabled in sync

### LINKED ISSUE #4122: Can we get `verify_ca = false` option for `kanidm-ipa-sync`?

As started in https://github.com/kanidm/kanidm/issues/4064, I'm not able to resolve my `TlsError` ([from here](https://github.com/kanidm/kanidm/issues/4064#issuecomment-3754352632)).

We already have the option `verify_ca = false` for tools, why not have it for `kanidm-ipa-sync` too?
It's even safer in this context.

I have made 3 attempts to resolve the error, but none have succeeded. It works fine with `1.7.4`, doesn't complain on my chain. I can't change the certificate of my IPA instance (it's still working, and there are no issues with it), but I have tried all possible combinations of chains that I send via `ipa_ca` and the error always persists.

### Additional context

Now I'm wondering, if I only want to get data out of FreeIPA, why do I care that much about the TLS error, except that there is no option to skip it?

Firstyear:
Fixes #4122

Checklist

### DIFF
--- a/book/src/entry_management.md
+++ b/book/src/entry_management.md
@@ -33,7 +33,7 @@ Attributes of an entry can be removed by setting them to `null`.
 ```
 
 This example is located in
-[examples/migrations](https://github.com/kanidm/kanidm/blob/master/examples/migrations/) in the repository.
+[examples/migrations](https://github.com/kanidm/kanidm/blob/master/examples/migrations/) in the repository. There are other migrations there you may find useful.
 
 ## Recommendations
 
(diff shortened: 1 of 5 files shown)


tools/iam_migrations/freeipa/src/config.rs
```diff
@@ -9,7 +9,9 @@ pub struct Config {
     pub schedule: Option<String>,
     pub status_bind: Option<String>,
     pub ipa_uri: Url,
-    pub ipa_ca: String,
+    #[serde(default)]
+    pub ipa_verify_ca: bool,
```
yaleman: So, defaulting to not verifying the CA? (on both)

Firstyear: OPPS that's not what I want.

yaleman (dismissed):

PR #4151, opened 2026-02-19
20260218 credential reset

Firstyear:
# Change summary

This adds the ability to send credential reset links to users. The reset link must be sent to a registered email of the account which we check during intent token creation. Additionally, since we now are sending links, document the mail-sender. Finally add the ability to revoke credential links from the sent mail. 

Checklist

### DIFF
--- a/.github/workflows/docker_build_kanidm.yml
+++ b/.github/workflows/docker_build_kanidm.yml
@@ -40,12 +40,12 @@ jobs:
         with:
           persist-credentials: false
       - name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@v3
+        uses: docker/setup-buildx-action@v4
       - name: Docker metadata
         id: meta
-        uses: docker/metadata-action@v5
+        uses: docker/metadata-action@v6
       - name: Build kanidm
-        uses: docker/build-push-action@v6
+        uses: docker/build-push-action@v7
         with:
           platforms: "linux/amd64"
           tags: ghcr.io/${{ needs.set_tag_values.outputs.owner_lc }}/kanidm:devel,ghcr.io/${{ needs.set_tag_values.outputs.owner_lc }}/kanidm:${{ needs.set_tag_values.outputs.ref_name}}
--- a/.github/workflows/docker_build_kanidmd.yml
+++ b/.github/workflows/docker_build_kanidmd.yml
@@ -40,12 +40,12 @@ jobs:
         with:
           persist-credentials: false
       - name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@v3
+        uses: docker/setup-buildx-action@v4
       - name: Docker metadata
         id: meta
-        uses: docker/metadata-action@v5
+        uses: docker/metadata-action@v6
       - name: Build kanidmd
-        uses: docker/build-push-action@v6
+        uses: docker/build-push-action@v7
         with:
           platforms: "linux/amd64"
           tags: ghcr.io/${{ needs.set_tag_values.outputs.owner_lc }}/kanidmd:devel,ghcr.io/${{ needs.set_tag_values.outputs.owner_lc }}/kanidmd:${{ needs.set_tag_values.outputs.ref_name}}
--- a/.github/workflows/docker_build_radiusd.yml
+++ b/.github/workflows/docker_build_radiusd.yml
@@ -40,14 +40,14 @@ jobs:
         with:
           persist-credentials: false
       - name: Set up QEMU
-        uses: docker/setup-qemu-action@v3
+        uses: docker/setup-qemu-action@v4
       - name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@v3
+        uses: docker/setup-buildx-action@v4
       - name: Docker metadata
         id: meta
-        uses: docker/metadata-action@v5
+        uses: docker/metadata-action@v6
       - name: Build radius
-        uses: docker/build-push-action@v6
+        uses: docker/build-push-action@v7
         with:
           platforms: linux/arm64,linux/amd64
           tags: ghcr.io/${{ needs.set_tag_values.outputs.owner_lc }}/radius:devel,ghcr.io/${{ needs.set_tag_values.outputs.owner_lc }}/radius:${{ needs.set_tag_values.outputs.ref_name}}
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -175,9 +175,9 @@ dependencies = [
 
 [[package]]
 name = "arc-swap"
-version = "1.8.1"
+version = "1.8.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "9ded5f9a03ac8f24d1b8a25101ee812cd32cdc8c50a4c50237de2c4915850e73"
+checksum = "f9f3647c145568cec02c42054e07bdf9a5a698e15b466fb2341bfc393cd24aa5"
 dependencies = [
  "rustversion",
 ]
@@ -311,9 +311,9 @@ dependencies = [
 
 [[package]]
 name = "async-compression"
-version = "0.4.39"
+version = "0.4.41"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "68650b7df54f0293fd061972a0fb05aaf4fc0879d3b3d21a638a182c5c543b9f"
+checksum = "d0f9ee0f6e02ffd7ad5816e9464499fba7b3effd01123b515c41d1697c43dad1"
 dependencies = [
  "compression-codecs",
  "compression-core",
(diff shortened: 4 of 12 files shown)


book/src/email_setup.md
```diff
@@ -0,0 +1,105 @@
+# Outgoing Email
+
+> "All services evolve to the point they eventually can send email" - some wise nerd, probably.
+
+Kanidm can be configured to optionally send messages to users. This is important for features such
+as sending credential reset links.
+
+## Architecture
+
+Kanidm internally maintains a message queue in it's the database. This allows all servers to queue
```
yaleman:
```suggestion
Kanidm maintains a message queue in its database. This allows all servers to queue
```

book/src/email_setup.md
```diff
@@ -0,0 +1,105 @@
...
+
+kanidm group add-members idm_message_senders mail-sender
+```
+
+The service account must have an api token with read-write privileges.
+
+```bash
+kanidm service-account api-token generate ACCOUNT_ID LABEL [EXPIRY] --readwrite
+kanidm service-account api-token generate mail-sender "mail sender token" --readwrite
+```
```
yaleman:
```suggestion

> [!WARNING]
>
> The mail-sender service account should only be a member of `idm_message_senders` - never add them to any other group - use another account for other purposes!
```

... this needs formatting, but I feel like we need to hit people over the head with it.

book/src/email_setup.md
```diff
@@ -0,0 +1,105 @@
...
+```bash
+kanidm system message-queue send-test-message TO_ACCOUNT
+kanidm system message-queue send-test-message ellie
+```
+
+Once inserted, the message can be viewed in the queue.
+
+
+```bash
+kanidm system message-queue list
```
yaleman:
```suggestion
$ kanidm system message-queue list
```

tools/cli/src/opt/kanidm.rs
```diff
@@ -400,6 +400,20 @@ pub enum AccountCredential {
...
+        account_id: String,
+
+        /// Optionally set how many seconds the reset token should be valid for.
+        /// Default: 3600 seconds
+        ttl: Option<u64>,
+
+        /// Optionally specify the email the token should be sent to. This email address
+        /// must exist on the account for the reset to be sent.
+        alternate_email: Option<String>,
+    },
```
CEbbinghaus, line 416: for the CLI it should be that an admin can send the email to an alternate address. Self serve online and CLI should always validate the email, But if the user has lost access to the email then an admin should be able to send an email to a different acount IMO.

This may just be me tho

server/lib/src/idm/credupdatesession.rs
```diff
@@ -969,6 +977,55 @@ impl IdmServerProxyWriteTransaction<'_> {
...
+
+        // If the email is set, validate it.
+
+        let to_email = if let Some(to_email) = event.email {
+            account.mail().contains(&to_email)
+                .then_some(to_email)
+                .ok_or_else(|| {
+                    error!(spn = %account.spn(), "Requested email address is not present on account, unable to send credential reset.");
+                    OperationError::CU0007AccountEmailNotFound
+                })
```
CEbbinghaus, line 997: Won't this take longer if no email matches thereby allowing the sidechannel? or is there some other mechanism that guarantees we always return in constant time?

tools/mail_sender/src/main.rs
```diff
@@ -66,19 +69,66 @@ struct MessageStatus {
...
+                format!(
+                    r#"Hi {display_name},
+
+This is your credential reset link:
+
+{reset_url}
+
+This link will expire at {pretty_expiry_time}
+
+If you did not request this message, please click the following link:
```
CEbbinghaus:
```suggestion
If you did not request this message, please revoke the reset using the following link:
```

Not sure that "revoke the reset" is great but we shouldn't leave it at just "Click the link".

Its already raising alarm bells for scams and phishing for me

CEbbinghaus (commented):
Less concrete problems, more opinions and questions

server/core/src/https/v1.rs
```diff
@@ -1231,6 +1231,7 @@ pub async fn person_id_credential_update_get(
     tag = "person/credential",
 )]
 // TODO: this shouldn't be a get, we're making changes!
+// fy: Ahhh yes, APIs - where regrets become enshrined in code.
```
yaleman, line 1234: isn't this new enough that we can just torch it and do it properly?

server/core/src/https/v1.rs
yaleman: plz put the body in ... I think this is it?
```suggestion
        (status=200, body=CUIntentToken),
```

server/lib/src/server/identity.rs
yaleman: kinda gives me the heebie-jeebies that we're losing the session along the way for auditing purposes

tools/mail_sender/src/config.rs
```diff
@@ -1,14 +1,18 @@
...
 #[derive(Debug, Deserialize)]
 pub struct Config {
     pub token: String,
     pub schedule: Option<String>,
-    // pub status_bind: Option<String>,
+
+    pub instance_display_name: String,
+    pub instance_url: Url,
+
+    // pub status_bind: Option<Address>,
```
yaleman:
```suggestion
```

tools/mail_sender/src/main.rs
```diff
@@ -66,19 +69,66 @@ struct MessageStatus {
...
+
+{revoke_url}
```
yaleman, line 123: I sure hope these links aren't poisoned by mail filters GET'ing them 😄

server/lib/src/idm/credupdatesession.rs
```diff
@@ -969,6 +977,55 @@ impl IdmServerProxyWriteTransaction<'_> {
...
+
+        let message = OutboundMessage::CredentialResetV1 {
+            display_name,
+            intent_id,
+            expiry_time,
+        };
+
+        self.qs_write.queue_message(
+            // Should we actually impersonate here? We probably need an account for internal sending
+            // that is disconnected from the act of creating the reset.
```
yaleman, line 1022: if we're only _**queueing**_ the message I'd say it's best to note it as the identity that triggered it

server/lib/src/idm/credupdatesession.rs
```diff
@@ -1040,15 +1112,71 @@ impl IdmServerProxyWriteTransaction<'_> {
...
-            expiry_time,
-        })
+    #[instrument(level = "debug", skip_all)]
+    pub fn revoke_credential_update_intent(
+        &mut self,
+        token: CredentialUpdateIntentTokenExchange,
+        _current_time: Duration,
+    ) -> Result<(), OperationError> {
+        let CredentialUpdateIntentTokenExchange { intent_id } = token;
+        // If given an intent_id, immediately transition it to consumed to prevent it's used.
```
yaleman:
```suggestion
        // If given an intent_id, immediately transition it to consumed to prevent reuse.
```

prevent reuse maybe?

server/core/src/https/v1.rs
Firstyear: That is an *old* existing api, it is internal to kani though.

server/lib/src/server/identity.rs
Firstyear: But there is no session internally, we aren't sourced from a users session in this flow.

tools/cli/src/opt/kanidm.rs
Firstyear: Then the admin can enrol the email to the account first, then hit the credential send button :)

tools/mail_sender/src/main.rs
Firstyear: I specifically wrote both forms so you have to click through them so they aren't poisoned :)

server/lib/src/idm/credupdatesession.rs
Firstyear: It's a permissions thing - users don't have the ability or permissions to queue messages, that's why we have to do the impersonate here as system to queue the message on behalf-of the account. The account at this point has to have been access control checked to ensure they are allowed to trigger a credential reset though.

server/lib/src/idm/credupdatesession.rs
Firstyear: You have to already have validated you have permission to reset the accounts permissions at this point. So you can't even start this process unless you have access to write their credentials.

Firstyear:
As a reminder, this isn't the *anonymous* reset version, this is the *admin triggered* and permission validated flow :)

tools/cli/src/opt/kanidm.rs
CEbbinghaus: hurr durrr but requires running two commands, I just want `kanidm work` ;)

Firstyear:
@yaleman Any other comments?

PR #4156, opened 2026-02-20
20260220 prevent migration accidents

Firstyear:
# Change summary

Prevent possible abuse of the new migration feature through introduction of an internal access control mechanism to limit what can be internally performed. 

Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -35,7 +35,7 @@ checksum = "b169f7a6d4742236a0a00c541b845991d0ac43e546831af1249753ab4c3aa3a0"
 dependencies = [
  "cfg-if",
  "cipher 0.4.4",
- "cpufeatures",
+ "cpufeatures 0.2.17",
 ]
 
 [[package]]
@@ -151,9 +151,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.101"
+version = "1.0.102"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5f0e0fee31ef5ed1ba1316088939cea399010ed7731dba877ed44aeb407a75ea"
+checksum = "7f202df86484c868dbad7eaa557ef785d5c66295e41b460ef922eca0723b842c"
 
 [[package]]
 name = "ar_archive_writer"
@@ -190,7 +190,7 @@ checksum = "3c3610892ee6e0cbce8ae2700349fcf8f98adb0dbfbee85aec3c9179d29cc072"
 dependencies = [
  "base64ct",
  "blake2",
- "cpufeatures",
+ "cpufeatures 0.2.17",
  "password-hash 0.5.0",
 ]
 
@@ -221,7 +221,7 @@ dependencies = [
  "rustc-hash 2.1.1",
  "serde",
  "serde_derive",
- "syn 2.0.114",
+ "syn 2.0.117",
 ]
 
 [[package]]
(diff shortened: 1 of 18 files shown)


server/lib/src/server/access/create.rs
```diff
@@ -1,39 +1,75 @@
+use super::migration::{migration_entry_attrs, MIGRATION_ENTRY_CLASSES, MIGRATION_IGNORE_CLASSES};
 use super::profiles::{
     AccessControlCreateResolved, AccessControlReceiverCondition, AccessControlTargetCondition,
 };
-use super::protected::PROTECTED_ENTRY_CLASSES;
+use super::protected::{PROTECTED_ENTRY_CLASSES, PROTECTED_MOD_PRES_ENTRY_CLASSES};
```
yaleman, line 5: We should make `kanidmd_lib::server:access::protected` public (`access` is `pub(crate)`) so that we can link to it on docsrs and go "this is what's blocked" and not have to keep it up to date. .. or move the constant or something so we can do it 😄

server/daemon/insecure_server.toml
```diff
@@ -14,7 +14,7 @@ tls_key = "/tmp/kanidm/key.pem"
 #   Defaults to "info"
 #
 # log_level = "info"
-log_level = "debug"
+# log_level = "debug"
```
yaleman, line 17: shocking

server/lib/src/server/access/migration.rs
```diff
@@ -0,0 +1,134 @@
+use crate::prelude::{Attribute, EntryClass};
+use std::collections::BTreeSet;
+use std::sync::LazyLock;
+
+/// These entry classes may be affected by migrations. All protection rules still
```
yaleman:
```suggestion
/// These entry classes can be modified by migrations. All protection rules still
```

server/lib/src/server/access/migration.rs
```diff
@@ -0,0 +1,134 @@
...
+            Attribute::LdapMaxQueryableAttrs,
+            Attribute::LdapAllowUnixPwBind,
+            Attribute::DomainDisplayName,
+        ]);
+    }
+
+    if classes.contains(EntryClass::Group.into()) {
+        allow_cls.clear();
+        allow_cls.extend([
+            Into::<&'static str>::into(EntryClass::Group),
```
yaleman: wwaaaaaayyyy less gross 😄 
```suggestion
            EntryClass::Group.as_ref(),
```

server/lib/src/server/identity.rs
```diff
@@ -68,12 +64,39 @@ pub struct IdentUser {
     // Other metadata?
 }
 
+#[derive(Debug, Clone)]
+/// The internal role being used for this operation.
+pub enum InternalRole {
+    /// The internal database system. This has unlimited power.
```
yaleman:
```suggestion
    /// The internal database system. This has unlimited crab power.
```

server/lib/src/server/identity.rs
```diff
@@ -169,11 +191,21 @@ impl Identity {
...
+        Identity {
+            origin: IdentType::Internal(InternalRole::Migration),
+            source: Source::Internal,
+            session_id: UUID_INTERNAL_SESSION_ID,
+            scope: AccessScope::ReadWrite,
+            limits: Limits::unlimited(),
+        }
+    }
+
     pub(crate) fn from_internal() -> Self {
```
yaleman, line 204: This should probably be renamed as system() or something similar to `migration()` since "internal" is no longer just "system" ?

server/lib/src/server/access/create.rs
Firstyear: Yeah, I think that will be a future goal though.

server/lib/src/server/identity.rs
Firstyear: Yes, we should, but that would be a *huge* PR in itself. So I didn't want to take that on here when I need to backport this.

PR #4166, opened 2026-02-24
Make Dockerfiles more reproducible

2tefan:
# Change summary

This PR pins the base-images to specific hashes which improves the reproducibility of the container images. Also, it specifies which Dockerfile frontend to use, which I think, makes sense, as the Dockerfiles use features like `--mount-type=cache` which need [dockerfile 1.2.0](https://github.com/moby/buildkit/releases/tag/dockerfile%2F1.2.0).

I also updated the dependabot config, so it should also update those hashes :)

More info:
- [Docs about Dockerfile frontend](https://docs.docker.com/build/buildkit/frontend/#stable-channel)
- [Docs about Image digests](https://docs.docker.com/dhi/core-concepts/digests/)

Checklist

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -272,3 +272,37 @@ updates:
       all: # group all the things
         patterns:
           - "*"
+  # Update Dockerimage digests
+  - package-ecosystem: "docker"
+    directory: "/tools"
+    schedule:
+      interval: weekly
+      time: "06:00"
+      timezone: Australia/Brisbane
+    open-pull-requests-limit: 99
+    groups:
+      all: # group all the things
+        patterns:
+          - "*"
+  - package-ecosystem: "docker"
+    directory: "/server"
+    schedule:
+      interval: weekly
+      time: "06:00"
+      timezone: Australia/Brisbane
+    open-pull-requests-limit: 99
+    groups:
+      all: # group all the things
+        patterns:
+          - "*"
+  - package-ecosystem: "docker"
+    directory: "/rlm_python"
+    schedule:
+      interval: weekly
+      time: "06:00"
+      timezone: Australia/Brisbane
+    open-pull-requests-limit: 99
+    groups:
+      all: # group all the things
+        patterns:
+          - "*"


Firstyear:
Kind of confused here, what problem does this solve?

yaleman:
Thanks for the PR but I'm going to reject it unless there's some real explanation why it's needed ('reproducibility' isn't enough)

- Specifying the Dockerfile syntax doesn't add anything unless you're using a pre-2020-ish buildkit version, in which case I'm sorry and that's on you
- Pinning versions like this just adds more noise/maintenance to our dev build pipelines when the repo digest/version and build time are already tagged into the images

2tefan:
Yeah, sorry, I should have properly added more arguments, why I think making the Dockerfiles more reproducible would make sense.

For example, Kanidmd is dynamically linked to these libs:

```sh
 ldd ./usr/src/kanidm/target/release/kanidmd
	linux-vdso.so.1 (0x00007faafa496000)
	libc.so.6 => /lib64/libc.so.6 (0x00007faaf63d6000)
	/lib64/ld-linux-x86-64.so.2 (0x00007faafa498000)
	libssl.so.3 => /lib64/libssl.so.3 (0x00007faaf62b2000)
	libcrypto.so.3 => /lib64/libcrypto.so.3 (0x00007faaf5c2c000)
	libgcc_s.so.1 => /lib64/libgcc_s.so.1 (0x00007faaf5bfd000)
	libm.so.6 => /lib64/libm.so.6 (0x00007faaf5b02000)
	libjitterentropy.so.3 => /lib64/libjitterentropy.so.3 (0x00007faaf5af7000)
	libz.so.1 => /usr/lib64/zlib-ng-compat/libz.so.1 (0x00007faaf5ace000)
```

They are from the base tumbleweed image. 
And, for example, my machine just happens to have cached an older version from October. 
Now, I and probably other contributors might test their changes with other versions of `libcrypto` or `libc` which could cause some confusion.

I kinda see this like adding a lockfile (like `Cargo.lock`) to a repository, because this ensures that everyone uses the exact same version.

The same also then applies when the CI is run.
Locally you might be using a good version, but in the CI just the newest image version is taken, which might have some issues.
You can't really know beforehand.
Imo it also helps with reducing supply-chain risk, as you know what you will be getting.


But at the end, it's your call if the noise/maintenance outweighs the benefits.
I'm still very happy that you provide this awesome project to the open-source community :)
Thanks for taking the time to review my PR

yaleman:
Thanks for your PR and thanks for understanding 💜

Firstyear:
> And, for example, my machine just happens to have cached an older version from October.

Docker/podman layers are based on hashes anyway, so an old cached "layer" won't matter, it should pull the layer version *we* built against.

Firstyear:
Also we copy the libs in during the build, so it's not taking from the "cached" layer you have, it's always precisely what is there at the time the container is built.

2tefan:
Yes, the image produced by the pipeline is perfectly fine, if you prefer to just taking the latest rolling release version, because of the noise it would otherwise create.

This issue about outdated caches was just about when developing something locally and then building a container locally for testing. Depending on when you build (and whether you’ve pulled), you may end up testing against different base libs

Firstyear:
Sorry I don't understand.

How can I end up testing against different base libs? If I build and test image A, the libs will never change in image A.

If I then do a new build and test for image B, the libs within image B will never change. Sure they may have updated since A, but image B still remains a self contained unit.

2tefan:
Yeah, those layers from A won't change, if you just work on B you're mostly fine. But problems can occur, when you already start with an outdated/broken image A. 

But also when you work on other projects, do a `docker/podman pull` somewhere else and go back to kanidm, you'll get properly a different base-image version and other libs. It's hard to go back to the image A you have been working with. (Also you got to rebuild your layers again, as the base image changed)

Firstyear:
Given how we do our builds and the processes around them, the situation you are suggesting here won't occur as we always pull images, and we aren't "swaping" around between different image sources etc.

PR #4176, opened 2026-02-26
Don't revert admin changes in some groups during migrcation

### LINKED ISSUE #4171: `idm_people_self_name_write` membership resets on upgrades
### I did this

Remove `idm_all_persons` from `idm_people_self_name_write`. Then upgrade Kanidm from v1.8.6 to v1.9.0.

### I expected the following

`idm_people_self_name_write` remains empty.

### This happened instead

`idm_all_persons` became a member of `idm_people_self_name_write`.

### Kanidm version details

- Output of `kanidm(d) version`: kanidm 1.9.0/kanidmd 1.9.0
- Are you running it in a container? If so, which image/tag?: latest (was 1.9.0 at the time)

### Any other comments

This bug was _not_ reproduced on upgrade from v1.9.0 to v1.9.1.

I observed this behavior once before as well but did not make note of the versions at that time.

Firstyear:
During migrations we were "setting" the content of Member on all groups. That meant that users who removed idm_all_persons from the person_self_write permission group found this change reverted on upgrade.

This splits up migrations of groups to "always create" members and "create once" members. This way if someone "undoes" a critical membership such as idm_admin, it can be reverted on domain remigrate, but a non-critical change is respected.

Fixes #4171

Checklist


yaleman (approved):

yaleman:
At some point we are going to have to start noting issue numbers (or documenting in general) in comments so we can explain all the various edge case handlers 😆

PR #4169, opened 2026-02-25
Fix bug where DEBUG is always true in RADIUS entrypoint.

BenJetson:
# Change summary

Howdy, KanIDM maintainers. I recently set up a new KanIDM instance and went to set up a RADIUS server today.

I noticed that the RADIUS server logs always had the debug output, even though I hadn't supplied the `DEBUG` environment variable like the docs suggested.

Upon looking at the entrypoint script for the RADIUS container, it became apparent that there was no way to set the `DEBUG` variable to false, since it defaults to true on both branches of that `if` statement.

This PR just changes the initial value of `DEBUG` to `False`, which matches what the current KanIDM book suggests the behavior should be, fixing this small bug.

Fixes #4168

Checklist

- [x] This PR contains no AI generated code
- [ ] book chapter included (if relevant)
- [ ] design document included (if relevant)

PR #4177, opened 2026-02-27
Disable multithreading on RADIUS when DEBUG is False.

### LINKED ISSUE #4168: RADIUS server always outputs debug logs regardless of DEBUG environment variable presence/value
### I did this

Set up a KanIDM RADIUS server instance, without setting the `DEBUG` variable on the container.

### I expected the following

No debug output from FreeRADIUS.

### This happened instead

Regardless of `DEBUG` environment variable value, the FreeRADIUS server always outputs debug level logs.

### Kanidm version details

- Output of `kanidm(d) version`: 1.9.1
- Are you running it in a container? If so, which image/tag?: `kanidm/radius:1.9.1`
- Operating System / Version (On Unix please post the output of `uname -a`):  `Linux optiplex 6.8.0-100-generic #100-Ubuntu SMP PREEMPT_DYNAMIC Tue Jan 13 16:40:06 UTC 2026 x86_64 x86_64 x86_64 GNU/Linux`

### Any other comments

There's a simple fix for this in the entrypoint script; I was about to open a pull request and decided I should open an issue first to document the bug first.

BenJetson:
# Change summary

I submitted the previous PR #4169 to address issue #4168, which addressed the issue that the RADIUS server always had `DEBUG` set to `True`. 

After further experimentation and testing, I think I discovered why it was that way to begin with. There appears to be an issue with the RADIUS server when multithreading is enabled, where the python3 module fails at the authorize step with `RLM_MODULE_FAIL` before even calling the Kanidm library function. Thus, the server completely breaks unless `DEBUG` is `True`.

This PR adds the `-t` flag when `DEBUG` is `False`, disabling multithreading (which is the case when passing `-X`).

This time, I have more thoroughly tested the change. See more info in [#4168, comment 3970196312](https://github.com/kanidm/kanidm/issues/4168#issuecomment-3970196312).

Fixes #4168 (again)

Checklist

### DIFF
--- a/rlm_python/radius_entrypoint.py
+++ b/rlm_python/radius_entrypoint.py
@@ -147,7 +147,9 @@ def run_radiusd() -> None:
     if DEBUG:
         cmd_args = ["-X"]
     else:
-        cmd_args = ["-f", "-l", "stdout"]
+        # NOTE: the python3 module for radiusd does not seem to work correctly
+        # when multithreading is enabled, hence -t. See #4168 for details.
+        cmd_args = ["-f", "-t", "-l", "stdout"]
     freeradius_bin = find_freeradius_bin()
     if freeradius_bin is None:
         print("Failed to find FreeRADIUS binary, quitting!", file=sys.stderr)


yaleman:
> where the python3 module fails at the authorize step with RLM_MODULE_FAIL before even calling the Kanidm library function. Thus, the server completely breaks unless DEBUG is True

wow, that's... fun?

PR #4179, opened 2026-02-28
Rust FreeRADIUS module

### LINKED ISSUE #305: Rewrite kanidmradius to rust

yaleman:
# Change summary

- reimplementing things in Rust, 'cause that's how we roll. 
- yes, there is a lot of random nonsense in there, this needs a solid review, I just threw code at the wall until it stuck. the test_radius script bounces radtest off it and indicates it works just as poorly as the python version if that counts? 😄 
- the rust version is a bunch smaller! 770-940MB

Fixes #305

aldnxy


book/src/integrations/radius.md
```diff
@@ -141,7 +141,7 @@ docker run --name radiusd \
     -p 1812:1812 \
     -p 1812:1812/udp \
     --interactive --tty \
-    --mount "type=bind,src=$(pwd)/examples/radius.toml,target=/data/kanidm" \
+    --mount "type=bind,src=$(pwd)/examples/radius.toml,target=/data/radius.toml" \
```
Firstyear, line 144: We still need certificates and such bound in, that's what we were doing with the /data/kanidm path yes?

unix_integration/rlm_kanidm/container/mods-available/python3
```diff
@@ -0,0 +1,66 @@
+#
```
Firstyear: Can delete this file since it won't be used.

unix_integration/rlm_kanidm/src/lib.rs
```diff
@@ -0,0 +1,773 @@
...
+fn kvpairs_to_attributes(
+    request_attrs: *const KVPair,
+    request_attrs_len: usize,
+) -> Result<RequestAttributes, ModuleError> {
+    if request_attrs.is_null() && request_attrs_len != 0 {
+        return Err(ModuleError::Other(
+            "request_attrs pointer is null with non-zero length".to_string(),
+        ));
+    }
+    let attrs_slice = if request_attrs_len == 0 {
```
Firstyear: Manually define the slice type annotation here to make it clearer what this is. Likely `: &[KVPair]`. It makes it a lot easier to understand especially when working near unsafe to be a bit more explicit about things.

unix_integration/rlm_kanidm/src/lib.rs
```diff
@@ -0,0 +1,773 @@
...
+    if !result.error.is_null() {
+        unsafe {
+            drop(CString::from_raw(result.error));
+        }
+    }
+}
+
+/// Helper to free an array of KVPair allocated in Rust and returned to C. This should be called for the `reply` and `control` fields of `AuthResultC` after the caller is done using them, to avoid memory leaks.
+fn free_kv_pairs(ptr_pairs: *mut KVPair, len: usize) {
+    if ptr_pairs.is_null() || len == 0 {
```
Firstyear: It's valid to have a non-null but zero len allocation in rust (minor nit). So in *theory* you shouldn't check the length here.

unix_integration/rlm_kanidm/src/lib.rs
```diff
@@ -0,0 +1,773 @@
...
+
+    pub fn user_id(&self) -> Option<&str> {
+        self.get(ATTR_TLS_SAN_DN_CN)
+            .or_else(|| self.get(ATTR_TLS_CN))
+            .or_else(|| self.get(ATTR_USER_NAME))
+    }
+}
+
+#[derive(Debug, Clone)]
+pub struct AuthorizeResult {
```
Firstyear: This should be an enum (since it's self contained within rust) and then instead of auth_result_from_pairs we can then do a match and build the Authorize Result from that. It saves the Vec -> KVPair dance, and makes the possible return states clear.

unix_integration/rlm_kanidm/src/lib.rs
```diff
@@ -0,0 +1,773 @@
...
+        match self.runtime.block_on(self.fetch_token_http(user_id))? {
+            Some(token) => {
+                self.insert_cache(user_id.to_string(), token.clone(), now);
+                Ok(Some(token))
+            }
+            None => Ok(None),
+        }
+    }
+
+    fn lookup_cache(&self, user_id: &str, now: Instant) -> Option<RadiusAuthToken> {
```
Firstyear: Split cache to it's own struct/module rather than making it part of "Module" so that it's logically distinct.

Firstyear (changes requested):
Okay this actually is less of a horror show than I was expecting (given your comments :P )

I will say I think that if I had any major change it's to layer things a little bit more.

I think a good set of layers is:

```
C Boundary <-> Conversion <-> Logic
```

So you have the C boundary in things like `freeradius.rs` where you should also have your other extern C functions like `rlm_kanidm_instantiate` and `authorize`

Then these functions should really take those inputs and call a "rust" conversion function that converts from C -> Rust types.

Then you have a separate Rust logic layer.

That way you can more easily test the Rust logic layer without having to go through all the conversion/boundaries, and could even be integrated with testkit etc. 

It also means you can then directly test the conversion functions too. 

Finally it will help just arrange the code better based on it's intent and purpose. Currently  Ithink there is a bit "too much' overlap between what is the Conversion <-> Rust layer, for example my comments about the AuthResult type where you almost have a "C style" structure, not a Rust Enum instead. That makes it feel/read awkwardly on the Rust side both in construction of the result, but also the conversion step becomes more awkward.

Anyway I hope that helps, I think that this looks like a great start though.

book/src/integrations/radius.md
yaleman: `/data/kanidm` was the config file being mounted in - a line down the `/tmp/kanidm` -> `/data` mount gets certs.

Given how *very* different the radius config became from the Kanidm config, and it'd been implied for a bit, I moved to being more obvious about the file paths.

Firstyear:
Okay, I'm not done yet - I've massively cleaned up the bootstrap, moved the whole code to rlm_kanidm (since this is not actually part of unix integration), massive docker cleanups. I've submitted a PR to opensuse to fix the header issue, and in the meantime, I just chucked the headers into include and use them rather than "yolo download at build". 

Next up I plan to finish the ffi side cleanups and improve the testing there.

Firstyear:
@yaleman This has been working reliably for me for a week now in production, with no memory leaks or resource leaks. So should be good to review now.

Firstyear:
The CI check I'm not completely sure about why it's failing though.....

rlm_kanidm/entrypoint/src/kanidm_radiusd.rs
```diff
@@ -0,0 +1,20 @@
...
+
+#[derive(Debug, Parser)]
+struct Cli {
+    #[arg(long, env = "KANIDM_RLM_CONFIG")]
+    config: Option<PathBuf>,
+}
+
+fn main() {
+    let cli = Cli::parse();
+    let debug = std::env::var_os("DEBUG").is_some();
```
yaleman: this should be a bool in the clap parser so it's discoverable

Firstyear:
Rebaed to 1.11 dev

rlm_kanidm/entrypoint/src/bootstrap.rs
```diff
@@ -0,0 +1,274 @@
...
+
+        // Anything else will fail, it can't be valid in a container
+        if file_type.is_file() {
+            fs::copy(&source_path, &destination_path)?;
+        }
+    }
+
+    Ok(())
+}
+
```
yaleman, line 163: we've pulled openssl out as a library but explicitly call it using a CLI command? this feels kinda gross, but I guess expected since it's freeradius nasty

rlm_kanidm/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+## Building it
+
+You need libtalloc development headers (packages are `libtalloc-dev` on Ubuntu and `libtalloc-devel` on OpenSUSE)
+
+```bash
+cargo build -p rlm_kanidm --features freeradius-module
+```
+
+If headers are in a non-standard path, or don't play nice, you can run `pull_headers.sh` which will yeet them into place - note this is designed for CI, so it's writing into `/usr/include/`
```
yaleman: This needs updating since the script's gone; also need to document specifics of where the current headers came from and how to update them.

rlm_kanidm/entrypoint/src/bootstrap.rs
Firstyear: I can change that, but you wrote that section ;)

Firstyear: Okay, so this concept of hashing/symlinking certs is a very "openssl concept". I don't see much value in recreating that ourselves, and given freeradius in the container already relies on openssl, I think this can stay.

yaleman (commented):
minor docs thing but otherwise woo

rlm_kanidm/README.md
Firstyear: It's pretty much out-of-the-box on opensuse, not sure re ubuntu but upstream freeradius did accept the pr to update their makefiles to it should "just work" in future too.

Firstyear (approved):

PR #4187, opened 2026-03-02
core/ui: reject unsafe /ui/unlock return locations

Oluwatobi-Mustapha:
## Change summary

- enforce local-path-only validation for `/ui/unlock` reauth return locations
- reject unsafe redirect targets such as `//host/...`, `*`, and non-absolute paths
- apply validation at both source (referer capture) and sink (final redirect)
- add unit regression coverage for accepted/rejected paths

Checklist

### DIFF
--- a/.github/workflows/docker_build_kanidm.yml
+++ b/.github/workflows/docker_build_kanidm.yml
@@ -59,7 +59,7 @@ jobs:
           # Must use OCI exporter for multi-arch: https://github.com/docker/buildx/pull/1813
           outputs: type=oci,dest=/tmp/kanidm-docker.tar
       - name: Upload artifact
-        uses: actions/upload-artifact@v6
+        uses: actions/upload-artifact@v7
         with:
           name: kanidm-docker
           path: /tmp/kanidm-docker.tar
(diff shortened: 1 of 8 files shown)


Firstyear:
Hi there,

Gonna reject this because it's not needed.

While on the surface it *seems* like there is a redirect injection here, if you observe https://github.com/kanidm/kanidm/blob/master/server/core/src/https/views/login.rs#L1144 which actually sets the session_context into the cookie, you'll notice this is signed with https://github.com/kanidm/kanidm/blob/master/server/core/src/https/views/cookies.rs#L46 using HMAC-SHA256

As a result, since this value can't be user-manipulated, we don't need to worry about the url because *we* always control that value.

Additionally, most urls in Kani aren't affected by things like a path-traversal since we don't allow access to the filesystem anyway, and we don't rely on url filtering/safety so there are no weird WAF rules to confuse etc. 

Hope that helps,

Firstyear:
A better option if you wanted to really harden this would be to make session context use the https://github.com/kanidm/kanidm/blob/master/server/core/src/https/views/constants.rs#L25 type so that there are predefined states and we aren't serialising a location, but a state that we translate into a url.

PR #4190, opened 2026-03-04
Policy management proposal

Foosec:
# Change summary
After a few discussions over this with @Firstyear im finally getting around to my promise of starting with a proposed design doc in hopes of starting a discussion on a potential implementation. 

-

Fixes #

Checklist


CEbbinghaus (commented):
Depending on the scope of a "Policy" (sounds to me kinda like configuration options), it may be prudent to offer this as an extension to kanidm in the same way that the smtp client is. This way only those that require "policy management" would need to host an instance of it.

Would it be worth while to explain in the design doc what exactly a "policy" is and what setting one entails / affects.

Firstyear:
I'll review this in coming days but I simply don't have time for now ... :(

Foosec:
> Depending on the scope of a "Policy" (sounds to me kinda like configuration options), it may be prudent to offer this as an extension to kanidm in the same way that the smtp client is. This way only those that require "policy management" would need to host an instance of it.
> 
> Would it be worth while to explain in the design doc what exactly a "policy" is and what setting one entails / affects.

Thats a very good point to add :) i'll add the policy definition part
Just wanted to get the general idea across and hopefully work together on a structure that makes sense
It would probably be prudent to design it so its not kanidm specific i.e someone else could implement it, to get "Implementor" adoption

Foosec:
> I'll review this in coming days but I simply don't have time for now ... :(

No worries, theres no rush :) 
Added the policy explainer

Firstyear (commented):
Hey mate, thanks for finally writing this up. I think this is a great start, and as you say "will there be questions" ... yes I have lots.

So to help give some info, Kanidm's client side tools have both an unprivileged component (unixd) which does all network communication, and a privileged and isolated component (unixd-tasks) that handles things like homedir creation, mounts etc. 

From this proposal it's not clear if unixd-tasks is the implementor (because it could be ....) or if unixd should just setup and set policy into known locations that other implementors can consume.

A lot of other things makes sense. The unixd daemon authenticates to Kanidm to fetch the policy, and makes it available in some way on the client, that's fine.

How should per user + machine policy be handled? Do you have thoughts about user policy too, similar to GPO's that apply to a user's session (contrast a machine policy).

WRT to policy snapshots, you have to be able to account for a similar issue like "registry tatooing". You apply some policy A initially, and then you apply A+B, but B fails. How do you roll back B to the original state? In the current design that's not accounted for, but that seems more like an "implementor" problem that it needs to account for and take snapshots as needed to perform the rollbacks. Similar, how do you *test* the policy applied in a successful manner? Still, that's somewhat outside of our scope, since we're just distributing stuff.

Worth pointing out that even if Kanidm is the distribution layer, we have a highly privileged set of tools, both at system level access, and the ability to provide account details, so while I get the background to why policy uploads are signed/attested, there still is the fundamental issue that if Kanidm gets owned, someone can just become root and take over the system. That reduces all the policy signing/attest to an audit trail rather than anything else, so I won't write it off, but just a reminder of that in the threat model. 

Anyway, probably more questions to come but thanks again for writing it up.

Foosec:
Hello!

Thank you for the response, questions are a good thing so i hope more to come :)

I'll tart with the combined question of: should unixd or unixd-tasks handle policies and the threat model of a root account being able to bypass them:

I believe it would be more prudent security wise to have unixd-tasks handle policy distribution, simply to up the privilege required by a possible attacker to bypass it, needing at least CAP_SYS_ADMIN then or well, root to mess with the daemon or its policies.
So that brings it back to, signing and attestation mean nothing, if an attacker has sufficient privileges which i totally agree with and its a threat scenario i've considered. My thoughts on it are, that in the grand scheme of things there isn't much userland can do to defend against this. 
We could save a digest into a TPM and enable implementors to self verify or something along those lines, but if you are root you can just replace the implementors anyway. (Or mess with the TPM)
There is certainly a lot that could be done with LSM's to defend unixd-tasks from attacks, like strict SELinux defintions or using IMA_EVM: https://en.opensuse.org/SDB:Ima_evm .

What might be a good option is to serve policies via dbus or unix sockets instead of dropping files, as that would be slightly harder to mess with? 

If it were to go DBUS route or Unix sockets, good polkit defintions would also be prudent. 

to that, yes unixd-tasks (or broadly, Kani) would (in my view) not serve as an implementor of the policies, just as a distribution and attestation service for them. 

I think user + machine policy handling is something we can and should discuss further, an idea would be that if a process requesting a policy be given a machine policy if running as a "machine" account (Root, www-data, etc)
But if the UID of the requesting process is a Kani user, then the process gets sent the User policy. 

But then again, that could be too rigid and you might want a machine policy to override any potential user policies in certain scenarios. Perhaps a user policy supersedes machine policies by default, unless specified otherwise by the machine policy? Or perhaps the best scenario is not dealing with this, and letting the Implementor decide how this should behave and always giving it machine and user policies on request. 

As for policy snapshots, its up for discussion, it could offer a way for an implementor to ACK that hey, i applied the policy, or respond that the policy is invalid etc. Or it could *just* be distribution. 

Looking forward to further discussion, lots to consider and lots of gotchas :)

Firstyear:
> I believe it would be more prudent security wise to have unixd-tasks handle policy distribution, simply to up the privilege required by a possible attacker to bypass it, needing at least CAP_SYS_ADMIN then or well, root to mess with the daemon or its policies. So that brings it back to, signing and attestation mean nothing, if an attacker has sufficient privileges which i totally agree with and its a threat scenario i've considered. My thoughts on it are, that in the grand scheme of things there isn't much userland can do to defend against this. We could save a digest into a TPM and enable implementors to self verify or something along those lines, but if you are root you can just replace the implementors anyway. (Or mess with the TPM) There is certainly a lot that could be done with LSM's to defend unixd-tasks from attacks, like strict SELinux defintions or using IMA_EVM: https://en.opensuse.org/SDB:Ima_evm .

Yep, that seems reasonable. The reason I'm pointing out this though is that kanidm-unixd is already in a position to arbitrate who is or is not root, which allows policy bypassing anyway. But as you note, SELinux confined users can defend from this, but by default SELinux doesn't really do very well to protect root / unconfined processes like user sessions.

> 
> What might be a good option is to serve policies via dbus or unix sockets instead of dropping files, as that would be slightly harder to mess with?

Such that a client has to request policies related to it? The risk/issue there, would be that then you need some form of access controls, where as file dropping at least lets us put things in places where we know it's going to the intended recipient. It also works then with things like systemd credentials. 

> 
> If it were to go DBUS route or Unix sockets, good polkit defintions would also be prudent.

Does polkit allow policy based on what the program is that is running? I thought it was mostly user/group based as well. 

> 
> to that, yes unixd-tasks (or broadly, Kani) would (in my view) not serve as an implementor of the policies, just as a distribution and attestation service for them.

Perfect, I agree with that then. 

> 
> I think user + machine policy handling is something we can and should discuss further, an idea would be that if a process requesting a policy be given a machine policy if running as a "machine" account (Root, www-data, etc) But if the UID of the requesting process is a Kani user, then the process gets sent the User policy.
> 
> But then again, that could be too rigid and you might want a machine policy to override any potential user policies in certain scenarios. Perhaps a user policy supersedes machine policies by default, unless specified otherwise by the machine policy? Or perhaps the best scenario is not dealing with this, and letting the Implementor decide how this should behave and always giving it machine and user policies on request.

I think that becomes an implementor questions, where we just serve "here are the machine policies" and "here are the user policies" and then the implementor is in the trusted position to merge those in a way that is relevant to the application. 

> 
> As for policy snapshots, its up for discussion, it could offer a way for an implementor to ACK that hey, i applied the policy, or respond that the policy is invalid etc. Or it could _just_ be distribution.

Pros and cons to both, but with any of these schemes there is always the risk that a bad policy can cause the system to go into a bad/undefined state. For example, you want to at least fail-safe rather than fail-open. 

So having policy snapshots at least allows a fallback to a former "known good" policy vs a fail-closed situation. But the fail-closed is probably the easier and simpler approach, and alerts people faster that a potential problem is being experienced. The risk is if that affects the ability to login and then recover/rescue the system from that state.

Foosec:
From Polkits manpage: 
```
   The Subject type
       The subject parameter passed to user functions is an object with information about the process being
       checked. It is of type Subject and has the following attributes

      int pid
           The process id.

       int uid
           The UID of the user of the subject.

       string user
           The user name.

       string[] groups
           Array of groups that user user belongs to.

       string seat
           The seat that the subject is associated with - blank if not on a local seat.

       string session
           The session that the subject is associated with.

       string system_unit
           The systemd unit that the subject's process is part of (if any). Note that this can only match on system
           units, as user units can be created with any name without privileges (unlike system units which require
           root to create). A process running in a user unit will return the user session unit in this attribute
           (e.g.: user-1000.service).

       boolean local
           Set to true only if seat is local.

       boolean no_new_privileges
           Set only if system_unit is not empty, and set to true only if the referenced systemd service unit has
           the NoNewPrivileges= setting enabled. This ensures that the process cannot gain any new privileges via
           executing setuid binaries.

       boolean active
           Set to true only if the session is active.

       The following methods are available on the Subject type:

       boolean isInGroup(string groupName);

       boolean isInNetGroup(string netGroupName);

       The isInGroup() method can be used to check if the subject is in a given group and isInNetGroup() can be
       used to check if the subject is in a given netgroup.
```

So using DBUS and Polkit it should be doable to check which process is requesting, but it would rely on an external authorization framework for that (Polkit). 

If we were to use raw Unix sockets, Kani could check what the PID of the connecting process is, but it would be harder to implement for the clients. (At least until libraries exist :) )

DBUS also exposes the PID: https://dbus.freedesktop.org/doc/dbus-java/api/org/freedesktop/DBus.html "GetConnectionUnixProcessID" (Its the java bindings, but i imagine it does so for rust too :) ) 

I think this opens up an interesting question; should a policy be able to choose in the manifest, if its exposed via dbus/unixd (whichever of those we decide on) or via being dropped as a file at a path? 
This way you could kinda monkeypatch dropping configs and dotfiles without the software strictly implementing support.

Food for thought!

> I think that becomes an implementor questions, where we just serve "here are the machine policies" and "here are the user policies" and then the implementor is in the trusted position to merge those in a way that is relevant to the application.

Perfect, i was leaning towards that myself. 

> Pros and cons to both, but with any of these schemes there is always the risk that a bad policy can cause the system to go into a bad/undefined state. For example, you want to at least fail-safe rather than fail-open.

> So having policy snapshots at least allows a fallback to a former "known good" policy vs a fail-closed situation. But the fail-closed is probably the easier and simpler approach, and alerts people faster that a potential problem is being experienced. The risk is if that affects the ability to login and then recover/rescue the system from that state.

Circling back to files vs dbus, i feel like having a feedback mechanism when exposed via dbus would be useful. 
In such a case it could automatically fall back to the previous policy, if the new policy isn't applicable. 

Perhaps letting the implementor decide in this case is also the correct approach? Ultimately if it wants to, it can cache older policies and apply them anyway, and Kani has no control over that as is in any scenario i can dream of.

PR #4202, opened 2026-03-10
Defer on some routes

### LINKED ISSUE #4192: FR: Disable log on /status route
I use `/status` route with livenessProbe and readinessProbe on k8s.

A clear description of what the problem is. Ex. I'm confused by, or would like to know how to...
Into logs, I can see the request on `/status` route. But I get a log every seconds

An option to disable trace on /status route

A description of what you'd expect to happen.
No logs on `/status` request

Set log to warm

Are there any alternative solutions or features you've considered.
 /

Firstyear:
# Change summary

- Match on some routes and mark their spans as defered. 
- Rather than string match we could make a second "make-span" function and then setup two trace-layers - one for everything that is "always log" and another for "defered". 

Fixes #4192

Checklist


server/core/src/https/trace.rs
```diff
@@ -22,6 +22,11 @@ impl<B> tower_http::trace::MakeSpan<B> for SpanCreator {
         // span to always be at the info level. If this is not done, then there
         // will not be an event uuid available which causes TONS of problems. Like
         // crashing.
+
+        // This marks some spans as defered on "noisy" routes that people often don't want
+        // to see.
+        let is_defered = request.uri() == "/" || request.uri() == "/status";
```
yaleman:
```suggestion
        let is_deferred = request.uri() == "/" || request.uri() == "/status";
```

server/core/src/https/trace.rs
```diff
@@ -22,6 +22,11 @@ impl<B> tower_http::trace::MakeSpan<B> for SpanCreator {
         // span to always be at the info level. If this is not done, then there
         // will not be an event uuid available which causes TONS of problems. Like
         // crashing.
+
+        // This marks some spans as defered on "noisy" routes that people often don't want
```
yaleman:
```suggestion
        // This marks some spans as deferred on "noisy" routes that people often don't want
```

server/core/src/https/trace.rs
```diff
@@ -34,6 +39,7 @@ impl<B> tower_http::trace::MakeSpan<B> for SpanCreator {
             status_code = tracing::field::Empty, // filled in later, used by tracing forest
             http.response.status_code = tracing::field::Empty, // filled in later, used by otel
             latency = tracing::field::Empty, // filled in later
+            defer = is_defered,
```
yaleman:
```suggestion
            defer = is_deferred,
```

yaleman (approved):

PR #4206, opened 2026-03-11
20260311 zxcvbn check

### LINKED ISSUE #4200: Password validation fails with misleading error when displayname is empty
### I did this
1. Create a person with empty displayname:
      kanidm person create testuser ""
2. Set an email:
      kanidm person update testuser --mail testuser@example.com
3. Generate a credential reset token:
      kanidm person credential create-reset-token testuser
4. Open the reset link and try to set any password, including completely random ones like `Zx9$kL2#mN5@pQ`

### I expected the following
Password is accepted, or a clear error indicating that displayname cannot be empty, or at least an error in the logs indicating that. **I suspect the empty string is actually being used to match in the password, and always matching the empty display name, causing the issue.**

### This happened instead
Error PasswordQuality([NamesAndSurnamesByThemselvesAreEasyToGuess, AvoidDatesAndYearsThatAreAssociatedWithYou]), completely misleading and unrelated to the actual problem. Even worse: misleading users into trying several different passwords and growing frustrated.

### Kanidm version details

- Output of `kanidm(d) version`: kanidmd 1.9.1
- Are you running it in a container? If so, which image/tag?:  docker.io/kanidm/server:latest
- Operating System / Version (On Unix please post the output of `uname -a`): Linux hypnotist 6.1.0-34-cloud-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.1.135-1 (2025-04-25) x86_64 GNU/Linux

### Any other comments
Fixed by simply setting the display name:
kanidm person update testuser --displayname "testuser"

Firstyear:
# Change summary

- Prevent empty strings in displayname which can confuse the zxcvbn password check. 

Fixes #4200

Checklist

### DIFF
--- a/book/src/examples/traefik.md
+++ b/book/src/examples/traefik.md
@@ -1,62 +1,70 @@
 # Traefik
 
-Traefik is a flexible HTTP reverse proxy webserver that can be integrated with Docker to allow dynamic configuration and
-to automatically use LetsEncrypt to provide valid TLS certificates. We can leverage this in the setup of Kanidm by
-specifying the configuration of Kanidm and Traefik in the same
-[Docker Compose configuration](https://docs.docker.com/compose/).
+Traefik is a flexible HTTP reverse proxy webserver that can be integrated with Docker to allow dynamic configuration and to automatically use LetsEncrypt to provide valid TLS certificates. 
+
+To leverage this with Kanidm, Traefik must be configured with a custom [ServersTransport](https://doc.traefik.io/traefik/reference/routing-configuration/http/load-balancing/serverstransport/) and routed to Kanidm's `HTTPS` scheme and port.
 
 ## Example setup
 
-Create a new directory and copy the following YAML file into it as `docker-compose.yml`. Edit the YAML to update the
-LetsEncrypt account email for your domain and the FQDN where Kanidm will be made available. Ensure you adjust this file
-or Kanidm's configuration to have a matching HTTPS port; the line
-`traefik.http.services.kanidm.loadbalancer.server.port=8443` sets this on the Traefik side.
+This example assumes that you have a working Traefik configuration with automatic TLS certificates.
 
 > [!NOTE]
 >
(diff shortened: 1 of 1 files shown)


server/lib/src/migration_data/dl14/schema.rs
```diff
@@ -757,7 +757,7 @@ pub static SCHEMA_ATTR_PATCH_LEVEL_DL7: LazyLock<SchemaAttribute> =
     LazyLock::new(|| SchemaAttribute {
         uuid: UUID_SCHEMA_ATTR_PATCH_LEVEL,
         name: Attribute::PatchLevel,
-        description: "".to_string(),
+        description: "The domains patch level".to_string(),
```
yaleman:
```suggestion
        description: "The domain's patch level".to_string(),
```

server/lib/src/entry.rs
```diff
@@ -2630,19 +2630,20 @@ impl<VALID, STATE> Entry<VALID, STATE> {
 
     pub(crate) fn get_display_id(&self) -> String {
         self.attrs
-            .get(&Attribute::Spn)
```
yaleman, line 2633: wouldn't SPN be preferred for clarity, given that get_display_id is used all over the place for identifying things in error messages and audit records? (especially once federation's a thing?)

Firstyear: Only a local item can have "name", remote items have "spn". That way local entries are just "admin" vs "admin@some.other.instance" but I can always make spn the default again.

yaleman: SPN feels less gross in general 🤔

Firstyear: Will revert then :)

Firstyear:
Made it spn only, since if you have a name, you have an spn.

server/lib/src/migration_data/dl14/schema.rs
yaleman: Oi! 😄

yaleman (approved):

PR #4219, opened 2026-03-20
OpenSSL shall be vanquished

### LINKED ISSUE #3476: Removal of OpenSSL

Firstyear:
# Change summary

- Remove OpenSSL from the Kanidm codebase
- Update all dependencies to OpenSSL free versions

Fixes #3476

Checklist


proto/src/cli.rs
```diff
@@ -48,7 +48,7 @@ pub struct KanidmdCli {
     pub db_arc_size: Option<usize>,
 
     #[clap(
-        env = "KANIDM_SERVER_ADMIN_BIND_PATH",
+        env = "KANIDM_ADMIN_BIND_PATH",
```
CEbbinghaus, line 51: unrelated change? how come?

server/lib/src/constants/mod.rs
```diff
@@ -193,7 +193,7 @@ pub const REPL_SUPPLIER_ADVANCE_WINDOW: Duration = Duration::from_secs(600);
 
 /// The number of days that the default replication MTLS cert lasts for when
 /// configured manually. Defaults to 4 years (including 1 day for the leap year).
-pub const REPL_MTLS_CERTIFICATE_DAYS: u32 = 1461;
+pub const REPL_MTLS_CERTIFICATE_DAYS: u64 = 1461 * 86400;
```
CEbbinghaus: Is that still measured in days???? 👀

proto/src/cli.rs
Firstyear: Because the env var colides with a compile time var, which means if you cargo run the server the admin bind path is always force set.

server/lib/src/constants/mod.rs
Firstyear: Could be renamed yeah

jinnatar:
You should also be able to nuke this and other similar deps from the debian build rules: https://github.com/kanidm/kanidm/blob/master/server/daemon/Cargo.toml#L69

server/lib/src/repl/supplier.rs
```diff
@@ -2,29 +2,144 @@ use super::proto::{
...
+        })?;
+
+        // Key Usage (server + client )
+        let eku_extension = ExtendedKeyUsage(vec![
+            oiddb::rfc5280::ID_KP_CLIENT_AUTH,
+            oiddb::rfc5280::ID_KP_SERVER_AUTH,
+        ]);
+
+        x509_builder.add_extension(&eku_extension).map_err(|err| {
+            error!(?err, "Unable to extended key usage extension");
```
yaleman:
```suggestion
            error!(?err, "Unable to add extended key usage extension");
```

yaleman (changes requested):
yaaaaaaay

Firstyear:
Renamed the mtls default, removed debian libssl deps.

PR #4220, opened 2026-03-20
Add shell diagnostics

### LINKED ISSUE #4209: User shell not working (probably symlink related)
### I did this
I created a testuser named `testuser` and added posix attributes and also set ssh-pubkey and password.
I set the shell of this user to `/run/current-system/sw/bin/fish`, which is a symlink since i am on NixOS.

### I expected the following
The shell is set and used for further logins (via ssh or `su` on the terminal).

### This happened instead
When i want to login via ssh the daemon shows that the shell does not exist.
When i login via `su` i am logged in with `bash`.

### Kanidm version details

- Output of `kanidm(d) version`: 1.9.1 on both
- If not a container, how'd you install it: via NixOS options
- Operating System / Version (On Unix please post the output of `uname -a`): Linux nixos 6.6.54 #1-NixOS SMP PREEMPT_DYNAMIC Fri Oct  4 14:30:05 UTC 2024 x86_64 GNU/Linux

### Any other comments

output of `/etc/shells`
`/run/current-system/sw/bin/fish
/nix/store/d305s1vfwd777ps7mz9sg1xfa2b82anf-fish-4.5.0/bin/fish
/run/current-system/sw/bin/bash
/run/current-system/sw/bin/sh
/nix/store/0550j0i8bmzxbcnzrg1g51zigj7y12ih-bash-interactive-5.3p9/bin/bash
/nix/store/0550j0i8bmzxbcnzrg1g51zigj7y12ih-bash-interactive-5.3p9/bin/sh
/bin/sh`

Firstyear:
# Change summary

- Improve shell diagnostics to be clearer during some error conditions, and provide the full file diag output 

Fixes #4209

Checklist

### DIFF
--- a/unix_integration/resolver/src/resolver.rs
+++ b/unix_integration/resolver/src/resolver.rs
@@ -401,17 +401,22 @@ impl Resolver {
 
         let requested_shell_exists = if let Some(shell_path) = maybe_shell.as_ref() {
             // Does the shell path as configured exist?
+            //
+            // Canonicalisation fails in two cases per https://doc.rust-lang.org/std/fs/fn.canonicalize.html#errors
+            //
+            // * Path does not exist
+            // * An intermediate element of the path is a file (not a directory/link).
+            //
             let mut exists = shell_path
                 .canonicalize()
                 .map_err(|err| {
                     warn!(
-                        "Failed to canonicalize path, using base path. Tried: {} Error: {:?}",
-                        shell_path.to_string_lossy(),
-                        err
+                        ?err,
+                        path = ?shell_path.display(),
+                        "Failed to canonicalise shell path",
                     );
                 })
-                .unwrap_or(Path::new(shell_path).to_path_buf())
-                .exists();
+                .is_ok();
 
             if !exists {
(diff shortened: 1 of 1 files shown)


unix_integration/resolver/src/resolver.rs
```diff
@@ -403,7 +404,7 @@ impl Resolver {
             let mut exists = shell_path
                 .canonicalize()
                 .map_err(|err| {
-                    debug!(
+                    warn!(
                         "Failed to canonicalize path, using base path. Tried: {} Error: {:?}",
```
yaleman: This'll get SUPER noisy and since it's not actually a blocking error then we should probably stop canonicalising the path in this case.

Firstyear: I added some extra comments and tidied it up, turns out canonicalise errors on non-existing files anyway, so this streamlines it a bit.

yaleman: yeah, canonicalize is a weird one sometimes.

PR #4224, opened 2026-03-21
Feat: Add Prompt Support

### LINKED ISSUE #2988: Add support for OIDC `prompt` key

I want to be able to force users to log in again when they access a specific oauth app.

[OIDC](https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest) defines a prompt parameter, that communicates the request to relogin via `prompt=login`.

None considered, we have one in the spec.

CEbbinghaus:
# Change summary

Since `prompt=login` was easier to add than `prompt=none`, I decided to just do both, and threw in `prompt=consent` for free.

Since we don't have the ability to select accounts, I left out the impl for `consent=select_account`. Thank god for SHOULD :3

since this is all RFC I don't think we need a book chapter or design doc

(here is hoping my ignorant little impl is sufficient)

Fixes #2988

Checklist


CEbbinghaus:
Re-Read the spec and gotta fix up the PR. Namely it doesn't support the space delaminated list which we need to fix. That also complicates the current impl so I'll have to rethink part of the implementation (particularly around Login/Consent. If all else fails we can just add support for only None / Login which since they are mutually exclusive allows us to proceed without worry.

I'm also gonna make the executive call that instead of throwing if the prompt value is "invalid" we will ignore instead of throw. This guarantees that authentication proceeds fine even if a new spec adds a key unknown to us.

server/lib/src/idm/oauth2.rs
```diff
@@ -2178,6 +2182,12 @@ impl IdmServerProxyReadTransaction<'_> {
             }
         };
 
+        // We could throw an error OR we could ignore it, what do we want to do?
```
Firstyear: Error is the correct answer.

proto/src/oauth2.rs
```diff
@@ -437,6 +438,17 @@ fn response_modes_supported_default() -> Vec<ResponseMode> {
     vec![ResponseMode::Query, ResponseMode::Fragment]
 }
 
+#[derive(Clone, Copy, Serialize, Deserialize, Debug, PartialEq, Eq)]
+#[serde(rename_all = "snake_case")]
+pub enum Prompt {
+    None,
```
Firstyear, line 455: It triped me up for a bit that this was a specific case separate to Option::None, so maybe we need to rename this and use serde rename to make in "none". That way this could be something like Prompt::NoInteraction or something? Not sure. But I'm not too hung up on it that we have to change it.

server/lib/src/idm/oauth2.rs
```diff
@@ -2317,6 +2327,21 @@ impl IdmServerProxyReadTransaction<'_> {
         // prompt - if set to login, we need to force a re-auth. But we don't want to
         // if the user "only just" logged in, that's annoying. So we need a time window for
         // this, to detect when we should force it to the consent req.
+        //
+        // Idk man, Sure its a little awkward but with how few apps actually use `prompt=login`
```
Firstyear, line 2352: Worth pointing out, that we should document in https://github.com/kanidm/kanidm/blob/master/server/core/src/https/views/login.rs#L328 that this only works because from https://github.com/kanidm/kanidm/blob/master/server/core/src/https/views/oauth2.rs#L182 we aren't actually checking/clearing the login state, so an existing session is ignored at this step.

server/lib/src/idm/oauth2.rs
```diff
@@ -2317,6 +2327,21 @@ impl IdmServerProxyReadTransaction<'_> {
...
+        // I think it is unlikely we will cause a significant amount of user friction by forcing
+        // Login on every request with this param.
+        // We can always add a login timestamp later on.
+
```
Firstyear, line 2356: To prevent needless login's we should add last_authentication https://github.com/kanidm/kanidm/blob/master/proto/src/internal/token.rs#L34 to UAT, and then if that last authentication is within the last say .... 5 minutes? We just allow you through, that way you don't get spammed with prompt=logins. 

For now, we just make last_authentication == issued_at in the authentication code, but we can add special re-auth logic later if we already have a session.

server/lib/src/idm/oauth2.rs
```diff
@@ -2368,7 +2401,12 @@ impl IdmServerProxyReadTransaction<'_> {
 
         let session_id = ident.get_session_id();
 
-        if consent_previously_granted || !o2rs.enable_consent_prompt() {
+        // OIDC Core 1.0 §3.1.2.1:
+        // prompt=consent - The Authorization Server SHOULD prompt the End-User for consent
+        let force_consent =
```
Firstyear: The logic here does read a little awkwardly? 

Maybe it needs to be

```
let consent_prompt_required = 
( !consent_previously_granted || matches(auth_req.prompt, Some(Prompt::Consent)) ) &&  o2rs.enable_consent_prompt();
```

The on the if consent_prompt_required path, we swap the two branches so the if doesn't need a not condition.

server/lib/src/idm/oauth2.rs
```diff
@@ -8121,4 +8189,247 @@ mod tests {
...
+
+        assert!(idms_prox_write.commit().is_ok());
+        new_ident
+    }
+
+    /// OIDC Core 1.0 §3.1.2.1:
+    /// > If an OP receives a prompt value outside the set defined above that it does not understand,
+    /// > it MAY return an error or it MAY ignore it
+    ///
+    /// Currently we return an error but for more compatibility/flexibility we could ignore it instead. Unsure what is best
```
Firstyear: As before, error is best.

server/lib/src/idm/oauth2.rs
```diff
@@ -8121,4 +8189,247 @@ mod tests {
...
+
+        let idms_prox_read = idms.proxy_read().await.unwrap();
+        let pkce_secret = PkceS256Secret::default();
+
+        let auth_req = auth_req_with_prompt(pkce_secret.to_request(), Some(Prompt::Invalid));
+
+        let result = idms_prox_read.check_oauth2_authorisation(Some(&ident), &auth_req, ct);
+
+        assert!(
+            result.unwrap_err() == Oauth2Error::InvalidRequest,
```
Firstyear: try assert_eq!( .... ) :)

server/lib/src/idm/oauth2.rs
CEbbinghaus: Disagree, lets hash this out :3

proto/src/oauth2.rs
CEbbinghaus: It has tripped me up too, I like the Prompt matching its potential values 1:1 so I'm keen to keep it. Perhaps with a comment to elaborate and link back to the RFC

server/lib/src/idm/oauth2.rs
yaleman: [3.1.2.1](https://openid.net/specs/openid-connect-core-1_0.html?#AuthRequest)

> If an OP receives a prompt value outside the set defined above that it does not understand, it MAY return an error or it MAY ignore it; in practice, not returning errors for not-understood values will help facilitate phasing in extensions using new prompt values.

Kanidm does error-on-bad. I say error.

yaleman: That error can be very clear about what's happening because it's not really a security issue, it's a "your app did something dumb" issue, so a "hey, your app sent the wrong value, you've got a bug, contact your admin" rather than a blanket vague "ERROR BAD UNAUTHORIZED BEEP BOOP" response. 😄

yaleman: If they can't spell `prompt` then they've got bigger issues. 😄

Firstyear: https://en.wikipedia.org/wiki/Robustness_principle postels law is a cancer - if you write a client and it emits shit, it's not our job to handle that. It's the callers bug, and I'm not gonna add in work arounds for it. So many ecosystems are now filled to the brim with hacks and workarounds because of postel instead of holding other systems to account to do better. And sometimes those hacks lead to security exploits. 

So we error here. If you give us bad input, that's not our problem.

CEbbinghaus: Well, an invalid value wouldn't trigger any of the functionality which we have implemented so it would just be skipped.

The main point of skipping invalid values (which the RFC points out) is to facilitate moving to new standards, A client written to target a newer RFC which includes a new `prompt` value would proceed with the login flow as normal, rather than throwing an error.

This is exactly what we **have** been doing thus far and it doesn't add any cost to implement since we would just treat it as if it had no Prompt at all.

clients based on newer RFC's can still login, and we only support those mechanisms we understand

CEbbinghaus: The alternative (not the end of the world) is that we just add the new `prompt` value as a enum variant and push that. Same effect of ignoring the behavior requested and proceeding with normal flow, just requires us to commit each time a client uses a new RFC rather than just supporting all unknowns implicitly

Firstyear: But the issue there is then the calling application expects us to handle the process in some way, and we'll likely do an auth flow, or return without a check or something else. So those unknowns in the callers side, trigger an unknown state, a flow that didn't work as intended. It's still much better to error here so that we know something isn't right, then we can make an informed decision to add the new keyword or handle it in a graceful way. Prompt terms also aren't added all the time, so the amount of times we'll hit this situation is strikingly rare.

Where it also matters is if a prompt key is misspelt - the calling application may send prompt=norne and never know that something has gone wrong if we silently ignore this. It also completely changes the expectation of the calling applications as we'd always force authentication in this case, when the client expects us to allow non-authenticated users to bypass this step.

CEbbinghaus: > But the issue there is then the calling application expects us to handle the process in some way

Well, while the application would like us to handle it a certain way, as per the spec it CANNOT expect us to:

> OIDC Core 1.0 §3.1.2.1: If an OP receives a prompt value outside the set defined above that it does not understand, it MAY return an error or it MAY ignore it

In fact, to me the whole `prompt` can be ignored if it is not supported by the OP, so no client could ever depend on its functionality. It was always a hint at "something I'd like you to do" 

Spec also explicitly recommends it 

> OIDC Core 1.0 §3.1.2.1: in practice, not returning errors for not-understood values will help facilitate phasing in extensions using new prompt values.

Since as you pointed out
> Prompt terms also aren't added all the time, so the amount of times we'll hit this situation is strikingly rare.

It is unlikely we will see this happening much, I'm not gonna die on this hill. Just wanted to have defended it a bit longer. So, if both of you are drawing this line in the sand, I'm happy to leave my code as written (although I would like to capture what the invalid value was to print into the logs).

I just like it when things Just Work™️so I'm keen on removing another potential friction point

Edit:

> Where it also matters is if a prompt key is misspelt

this is also one of those things, where it is really the client applications fault, if you mispell the prompt that raises bigger questions about your app... But you cannot depend on it functioning correctly at that point, so both an error and skipping the behavior are valid paths. I just think that while we should log invalid values, we shouldn't fail the request.

Firstyear: > OIDC Core 1.0 §3.1.2.1: If an OP receives a prompt value outside the set defined above that it does not understand, it MAY return an error

Says right there, we may return an error. :)

CEbbinghaus: :)

Firstyear: There have been security issues for less, lets always err on the side of "fail fast" rather than "move fast and break things :P "

server/lib/src/idm/oauth2.rs
CEbbinghaus: Fixed because we are no longer supporting consent prompt, since that requires a much more in depth implementation specifically around supporting `prompt=login%20consent` (we would need to reset the consent, for only the 1 login we are actively redirecting to)

server/lib/src/idm/oauth2.rs
CEbbinghaus: Problem is that this is not spec compliant and I am not overly keen on potentially allowing a stolen access token to bypass `prompt=login` so long as it was issued within the last 5 minutes.

So if at all possible I'd like to leave the UX improvements for a followup PR rather than "spec implementation"

CEbbinghaus:
HashSet is a forbidden type but we cannot use BTreeSet since it requires Ord which we would need to manually implement (which I didn't want to do). @Firstyear Please LMK what you'd prefer in terms of capturing the multiple values.

In terms of Implementaion however, this now properly implements the spec apart from the `consent` prompt which I also think is the least used/requested.

proto/src/oauth2.rs
Firstyear: That works for me too.

server/lib/src/idm/oauth2.rs
Firstyear: At the least, we should open an issue to review this so it's not forgotten.

Firstyear:
> HashSet is a forbidden type but we cannot use BTreeSet since it requires Ord which we would need to manually implement (which I didn't want to do). @Firstyear Please LMK what you'd prefer in terms of capturing the multiple values.

std::collections::HashSet is forbidden, but hashbrown::HashSet can be used (that's what the check is for). 

However, HashSet here will likely be MUCH slower, and is probably the wrong choice. 

What's the reason why you can't derive Ord/PartialOrd on Prompt?

server/lib/src/idm/oauth2.rs
yaleman: Prompt is prompt TBH, if the client app's being a pain then it's on them.

CEbbinghaus:
Ord automatic because `Invalid(String)` variant, We can do a PartialOrd ourselves and just do a string comparison if that is what you want, Wasn't sure what the desired path forward would be

Firstyear:
> Ord automatic because `Invalid(String)` variant, We can do a PartialOrd ourselves and just do a string comparison if that is what you want, Wasn't sure what the desired path forward would be

A simpler answer is just use a Vec, and use .contains on that. Yes it's a linear search, but there shouldn't be more than a small number of simultaneous prompt strings anyway. I'd say limiting it to 4 prompts and erroring if there are too many, that way we have an upper bound.

CEbbinghaus:
Changed it to a vec and Added Tests for Deserialization, We cannot limit the vec via Serde which means that if someone were to provide 5000 prompt values, we would have to first deserialize all of them before we limited the list. so it seems a little silly to go down that route. LMK what else needs to change

Firstyear:
> Changed it to a vec and Added Tests for Deserialization, We cannot limit the vec via Serde which means that if someone were to provide 5000 prompt values, we would have to first deserialize all of them before we limited the list. so it seems a little silly to go down that route. LMK what else needs to change

but that's okay, because the cost is in the *search* of the prompt vec, not in the deserialise of it. There are other limits in place to prevent that.

server/lib/src/idm/oauth2.rs
```diff
@@ -2178,6 +2182,25 @@ impl IdmServerProxyReadTransaction<'_> {
             }
         };
 
+        let invalid_prompts: Vec<&str> = auth_req
```
Firstyear, line 2193: Before this line, you need to have a check for prompt.len() > 5 and error if true.

Firstyear:
Apparently I had some forgotten comments hang around, but yeah, looking good. :)

server/lib/src/idm/oauth2.rs
CEbbinghaus: Will do, that is easy. Just wanted confirmation that it is still required. Sounds good to limit it for searching specifically.

server/lib/src/idm/oauth2.rs
CEbbinghaus: Done, Tracking issue raised #4273

CEbbinghaus:
No, Thank **you**, This was a blast and had me implementing OIDC Spec, something I will wear as a badge forever :3

ToxicMushroom:
I don't fully understand why but this seems to break docs on master https://github.com/kanidm/kanidm/actions/runs/24493455034/job/71583148614 

Though it succeeded in this PR here https://github.com/kanidm/kanidm/actions/runs/24493043935/job/71581836668

CEbbinghaus:
Wild, Yeah that is absolutely MB (haven't done much docs generation in Rust yet). Should have been caught by CI before merge

Firstyear:
@CEbbinghaus Can I leave you to make a follow up PR to fix it?

CEbbinghaus:
@Firstyear #4278 Ready for Review

Firstyear:
Thank you <3

PR #4239, opened 2026-03-30
add dependency data to released containers

### LINKED ISSUE #2432: Add compatibility to trivy scans

I would like to use the [trivy CVE scanner](https://trivy.dev/) which is wired to my company's [harbor docker repository](https://goharbor.io/) to ensure a certain security level is in place. You can set security limits to restrict pull access to that image then. Right now the scan yields no problems but that is due to the software meta data is not present for trivy to check.

According to https://aquasecurity.github.io/trivy/v0.48/docs/coverage/language/rust/ it would work if binaries are build with https://github.com/rust-secure-code/cargo-auditable.

Alternatively I have to check out the Git tag for docker image I use and scan it with trivy. This solution is quite manually and not "baked" in to the process of using the docker image.

**Additional context**

oliviacrain:
# Change summary

The tool and server containers now use cargo-auditable to build rust binaries in a way that exposes dependency data to vulnerability scanners.

Fixes #2432

## Testing
- Extracted built binaries from both containers, ran `cargo audit bin <path to binary>` on each binary, and saw output indicating `cargo auditable` data  was found in each binary.
- Ran `trivy image <local docker tag>` on both images and saw that the tool included our built binaries in its report summary.

## Checklist

### DIFF
--- a/server/Dockerfile
+++ b/server/Dockerfile
@@ -21,6 +21,7 @@ RUN \
     zypper install -y --no-recommends \
         sccache \
         cargo \
+        cargo-auditable \
         clang \
         gcc \
         cmake \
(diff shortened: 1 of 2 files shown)


oliviacrain:
Side note: I found it interesting that `cargo audit bin` found CVE-2023-49092, but `trivy` missed it. I thought this might be an issue with `trivy` not picking up the `cargo-auditable` data, but running `trivy image --format spdx-json <container image>` outputs a perfectly cromulent SBOM with all the dependency data, so :woman_shrugging:

Firstyear:
FYI https://github.com/advisories/GHSA-c38w-74pg-36hr isn't a super critical thing. We rarely use RSA keys anyway, and where we do we don't allow attacker controlled data (it's used in signing oauth2 tokens only which is server controlled data).

PR #4248, opened 2026-04-05
Log cleanup

yaleman:
# Change summary

Reduce the incredible number of info-level logs that the platform's emitting. Log line volume went up ~40% between 1.5 and 1.9 for zero functional benefit, this reduces the level a *lot*.

Fixes #

Checklist


proto/src/v1/auth.rs
```diff
@@ -141,6 +141,15 @@ pub enum AuthIssueSession {
     Cookie,
 }
 
+impl AsRef<str> for AuthIssueSession {
```
Firstyear: Shouldn't this be fmt display instead?

server/lib/src/idm/authsession/mod.rs
```diff
@@ -1640,9 +1640,9 @@ impl AuthSession {
                     }
                 };
 
-                security_info!(
-                    "Issuing {:?} session ({:?}) {} for {} {}",
-                    self.issue,
+                info!(
```
Firstyear: I think this is security info as it has security relevant details for the session issuance.

server/lib/src/idm/authentication.rs
```diff
@@ -89,7 +90,9 @@ impl fmt::Debug for AuthState {
             AuthState::Continue(allow) => write!(f, "AuthState::Continue({allow:?})"),
             AuthState::External(allow) => write!(f, "AuthState::External({allow:?})"),
             AuthState::Denied(reason) => write!(f, "AuthState::Denied({reason:?})"),
-            AuthState::Success(_token, issue) => write!(f, "AuthState::Success({issue:?})"),
+            AuthState::Success(_token, issue) => {
+                write!(f, "AuthState::Success({})", issue.as_ref())
```
Firstyear: Yeah, as per above, should be fmt display, not asref str.

server/lib/src/idm/authsession/mod.rs
yaleman: I thought we were getting away from security_info because it's just noisy and weird?

Firstyear: Yeah, but you also still used it in this PR ;) 

Anyway, I think security_info might be the exception, it's the admin_warn/debug that I want to mostly remove.

proto/src/v1/auth.rs
yaleman: Yeah I was trying to avoid another layer of invoking `write!` but fixed

server/lib/src/idm/authsession/mod.rs
yaleman: Definitely a discussion for another thread, but it seems like noise for noise sake, it logs an event_tag_id which isn't documented or mentioned anywhere, and also double-logs the level as its own string

eg:

`9c84d65d-2c9f-4622-8eb7-5e47886b65aa **INFO**        ┕━ ｉ [info]: search | event_tag_id: 10 | initiator: User( anonymous@example.com, 00000000-0000-0000-0000-ffffffffffff ) (ec13f369-0d12-49c4-af3f-6849b2d994a2, read only)`

anyway applied

PR #4259, opened 2026-04-10
20260331 send pw reset emails

### LINKED ISSUE #2299: FR: Add password reset form in the Web UI

Firstyear:
# Change summary

- This adds a feature to enable account recovery emails to be sent to accounts, provided they have a registered email and prove knowledge of it.
- Account recovery is protected by a CSRF system that was described as "what if CSRF hurt to touch". It relies on SameSite=strict tokens, and demands a proof of work from the clients browser as a minimal anti-spam measure. 
- This feature lays out the majority of the ground work for user-self-registration. 

cc @tacerus 

Fixes #2299 

Checklist


Firstyear:
In the PR I have been "somewhat" inconsistent with naming of the feature, so I'd welcome some input to rename the structs/function. Should it be account-recovery? Anonymous credential reset? Something else?

book/src/accounts/authentication_and_credentials.md
```diff
@@ -87,6 +89,14 @@ kanidm person credential create-reset-token demo_user 86400 --name idm_admin
 If the user wishes you can direct them to `https://idm.mydomain.name/ui/reset` where they can manually enter their token
 value.
 
+If you have [configured outbound email](/email_setup.md) you can send the reset link directly to the user's email with:
+
+```bash
+kanidm person credential send-reset-token <account_id> [TTL] [ALTERNATE_EMAIL]
```
CEbbinghaus, line 95: I feel like TTL should be a `--ttl` option rather than an argument, so it becomes:
```
kanidm person credential send-reset-token [--ttl] <account_id> [ALTERNATE_EMAIL]
```

Allowing for quick and easy "send this user a reset email under an alternate address using the default TTL"

libs/client/src/domain.rs
```diff
@@ -17,6 +19,20 @@ impl KanidmClient {
...
 
+    pub async fn idm_set_domain_allow_credential_reset_email(
+        &self,
+        enable: bool,
+    ) -> Result<(), ClientError> {
+        self.perform_put_request(
+            &format!(
+                "{}{}",
+                "/v1/domain/_attr/", ATTR_DOMAIN_ALLOW_CREDENTIAL_RESET_EMAIL
+            ),
```
CEbbinghaus, line 30:
```suggestion
            &format!(
                "/v1/domain/_attr/{}",
                ATTR_DOMAIN_ALLOW_CREDENTIAL_RESET_EMAIL
            ),
```

How come not embed the path into the format string?

server/core/src/actors/write_action.rs
```diff
@@ -0,0 +1,35 @@
...
+        &self,
+        email: String,
+        eventid: Uuid,
+    ) -> Result<(), OperationError> {
+        let ct = duration_from_epoch_now();
+        let mut idms_prox_write = self.idms.proxy_write(ct).await?;
+
+        let event = CredentialUpdateAnonymousAccountRequest {
+            email,
+            max_ttl: None,
```
CEbbinghaus, line 22: Dumb question but shouldn't this be filled via argument?

CEbbinghaus: Well, I now realize that we call the "anonymous" variant which obviously wouldn't recieve the TTL. But that begs the question of why this method is called `action_credential_reset_link` instead of something like `action_anonymous_credential_reset_link`

server/lib/src/idm/credupdatesession.rs
```diff
@@ -977,6 +980,55 @@ impl IdmServerProxyWriteTransaction<'_> {
...
+            PartialValue::EmailAddress(event.email.clone())
+        ));
+
+        let mut entries = self
+            .qs_write
+            .impersonate_search(filter.clone(), filter, &ident)?;
+
+        let entry = entries
+            .pop()
+            .and_then(|entry| entries.is_empty().then_some(entry))
```
CEbbinghaus, line 1009: TIL, neat trick

server/lib/src/idm/credupdatesession.rs
```diff
@@ -977,6 +980,55 @@ impl IdmServerProxyWriteTransaction<'_> {
...
+
+        // Look up the account.
+        let filter = filter!(f_eq(
+            Attribute::Mail,
+            PartialValue::EmailAddress(event.email.clone())
+        ));
+
+        let mut entries = self
+            .qs_write
+            .impersonate_search(filter.clone(), filter, &ident)?;
```
CEbbinghaus, line 1005: This is where a timing attack would occur, I presume that since we aren't strcmp'ing the emails that our query engine internally already ensures that no sidechannel is possible? How does it do this?

server/core/src/https/views/csrf.rs
```diff
@@ -0,0 +1,178 @@
...
+    jar: &CookieJar,
+
+    related_input: &[u8],
+    solution: &CsrfSolution,
+
+    current_time: Duration,
+) -> Result<(), ()> {
+    let nonce: Nonce = cookies::get_signed(state, jar, COOKIE_CSRF_NONCE).ok_or(())?;
+
+    warn!(?related_input, ?solution);
```
CEbbinghaus: ?

CEbbinghaus (changes requested):
Less comments to fix, more questions

book/src/accounts/authentication_and_credentials.md
Firstyear: Yep that makes sense.

libs/client/src/domain.rs
Firstyear: Great question.

server/core/src/actors/write_action.rs
Firstyear: Yeah, this comes back to the naming question of the feature.

server/lib/src/idm/credupdatesession.rs
Firstyear: Yeah, this is where it would occur (well actually, it would occur higher up but anyway). Given the POW on the submission, as it stands today it'll be pretty computationally expensive to pull of an attack like that today, but we could still wrap this with a fixed timeout.

server/core/src/https/views/csrf.rs
Firstyear: Debugging, I can remove it.

Firstyear:
What about the name "AccountRecoveryRequest" for the feature? I think that makes the most sense.

CEbbinghaus:
More than happy with `AccountRecoveryRequest`, or `AccountRecovery` for short. Short sweet and simple.

proto/src/internal/error.rs
```diff
@@ -457,8 +459,11 @@ impl OperationError {
             Self::CU0004SessionInconsistent => Some("The session is unable to be committed due to unresolved warnings.".into()),
             Self::CU0005IntentTokenConflict => Some("The intent token used to create this session has been reused in another browser/tab and may not proceed.".into()),
             Self::CU0006IntentTokenInvalidated => Some("The intent token has been invalidated/revoked before the commit could be accepted. Has it been used in another browser or tab?".into()),
-            Self::CU0007AccountEmailNotFound => Some("The requested email to send the credential update intent token to is not registered to the account.".into()),
+            Self::CU0007AccountEmailNotFound |
+            Self::CU0009AccountEmailNotFound
+            => Some("The requested email to send the credential update intent token to is not registered to the account.".into()),
```
yaleman, line 464:
```suggestion
            => Some("The target email for a credential update intent token send request is not registered on the account.".into()),
```

server/core/static/csrf.js
```diff
@@ -0,0 +1,106 @@
...
+
+async function protectedSubmitHandler(event) {
+    event.preventDefault();
+
+    const related_input_id = document.getElementById("protected_form_related_input_id").innerText;
+
+    document.getElementById("protected_form_submit").disabled = true;
+
+    const nonce = value_from_hex("protected_form_nonce_hex");
+    const mask = value_from_hex("protected_form_mask_hex");
```
yaleman, line 16: the use of innerText/p elements to store data here's ... far from best-practice.

[data attributes should be used](https://developer.mozilla.org/en-US/docs/Web/HTML/How_to/Use_data_attributes)

so the form can have

```html
<form id="protected_form"
  data-nonce-hex="(( csrf.nonce_hex ))"
  data-mask-hex="(( csrf.mask_hex ))"
... etc
```

and in the javascript you can load/refer to it thusly:

```javascript

const form = document.querySelector("#protected_form");

const nonce = value_from_hex(form.dataset.nonceHex);
```

server/core/static/csrf.js
```diff
@@ -0,0 +1,106 @@
...
+
+    let expected_buffer = new ArrayBuffer(4);
```
yaleman, line 18:
```suggestion
    const expected_buffer = new ArrayBuffer(4);
```

server/core/static/csrf.js
```diff
@@ -0,0 +1,106 @@
...
+    let expected_view = new DataView(expected_buffer);
```
yaleman, line 19:
```suggestion
    const expected_view = new DataView(expected_buffer);
```

server/core/static/csrf.js
```diff
@@ -0,0 +1,106 @@
...
+
+        let signature_view = new DataView(signature_bytes.buffer);
+        const signature_u32 = signature_view.getUint32();
+
+        const signature_u32_masked = signature_u32 & mask_u32;
+
+        const matches = expected_u32 == signature_u32_masked;
+        if (matches) {
+            break;
+        }
```
yaleman, line 79:
```suggestion
        if (expected_u32 === signature_u32_masked) {
            break;
        }
```

server/core/static/csrf.js
```diff
@@ -0,0 +1,106 @@
...
+    const nonce = value_from_hex("protected_form_nonce_hex");
+    const mask = value_from_hex("protected_form_mask_hex");
+
+    let expected_buffer = new ArrayBuffer(4);
+    let expected_view = new DataView(expected_buffer);
+    // DataView get/set always writes BigEndian
+    expected_view.setUint32(0);
+    const expected_u32 = expected_view.getUint32();
+
+    let mask_view = new DataView(mask.buffer);
```
yaleman, line 24:
```suggestion
    const mask_view = new DataView(mask.buffer);
```

server/core/static/csrf.js
```diff
@@ -0,0 +1,106 @@
...
+    const mask_u32 = mask_view.getUint32();
+
+    let key = await window.crypto.subtle.importKey(
```
yaleman, line 27:
```suggestion
    const key = await window.crypto.subtle.importKey(
```

server/core/static/csrf.js
```diff
@@ -0,0 +1,106 @@
...
+        true,
+        ["sign"],
+    );
+
+    const email_value = document.getElementById(related_input_id).value;
+    const encoder = new TextEncoder();
+    const email_bytes = encoder.encode(email_value);
+
+    let timestamp_buffer = new ArrayBuffer(8);
+    let timestamp_view = new DataView(timestamp_buffer);
```
yaleman, line 43:
```suggestion
    const timestamp_buffer = new ArrayBuffer(8);
    const timestamp_view = new DataView(timestamp_buffer);
```

server/core/static/csrf.js
```diff
@@ -0,0 +1,106 @@
...
+    // DataView get/set always writes BigEndian
+    let timestamp = BigInt(Date.now());
```
yaleman, line 45:
```suggestion
    const timestamp = BigInt(Date.now());
```

server/core/static/csrf.js
```diff
@@ -0,0 +1,106 @@
...
+        var mergedArray = new Uint8Array(email_bytes.length + solution_bytes.length + timestamp_bytes.length);
+        mergedArray.set(email_bytes);
+        mergedArray.set(solution_bytes, email_bytes.length);
+        mergedArray.set(timestamp_bytes, email_bytes.length + solution_bytes.length);
+
+        let signature = await window.crypto.subtle.sign("HMAC", key, mergedArray);
+
+        let signature_bytes = new Uint8Array(signature);
+
+        let signature_view = new DataView(signature_bytes.buffer);
```
yaleman, line 71:
```suggestion
        const signature = await window.crypto.subtle.sign("HMAC", key, mergedArray);

        const signature_bytes = new Uint8Array(signature);

        const signature_view = new DataView(signature_bytes.buffer);
```

server/core/static/csrf.js
```diff
@@ -0,0 +1,106 @@
...
+    var solution = new Uint32Array([0]);
+
+    while (true) {
+        // Oh Javascript, how we love theee.
+        const buffer = new ArrayBuffer(4);
+        new DataView(buffer).setUint32(0, solution[0]);
+        // This is what we have to send back.
+        const solution_bytes = new Uint8Array(buffer);
+
+        var mergedArray = new Uint8Array(email_bytes.length + solution_bytes.length + timestamp_bytes.length);
```
yaleman, line 62:
```suggestion
        const mergedArray = new Uint8Array(email_bytes.length + solution_bytes.length + timestamp_bytes.length);
```

server/core/templates/recover_form.html
```diff
@@ -0,0 +1,41 @@
...
+    src="/pkg/csrf.js?v=((crate::https::cache_buster::get_cache_buster_key()))"
+    defer></script>
+(% endblock %)
+
+(% block content %)
+		<p>Enter your email to recover your account</p>
+
+		<p id="protected_form_nonce_hex" hidden>(( csrf.nonce_hex ))</p>
+		<p id="protected_form_mask_hex" hidden>(( csrf.mask_hex ))</p>
+		<p id="protected_form_related_input_id" hidden>(( csrf.related_input_id ))</p>
```
yaleman, line 14: as noted in csrf.js these should be data elements on the form

yaleman (changes requested):
buncha javascript/html fixing needed on this one 😄

Firstyear:
Okay, renamed everything to be consistent, and resolved all the JS errors. I might have missed something so I plan to self-review this tomorrow morning too.

Firstyear:
Okay, I did the self-review and found a few things I'm missed, but should be good to go now :)

server/core/templates/login.html
```diff
@@ -61,5 +61,11 @@
 			class="btn btn-primary"
 		>Begin</button>
 	</div>
+	(% if display_ctx.domain_info.allow_account_recovery() -%)
+	<div class="input-group mb-3 justify-content-md-center">
+		<a href="(( Urls::Recover ))"
+		>Recover Account</a>
```
CEbbinghaus:
```suggestion
		<a href="(( Urls::Recover ))">Recover Account</a>
```

server/core/templates/login.html
```diff
@@ -61,5 +61,11 @@
 			class="btn btn-primary"
 		>Begin</button>
 	</div>
+	(% if display_ctx.domain_info.allow_account_recovery() -%)
```
CEbbinghaus, line 64:
```suggestion
	(% if display_ctx.domain_info.allow_account_recovery() %)
```

server/core/static/csrf.js
```diff
@@ -0,0 +1,99 @@
...
+
+    // We have to do a dance to get back to BE for xmit.
+    const buffer = new ArrayBuffer(4);
+    new DataView(buffer).setUint32(0, solution[0]);
+
+    // This is what we have to send back.
+    const solution_bytes = new Uint8Array(buffer);
+    const solution_hex = solution_bytes.toHex();
+
+    console.log(solution_hex);
```
CEbbinghaus: We keeping this in?

CEbbinghaus (dismissed):
God I love the new name. It feels so right

server/core/templates/login.html
CEbbinghaus: not 1000% certain on this, but its in none of the other blocks

Firstyear: It means that the line is removed in the final output rather than being an empty line that is rendered.

CEbbinghaus (approved):
I was wondering why that function was still using the old name.

Super happy with how it turned out, (custom csrf impl is fun too).

yaleman (approved):

Firstyear:
Woohoo! Thank you both so much for your invaluable feedback. <3

PR #4271, opened 2026-04-15
Handle multivalue URLs in SCIM

### LINKED ISSUE #4268: SCIM migration rejects array-valued oauth2_rs_origin with SC0007UrlSyntaxInvalid
### I did this

Declared an OAuth2 resource server migration in HJSON with a multivalue `oauth2_rs_origin` array to add multiple redirect URIs (including mobile app custom scheme URIs):

```hjson
{
  "id": "9cc52cb2-b7ff-4c1f-b513-f0bd35ce30d0",
  "assertions": [
    {
      "state": "present",
      "id": "c51de6b3-bf98-4a59-9620-6bef56fa2339",
      "oauth2_rs_origin": [
        "https://photos.example.com/auth/login",
        "https://photos.example.com/user-settings",
        "https://photos.example.com/api/oauth/mobile-redirect",
        "app.immich:/",
        "app.immich:///oauth-callback"
      ]
    }
  ]
}
```

The same redirect URIs work when added imperatively via `kanidm system oauth2 add-redirect-url`.

### I expected the following

The migration should apply the multivalue `oauth2_rs_origin` entries to the existing OAuth2 resource server, matching the behavior of `kanidm system oauth2 add-redirect-url`.

### This happened instead

The migration fails with `SC0007UrlSyntaxInvalid`:

```
ERROR ┕━ 🚨 [error]: SCIM URL syntax invalid | err: Error("invalid type: sequence, expected a string representing an URL", line: 0, column: 0)
ERROR ┕━ 🚨 [error]: Failed to apply migration | err: SC0007UrlSyntaxInvalid | path: /run/kanidm/migrations/35-immich-mobile-redirects.hjson
```

The SCIM migration deserializer expects `oauth2_rs_origin` to be a single string, not an array. This causes the entire migration batch to roll back, preventing any new entries declared in other migration files from being applied.

Using a single string value for `oauth2_rs_origin` works, but there is no way to declaratively add multiple redirect URIs via the migration path.

### Kanidm version details

- Output of `kanidm(d) version`: kanidm 1.9.0
- Are you running it in a container? If so, which image/tag?: No
- If not a container, how'd you install it: NixOS package (nixpkgs)
- Operating System / Version: NixOS 25.11 (Linux beefcake 6.12.x x86_64)

### Any other comments

Workaround: add the extra redirect URIs imperatively with `kanidm system oauth2 add-redirect-url <client> <url>` and keep only a single `oauth2_rs_origin` string in the migration file. This is fragile since the imperative state isn't tracked in version control.

Ideally the SCIM migration path would accept `oauth2_rs_origin` as either a string (single) or array of strings (multiple), matching what the REST API and CLI already support.

Firstyear:
# Change summary

- Correctly handle multivalue inputs in SCIM urls. 
- Add tests to ensure this doesn't happen again. 

Fixes #4268

Checklist

### DIFF
--- a/proto/src/scim_v1/client.rs
+++ b/proto/src/scim_v1/client.rs
@@ -13,6 +13,7 @@ use std::collections::{BTreeMap, BTreeSet};
 use std::num::NonZeroU64;
 use time::format_description::well_known::Rfc3339;
 use time::OffsetDateTime;
+use url::Url;
 use utoipa::ToSchema;
 use uuid::Uuid;
 
@@ -270,6 +271,10 @@ pub struct ScimEntryPutKanidm {
 #[serde_as]
 #[derive(Deserialize, Serialize, Debug, Clone)]
 pub struct ScimStrings(#[serde_as(as = "OneOrMany<_, PreferMany>")] pub Vec<String>);
+
+#[serde_as]
+#[derive(Deserialize, Serialize, Debug, Clone)]
+pub struct ScimUrls(#[serde_as(as = "OneOrMany<_, PreferMany>")] pub Vec<Url>);
 
 #[derive(Debug, Clone, Serialize, Deserialize, Default, ToSchema)]
 pub struct ScimEntryPostGeneric {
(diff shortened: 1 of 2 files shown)


CEbbinghaus (approved):
I feel like I'm too stupid to understand that test implementation (no assert?) but the rest all looks good.

Firstyear:
For reference, the asserts are here: https://github.com/kanidm/kanidm/blob/master/server/lib/src/valueset/mod.rs#L1075 

It tests that given a ValueSet it can be serialised and compares to a known string version of the same serialised content.

PR #4279, opened 2026-04-17
Add missing kanidm-mail-sender binary to kanidm/tools container

sfnke:
# Change summary

While reviewing the Kanidm documentation for sending password reset emails, I noticed that the required commands could not be executed because the corresponding binary was not present in the container. Although the binary is built during the builder stage, it is never copied into the final image.

Checklist


yaleman (approved):
Thanks!

Firstyear (approved):

Issue #4249, opened 2026-04-06
[WebUI] Group member management | silent 404 "nomatchingentries" error


ToxicMushroom:
### I did this
I added idm_admin as a member to a group via the admin webui. 

### I expected the following
Server responds with a toast saying saved, since they were not a member yet.

### This happened instead
No toast or feedback, server responded with 404 "nomatchingentries"

### Kanidm version details

- Output of `kanidm(d) version`: 7418fd5ea82703b3a37f0d79013322034b262b08
- Are you running it in a container? no
- If not a container, how'd you install it: ./run_insecure_dev_server.sh
- Operating System / Version (On Unix please post the output of `uname -a`): arch

### Any other comments
The cause seems to be this snippet here: https://github.com/ToxicMushroom/kanidm/commit/50bd47451a5f7bfab4951ad47b1c6b964c3b9b86#diff-8ffd18764ae406d2638c2eb604078c620bf062021f40105aa084432327aa1128R348-R358

- It returns this error when I try to retrieve `idm_admin` or `admin`, both via **name** or **uuid**.
Manually created service accounts e.g. `kanidm service-account create nonsystem NonSystem idm_admin` do not have this issue.

<details>

<summary>related trace logs: </summary>

```
b34aa827-1ecf-47e7-bd6e-f1a183df419e INFO     ┕━ scim_entry_id_get [ 5.83ms | 0.34% / 14.06% ]
b34aa827-1ecf-47e7-bd6e-f1a183df419e DEBUG       ┝━ proxy_read [ 42.8µs | 0.08% / 0.10% ]
b34aa827-1ecf-47e7-bd6e-f1a183df419e DEBUG       │  ┕━ read_acquire_ticket [ 8.56µs | 0.02% ]
b34aa827-1ecf-47e7-bd6e-f1a183df419e INFO        ┝━ validate_client_auth_info_to_ident [ 4.59ms | 10.50% / 11.07% ]
b34aa827-1ecf-47e7-bd6e-f1a183df419e DEBUG       │  ┝━ 🐛 [debug]:  | kid: KeyId { id: "254027c275ce" }
b34aa827-1ecf-47e7-bd6e-f1a183df419e TRACE       │  ┝━ 📍 [trace]: Session not yet expired | ct_odt: 2026-04-06 13:31:47.830753267 +00:00:00 | exp: 2026-04-07 12:20:55.0 +00:00:00
[...]

PR #4285, opened 2026-04-19
Group UI polish

ToxicMushroom:
# Change summary

- Swap 403 errors, used for showing unauthorized toast in the admin-ui.
This change concerns me of unintended effects, I'm not sure what all throws a 403 and if it should always have their responses swapped into the UI. 
On the other hand, I'm not a fan of hiding failed requests of the user and they should probably be swapped via a toast or some other communication mechanism.

- CSS changes to make it look more even/aligned.
- Give feedback when the user cannot modify any attributes, but CAN read-write.

## notes
1. When trying to add a member that doesn't exist, kanidm throws a `Plugin(ReferentialIntegrity("Uuid referenced not found in database"))` OperationError. See #4249 

Checklist


server/core/templates/admin/admin_group_view_partial.html
```diff
@@ -44,14 +44,19 @@
...
-        (% if can_rw -%)
+   
+    (% if can_rw -%)
+        (% if can_modify_any_attr -%)
             <button type="submit" class="btn btn-primary">Save</button>
         (% else -%)
-            <a href="/ui/unlock">
-                <button type="button" class="btn btn-secondary">Unlock Edit 🔒</button>
-            </a>
+            <div class="alert alert-secondary" role="alert">
```
Firstyear: I don't know if this is needed, since if all the fields are disabled, it should be obvious?

ToxicMushroom: Sure, I'll remove it

ToxicMushroom:
@Firstyear Do I add a check to work around #4249 or do I just leave it for a future proper fix ?

Firstyear:
yeah, leave it for now :)

Firstyear (approved):
All works as promised!

PR #4312, opened 2026-05-06
Resolve base64 encoding of webauthn fields

Firstyear:
Webauthn 6.1-dev enforces stricter rules about base64 encoding. The JS on the credential creation page was encoding with standard/padded rather than url safe unpadded. This caused webauthn to refuse to create credentials.

- Initially I thought it was due to uri encoding, but it wasn't. I did change the forms to use base64 instead, which I'm happy to revert, I think the uri encoding worked fine. 

Checklist

### DIFF
--- a/server/core/src/https/views/login.rs
+++ b/server/core/src/https/views/login.rs
@@ -8,6 +8,7 @@ use crate::https::{
 };
 use askama::Template;
 use askama_web::WebTemplate;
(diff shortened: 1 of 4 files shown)


Firstyear:
Only affects 1.10, not 1.9

yaleman (approved):
Yaaaaay the many kinds of encoding.

PR #4315, opened 2026-05-07
20260507 schema in memory

Firstyear:
# Change summary

This change at it's most simple, moves schema from a property of the database, to an in memory property based on the current domain level.

This change greatly improves test performance, as well as server startup speed. 

So the natural question is "why make this change"?

Once upon a time, when I started Kanidm I had just come out of working heavily on LDAP, and a little bit of Samba 4. Both of these were large inspirations for the project, both in what they did *well* but also avoiding what they did *badly*. Because of this, something that 389-ds did poorly was schema management. It was defined by configuration files that were unique to each server in replication. While schema *could* be replicated, it was more often than not, unreliable. Samba 4 (AD) made schema a first class database object - it was replicated along with all other entries, defined a really good set of propteries about attributes and classes, and generally was really hard to mess up.

Kanidm's schema since it's inception was actually modeled on AD - and to this day it has served that purpose amazingly.

And so now you still are asking "why make this change"?

Because having schema either as config files or in the database, is something needed if you expect the schema to be *customised*.

As we have evolved, it's become clear that people are interested in custom attributes, but they aren't interested in making new "generic ldap-esque objects" or modifying the properties of what already exists. In fact, we've tried to STOP people from realising that they could be making these changes! This is because we never have good strategies in place to handle custom access controls, schema, migrations, etc. 

As a result, I think it's better to have a clear split - there is schema that we as the Kanidm project ship. It's static, defined, well tested, and comes with built in access controls and objects. It solves 95% of what users want. And in future when we define and allow custom attributes, they can extend things in a controlled and methodical way that doesn't run risks regarding migration. 

This change reflects that - if we take customisation off the table there is no need to have the schema in the DB. Since each server knows what the schema "should be" for a given domain level, they can always end up with the same "view" of how to enforce entry structure. 

Today this PR does make things a little more complex because we have to support both DB schema and in memory. This is so we can properly test the migration/upgrade paths. But in future we'll be able to remove those, and as a result remove a lot of code. 

Checklist


server/lib/src/server/migrations.rs
```diff
@@ -910,6 +917,15 @@ impl QueryServerWriteTransaction<'_> {
         // new indexes, or this is a bootstrap and we have no indexes yet.
         self.reindex(false)?;
 
+        // Delete all existing DB contained schema.
+
+        let filter = filter!(f_and(vec![
+            f_eq(Attribute::Class, EntryClass::ClassType.into()),
+            f_eq(Attribute::Class, EntryClass::AttributeType.into()),
+        ]));
```
yaleman, line 925: Seems weird to do an reindex and THEN delete, since the delete would leave less to index? 🤔

server/lib/src/migration_data/system/schema.rs
```diff
@@ -0,0 +1,1223 @@
+use crate::prelude::*;
+use crate::schema::Replicated;
+
```
yaleman, line 3: Since we're doing a massive change here, and as I understand it...  SchemaAttributes are just "the schema for a given variant of `Attribute`" - why don't we take the opportunity to make a MigrationTrait/`schema()`/something that returns the schema for a given attribute variant? 

That'd avoid all the LazyLock stuff, statics etc, as we're cloning them on use at the moment either way; it also ensures we're building a schema for every attribute unless we yeet `_ => kekeke` at the end

yaleman (changes requested):
I have thoughts 😄 see the comment about schema

server/lib/src/server/migrations.rs
Firstyear: If we don't reindex then we have to full-table-scan to do the delete.

server/lib/src/migration_data/system/schema.rs
Firstyear: Yes, ultimately that could be a good approach - but that's a *massive* shift and change in how the whole schema / database validation works today. So this is an incremental step.

yaleman (approved):

Issue #4010, opened 2025-12-10
Prevent custom OIDC Claim mappings from having the same name as built-in claims


kaivol:
### Is your feature request related to a problem? Please describe

I created a custom OIDC Claim map with the name `groups` without realizing that this claim is already added by the built-in `groups` scope, which caused me some confusion.  

### Describe the solution you'd like

Don't allow creating custom claim maps for a claim that belongs to a built-in scope, or at least show a warning message. 

### Describe alternatives you've considered

Don't change anything 🤷

However, the consequences of having a custom claim map with the same name as a built-in claim should probably be documented in that case.

yaleman:
marking as a bug because the system should *totally* block that IMO

trissylegs:
Looking at this to try and have a go at contributing.

Firstyear:
@trissylegs Sounds good - I think the best place to start is https://github.com/kanidm/kanidm/blob/master/server/lib/src/valueset/oauth.rs#L753

trissylegs:
// Now perform our custom claim's from scopes. We do these second so that
    // a user can't stomp our claim names.

https://github.com/kanidm/kanidm/blob/c070a41192e9f82f0b58985f02e6055980ba7929/server/lib/src/idm/oauth2.rs#L3278

Well at least it was thought about.

Firstyear:
We do try to think about things :) 

The big issue with trying to prevent user defined claims stomping system ones is we can't control what the external application expects the claim name to be.

For example, look at zfs properties - you can define custom user properties but they HAVE to be `:` namespaced. So a user zfs property is `user:propertyname` which by definition can never conflict since no base zfs props have a `:` in them.

We can't do that for user claims because the external application expects a specific set of names of the claim. So we can't just force them all to have a specific delimiter. 

So then we have to ask what do we do? Do we rename our claims? But there are some "standard" claims people rely on. Do we force a specification of the claim name?

Or do we just "block" the specific claim names that we define one-at-a-time. What happens if we add new claims? Do we then break an existing claim value a user set? 

Lots of questions :)

Gabgobie:
> The big issue with trying to prevent user defined claims stomping system ones is we can't control what the external application expects the claim name to be.

This is the most important aspect for me. I already was (still am) in a situation where the external application expected to receive the `groups` claim and I absolutely needed to stomp[^1] it.

[...]

Issue #2641, opened 2024-03-09
On richer custom claim values


dvv:
Hi!

I'm exploring the ability to generate tokens with richer custom claims.
Say, I would like for app to define sub's access to a [mercure](https://mercure.rocks/docs/mercure) hub:

```
{
  "sub": "MY-SUB-HERE",
  ...
  "mercure": {
    "publish": [ "kanidm" ],
    "subscribe": [ "kanidm" ]
  }
}
```

Today it's not possible to collect object-like claims. And if I go with `kanidm system oauth2 update-claim-map app claim group '{"x":"y"}'` the value is rejected due to OAUTHSCOPE_RE regex applied to the claim value.

First, I wonder if it's worth for rules on claim values to be relaxed?
Second, will it be possible to introduce kinda basic template expansion for claim values so that the following worked:
```
{
  "sub": "MY-SUB-HERE",
  ...
  "mercure": {
    "publish": [ "kanidm" ],
    "subscribe": [ "kanidm/MY-SUB-HERE" ] # NB: target sub is interpolated here
  }
}
```

TIA

dvv:
I temporarily tweaked claim value regexp check and the resulting token reads:
```
"mercure": "{\"publish\":[\"kanidm\"],\"subscribe\":[\"kanidm\"]}",
```
which is indeed not what expected.

May be we could have "json" join type which collects custom claims to object-like token claim?
However, this can be treated as a corner case, so I doubt it's worth it.

Still, I leave the issue for at least discussion purpose.

Firstyear:
Okay, after reviewing the oauth2 rfc, oauth2 token introspection rfc and openid connect specification, none of them apply restrictions to the claim values. This means we are free to expand this. However, I do want to continue to be conservative for the moment in what we allow in claim values, as I could forsee low-quality client applications being vulnerable to issues if we were to allow full character sets in the values.

For now I think we expand this a little, and allow future changes potentially. Small steps.

Firstyear:
@dvv What changes to the claim value regex are you considering?

dvv:
I would like to test the following cases:
1. Some basic interpolation. Here we could define a set of safe values that can be stuffed in the resulting string. I would start with `${sub}` (safe as it's always a non-empty uuid); `${group}` (the group which defines the claim, safe as it's regexp-checked); `${app}`.
```
kanidm system oauth2 update-claim-map a1 can g1 update:${group} # { "can": ["update:g1"] }
[...]

PR #4324, opened 2026-05-09
20260424 templating

Firstyear:
# Change summary

This adds a minimal templating specification that supports customisable rendering and conditionals. 

Importantly it's done with some generics, and it's designed so that we can support customisable lookups, conditions and more. 

For example, we can make a `TemplateClaimAttribute` type that only matches to controlled and specific values, and then use that with `Context` to map those `TemplateClaimAttributes` to our internal entry values and types without giving up control of waht attributes are valid. We can then render them using our own defined strategies. 

Relates #4010 #2641

Checklist


Firstyear:
Fixes up the tests,

CEbbinghaus (commented):
I had a read. The tests seem to be quite surface level for assuring templating works. Too tired to go through all of the code with a fine toothed comb but in recon I can whip up some extra test cases to help silence the small voices screaming Wolf.

libs/templating/src/lib.rs
```diff
@@ -0,0 +1,453 @@
...
+            &self,
+            _attribute: &Self::Attribute,
+            _ctx: &Self::Context,
+            buffer: &mut String,
+        ) -> Result<(), Self::Error> {
+            match self {
+                Self::Direct => buffer.push_str("direct ident"),
+                Self::Json => serde_json::to_string(&["json", "ident"])
+                    .map(|out| buffer.push_str(&out))
+                    .map_err(|_err| ())?,
```
yaleman, line 202: you can avoid the serde_json::to_string and associated error handling here with `serde_json::json!(["json", "ident"]).to_string()`

libs/templating/src/lib.rs
```diff
@@ -0,0 +1,453 @@
...
+        let x: TemplateTest = template::parse("{{ ident | json }}").unwrap();
+
+        tracing::trace!(?x);
+
+        let ctx = ();
+        let y = x.render(&ctx).unwrap();
+
+        tracing::trace!(?y, "rendered");
+
+        assert_eq!(y.as_deref(), Some("[\"json\",\"ident\"]"));
```
yaleman: similar to above you could/should use `serde_json::json!(["json", "ident"]).to_string()`

libs/templating/src/lib.rs
```diff
@@ -0,0 +1,453 @@
...
+
+        let x: TemplateTest = template::parse("{{ ident if memberof(abc) | json }}").unwrap();
+
+        tracing::trace!(?x);
+
+        let y = serde_json::to_string(&x).unwrap();
+
+        tracing::trace!(?y, "serialised");
+    }
+}
```
yaleman, line 455: Given `Template` implements serde::Serialize this'll always pass if it compiles; could be worth checking intended content maybe?

libs/templating/src/lib.rs
Firstyear: It doesn't really save that much because we still need to push to the buffer. End of the day the json! is just three unwraps in a trench coat anyway.

yaleman (approved):

PR #4325, opened 2026-05-10
WIP/POC: Add fluent i18n/l10n

### LINKED ISSUE #2105: i18n in kanidm
Hi, I noticed a lack of i18n in kanidm and @Firstyear pointed me to open an issue here for conversation

My usual toolchain for introducing i18n to projects that haven't had them yet (like for instance https://github.com/Pryaxis/TShock/ or https://github.com/BlueMoonVineyard/BreweryNG) is gettext (or other .po-consuming system) + crowdin (propietary) or weblate (foss) for translation management

I'n not the most familiar with the situation of things in Rust, so i'll be doing some investigation on that; firstyear mentioned no non-rust dependencies

maciek134:
# Change summary

- POC of i18n/l10n with fluent

Fixes #2105

Checklist

Very basic start, mostly to get some feedback on the ergonomics. Askama having no variadic functions/macros makes it a bit hard, but I think the `with_arg` variant looks ok?

I also tried to show some fluent features like attributes and referencing other messages, makes sense to think about if/how those would be used to have some guidelines for translators later.

The various options are described in more detail in comments in `server/core/src/https/views/filters.rs‎`.

### DIFF
--- a/server/core/src/https/extractors/i18n.rs
+++ b/server/core/src/https/extractors/i18n.rs
@@ -1,0 +1,37 @@
+use axum::extract::FromRequestParts;
+use hyper::header::ACCEPT_LANGUAGE;
+use unic_langid::LanguageIdentifier;
+
+pub(crate) struct AcceptLanguage(pub Vec<LanguageIdentifier>);
+
+impl<S> FromRequestParts<S> for AcceptLanguage
+where
+    S: Send + Sync,
+{
+    type Rejection = ();
+
+    async fn from_request_parts(
+        parts: &mut axum::http::request::Parts,
+        _state: &S,
+    ) -> Result<Self, Self::Rejection> {
+        let mut languages: Vec<(LanguageIdentifier, f32)> = parts
+            .headers
+            .get(ACCEPT_LANGUAGE)
+            .map(|values| values.to_str().ok())
+            .flatten()
+            .unwrap_or_default()
+            .split(',')
+            .map(|value| value.split_once(';').unwrap_or((value, "1")))
+            .filter_map(|(language, quality)| language.parse().ok().zip(quality.parse().ok()))
+            .collect();
+
+        languages.sort_by(|a, b| b.1.total_cmp(&a.1));
+
+        Ok(Self(
+            languages
+                .into_iter()
+                .map(|(language, _)| language)
+                .collect(),
+        ))
+    }
+}
--- a/server/core/src/https/i18n.rs
+++ b/server/core/src/https/i18n.rs
@@ -1,0 +1,93 @@
+use std::{any::Any, collections::HashMap, sync::LazyLock};
+
+use askama::Template;
+use axum::{
+    body::Body,
+    http::HeaderValue,
+    response::{IntoResponse as _, Response},
+};
+use bytes::Bytes;
+use fluent::{concurrent::FluentBundle, FluentResource};
+use hyper::{header::CONTENT_TYPE, StatusCode};
+use unic_langid::{langid, LanguageIdentifier};
+
+use crate::https::middleware::i18n::I18nCtx;
+
+/// One of the options, seemed like a good fit with askama, but a build script can do this without an
+/// extra dependency too
+#[derive(rust_embed::Embed)]
+#[folder = "i18n/"]
+struct I18nAssets;
+
+pub(crate) const DEFAULT_LANGUAGE: LanguageIdentifier = langid!("en-AU");
+/// tl;dr: quickly thrown together, needs to be done better in the final implementation
+///
+/// This is a bit messy, LazyLock because this is accessed from async code, but
+/// `fluent::concurrent::FluentBundle` also has an internal Mutex because it uses a cache for
+/// formatters (which is generally not a great idea for a web server imo). Keeping this static, it
+/// could be moved to a thread local to lose both the LazyLock and internal Mutex at the cost of
+/// memory usage (not much though, FluentBundle can take references to FluentResource and resources
+/// are not modified).
+/// Alternatively, this could be put in the server state, but due to how askama works we'd need to
+/// either:
+/// - put the bundle into every template struct; or
+/// - put the bundle into askama::Values instead of the language code
+/// The first is going to make a mess of the code, the second is doable (probably following the the
+/// singletons already in the state) but more effort than a poc should get
+pub(crate) static I18N_BUNDLES: LazyLock<
+    HashMap<LanguageIdentifier, FluentBundle<FluentResource>>,
+> = LazyLock::new(|| {
+    let mut bundles = HashMap::new();
+
+    for path in I18nAssets::iter() {
+        let (lang, _) = path.split_once('/').unwrap();
+        let lang: LanguageIdentifier = lang.parse().unwrap();
+
+        if !bundles.contains_key(&lang) {
+            bundles.insert(
+                lang.clone(),
+                FluentBundle::new_concurrent(vec![lang.clone()]),
+            );
+        }
+
+        let cont = I18nAssets::get(&path).unwrap().data;
+        let cont = String::from_utf8(cont.to_vec()).unwrap();
+        let res = FluentResource::try_new(cont).unwrap();
+        bundles.get_mut(&lang).unwrap().add_resource(res).unwrap();
+    }
+
+    bundles
+});
+
+/// This is intended to replace `#[derive(WebTemplate)]` since that doesn't render with environment
+/// variables. I think it can be done without having every render call `into_response_with_i18n`
+/// manually, but I didn't want to spend too much time on it now.
+pub trait IntoResponseWithI18n {
+    fn into_response_with_i18n(self, ctx: I18nCtx) -> Response;
+}
+
+impl<T> IntoResponseWithI18n for T
+where
+    T: Template,
+{
+    fn into_response_with_i18n(self, ctx: I18nCtx) -> Response {
+        let mut values = HashMap::<&str, Box<dyn Any>>::new();
+        values.insert("lang", Box::new(ctx.lang));
+        let (status, content_type, body) = match self.render_with_values(&values) {
+            Ok(body) => (StatusCode::OK, HTML, Bytes::from_owner(body)),
+            Err(err) => {
+                error!(?err, "Error rendering with i18n");
+                (StatusCode::INTERNAL_SERVER_ERROR, TEXT, FAIL)
+            }
+        };
+
+        let mut resp = Body::from(body).into_response();
+        *resp.status_mut() = status;
+        resp.headers_mut().insert(CONTENT_TYPE, content_type);
+        resp
+    }
+}
+
+const HTML: HeaderValue = HeaderValue::from_static("text/html");
+const TEXT: HeaderValue = HeaderValue::from_static("text/plain");
+const FAIL: Bytes = Bytes::from_static("INTERNAL SERVER ERROR".as_bytes());
--- a/server/core/src/https/views/filters.rs
+++ b/server/core/src/https/views/filters.rs
@@ -78,7 +78,7 @@ macro_rules! fl_args {
 /// (( "translation-key" | with_arg("arg1", arg1_var) | with_arg("arg2", obj.arg2) | trans ))
 /// ```
 #[askama::filter_fn]
-pub fn trans(input: impl MsgId, env: &dyn askama::Values) -> askama::Result<String> {
+pub fn trans_with(input: impl MsgId, env: &dyn askama::Values) -> askama::Result<String> {
     let lang = env.get_value("lang").ok_or(askama::Error::ValueMissing)?;
     let lang: &LanguageIdentifier = lang.downcast_ref().ok_or(askama::Error::ValueType)?;
 
@@ -148,3 +148,20 @@ where
     );
     Ok(MsgIdWithArgs { msg_id, args })
 }
+
+/// Option 3: no macro, no with_arg, no having to reassign variables
+/// Code is generated in build.rs, but it's a huge mess since I just wanted to get something
+/// working. It could also potentially be made to make clippy error out for unused translations.
+///
+/// Since askama has no way of calling `.into` on filter arguments I had to use `&dyn ToString` to
+/// avoid having to call `.into` manually in templates, or wrapping every used value in `Some()`.
+///
+/// One funny thing I could also try is making most of this compile-time - instead of generating the
+/// functions that use a static FluentBundle reference just use the AST to make them output strings.
+/// The formatters would have to stay as statics.
+mod option_3 {
+    use super::*;
+    include!(concat!(env!("OUT_DIR"), "/gen_fluent_i18n.rs"));
+}
+
+pub use option_3::trans;
--- a/server/core/src/https/views/filters.rs
+++ b/server/core/src/https/views/filters.rs
@@ -1,0 +1,150 @@
+use std::collections::HashMap;
+
+use fluent::{FluentArgs, FluentValue};
+use unic_langid::LanguageIdentifier;
+
+use crate::https::i18n::I18N_BUNDLES;
+
+/// Option 1: arguments passed as an optional filter argument
+/// This basically needs the macro below to be usable (afaik askama doesn't
+/// let you run random code in the template to fill out a HashMap, and doesn't have variadic
+/// functions/macros), which means copying the variables from scope:
+///
+/// ```
+/// (% arg1 = arg1_var %)
+/// (% arg2 = obj.arg2 %)
+/// (( "translation-key" | trans_args(fl_args!("arg1" => arg1, "arg2" => arg2)) ))
+/// ```
+#[askama::filter_fn]
+pub fn trans_args(
+    input: &'static str,
+    env: &dyn askama::Values,
+    #[optional(None)] args: Option<HashMap<&'static str, FluentValue<'static>>>,
+) -> askama::Result<String> {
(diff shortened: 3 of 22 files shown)


server/core/templates/login_base.html
```diff
@@ -18,11 +18,13 @@
 	<h3>(( display_ctx.domain_info.display_name() ))</h3>
 	(% if let Some(reauth) = display_ctx.reauth %)
 	<div class="alert alert-info" role="alert">
-		Reauthenticating as (( reauth.username )) to access (( reauth.purpose ))
+		(% set username = reauth.username %)
+		(% set purpose = reauth.purpose %)
+		(( "login.reauth" | trans_args(fl_args!("username" => username, "purpose" => purpose)) ))
```
CEbbinghaus, line 23:
```suggestion
		(( "login.reauth" | trans_args(fl_args!("username" => reauth.username, "purpose" => reauth.purpose)) ))
```

The ergonomics requiring `set x = y` sucks a little, would be nice to be able to do the above

maciek134: Sadly I don't think that's possible, askama runs Rust macros as-is - https://askama.rs/en/stable/template_syntax.html#calling-rust-macros

> One important thing to note is that contrary to the rest of the expressions, Askama cannot know if a token given to a macro is a variable or something else

And askama macros don't have variadic args so can't do that.

One thing I didn't think about earlier would be to generate the filter at build time, adding all the possible parameters as optional arguments, eg. for the strings I touched in this PR it output something like this:

```rs
#[askama::filter_fn]
pub fn trans(
    input: &'static str,
    env: &dyn askama::Values,
    #[optional(None)] username: FluentValue<'static>,
    #[optional(None)] purpose: FluentValue<'static>,
)
```
Then usage would be pretty nice:

```
(( "login.reauth" | trans(username = reauth.username, purpose = reauth.purpose) ))
```

That would also make it natural to have build-time checks for all translation strings. One drawback I can think of immediately is that it would most likely crash rust-analyzer at some point, and/or editor autocomplete in some editors - but this is also easy to verify.

If it generated a helper function for every translation string we'd also get a nice clippy error if any string was left unused (that would mean any strings used purely as references would have to be [terms](https://projectfluent.org/fluent/guide/terms.html) instead, which is an added benefit imo).

maciek134: [`8068112` (this PR)](https://github.com/kanidm/kanidm/pull/4325/commits/80681120d3fde168aea612fa7c7cf6b6ed701a15) has the poc implementation of what I described above. Example output:

<details>
<summary>gen_fluent_i18n.rs</summary>

```rs
fn i18n_message_login_username(bundle: &::fluent::concurrent::FluentBundle<::fluent::FluentResource>, ) -> ::askama::Result<String> {
    let pattern = bundle.get_message("login-username").ok_or(::askama::Error::ValueMissing)?.value().ok_or(::askama::Error::ValueMissing)?;
        
   let args = None;

    let mut _errors = Vec::new();
    let value = bundle.format_pattern(&pattern, args, &mut _errors);
    Ok(value.into())
}
fn i18n_message_login(bundle: &::fluent::concurrent::FluentBundle<::fluent::FluentResource>, ) -> ::askama::Result<String> {
    let pattern = bundle.get_message("login").ok_or(::askama::Error::ValueMissing)?.value().ok_or(::askama::Error::ValueMissing)?;
        
   let args = None;

    let mut _errors = Vec::new();
    let value = bundle.format_pattern(&pattern, args, &mut _errors);
    Ok(value.into())
}
fn i18n_message_login_to_access(bundle: &::fluent::concurrent::FluentBundle<::fluent::FluentResource>, purpose: FluentValue<'static>) -> ::askama::Result<String> {
    let message = bundle.get_message("login").ok_or(::askama::Error::ValueMissing)?;
    let pattern = message.get_attribute("to-access").ok_or(::askama::Error::ValueMissing)?.value();
        

    let mut args = HashMap::new();
    args.insert("purpose", purpose);
    let args = Some(&::fluent::FluentArgs::from_iter(args));
        

    let mut _errors = Vec::new();
    let value = bundle.format_pattern(&pattern, args, &mut _errors);
    Ok(value.into())
}
fn i18n_message_login_reauth(bundle: &::fluent::concurrent::FluentBundle<::fluent::FluentResource>, username: FluentValue<'static>, purpose: FluentValue<'static>) -> ::askama::Result<String> {
    let message = bundle.get_message("login").ok_or(::askama::Error::ValueMissing)?;
    let pattern = message.get_attribute("reauth").ok_or(::askama::Error::ValueMissing)?.value();
        

    let mut args = HashMap::new();
    args.insert("username", username);
    args.insert("purpose", purpose);
    let args = Some(&::fluent::FluentArgs::from_iter(args));
        

    let mut _errors = Vec::new();
    let value = bundle.format_pattern(&pattern, args, &mut _errors);
    Ok(value.into())
}
fn i18n_message_login_authenticate(bundle: &::fluent::concurrent::FluentBundle<::fluent::FluentResource>, purpose: FluentValue<'static>) -> ::askama::Result<String> {
    let message = bundle.get_message("login").ok_or(::askama::Error::ValueMissing)?;
    let pattern = message.get_attribute("authenticate").ok_or(::askama::Error::ValueMissing)?.value();
        

    let mut args = HashMap::new();
    args.insert("purpose", purpose);
    let args = Some(&::fluent::FluentArgs::from_iter(args));
        

    let mut _errors = Vec::new();
    let value = bundle.format_pattern(&pattern, args, &mut _errors);
    Ok(value.into())
}
fn i18n_message_login_remember_me(bundle: &::fluent::concurrent::FluentBundle<::fluent::FluentResource>, ) -> ::askama::Result<String> {
    let message = bundle.get_message("login").ok_or(::askama::Error::ValueMissing)?;
    let pattern = message.get_attribute("remember-me").ok_or(::askama::Error::ValueMissing)?.value();
        
   let args = None;

    let mut _errors = Vec::new();
    let value = bundle.format_pattern(&pattern, args, &mut _errors);
    Ok(value.into())
}

#[::askama::filter_fn]
pub fn trans(
    input: &'static str,
    env: &dyn ::askama::Values,
    #[optional(&0)] purpose: &dyn ToString,
    #[optional(&0)] username: &dyn ToString
) -> ::askama::Result<String> {
    let lang = env.get_value("lang").ok_or(::askama::Error::ValueMissing)?;
    let lang: &::unic_langid::LanguageIdentifier = lang.downcast_ref().ok_or(::askama::Error::ValueType)?;

    let bundle = I18N_BUNDLES.get(lang).ok_or(askama::Error::ValueMissing)?;

    match input {
        "login" => i18n_message_login(bundle, ),
        "login.reauth" => i18n_message_login_reauth(bundle, ToString::to_string(username).try_into().unwrap(), ToString::to_string(purpose).try_into().unwrap()),
        "login-username" => i18n_message_login_username(bundle, ),
        "login.authenticate" => i18n_message_login_authenticate(bundle, ToString::to_string(purpose).try_into().unwrap()),
        "login.to-access" => i18n_message_login_to_access(bundle, ToString::to_string(purpose).try_into().unwrap()),
        "login.remember-me" => i18n_message_login_remember_me(bundle, ),
        _ => unimplemented!()
    }
}
```

</details>

maciek134: Oooh, I should have just looked at what the `askama::filter_fn` macro does, this can be done a lot better if I don't use it:
- arguments would only work for the message they belong to
- should be able to avoid the unnecessary `ToString`
- better errors
- no unnecessary unsafe, not sure why they do that:
```rs
let username = unsafe { self.username };
let purpose = unsafe { self.purpose };
```

The `rust-analyzer` crash prediction above also won't happen, because the filter is not actually a function.

server/core/src/https/i18n.rs
```diff
@@ -0,0 +1,93 @@
...
+        let cont = String::from_utf8(cont.to_vec()).unwrap();
+        let res = FluentResource::try_new(cont).unwrap();
+        bundles.get_mut(&lang).unwrap().add_resource(res).unwrap();
+    }
+
+    bundles
+});
+
+/// This is intended to replace `#[derive(WebTemplate)]` since that doesn't render with environment
+/// variables. I think it can be done without having every render call `into_response_with_i18n`
```
yaleman, line 63: What environment variables is it using in the rendering? There should be no environment variables used at all.  Do you perhaps mean the axum extensions?

server/core/templates/login_base.html
yaleman: When we get into building huge swaths of code at build-time with both Rust and Askama macros like that, it smells like we're trying to be WAY too "tricky" and need to look at better approaches. 

You could pass the translated `Reauthenticating as (( reauth.username )) to access (( reauth.purpose ))` as a string from the Rust code and avoid all the macro-ninja tricks. There's FAR too much Rust code in the Askama templates already.

ie - replace the `display_ctx.reauth: Option<struct>` with a `display_ctx.reauth_msg: Option<String>` and just use it.

server/core/src/https/i18n.rs
maciek134: Uh, not sure why I wrote environment there, I meant askama's runtime variables.

server/core/templates/login_base.html
maciek134: I thought the whole point of using askama was to generate things at build-time? If it's not then there are more ergonomic templating solutions out there (and in that case I definitely went into overly "clever" solutions here).

Option 2 has no Rust code in the templates and no generated code if that's the goal. Option 3 generates some code, but imo has the best DX and allows for much better compile-time checks.

I wanted to avoid adding/changing data in every template struct. I think it's a bit unorthodox to separate the translations from the templating system, but I guess with some care to keep it consistent it could be nice too:

```rs
pub struct LoginDisplayCtx {
    pub domain_info: DomainInfoRead,
    pub reauth_msg: Option<String>,
    pub oauth2_msg: Option<String>,
    pub error_msg: Option<String>,
}
```

```rs
let display_ctx = LoginDisplayCtx {
    reauth_msg: fl!(lang, "login.reauth", {
        "username" => uat.spn.clone(),
        "purpose" => ReauthPurpose::ProfileSettings
    }),
    // ...
};
```

```
(% if let Some(reauth_msg) = display_ctx.reauth %)
<div class="alert alert-info" role="alert">
    (( reauth_msg ))
</div>
(% else if ... %)
```

No dynamic dispatch either and no need for `.into_response_with_i18n()` :confetti_ball: and if a proc macro (or build.rs-generated macro_rules) is ok then it also allows for full compile-time checks - to ensure the used translation strings exist (and are all used), arguments are correct, etc.

And at least in this case the messages could be merged into something like `alert_msg`, or even combined with `error` too and make the alert class depend on a flag. Or, at the risk of trying to be too clever again, go one step further and have an alert struct:

```rs
pub struct Alert {
    pub is_error: bool,
    pub message: String,
}
```

```rs
pub struct LoginDisplayCtx {
    pub domain_info: DomainInfoRead,
    pub alert: Option<Alert>,
}
```

```rs
let display_ctx = LoginDisplayCtx {
    domain_info,
    alert: Some(Alert {
        is_error: false,
        message: fl!(lang, "login.reauth", {
            "username" => uat.spn.clone(),
            "purpose" => ReauthPurpose::ProfileSettings
        })
    }),
};
```

```
(% if let Some(alert) = display_ctx.alert %)
<div class="alert alert-(% if alert.is_error %)error(% else %)info(% endif %)" role="alert">
    (( alert.message ))
</div>
(% endif %)
```

server/core/src/https/extractors/i18n.rs
```diff
@@ -0,0 +1,37 @@
...
+
+pub(crate) struct AcceptLanguage(pub Vec<LanguageIdentifier>);
+
+impl<S> FromRequestParts<S> for AcceptLanguage
+where
+    S: Send + Sync,
+{
+    type Rejection = ();
+
+    async fn from_request_parts(
```
Firstyear, line 13: Is there a case where we want the users language to be an account preference rather than a browser one? I guess probably not ....

server/core/src/https/views/filters.rs
```diff
@@ -0,0 +1,167 @@
...
+        let mut args = ::std::collections::HashMap::<&str, ::fluent::FluentValue>::new();
+        fl_args!(args, $key => $val, $($rest)*);
+        Some(args)
+    }};
+}
+
+/// Option 2: no macro, arguments composed with another filter
+///
+/// ```
+/// (( "translation-key" | with_arg("arg1", arg1_var) | with_arg("arg2", obj.arg2) | trans ))
```
Firstyear, line 78: This syntax does feel nicer tbh.

Firstyear: Another option, could we do something like 

```
(( translation-key-pre ))(( var ))(( translationkey.post ))
```

So that in cases where a value is populated between two things we have a pre-post, and then in other cases we can just have empty pre/post instead? This way we leave formatting to askama and the transaltions become more "static"? 

Does that hurt localisation?

Firstyear (changes requested):
Hey mate, thanks so much for this PR. This is a pretty big one to tackle, so thanks for reviewing all our comments. But it's very exciting to see this happening!

server/core/src/https/extractors/i18n.rs
CEbbinghaus: Personally I believe language should be an optional account setting which overrides system default.

server/core/src/https/views/filters.rs
yaleman: There's a lot of situations where nouns etc are in different positions in a sentence, and we've already got one example where we need to handle multiple variables, which would break this too. (it's pretty tough ergonomically as well)

server/core/src/https/extractors/i18n.rs
maciek134: Yeah, the header can be wrong when travelling for example. Also people sometimes want specific services in a different language, especially in countries that use multiple languages. 
For now this can be overwritten with a query param (should get a picker in the UI), should be easy to make it a user attribute too.

server/core/src/https/views/filters.rs
maciek134: Breaks with Slavic and Asian languages for sure since they have a different sentence order than English

maciek134:
A more direct comparison of the currently discussed options. The order is arbitrary / chronological, not preference.

<table>
<tr>
 <td>1</td>
 <td>

```django
(% set username = reauth.username %)
(% set purpose = reauth.purpose %)
(( "login.reauth" | trans(fl_args!(
    "username" => username,
    "purpose" => purpose
)) ))
```
</td>
</tr>
<tr>
 <td>2</td>
 <td>

```django
(( "login.reauth"
   | with_arg("username", reauth.username)
   | with_arg("purpose", reauth.purpose)
   | trans ))
```
</td>
</tr>
<tr>
 <td>3</td>
 <td>

```django
(( "login.reauth" | trans(
    username = reauth.username,
    purpose = reauth.purpose
) ))
```
</td>
</tr>
<tr>
 <td>4</td>
 <td>

```rs
let display_ctx = LoginDisplayCtx {
    alert_msg: fl!(lang, "login.reauth", {
        "username" => uat.spn.clone(),
        "purpose" => ReauthPurpose::ProfileSettings
    }),
    // ...
};
```
</td>
</tr>
</table>

1. is the naive attempt and doesn't really fit with askama imo, the variable reassignment is going to be messy. 
2. is a bit verbose, but the implementation is very simple and has no macros or code generation (which tbf is a con if you want compile-time checks for the translations).
3. is cleaner on usage and supports build-time checks, but needs the code to be generated which complicates the implementation and makes it less obvious how things work.
4. is trivial to implement, requires least changes in the way templates are rendered and doesn't need dynamic dispatch - at the cost of moving all translations outside the templates which makes it harder to distinguish between data and content. At the same time it would be the same way any Rust strings would have to be translated (API responses or logs for example).

General info that applies to any way this is implemented / fluent specifics:
- variable positions can't be static, for example `Reauthenticating as { $username } to access { $purpose }` → `{ $purpose }にアクセスするに{ $username }として再認証` (hopefully I didn't butcher it too much :sweat_smile:)
- enums should be handled inside translations - this helps avoid easy to make assumptions about the variants being translated the same in every context - fluent still lets you define this in one term and reference it everywhere for DRY, but makes it possible for translators to diverge as needed
```rs
match self {
    Self::InvalidUsername => write!(f, "invalid_username"),
    Self::SessionExpired => write!(f, "session_expired")
}
```
```fluent
login-error = { $error ->
    [invalid_username] Invalid username
    [session_expired] Session expired
   *[others] Unknown error
}
```
- fluent handles more than just translations when the built-in functions are enabled - it can localize dates and numbers too

Firstyear:
Syntax wise 2 or 3 look the most intuitive to use. Do you think the compile time checks have a strong value in the implementation that justifies the extra complexity?

maciek134:
I think yes, mostly because it makes maintenance easier - one less thing to worry about during review for example. But it can be done multiple ways and now I think I can do it for all options, so it doesn't have to impact the syntax choice.

Firstyear:
Okay, if it makes maintenance easier then I'm okay with a bit more upfront cost/work to achieve that.

maciek134:
Then I guess it's a matter of preference as to the syntax at this point, I don't think there is much technical difference between 2 and 3 (both need a build script or proc macro to implement the checks).

Any preference when it comes to build script vs proc macro? I'd generally go with a script since the output is easier to inspect.

CEbbinghaus:
Im leaning towards 3 to cut down on boilerplate "with_arg"

maciek134:
Unless in the meantime one of the options looks like the one to go with, I'll prepare two separate branches for options 2 and 3. Still with a single file but with all the compile-time setup for easier comparison regarding complexity / usage.

Issue #4329, opened 2026-05-11
1.10.x breaks oidc client login due to client_id in jwt header


TomGrozev:
### I did this
When trying to login using OIDC through some client, some clients don't allow custom jwt headers (i.e. joserfc).

### I expected the following
Be able to login and the JWT be accepted by the client.

### This happened instead
The JWT is rejected due to the custom header that is unrecognized.

### Kanidm version details
1.10.x

- Output of `kanidm(d) version`: kanidm 1.10.1
- Are you running it in a container? If so, which image/tag?: yes - docker.io/kanidm/server:1.10.1
- If not a container, how'd you install it: N/A
- Operating System / Version (On Unix please post the output of `uname -a`): N/A

### Any other comments
Clients such as joserfc don't allow custom headers as a security feature and client_id would need to be explicitly enabled on every single app that uses it. Is there an option to disable this in kanidm at all?

Example of issue this causes: https://github.com/christiaangoossens/hass-oidc-auth/issues/307

jinnatar:
Having run into this myself, as far as my reading of the various RFCs around this goes, there's nothing in them that defines "only allow these". There is a "Don't allow these" list (does not contain client_id) and then definitions for the common ones. Also as mentioned on the hass-oidc side, during discovery critical headers are the ones that should be scrutinized, and this isn't marked as such. I have a feeling joserfc does not implement that part of discovery, but I haven't looked into it yet.

So ultimately in my opinion the best path to take here is to open an issue on joserfc for them to just not do this kind of overly strict non-RFC required or even recommended checking, at least not by default. But, probably better to re-read all the relevant bits first and confirm we're all on the same page.

Edit: It would also be good to understand *why* Kanidm is putting the client_id in the header as it's not quite clear to me from #4230. I didn't find any hints of that "being a thing" in any other implementation while spelunking a while back, but perhaps it was just somehow the easiest way to improve introspection? Anyway, it's "weird" and it's also bad that other libraries are penalizing for "weird".

lepture:
when using `joserfc`, you can add additional headers via: https://jose.authlib.org/en/guide/registry/#additional-headers

Firstyear:
@lepture This comment doesn't help and isn't constructive - we know that the library can be directed to accept additional headers, but that's not possible when it's buried as a dependency of another project.

Additionally we have already issued 1.10.2 which reverts the header change to make Kanidm compatible with joserfc.

PR #4332, opened 2026-05-11
fix(oauth2): prevent re-authentication loop in prompt=login flows after successful auth

grisia:
I hit the same re-authentication loop from #4329 with Stalwart as the client.
The important constraint for this patch was to avoid mutating request parameters or cookies. I did not want to remove `prompt=login` from the original request or rewrite the stored OAuth2 request cookie. Instead, the fix makes `/ui/oauth2/resume` continue via internal server-side resume context, so the original request remains unchanged while the flow can complete successfully after login.

### DIFF
--- a/server/core/src/actors/v1_read.rs
+++ b/server/core/src/actors/v1_read.rs
@@ -1353,6 +1353,37 @@ impl QueryServerReadV1 {
         auth_req: AuthorisationRequest,
         eventid: Uuid,
     ) -> Result<AuthoriseResponse, Oauth2Error> {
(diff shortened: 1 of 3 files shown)


Firstyear:
Hey there,

thanks for the PR. This issue was brought up earlier in the features development and there is actually a different way we want to "break the cycle" in this case. So I'm going to close this PR in favour of the alternate path, but thank you for taking the time to investigate and submit it regardless (even if we didn't accept it this time)

By the way, in future we submit fixes to master, and then we backport them separately to the maintenance branch (rather than submitting directly to the maint branch).

PR #4334, opened 2026-05-12
Revert ClientID header

### LINKED ISSUE #4329: 1.10.x breaks oidc client login due to client_id in jwt header

Firstyear:
Some external libraries do not follow the JWS RFC and are rejecting JWS with non-critical header extensions.

This reverts the client_id header change from the protected header and instead uses a key-id to client_id map internally to perform the token linking.

Fixes #4329

Checklist

### DIFF
--- a/server/lib/src/idm/oauth2.rs
+++ b/server/lib/src/idm/oauth2.rs
@@ -9,7 +9,9 @@ use crate::idm::server::{
     IdmServerProxyReadTransaction, IdmServerProxyWriteTransaction, IdmServerTransaction, Token,
 };
 use crate::prelude::*;
-use crate::server::keys::{KeyObject, KeyProvidersTransaction, KeyProvidersWriteTransaction};
+use crate::server::keys::{
+    KeyId, KeyObject, KeyProvidersTransaction, KeyProvidersWriteTransaction,
+};
 use crate::utils;
 use crate::value::{Oauth2Session, OauthClaimMapJoin, SessionState, OAUTHSCOPE_RE};
 use base64::{engine::general_purpose, Engine as _};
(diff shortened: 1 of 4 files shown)


server/lib/src/idm/oauth2.rs
```diff
@@ -1134,7 +1172,16 @@ impl IdmServerProxyWriteTransaction<'_> {
                 warn!("OAuth2 Client Authentcation Required");
             })?;
 
-        let o2rs = self.get_client(&client_auth.client_id)?;
+        let o2rs = self
+            .oauth2rs
+            .inner
+            .rs_set_get(&client_auth.client_id)
+            .ok_or_else(|| {
+                warn!("Invalid OAuth2 client_id {}", &client_auth.client_id);
```
yaleman: This should be a debug rather than a warn
```suggestion
                debug!("Invalid OAuth2 client_id {}", &client_auth.client_id);
```

server/lib/src/idm/oauth2.rs
```diff
@@ -2131,17 +2160,17 @@ impl IdmServerProxyWriteTransaction<'_> {
...
         })?;
 
         let o2rs = self
             .oauth2rs
             .inner
-            .rs_set_get(unverified_client_id)
+            .rs_from_kid(unverified_kid)
             .ok_or_else(|| {
-                warn!("Invalid OAuth2 client_id");
+                warn!("Invalid OAuth2 key id");
```
yaleman: Another should-be-a-debug TBH

```suggestion
                debug!("Invalid OAuth2 key id");
```

yaleman (dismissed):
Minor niggles around log levels, but otherwise weeeee

yaleman (approved):

Issue #2824, opened 2024-06-06
Support multiple signed-in accounts


Kladki:
### Is your feature request related to a problem? Please describe

Sometimes A user might want to be able to be signed into multiple different accounts, such as for different access levels, [plural users](https://morethanone.info/), separate work and person accounts, etc.

### Describe the solution you'd like

The ability to be signed into multiple accounts in a single browser, while also having a prompt to select account and/or login to a new account when authorizing apps. 

### Describe alternatives you've considered

Using separate browsers / container tabs, but that can become tedious, especially on mobile.

Firstyear:
There are a few use cases here but we aren't sure they fully apply to Kanidm.

Something to keep in mind is that we do want privilege separation to exist, so if a user can swap between account A and account B without re-authentication in between, then there is no separation, you may as well have assigned all the privileges to the single account. So to change between accounts, if we were to implement this, then we would enforce authentication for each switch.

With regard to access levels, I think that this could be done without whole separate accounts. You could have roles that you acquire and release in a session while retaining a single account presence. This way each role change could be accompanied by authentication challenges, else the session may as well just have full access all the time, which as above we don't want.

I can see why work vs personal matters in the context of say a google account or a github account, but in the places where Kanidm is currently used (businesses or homelabs etc), there isn't an overlap where a person has a separate work and personal instance, so I think this is less compelling for us. 

Currently I think it's unlikely we'll add fast account switching for the above reasons, we would rather have the account able to acquire and release privileges/roles as required. We would need to put in a fair bit of thought to make that happen.

cosmoswafer:
How about using two different browsers, which is what I’m currently doing.

Firstyear:
Closing today because we think this use case isn't supportable. We are happy to reconsider if there is a compelling use case for the feature.

PR #4336, opened 2026-05-12
Add max_age / improve prompt=login - now without grace periods! 

### LINKED ISSUE #4321: Issues with prompt / login_hint
### I did this

I am trying to connect stalwart mail server webui to kanidm. 
Stalwart unfortunately always adds prompt=login&login_hint=user@example.com to the url, where user@example.com is the mail address and not my kanidm user.

### I expected the following

- change username to user
- login with passkey
- redirect to stalwart

### This happened instead

- change username to user
- login with passkey
- redirect to login page (Username prompt) on URL /ui/oauth2/resume
- change username to user
- login with passkey
- redirect to login page (Username prompt) on URL /ui/oauth2/resume
- ...

### Kanidm version details

- Output of `kanidm(d) version`: kanidm 1.10.1
- Are you running it in a container? If so, which image/tag?: yes 1.10.1

### Any other comments

As I understand it, login_hint should be a suggestion, not a restriction. But even if it was, there should be an error message.

After removing login_hint/prompt from the url by patching the stalwart webui, login works.

Firstyear:
This improves prompt=login, importantly cycle breaking when the login is performed from the session resumption.

Fixes #4321

Checklist

### DIFF
--- a/server/core/src/https/views/oauth2.rs
+++ b/server/core/src/https/views/oauth2.rs
@@ -1,20 +1,16 @@
+use super::login::{LoginDisplayCtx, Oauth2Ctx};
+use super::{cookies, UnrecoverableErrorView};
+use crate::https::views::{
+    errors::HtmxError,
+    login::{Reauth, ReauthPurpose},
+};
 use crate::https::{
     extractors::{DomainInfo, DomainInfoRead, VerifiedClientInformation},
     middleware::KOpId,
     ServerState,
 };
-use kanidmd_lib::idm::oauth2::{AuthorisationRequest, AuthoriseResponse, Oauth2Error};
-use kanidmd_lib::prelude::*;
-
-use kanidm_proto::internal::COOKIE_OAUTH2_REQ;
-
-use std::collections::BTreeSet;
-
 use askama::Template;
 use askama_web::WebTemplate;
-
-#[cfg(feature = "dev-oauth2-device-flow")]
-use axum::http::StatusCode;
 use axum::{
     extract::{Query, State},
     http::header::ACCESS_CONTROL_ALLOW_ORIGIN,
@@ -23,10 +19,17 @@ use axum::{
 };
 use axum_extra::extract::cookie::{CookieJar, SameSite};
 use axum_htmx::HX_REDIRECT;
+use kanidm_proto::internal::UserAuthToken;
+use kanidm_proto::internal::COOKIE_OAUTH2_REQ;
+use kanidmd_lib::idm::oauth2::{
+    AuthorisationRequest, AuthorisationRequestContext, AuthoriseResponse, Oauth2Error,
+};
+use kanidmd_lib::prelude::*;
 use serde::Deserialize;
+use std::collections::BTreeSet;
 
-use super::login::{LoginDisplayCtx, Oauth2Ctx};
-use super::{cookies, UnrecoverableErrorView};
+#[cfg(feature = "dev-oauth2-device-flow")]
+use axum::http::StatusCode;
 
 #[derive(Template, WebTemplate)]
 #[template(path = "oauth2_consent_request.html")]
@@ -44,6 +47,14 @@ struct AccessDeniedView {
     operation_id: Uuid,
 }
 
+#[derive(Default)]
+enum AuthReqState {
+    #[default]
+    None,
+    Initial(AuthorisationRequest),
+    Resumed(AuthorisationRequest),
+}
+
 pub async fn view_index_get(
     State(state): State<ServerState>,
     Extension(kopid): Extension<KOpId>,
@@ -58,7 +69,7 @@ pub async fn view_index_get(
         client_auth_info,
         domain_info,
         jar,
-        Some(auth_req),
+        AuthReqState::Initial(auth_req),
     )
     .await
 }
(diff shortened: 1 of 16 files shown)


grisia:
Please excuse me, but I think I would like to briefly raise a concern here.

I am not sure it should be Kanidm's responsibility to decide whether an authentication is "fresh enough". It seems to me that only the client should decide whether it wants to require a fresh authentication.

The current implementation might introduce a security concern, because within that window no re-authentication is required. It also seems possible that a user may intentionally want to sign in as a different account. In that case, would they have to wait 300 seconds before being able to proceed?

The closed PR:https://github.com/kanidm/kanidm/pull/4332 I proposed, although it may not be the most elegant solution, takes a different approach: it keeps track of whether authentication has just been completed successfully in the current flow. Only in that case is the flow allowed to continue.

Firstyear:
> Please excuse me, but I think I would like to briefly raise a concern here.

No need to feel sorry, when we make decisions we need to be able to explain and justify them. That's part of a good project.

> I am not sure it should be Kanidm's responsibility to decide whether an authentication is "fresh enough". It seems to me that only the client should decide whether it wants to require a fresh authentication.

I think that it is very much within Kanidm's responsibility as we are the identity provider, charged with making security decisions that clients (oidc relying parties) rely upon. 

Now more importantly in this case prompt=login is not a security control, it is a UI guidance control. This can be easily demonstrated - if prompt=login is removed from the GET query, then the RP has no signal or knowledge that it was removed. 

The strict version of this is the "max_age" parameter which *does* enforce a time window for freshness of the session - exactly the same as what we are doing in our code here. There is a max_age=0 but that has it's own issues. Additionally, if max_age is set, then the IDP must add the auth_time claim - today we don't actually support this flow however, but we should. 

>
> The current implementation might introduce a security concern, because within that window no re-authentication is required. It also seems possible that a user may intentionally want to sign in as a different account. In that case, would they have to wait 300 seconds before being able to proceed?

The current implementation of prompt=login is using our existing auth code, not the re-auth code. That's why it falsely presents you with the ability to change accounts. It shouldn't do that. 

I think looking at it, we may have released prompt=login support too soon as it seems to be causing more issues for users, as well as confusion, and logically it has some mistakes that interact with things like max_age. 

I'm going to propose it is disabled in 1.10 so that we can improve it in 1.11.

Firstyear:
@grisia We have accepted the removal of prompt=login for 1.10 in the meantime. The update to this PR takes some inspiration from your PR, but goes further to improve the reauth session handling and validation.

Firstyear:
@grisia You are also welcome to review, I included ideas from your PR into this.

grisia:
This looks good to me overall and matches what I would expect from `prompt=login`.

The one edge case I hit in a live test was being already signed into Kanidm as one user, then opening the Stalwart admin UI as another. With a mismatched `login_hint`, I think a fresh login flow is more correct than reauthenticating the existing session.

This also feels adjacent to the multi-session/account-switching discussion in #2824.

I opened a small follow-up for that case here:
https://github.com/Firstyear/kanidm/pull/1

Firstyear:
> The one edge case I hit in a live test was being already signed into Kanidm as one user, then opening the Stalwart admin UI as another. With a mismatched `login_hint`, I think a fresh login flow is more correct than reauthenticating the existing session.

We don't support login's with email addresses today, so we have no way to know if the login_hint mismatches to the account. I think that will have to be a follow up issue as there are other changes to the login flow that would need to occur to make that work correctly.

grisia:
Just to clarify: this is not about email-based login. I am aware Kanidm does not support that today, so in Stalwart I already only use the plain login name as `login_hint`.

The behavior that still feels odd to me is when I already have a Kanidm session as Alice, but a client sends `prompt=login` with `login_hint=bob`, and Kanidm then reauthenticates Alice.

For matching hints, reauth seems correct.
For a different hinted login name, I would expect Kanidm to fall back to a fresh authentication flow instead.

In the follow-up I tried to correct that behavior, and in my local setup it behaves as expected with that change.

server/core/src/https/views/oauth2.rs
```diff
@@ -90,36 +103,46 @@ async fn oauth2_auth_req(
     client_auth_info: ClientAuthInfo,
     domain_info: DomainInfoRead,
     jar: CookieJar,
-    maybe_auth_req: Option<AuthorisationRequest>,
+    // maybe_auth_req: Option<AuthorisationRequest>,
```
yaleman:
```suggestion
```

server/lib/src/constants/mod.rs
```diff
@@ -191,6 +191,9 @@ pub const OAUTH_REFRESH_TOKEN_EXPIRY: u32 = 3600 * 16;
 /// of the refresh token, which is bound to the issuing session.
 pub const OAUTH2_ACCESS_TOKEN_EXPIRY: u32 = 15 * 60;
 
+/// The absolute maximum that can be requested for max-age requests in OIDC.
+pub const OAUTH2_OIDC_MAX_AGE_CLAMP: i64 = 86400;
```
yaleman, line 195: why don't we just alias this to `DEFAULT_AUTH_SESSION_EXPIRY` so that we don't end up with a mismatch later?

yaleman (changes requested):
Minor questions, yay for dumping the grace period which seemed like a bad pattern.

server/lib/src/constants/mod.rs
Firstyear: Because the auth session length could be extended by account policy. I think the max_age here should be detached from that though, really this clamp is just here to ensure that max_age isn't set to a ridiculous value.

yaleman (approved):

PR #4338, opened 2026-05-12
Improve handling of alerts in cred update dialog

### LINKED ISSUE #4320: Just entering ui/update_credentials already enables the save button & nav warnings
### I did this
1. Added a passkey, saved. This dropped me to the profile details page, which confused me a bit.. "Did it save?"
2. Navigated back to ui/update_credentials to see if things actually did save
3. Everything seemed all right, so I tried to navigate back to Profile

### I expected the following
Easy nav back to profile, because I hadn't changed anything.

### This happened instead
Browser alert: "This page is asking you to confirm that you want to leave — information you’ve entered may not be saved."

.. That in turn re-triggered the confusion, "wait, it didn't save after all? Let me hit save again because it's green & enabled" .. But in fact this is an infinite loop, it's always in a state of "can save, danger to navigate away".

I feel this will feed into the confusion around temporal changes that haven't been saved yet, because now you can't really confirm anything was saved without a refresh and clicking through the scary warning.

### Kanidm version details

- Output of `kanidm(d) version`: Both 0.10.0 & 0.10.1 have this
- Are you running it in a container? If so, which image/tag?: standard image

### Any other comments
I don't think it's always been like this, might be a regresssion in 0.10.0?

Firstyear:
This improves the handling of the alerts so that only if the credential update session is dirty, are the handlers used to notify the user of potential unsaved changes.

Fixes #4320

Checklist


ToxicMushroom (approved):
Seems good from my tests :)

PR #4348, opened 2026-05-15
feat: add live server password feedback

ToxicMushroom:
# Change summary

- Adds live password feedback after a 500ms of keyboard silence.
- Added for the password and unix password forms.
 
Supersedes #3650 

Checklist

### DIFF
--- a/server/core/src/https/views/reset.rs
+++ b/server/core/src/https/views/reset.rs
@@ -152,6 +152,11 @@ pub(crate) struct NewPassword {
 }
 
 #[derive(Deserialize, Debug)]
+pub(crate) struct PasswordStrengthCheck {
+    new_password: String,
+}
+
+#[derive(Deserialize, Debug)]
 pub(crate) struct NewPublicKey {
     title: String,
     key: String,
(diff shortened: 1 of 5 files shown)


server/core/src/https/views/reset.rs
```diff
@@ -756,6 +761,76 @@ pub(crate) async fn view_new_pwd(
...
+        Err(operr) => {
+            return Err(ErrorResponse::from(HtmxError::new(
+                &kopid,
+                operr,
+                domain_info,
+            )))
+        }
+    };
+
+    let check_res = PwdCheckResult::Failure {
```
Firstyear: Why not change PwdCheckResult to have another enum variant that only encodes warnings? 

Or alternately, given that this could check that the passwords are or aren't equal, wouldn't this remove the need for the js hooks? 

Also you can probably have some template logic that if the second input is empty, then don't display the pwd not equal warning?

Firstyear (commented):
Mostly looks really good and simple! Some suggestions but I don't think they'll be too hard to action :)

server/core/src/https/views/reset.rs
ToxicMushroom: Yes, certainly possible for this to replace the JS pwd equality check.
Main drawbacks are that it'll feel slow due to the 500ms delay, and the outline of the inputs wont be red anymore as I can't swap the <input> elems.

server/core/templates/credential_update_set_unixcred_partial.html
```diff
@@ -1,53 +1,68 @@
...
+            (% if !pwd_equal -%)
+                (% set potentially_invalid_reinput_class = "is-invalid" -%)
+            (% else -%)
+                (% set potentially_invalid_reinput_class = "" -%)
+            (% endif -%)
+        (% else -%)
+            (% set potentially_invalid_input_class = "" -%)
+            (% set potentially_invalid_reinput_class = "" -%)
+            (% set pwd_equal = true -%)
+        (% endif -%)
```
yaleman: The original had been setting the defaults and only changing when `PwdCheckResult::Failure` needed to, and now there's a decl / set pair? It seems like a weird change.

ToxicMushroom: The old code just didn't work anymore (the variables stayed on their defaults), probably after an askama update ?

ToxicMushroom: But indeed I'm not a fan of the new code here

yaleman: All of this could be done on the Rust side too instead of in the template 😄

Firstyear: Yeah, anything we can do in the template is good.

ToxicMushroom:
@Firstyear Can I ignore the rust 1.97 clippy lints?

ToxicMushroom:
I simplified the templates and cleaned up the enum.

Firstyear (dismissed):
Looks great, thank you!

Firstyear (approved):

PR #4349, opened 2026-05-16
20260516 oauth2 tester

Firstyear:
# Change summary

Adds an OAuth2 tester with some configurable form parameters. I plan to merge in an OAuth2 version, and move the device flow code into the this tester too. 

It looks like the oauth2/openidconnect crate are a bit problematic, they have ancient versions of a bunch of deps. We were using them in one place for a test case to do pkce - something that we have already.

So I have moved this tester out of the workspace to avoid polluting our workspace with the deps such as axum-server, tower, oauth2, etc. 

I wonder if it would make more sense for this to go as a separate repo rather than here? 

Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@ altnames.cnf
 .backup*_test.json
 .backup_test.db
 **/target/**
+tools/oauth2-tester/Cargo.lock
 /insecure
 **/*.rs.bk
 test.db
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1328,7 +1328,7 @@ dependencies = [
  "kanidm_utils_users",
  "kanidmd_core",
  "prctl",
- "reqwest 0.13.4",
+ "reqwest",
  "sd-notify",
  "serde",
  "serde_json",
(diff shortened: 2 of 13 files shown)


Firstyear:
(I'm leaning to this being a separate repo potentially?)

yaleman:
Played around a bit - compare my PR-on-your-PR here for diffs

https://github.com/Firstyear/kanidm/compare/20260516-oauth2-tester...yaleman:kanidm:20260516-oauth2-tester?expand=1

I'd say swap to using `axum-oidc` instead of directly using `openidconnect` which avoids a lot of weird backflips (and use the 1.x dev versions)

Firstyear:
@yaleman This just adds more dependencies - axum-oidc itself uses openidconnect and it's using version 3 of the library which relies on an even older version of reqwest. So I think that the approach I took here, as annoying as the new types are is better as openidconnect does plan to at least update to the latest reqwest sooner-than-later.

yaleman (dismissed):

yaleman (approved):

PR #4356, opened 2026-05-19
Resolve incorrect location of commited flag

Firstyear:
The flag on DbTxn's in the cache was incorrectly set before the commit was performed. This meant that a failure to commit (such as foreign key violation which is only checked at commit by sqlite) would result in the transaction being marked as committed, but still not complete. This left the DB with an uncommited txn present that prevented the creation of more transactions.

Thanks to @kmq for reporting.

Checklist


yaleman (approved):

Firstyear:
backported to 1.10.0 branch, but I don't think there is a rush to release so I'll wait to see if any other issues arise.

PR #4370, opened 2026-05-28
20260528 radius

### LINKED ISSUE #4368: Add group support to RADIUS
# Is your feature request related to a problem? Please describe

Many network devices like Switches, Router, Firewalls, etc. still only support RADIUS as an SSO provider for management logins. 

# Describe the solution you'd like

I'd like to see *some* sort of support for per user or per-group RADIUS group. 

MikroTik's RouterOS for example, lets you set a default group for RADIUS logins, but then pass a specific group via the `MikroTik-Group` attribute. I'm not sure the best solution - attributes per user, per group, custom policies to map data to custom attributes, etc.

The complication of course, is that groups are single key-pair value pretty consistently across all vendors. There's no nuance for passing an entire list and then granting perms for anything that matches. 

### MikroTik
https://help.mikrotik.com/docs/spaces/ROS/pages/328097/RADIUS
`MikroTik-Group = "full"`

### HP
`H3C-Exec-Privilege = 15`

### Aruba
`Aruba-Admin-Role = administrators`

### Cisco
`cisco-av-pair = "role=SUPER-ADMIN-ROLE"`
# Describe alternatives you've considered

I've tried mounting a custom `/etc/raddb/sites-available/inner-tunnel ` on top of the default one in the container, but as Kanidm doesn't expose any group info at all in RADIUS, that's not really scalable *if* I was able to get it to work. The best I could do is match on the username or just set a static "if you are allowed to use RADIUS at all, you get this role 100% of the time". But I'd like to be able to have network admins login to network devices while still having limited RADIUS access for users to be able to authenticate to WiFi etc without getting full admin access to network devices.

Firstyear:
# Change summary

- Allows specifying custom reply attributes that are associated to group memberships.

Fixes #4368

Checklist


Firstyear:
The only change I'd consider making is the reply-attribute keys as an enum - FreeRadius only allows certain keys to be valid, so we may want to more strictly control what valid values *are* that could be added here.

yaleman (approved):

PR #4372, opened 2026-05-29
Avoid LSE instructions in arm64 Linux builds

### LINKED ISSUE #4371: Kanidm 1.10.3 docker arm exits with code 132
### I did this

confirmed the upgradeability following [docs](https://kanidm.github.io/kanidm/stable/server_updates.html)
updated to `kanidm/server:latest` / `1.10.3` on my homelab PI 4
ran the fresh kanidm version

minimal compose.yml:
```yaml
services:
  kanidm:
    container_name: 'kanidm'
    image: docker.io/kanidm/server:1.10.3
    networks:
      proxy:
    volumes:
      - './data:/data'
      - '../00-certs/kanidm:/certs'
      #.....
```

### I expected the following

system runs and migrates to new version or at least logs something

### This happened instead

podman log-less exit with code 132

```json
{
  "OciVersion": "1.2.1",
  "Status": "exited",
  "Running": false,
  "Paused": false,
  "Restarting": false,
  "OOMKilled": false,
  "Dead": false,
  "Pid": 0,
  "ExitCode": 132,
  "Error": "",
  "StartedAt": "2026-05-28T18:29:39.028022598+02:00",
  "FinishedAt": "2026-05-28T18:29:39.605127551+02:00",
  "CheckpointedAt": "0001-01-01T00:00:00Z",
  "RestoredAt": "0001-01-01T00:00:00Z"
}
```

### Kanidm version details

- Are you running it in a container? If so, which image/tag?: kanidm/server:latest or kanidm/server:1.10.3
- If not a container, how'd you install it: n/a
- Operating System / Version (On Unix please post the output of `uname -a`): `Linux rpi-homelab 6.18.26 #1-NixOS SMP Thu Apr 30 09:13:53 UTC 2026 aarch64 GNU/Linux`

### Any other comments

version 1.10.2 seems to work normaly and running `sudo podman run --rm kanidm/server:1.10.2`
returns errors:
```
WARNING: No configuration path was provided, relying on environment variables.
ERROR: Tls Private Key and Certificate Chain are required.
ERROR: Unable to build server configuration from provided configuration inputs.
```

while running `sudo podman run --rm kanidm/server:1.10.3`
produces no output

caniko:
## Summary

Linux `aarch64` builds now apply conservative ARMv8.0-A flags through the shared Cargo/build-profile path rather than only inside the server container Dockerfile. This keeps release outputs compatible with Raspberry Pi 4 / Cortex-A72 class systems that do not support ARMv8.1-A LSE atomics.

Concretely, the aarch64 Linux target adds:

- a profile-side `armv8_a` default for native aarch64 Linux builds
- `-Ctarget-feature=-lse` from that profile
- `CFLAGS_aarch64_unknown_linux_gnu=-march=armv8-a`
- `CXXFLAGS_aarch64_unknown_linux_gnu=-march=armv8-a`

The C/C++ flags are required because the observed illegal instruction comes from `libmimalloc-sys`, whose build script requests `-march=armv8.1-a` and can emit LSE instructions such as `casal`/`ldaddal` on arm64 builders with newer CPUs.

Closes #4371.

## Validation

- `nix shell nixpkgs#cargo -c cargo metadata --no-deps --format-version 1`
- `nix shell nixpkgs#cargo nixpkgs#rustfmt -c cargo fmt --check`
- `git diff --check`
- `nix shell nixpkgs#cargo nixpkgs#rustc nixpkgs#pkg-config nixpkgs#clang nixpkgs#lld -c cargo check -p kanidm_build_profiles`
- Earlier container-only version: `docker build --target repos -f server/Dockerfile -t kanidm-dockerfile-syntax-check .`
- Separately smoke-tested an equivalent 1.10.3 rebuild on an RK3399 ARMv8.0-A host: `kanidmd version` prints `kanidmd 1.10.3` instead of exiting with SIGILL.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -8,7 +8,23 @@ rustflags = ["-C", "link-arg=-fuse-ld=lld", "-C", "force-frame-pointers=yes"]
 
 [target.aarch64-unknown-linux-gnu]
 linker = "clang"
-rustflags = ["-C", "link-arg=-fuse-ld=lld", "-C", "force-frame-pointers=yes"]
+rustflags = [
+    "-C",
+    "link-arg=-fuse-ld=lld",
+    "-C",
+    "force-frame-pointers=yes",
+    "-C",
+    "target-cpu=generic",
+    "-C",
+    "target-feature=-lse",
+]
+
+[env]
+# Keep Linux arm64 release builds compatible with ARMv8.0-A systems such as
+# Raspberry Pi 4. libmimalloc-sys otherwise requests -march=armv8.1-a, which
+# can emit LSE atomics that SIGILL on older arm64 hardware.
+CFLAGS_aarch64_unknown_linux_gnu = { value = "-march=armv8-a", force = false }
+CXXFLAGS_aarch64_unknown_linux_gnu = { value = "-march=armv8-a", force = false }
 
 [target.powerpc64le-unknown-linux-gnu]
 linker = "clang"
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -8,16 +8,7 @@ rustflags = ["-C", "link-arg=-fuse-ld=lld", "-C", "force-frame-pointers=yes"]
 
 [target.aarch64-unknown-linux-gnu]
 linker = "clang"
-rustflags = [
-    "-C",
-    "link-arg=-fuse-ld=lld",
-    "-C",
-    "force-frame-pointers=yes",
-    "-C",
-    "target-cpu=generic",
-    "-C",
-    "target-feature=-lse",
-]
+rustflags = ["-C", "link-arg=-fuse-ld=lld", "-C", "force-frame-pointers=yes"]
 
 [env]
 # Keep Linux arm64 release builds compatible with ARMv8.0-A systems such as
--- a/libs/profiles/src/lib.rs
+++ b/libs/profiles/src/lib.rs
@@ -11,6 +11,7 @@ use std::env;
 #[allow(non_camel_case_types)]
 enum CpuOptLevel {
     apple_m1,
+    armv8_a,
     none,
     native,
     neon_v8,
@@ -25,6 +26,10 @@ impl Default for CpuOptLevel {
             CpuOptLevel::x86_64_v2
         } else if cfg!(target_arch = "aarch64") && cfg!(target_os = "macos") {
             CpuOptLevel::apple_m1
+        } else if cfg!(target_arch = "aarch64") && cfg!(target_os = "linux") {
+            // Keep native arm64 Linux release builds on the ARMv8.0-A baseline.
+            // Dependency C/C++ flags for cross builds are enforced in .cargo/config.toml.
+            CpuOptLevel::armv8_a
         /*
         } else if cfg!(target_arch = "aarch64") && cfg!(target_os = "linux") {
             // Disable neon_v8 on linux - this has issues on non-apple hardware and on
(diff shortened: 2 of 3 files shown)


jinnatar:
This would apply a fix to only the container builds which leaves all other build mechanisms still potentially with bad binaries. Just as a thought, I wonder if it would be possible to do similar tweaking in the build profiles instead: https://github.com/kanidm/kanidm/tree/master/libs/profiles .. from those the change would propagate to all release mechanisms. Right now all release profiles are set to auto, so I assume there may be a way to do "slightly less auto" in the case of arm64 so we're not building to the capability of the GHA runners.

yaleman:
Yes this should be done in profiles so it's not specifically targeting one thing.

caniko:
Moved this out of the server container build in 43b8e4b0f.

The PR now sets the conservative aarch64 Linux baseline in two places:

- `libs/profiles` now has an `armv8_a` default for native aarch64 Linux profile builds, so the profile metadata no longer implies build-host auto-detection.
- `.cargo/config.toml` sets target-specific Rust flags and `CFLAGS_aarch64_unknown_linux_gnu` / `CXXFLAGS_aarch64_unknown_linux_gnu` for all aarch64 Linux Cargo builds from this repo.

I kept the C/C++ flags in Cargo config because the actual SIGILL is from `libmimalloc-sys` C code, not Kanidm Rust code. A profile build script cannot reliably push `CFLAGS` into transitive dependency build scripts, while Cargo target env does reach `cc` users like mimalloc across containers, release make targets, and other Cargo-based release builds.

.cargo/config.toml
```diff
@@ -8,7 +8,23 @@ rustflags = ["-C", "link-arg=-fuse-ld=lld", "-C", "force-frame-pointers=yes"]
...
 [target.aarch64-unknown-linux-gnu]
 linker = "clang"
-rustflags = ["-C", "link-arg=-fuse-ld=lld", "-C", "force-frame-pointers=yes"]
+rustflags = [
+    "-C",
+    "link-arg=-fuse-ld=lld",
+    "-C",
+    "force-frame-pointers=yes",
+    "-C",
+    "target-cpu=generic",
```
Firstyear: This isn't needed as it comes from profiles.

libs/profiles/src/lib.rs
```diff
@@ -116,6 +122,9 @@ pub fn apply_profile() {
 
     match profile_cfg.cpu_flags {
         CpuOptLevel::apple_m1 => println!("cargo:rustc-env=RUSTFLAGS=-Ctarget-cpu=apple_m1"),
+        CpuOptLevel::armv8_a => {
+            println!("cargo:rustc-env=RUSTFLAGS=-Ctarget-cpu=generic -Ctarget-feature=-lse")
```
Firstyear: I think you don't need target-cpu here, only target features is needed.

Firstyear (changes requested):

.cargo/config.toml
caniko: Removed the duplicate Rust target flags from `.cargo/config.toml`; the Rust-side `-lse` now comes only from `libs/profiles`. I left the target-specific `CFLAGS` / `CXXFLAGS` there because those are for `cc` users like `libmimalloc-sys`, not Kanidm Rust code.

libs/profiles/src/lib.rs
caniko: Updated in `5f758e6de`: `armv8_a` now emits only `-Ctarget-feature=-lse`.

Firstyear (approved):

Firstyear:
Thank you!

PR #4388, opened 2026-06-04
WIP: Implement person/group creation/deletion from admin UI

hut:
### NOTE: THIS PR IS A WORK IN PROGRESS, NOT READY FOR MERGE

# Change summary

My team and me are working on enhancing the Kanidm admin UI.

Update: By now this PR allows you to create/delete persons and groups.

The original PR message was this, for reference:

> The first step is to add a feature to create new users (specifically Persons) in the admin UI. For later, we also plan implementing person deletion, group creation/deletion, and more.
>
> This is my WIP of person creation. The following works:
>
> - [X] Form with username and optional email address
> - [X] Person created successfully via `handle_create`
> - [X] On various errors, error messages are shown in a toast

What is missing:

- [X] More error messages (e.g. if email address is invalid)
- [X] The user that gets swapped in via HTMX into the user table has a wrong Uuid, this is not implemented yet.
    - This was "fixed" by removing the UUID from the interface in favor of the email address.
- [ ] No toast on success
    - I'd appreciate some help with this. There's a bit of commented-out code for a `PersonEntryResponse` that both updates the user table via HTMX and shows a success toast, copied from the "add member to group" feature. But if I use this, simply nothing happens in the UI when I create a user, even though according to the network tab of firefox developer tools, the HTMX response is properly received, and the console log is free of error messages.

Since this is my first contribution, I would love to hear your thoughts on the implementation.

- Are you interested in merging such a feature?
- Am I on the right track with the implementation?
- Anything critical that is missing?

Thank you for your feedback and for this beautiful project. <3

Checklist

### DIFF
--- a/server/core/src/https/views/admin/persons.rs
+++ b/server/core/src/https/views/admin/persons.rs
@@ -3,23 +3,28 @@ use crate::https::extractors::{DomainInfo, VerifiedClientInformation};
 use crate::https::middleware::KOpId;
 use crate::https::views::errors::HtmxError;
 use crate::https::views::navbar::NavbarCtx;
-use crate::https::views::Urls;
+use crate::https::views::{ErrorToastPartial, Urls};
 use crate::https::ServerState;
 use askama::Template;
 use askama_web::WebTemplate;
-use axum::extract::{Path, State};
+use axum::extract::{Form, Path, State};
 use axum::response::{IntoResponse, Response};
 use axum::Extension;
 use axum_htmx::{HxPushUrl, HxRequest};
+use kanidm_proto::constants::{ATTR_NAME, ATTR_MAIL, ATTR_DISPLAYNAME};
 use kanidm_proto::attribute::Attribute;
-use kanidm_proto::internal::{OperationError, UserAuthToken};
+use kanidm_proto::internal::{OperationError, UserAuthToken, SchemaError};
 use kanidm_proto::scim_v1::server::{
     ScimEffectiveAccess, ScimEntryKanidm, ScimListResponse, ScimPerson,
 };
 use kanidm_proto::scim_v1::ScimEntryGetQuery;
 use kanidm_proto::scim_v1::ScimFilter;
+use kanidm_proto::v1::Entry;
+use kanidm_proto::internal::CreateRequest;
 use kanidmd_lib::constants::EntryClass;
 use kanidmd_lib::idm::authentication::ClientAuthInfo;
+use serde::{Deserialize, Serialize};
+use std::collections::BTreeMap;
 use uuid::Uuid;
 
 pub const PERSON_ATTRIBUTES: [Attribute; 9] = [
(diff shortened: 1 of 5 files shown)


server/core/src/https/views/admin/persons.rs
```diff
@@ -93,6 +98,143 @@ pub(crate) async fn view_person_view_get(
...
+                    return Err(HtmxError::new(&kopid, error, domain_info.clone()))?;
+                }
+            }
+            OperationError::SchemaViolation(ref schemaerror) => {
+                match schemaerror {
+                    SchemaError::InvalidAttributeSyntax(details) => {
+                        if details == "displayname" {
+                            return Ok((ErrorToastPartial {
+                                err_code: OperationError::UI0006MissingDisplayName,
+                                operation_id: kopid.eventid,
```
janhaa, line 212: we should use a static map (attributes -> OperationError) here to resolve from to have less repeated code and better maintainability, e.g.:

```
Attribute::Name -> OperationError::UI0005PersonAlreadyExists,
Attribute::Mail -> OperationError::UI0007DuplicateEmail
....
```

janhaa (commented):
nice :)

Firstyear:
Yeah I think error handling is something we definitely need to improve but I think that's more of a server internal problem than anything. Rather, it's not error handling as much as how we can community better when errors occur. 

Otherwise this is a good start, I'll do a deeper review later but thank you!

Firstyear:
@hut Is there wore you want to add to this PR?

hut:
@Firstyear Yes, definitely.

Here's a more polished version that is used in production on my organization's server.

I've tested it only on 1.10.4, but the version I pushed is based on master. Will test if it still works and report back.

hut:
It includes group addition/deletion too btw.

server/core/src/https/views/admin/persons.rs
```diff
@@ -95,6 +100,168 @@ pub(crate) async fn view_person_view_get(
...
+    State(state): State<ServerState>,
+    Extension(kopid): Extension<KOpId>,
+    DomainInfo(domain_info): DomainInfo,
+    VerifiedClientInformation(client_auth_info): VerifiedClientInformation,
+    Form(query): Form<AddPersonForm>,
+) -> axum::response::Result<Response> {
+    let username = query.name.to_string();
+    let email = query.email.to_string();
+
+    let mut new_acct = Entry {
```
Firstyear: A better way to build this would be a https://github.com/kanidm/kanidm/blob/master/proto/src/scim_v1/client.rs#L280 which you pass to https://github.com/kanidm/kanidm/blob/master/server/core/src/actors/v1_scim.rs#L190

server/core/src/https/views/admin/persons.rs
Firstyear: Realistically the process we go through to validate the entry is the problem, if we made a proper ScimPersonPut type or similar, and then we can pre-validate requirements before we send them down to the server, that way we can have better error detection earlier on.

Firstyear (changes requested):
I think this is looking really good, I think if we swap to the ScimEntryPost variant, or we add a ScimGroupPost + some validators earlier then it'd probably be nicer. But I'm happy to just start with the ScimEntryPost change and we can improve the more specific apis later.

Firstyear:
@hut Something to keep in mind is that the admin ui is still feature gated, so we can merge this even if it's not in a fully-complete state.

hut:
> @hut Something to keep in mind is that the admin ui is still feature gated, so we can merge this even if it's not in a fully-complete state.

That's relieving to hear, because since I built this, I've moved on to another project and currently don't have time to spare to polish this PR.

If you could take it from here, I would appreciate it.

P.S.:

> I've tested it only on 1.10.4, but the version I pushed is based on master. Will test if it still works and report back.

Tested it, and it works fine. Though there should be more error messages for all the different reasons this might fail.

Firstyear:
No problem, I'm happy to do that.

Firstyear (approved):

Firstyear:
It's not perfect, but it's a great start, and I fixed up the entry create now to use the scim variants.

PR #4393, opened 2026-06-08
docs: document oauth key rotation

bdolgov:
# Change summary

I was researching key rotation because I am moving my instance from a relatively not trusted environment (playground in public cloud) into a relatively trusted environment. I was surprised to find no documentation or recommendation about rotation in the docs, especially given that the keys are stored not in HSMs. I found that keycloak [recommends](https://www.keycloak.org/docs/latest/server_admin/index.html#rotating-keys) periodic rotation, and zitadel [explains](https://zitadel.com/docs/guides/integrate/login/oidc/webkeys#rotation-example) why you should not immediately activate new keys.

Do these recommendations make sense? Does including them in the docs (and on this page) make sense? 

Checklist

- [x] This PR contains no AI generated code **(but the docs were AI-generated)**

### DIFF
--- a/book/src/integrations/oauth2.md
+++ b/book/src/integrations/oauth2.md
@@ -414,6 +414,55 @@ kanidm system oauth2 warning-enable-legacy-crypto <client name>
 
 In this mode, Kanidm will not offer `ES256` support for the client at all.
 
+## Rotating Token Signing Keys
+
+Each OAuth2 client has its own unique cryptographic keys, used to sign the `id_token` and access tokens it issues.
+Kanidm stores these keys in its database and publishes the public portion at the client's JWKS endpoint
+(`https://idm.example.com/oauth2/openid/:client_id:/public_key.jwk`) so that clients can validate the tokens they
+receive. Kanidm does not rotate signing keys automatically. A client's keys remain in use until you rotate or revoke
+them.
+
+To rotate the signing and encryption keys for a client, specify the time the rotation should take effect as an
+[RFC 3339][rfc3339] timestamp. As soon as you run the command the new public key is published in the JWKS endpoint, but
+it does not begin signing tokens until the time you specified:
+
+```bash
+kanidm system oauth2 rotate-cryptographic-keys <client name> <rotation time in the future>
+```
+
+After the rotation takes effect, newly issued tokens are signed with the new key. The previous key is retained and
+continues to be published in the JWKS endpoint, so tokens that were already issued remain valid until they expire. This
+allows a key rotation to occur without disrupting active sessions.
+
+> [!IMPORTANT]
+>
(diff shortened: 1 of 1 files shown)


Firstyear:
>     * [x]  This PR contains no AI generated code **(but the docs were AI-generated)**

Ahh, we can't accept any AI generated *content*. not just code. So this will need to be reworked.

> I was researching key rotation because I am moving my instance from a relatively not trusted environment (playground in public cloud) into a relatively trusted environment. I was surprised to find no documentation or recommendation about rotation in the docs, especially given that the keys are stored not in HSMs.

Whenever we discuss something like this, we need to do so with a risk/threat model. 

In the keycloak case, this appears to be used to cause token *expiration* to occur. Specifically refresh tokens in this case, as reading between the lines they say " If a user was inactive in the period between the new keys being added and the old keys being removed, that user will have to re-authenticate."

We have separate methods to ensure that refresh tokens have an expiry that doesn't rely on private-key rotation. So this doesn't apply.

The zitadel docs don't describe why rotation should be undertaken.

If an admin is rotating keys because they think they *might* be compromised, then they should be revoking the keys instead, not rotating them. Just rotating keys on a schedule, seems like a very specific need. 

So I think while we should document *how* to rotate and revoke keys, we shouldn't be prescribing a schedule for key rotation to occur. Things like that always have to be done with some kind of goal in mind. 

Anyway, we do want to support HSMs in future, the private key handling internal to Kanidm would actually allow it. we just don't have any HSM's available to do the development work against today!

Firstyear (approved):

Firstyear (changes requested):
Revert approval

book/src/integrations/oauth2.md
```diff
@@ -414,6 +414,55 @@ kanidm system oauth2 warning-enable-legacy-crypto <client name>
...
+> You should almost always schedule a rotation for a point in the **future**, not `now`. Relying parties cache the JWKS
+> endpoint, and in a [replicated](../repl/) deployment the new key must propagate to all Kanidm nodes. If a key begins
+> signing tokens before clients and peers have discovered it, those clients will reject the tokens as having an unknown
+> signing key until their cache refreshes. Staging the rotation in the future gives the new public key time to be
+> discovered everywhere before it is first used to sign. Allow a window comfortably longer than your clients' JWKS cache
+> lifetime — a few hours, or a day, is a safe choice.
+
+Rotating keys periodically is good practice and limits the window in which a leaked key could be abused. A rotation
+every three to six months is a reasonable starting point for most deployments. Note that rotation does not revoke the
+previous key: it is retained so that already-issued tokens remain valid, and it must be revoked manually once it is no
```
yaleman, line 448:
```suggestion
previous key; it is retained so that already-issued tokens remain valid, and it must be revoked manually once it is no
```

yaleman (changes requested):

PR #4415, opened 2026-06-25
change enroll to std spellling

solinium:
change to standard enroll spelling


micolous (changes requested):
It is already correct: https://en.wiktionary.org/wiki/enrol#English

"enroll" is the American spelling.

This PR also renames many internal identifiers, which I suspect would break existing Kanidm deployments.

yaleman:
Thanks but I'm going to close this one, the spelling is correct per our usual en-AU spelling standards.

PR #4419, opened 2026-06-27
LDAP session ID and application password cleanup

yaleman:
- Replace UnixBind with ApplicationPasswordBind so the application UUID is preserved in the session. 
- Refactored LdapSession::ApplicationPasswordBind to a struct variant with named fields for clarity.
- Thread session_id through from application_auth_ldap into verify_application_password and validate_ldap_session instead of generating it independently in multiple places.

Checklist


libs/crypto/src/lib.rs
```diff
@@ -30,6 +30,7 @@ use std::fmt;
 use std::fmt::Display;
 use std::num::ParseIntError;
 use std::time::{Duration, Instant};
+use subtle::ConstantTimeEq;
```
Firstyear, line 33: We should expose this via crypto-glue instead IMO

server/lib/src/be/mod.rs
```diff
@@ -633,7 +635,18 @@ pub trait BackendTransaction {
         };
 
         if idl.len() > FILTER_SUBSTR_TEST_THRESHOLD {
+            let mut trigraph_count = 1;
             for idx_key in grapheme_iter {
+                trigraph_count += 1;
```
Firstyear, line 640: This is an invalid change, and may break substring queries. Please revert this one. If the goal is to limit thelength of a substring query there are better ways to do it I think.

server/lib/src/idm/application.rs
```diff
@@ -185,29 +185,7 @@ impl IdmServerAuthTransaction<'_> {
...
-                    session_id,
-                    effective_session: LdapSession::UnixBind(account.uuid),
-                }))
-            }
-            None => {
-                security_info!("Account does not have a configured application password.");
-                Ok(None)
-            }
-        }
+        account.verify_application_password(lae.eventid, application, &lae.cleartext)
```
Firstyear, line 188: This change is also invalid - the event id is not the same as the session id. The session id may persist over a number of events.

Firstyear (changes requested):
Okay, I have some initial comments but over all this PR is touching too much, and seems rlike it might introduce a lot of regressions. I think it needs to be split up and much more careful in what we touch in smaller sections to ensure we aren't breaking anything.

server/lib/src/server/access/mod.rs
```diff
@@ -3118,7 +3136,63 @@ mod tests {
...
         let ex_b = vec![];
 
-        test_acp_search!(&se_b, vec![], r_set, ex_b);
+        test_acp_search!(&se_b, vec![acp], r_set, ex_b);
+    }
+
+    #[test]
+    fn test_access_oauth2_no_acp_denied() {
+        sketching::test_init();
+        // Test that group membership alone is NOT sufficient to see an OAuth2 RS entry
```
Firstyear, line 3145: This change is questionable, because if you can read the entry you can query it in different ways. So I think this actually isn't valid.

yaleman:
ignore this one wrong git base!

PR #4421, opened 2026-06-27
LDAP session ID and application password cleanup

yaleman:
tl;dr found some weirdness, wasn't clear how things worked so this is really just a developer UX refactor

Changes:

- Refactored LdapSession::ApplicationPasswordBind to a struct variant with named fields for clarity.
- application_auth_ldap was previously resolving down to a LdapSession::UnixBind instead of an LdapSession::ApplicationPasswordBind - the resulting policy decisions are the same, because the system really only acts on the user's uuid but now it's not using the wrong struct just 'cause.
- "eventid" is now request_id for the lifetime to denote it's an individual request which matches the function names a little clearer

Checklist

### DIFF
--- a/server/core/src/actors/v1_read.rs
+++ b/server/core/src/actors/v1_read.rs
@@ -1540,13 +1540,13 @@ impl QueryServerReadV1 {
     #[instrument(
         level = "info",
         skip_all,
-        fields(uuid = ?request_id)
+        fields(uuid = ?eventid)
     )]
     /// Retrieve a public jwk
     pub async fn handle_public_jwk_get(
         &self,
         key_id: String,
-        request_id: Uuid,
+        eventid: Uuid,
     ) -> Result<Jwk, OperationError> {
         let mut idms_prox_read = self.idms.proxy_read().await?;
 
--- a/server/core/src/actors/v1_read.rs
+++ b/server/core/src/actors/v1_read.rs
@@ -1540,13 +1540,13 @@ impl QueryServerReadV1 {
     #[instrument(
         level = "info",
         skip_all,
-        fields(uuid = ?eventid)
+        fields(uuid = ?request_id)
     )]
     /// Retrieve a public jwk
     pub async fn handle_public_jwk_get(
         &self,
         key_id: String,
-        eventid: Uuid,
+        request_id: Uuid,
     ) -> Result<Jwk, OperationError> {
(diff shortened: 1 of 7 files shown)


CEbbinghaus (dismissed):

server/core/src/actors/v1_read.rs
```diff
@@ -1540,13 +1540,13 @@ impl QueryServerReadV1 {
     #[instrument(
         level = "info",
         skip_all,
-        fields(uuid = ?eventid)
+        fields(uuid = ?request_id)
```
Firstyear: Every other path in the server calls this eventid. 'Event' is a standard term anyway for this, so I think this change will only cause more confusion - not less.

server/lib/src/idm/server.rs
```diff
@@ -1063,22 +1063,24 @@ pub trait IdmServerTransaction<'a> {
     #[instrument(level = "debug", skip_all)]
     fn validate_ldap_session(
         &mut self,
-        session: &LdapSession,
+        ldap_bound: &LdapBoundToken,
```
Firstyear, line 1066: Why is this being changed?

Firstyear (changes requested):
I'm not super keen on most of this PR, there are some nice changes like the named struct members for some parts, and the LdapSession default, but mostly I think this is only going to introduce developer confusion,and some changes dont seem to have any purpose. I think maybe it would be better to take this one back to the drawing board and target one thing at a time.

Firstyear:
https://en.wikipedia.org/wiki/Event-driven_finite-state_machine <- this is where the term Event comes from btw.

server/core/src/actors/v1_read.rs
yaleman: Uh, not sure how I even edited this one, probably just working too quickly while reverting the other stuff. Ignore this - reverted.

server/lib/src/idm/server.rs
yaleman: Because on line 1073 it needs the session ID from the `LdapBoundToken` for the `Identity` that's returned.

So this aligns with the concept that an `LdapBoundToken` keeps that session id until it's deauthed - matching the flow from the `process_uat_to_identity`

Firstyear:
I think this one needs to be redone in smaller chunks. Sorry :(

PR #4442, opened 2026-07-03
Limit password max length

### LINKED ISSUE #4414: Maximum password length
Within Kanidm we implement and enforce a *minimum* password length. This is to reduce the risk of bruteforce, dictionary attacks, and other forms of weak passwords. 

However, at the opposite end we do not enforce a *maximum* length. I am proposing that we add one.

The major reason for this is to prevent a style of DOS attack that relies on an attacker submitting a large password input for verification that then overwhelms the hash algorithm. This is the reason why bcrypt implements a 72 byte maximum for passwords. 

The second is TPMs. TPMs allow passwords for a variety of functions (PIN values). These are limited in length to the block size of the HMAC algorithm used, commonly HMAC-SHA256 which limits to 32 bytes. 

I am proposing one of two paths:

1) we limit passwords to 32 bytes in length. At even 28 characters, a lowercase-alpha only password still has more entropy than an aes-128 key. Once we expect that uppercase and numeric is possible, we exceed this greatly. As an example, the famous "correct horse battery staple" at 28 characters would still fit in this scheme, but it does get pretty tight at this limit. Regardless more commonly we see passwords < 16 characters, so this would likely not impact users. 

2) we limit passwords to 64 bytes. This prevents the possible limits on longer passwords such as the "correct horse battery staple" style. It does mean that we no longer fit within a TPM PIN limit, but we can work around this through doing an HMAC of the PW before we use it as the PIN (to  prevent hash-shucking)

I would prefer the 32 byte limit as it provides the simplest path forward for TPM integration and interop, and we can *expand* the limit later. However, I can also see why people would prefer 64. 

So I'm open to questions/comments/thoughts.

Firstyear:
This checks new passwords that they are not in excess of 64 bytes. As some users may have created passwords in the past that are longer than this, existing passwords up to 128 bytes can continue to be validated.

Fixes #4414

Checklist

### DIFF
--- a/libs/crypto/src/lib.rs
+++ b/libs/crypto/src/lib.rs
@@ -38,7 +38,7 @@ pub use sha2;
 
 // Per https://pages.nist.gov/800-63-4/sp800-63b.html max length should be 64. This is
 // measured in GRAPHEMES, not bytes.
-pub const PW_MAX_LENGTH_NIST: u32 = 64;
+pub const PW_MAX_LENGTH_NIST: u32 = 128;
 
 // Single factor passwords have a greater minimum length requirement per nist.
 pub const PW_SFA_MIN_LENGTH_NIST: u32 = 15;
--- a/libs/crypto/src/lib.rs
+++ b/libs/crypto/src/lib.rs
@@ -36,7 +36,8 @@ mod crypt_md5;
 
 pub use sha2;
 
-// Per https://pages.nist.gov/800-63-4/sp800-63b.html max length should be 64.
+// Per https://pages.nist.gov/800-63-4/sp800-63b.html max length should be 64. This is
+// measured in GRAPHEMES, not bytes.
 pub const PW_MAX_LENGTH_NIST: u32 = 64;
 
 // Single factor passwords have a greater minimum length requirement per nist.
@@ -46,8 +47,8 @@ pub const PW_SFA_MIN_LENGTH_NIST: u32 = 15;
 pub const PW_MFA_MIN_LENGTH: u32 = 10;
 
 // Since we added a max length constraint later, we should continue to allow long
-// passwords to be validated.
-pub const PW_MAX_LENGTH_CHECK: usize = PW_MAX_LENGTH_NIST as usize * 2;
+// passwords to be validated. NOTE this is BYTES not GRAPHEMES.
+pub const PW_MAX_LENGTH_CHECK: usize = PW_MAX_LENGTH_NIST as usize * 4;
 
 // NIST 800-63.b salt should be 112 bits -> 14  8u8.
 const PBKDF2_SALT_LEN: usize = 24;
--- a/libs/crypto/src/lib.rs
+++ b/libs/crypto/src/lib.rs
@@ -36,6 +36,19 @@ mod crypt_md5;
 
 pub use sha2;
 
+// Per https://pages.nist.gov/800-63-4/sp800-63b.html max length should be 64.
+pub const PW_MAX_LENGTH_NIST: u32 = 64;
+
+// Single factor passwords have a greater minimum length requirement per nist.
+pub const PW_SFA_MIN_LENGTH_NIST: u32 = 15;
+// When used with MFA, the MIN length can be relaxed. Per NIST 8 is acceptable, but we
+// have used 10 for this value.
+pub const PW_MFA_MIN_LENGTH: u32 = 10;
+
+// Since we added a max length constraint later, we should continue to allow long
+// passwords to be validated.
+pub const PW_MAX_LENGTH_CHECK: usize = PW_MAX_LENGTH_NIST as usize * 2;
+
 // NIST 800-63.b salt should be 112 bits -> 14  8u8.
 const PBKDF2_SALT_LEN: usize = 24;
 
@@ -920,6 +933,11 @@ impl Password {
         cleartext: &str,
         hsm: Option<(&mut dyn TpmHmacS256, &HmacS256Key)>,
     ) -> Result<bool, CryptoError> {
+        if cleartext.len() > PW_MAX_LENGTH_CHECK {
+            error!("Cleartext input exceeds safe KDF length {PW_MAX_LENGTH_CHECK}, refusing to proceed.");
+            return Ok(false);
+        }
+
         match (&self.material, hsm) {
             (
                 Kdf::TPM_ARGON2ID {
--- a/proto/src/internal/credupdate.rs
+++ b/proto/src/internal/credupdate.rs
@@ -336,6 +336,7 @@ pub enum PasswordFeedback {
     CommonNamesAndSurnamesAreEasyToGuess,
     // Custom
     TooShort(u32),
+    TooLong(u32),
     BadListed,
     DontReusePasswords,
 }
@@ -423,6 +424,10 @@ impl fmt::Display for PasswordFeedback {
             PasswordFeedback::TooShort(minlength) => write!(
                 f,
                 "Password was too short, needs to be at least {minlength} characters long."
+            ),
+            PasswordFeedback::TooLong(maxlength) => write!(
+                f,
+                "Password was too long, must not be greater than {maxlength} characters long."
             ),
             PasswordFeedback::UseAFewWordsAvoidCommonPhrases => {
                 write!(f, "Use a few words and avoid common phrases.")
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3826,6 +3826,7 @@ dependencies = [
  "time",
  "tokio",
  "tracing",
+ "unicode-segmentation",
  "url",
  "utoipa",
  "uuid",
@@ -7321,6 +7322,12 @@ checksum = "5fd4f6878c9cb28d874b009da9e8d183b5abc80117c40bbd187a1fde336be6e8"
 dependencies = [
  "tinyvec",
 ]
+
+[[package]]
+name = "unicode-segmentation"
+version = "1.13.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c6f5d3c3b1bf09027a88a6bc961fc00497d651009560b5463668dc81b0fa87a8"
 
 [[package]]
 name = "unicode-width"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -296,6 +296,7 @@ tracing = { version = "^0.1.44", features = [
 ] }
 tracing-subscriber = { version = "^0.3.23", features = ["env-filter"] }
 tracing-forest = { version = "^0.3.1", features = ["defer"] }
+unicode-segmentation = "1.13.3"
 url = "^2.5.8"
 urlencoding = "2.1.3"
 utoipa = { version = "5.5.0", features = ["url", "uuid"] }
--- a/server/lib/Cargo.toml
+++ b/server/lib/Cargo.toml
@@ -65,6 +65,7 @@ tokio = { workspace = true, features = ["net", "sync", "time", "rt"] }
 nonempty = { workspace = true, features = ["serialize"] }
 
 tracing = { workspace = true, features = ["attributes"] }
+unicode-segmentation = { workspace = true }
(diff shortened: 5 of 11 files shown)


proto/src/internal/credupdate.rs
```diff
@@ -424,6 +425,10 @@ impl fmt::Display for PasswordFeedback {
                 f,
                 "Password was too short, needs to be at least {minlength} characters long."
             ),
+            PasswordFeedback::TooLong(maxlength) => write!(
+                f,
+                "Password was too long, must not be greater than {maxlength} characters long."
```
Gabgobie, line 430: I've unfortunately seen this just now and added my thoughts to the linked issue. Nonetheless, I did try to give this a short review.

This feedback message is confusing when your password _is_ <=64 _characters_ long but >64 bytes. Depending on your language, this difference may be drastic. I don't really have a better suggestion though. We can't reasonably teach a user about UTF-8 and the byte length of their characters/password.

yaleman, line 430:
```suggestion
                "The password entered encodes to more than {maxlength} bytes, please use less characters."
```

yaleman (dismissed):
The user-facing wording's always going to be hard, but added a suggestion.

libs/crypto/src/lib.rs
```diff
@@ -36,6 +36,19 @@ mod crypt_md5;
 
 pub use sha2;
 
+// Per https://pages.nist.gov/800-63-4/sp800-63b.html max length should be 64.
```
Gabgobie: I'd like to suggest using a _significantly_ higher limit.

NIST is not saying that the max length should be 64 but that it should be _at least_ 64. - It's a lower limit.

We should definitely not let people upload their favorite movie as a password but this will cause all kinds of UX issues I don't want to deal with and I don't want to have to explain to my peers.

Firstyear: As mentioned in the ticket, once you are past 28 bytes you already exceed an AES-128 key, and even in 64 characters, you can fit a whole AES256 key in hex. So initially I want to start with this as a limit because it is more than enough for a secure password.

Firstyear: Also PS you haven't defined a threat model - what is the risk you are worried about. Online bruteforce? We have rate limits for that. Password hash disclosure? Well that's what the PW minimum length is for as well as argon2id. Like what's the actual concern besides "bigger is better"?

proto/src/internal/credupdate.rs
Firstyear: I made the code check for graphemes not bytes, so this should make it a bit easier.

Firstyear: Made the code check for graphemes, not bytes.

libs/crypto/src/lib.rs
Gabgobie: > threat model - what is the risk you are worried about

I terms of threat, this is mostly about the unix password since I use WebAuthN otherwise.

Until service passwords are a thing, the unix password is fairly powerful[^1] and has to be something I can type in. That means I have to take a passphrase.[^2] I am working in a busy environment with poor/no visual protection so my worry is fragments of the password leaking. Going by unixd's offline capabilities, I'm guessing the unix password's hash can be freely fetched and subjected to an offline attack. More words reduce the percentage a fragment represents and probability for a full reconstruction.

What I am more worried about, however, is the UX. Given the previously outlined threat model, I frequently change the passphrase. It is in the nature of this kind of password to vary in length and the worst part of password rotation is length/character rejection and regeneration. From experience with other services, this process can easily cost an additional 10min of sitting there, trying if a password is accepted and regenerating in one big frustrating loop. With less words, I'd have to do this even more frequently.

I am 100% on board that a limit is necessary so no single user can overload the instance but on the other hand, I am convinced that this limit should _never_ be encountered by _any_ user during normal operation.

Aside from the mathematical likelihood of a breach, there are also a few social aspects to this, one of which is that I spent _a lot_ of time convincing my peers to even start using a password manager and more importantly to use the generator thereof. Part of this process was telling them to just use the highest settings because they won't be able to and don't have to remember the password anyway. This has helped tremendously in discouraging the use of a password manager as a bookmark list with immense amounts of password duplication. Given that security is always a `>=`, this also has the nice side effect, that no average user needs to be able to assess password strength. It will always be (more than) sufficient at identical cost. I'd rather not start undoing this. This is not a bigger is better in disguise. This is supposed to highlight the wide-reaching consequences of facing a user with the prompt to take a less secure password.

Maybe there is value in asking the other way around:

- How many Bytes from a single user does it take to hurt us? Then maybe consider the same with 100 coordinated users and go from there. The password can only be validated after a successful TOTP challenge so we a) know the attacker's identity and b) this probably mostly concerns the account update operation
- We already don't do `32 Bytes` so we need to use an HMAC for TPM integration. What limitations does this pose?

From my point of view, there needs to be a stronger reason for disturbing the user than "less is secure enough, go and try again", especially for an artificial limitation. Otherwise we would be doing more harm than we prevent. Off the top of my head, harm in the form of wasted time for both users and support staff, reputation ("I used this application and it doesn't support secure passwords!" or "We can't deploy this application even though it is reasonable because our users wouldn't understand.") and - even if unreasonable - comfort/peace of mind.

To leave out the entire user debate and insert some personal words for a second: This directly impacts me. My password being grandfathered in doesn't help much given that I am going to rotate it soon enough. I will have a terrible time during every rotation and I will likely have to deal with user complaints about this. I don't want any of that unless there is a technical limitation that doesn't allow for a higher limit.

Besides - is it really our place to dictate what other people should feel comfortable with? I can get behind enforcing a minimum. I support protecting the server with a maximum. But there's a difference between protecting the server and being intrusive.

---

When only considering my own issues and looking for the simplest compromise, it would already be sufficient to allow 256 Bytes (as is technically the case since de8facf) and report that 64 characters are allowed in the error message. That's more or less exactly what the current PR does except that it's slightly less restrictive and probably more performant. Think this is a good middle-ground?

[^1]: I think to remember it currently being (mis-)used for e-mail accounts and other ldap-only applications. This will hit me as soon as I find the capacity to integrate those services with Kani as well.
[^2]: Which in my perspective invalidates the comparison to the AES keys since we are no longer dealing with a random bit sequence.

Firstyear: This isn't a threat model. 

A threat or risk is things like:

* An attacker is attempting to online bruteforce the password
* An attacker with access to the hash as they have root on the Kanidm system
* An attacker who gains the hash via a backup dump from the server. 

Each of these has different things present to protect the password. 

First, an attacker attempting an online bruteforce will run into our softlocking and rate limiting mechanisms so that they are unable to attempt more than ~100 attempts per day. This makes online bruteforce infeasible.

Second, and attacker with the hash needs because they have root on the server, doesn't need to attack the hash because they can just change any credential they want. 

Third, if they have the hash from a backup dump, we use argon2id which is both cpu and memory hard, making gpu bruteforce difficult (see also cuda hashcat), while also consuming a lot of resources. I'm also not sure of argon2id rainbow tables etc. 


Now consider these wordlists you keep talking about - even with a wordlist, you still have to face either A) the bruteforce preventions or B) the argon2id cost penalty to validate the password. In fact when we design these schemes we take wordlists into account, as does things like cuda hashcat!!!


More importantly when we discuss these the conversation always goes to the *minimums*. We already use zxcvbn to improve password entropy, we have minimum length requirements, and we store passwords in formats that make them difficult to attack.

In other words - we have to consider that a password of 10 characters (today) that has passed through these checks is secure. 

So if we assume that the *minimum* password is secure, what do we gain from security-maxing? It's nothing more than psychological / emotional validation, not true security. 

This is why I picked 64 - if 10 is already "secure" then 64 *surely* is. If you want to make an argument, it should be that we should increase our minimums. 

In addition, I also changed the code to accept this as 64 characters (up to 256 bytes). 256 bytes is the maximum input limit for Entra AD from my research. Other schemes like bcrypt limit to 72 bytes and are considered secure.

So I think these are reasonable values, and have supporting evidence. Remember, the problem here is that argon2id takes cpu and memory time from the server too - so the longer we allow the input, the more we make degrade performance of the system. That's what this cap is intended to solve (and in fact, is why bcrypt limits to 72 bytes.).

Firstyear:
I've made the limit 128 chars (512 bytes).

libs/crypto/src/lib.rs
Gabgobie: <details><summary>My personal issues have been resolved as of 3ed7118 but I can't speak for everyone. In that sense I'd be fine dropping this thread.</summary>

> This isn't a threat model.

<https://owasp.org/www-community/Threat_Modeling#threat-modeling-four-question-framework>

- [x] Assess Scope
  - unixd machines
  - mail
  - other non-oauth-services
- [x] Identify what can go wrong
  - Fragments of the passphrase leak
  - The passphrase may be brute-forced offline with the help of the leaked fragments
- [x] Identify countermeasures or manage risk
  - Frequently rotate passphrase to mitigate leaked fragments
  - Use a passphrase with many segments to reduce leaked percentage
- [ ] Assess your work

Did you want a $\LaTeX$ PDF going into detail about my environment?

---

> In other words - we have to consider that a password of 10 characters (today) that has passed through these checks is secure.

But we are in agreement that the entropy of a few words is hardly comparable to the entropy of a string of random characters with the same length, right? And as soon as part of the password is leaked, the difference becomes even more apparent.

When there's no room for enough words, this quickly drops into an uncomfortable realm.

---

In both the issue and this PR, all responses have been a "you don't need long passwords" when I explicitly stated _multiple times_ that my bigger concern is the resulting UX and social implications as well as the missing/never assessed benefit compared to these drawbacks.

If this is a long-made decision, I'll remind you that this is _your_ project and that you are in your right to make this decision unilaterally but I'd appreciate if that was made clear so we don't have to waste time on an argument that doesn't address my main concerns.

Otherwise, I'll reiterate my main points that are still to be addressed:

---

> [Encouraging the use of _very_ complex passwords...] Given that security is always a `>=`, this also has the nice side effect, that no average user needs to be able to assess password strength. It will always be (more than) sufficient at [in every practical sense] identical cost. I'd rather not start undoing this.

---

> Remember, the problem here is that argon2id takes cpu and memory time from the server too - so the longer we allow the input, the more we make degrade performance of the system.

Which is why I support adding a limit. But this limit should be based on projected load and actual harm that comes to the deployment.

> Maybe there is value in asking the other way around:
>
> - How many Bytes from a single user does it take to hurt us? Then maybe consider the same with 100 coordinated users and go from there. The password can only be validated after a successful TOTP challenge so we a) know the attacker's identity and b) this probably mostly concerns the account update operation
> - We already don't do 32 Bytes so we need to use an HMAC for TPM integration. What limitations does this pose?
> - [Are there any other factors to consider that make strong points for a lower limit]
>
> From my point of view, there needs to be a stronger reason for disturbing the user than "less is secure enough, go and try again", especially for an artificial limitation. Otherwise we would be doing more harm than we prevent.

---

> Besides - is it really our place to dictate what other people should feel comfortable with? I can get behind enforcing a minimum. I support protecting the server with a maximum. But there's a difference between protecting the server and being intrusive.

---

> It would be a breaking change for every user that does use very complex passwords.

It would be _extremely_ unlikely that this happens to a user but as demonstrated in #4414, `limit * 2` is not a guarantee for smooth sailing. This should be documented as such.

</details>

Since this is seeking to protect the server, I still think that a soft 128 character limit (512 Bytes instead of 128 graphemes) would be easier and sufficient.

---

> This is why I picked 64 - if 10 is already "secure" then 64 surely is. If you want to make an argument, it should be that we should increase our minimums.

I'd probably be on board with that. Most of the poor passwords as per `correct horse battery staple` I know of are 10 chars and rejecting them could push more people towards PWMs or passphrases.

Then again, [the minimum is configurable via policy](https://kanidm.github.io/kanidm/stable/accounts/account_policy.html#password-minimum-length) so I don't see the need to be prescriptive here. This may raise the bar too high for some people and be a hard barrier to entry. A higher default may be the best way to go.

yaleman (approved):

PR #3700, opened 2025-06-27
20250621 application passwords

Firstyear:
# Change summary

- Add some more scim apis for application passwords
- helper function so a test user can use a passkey
- access control support so that a user can list applications for their own account

Relates #41 

Checklist

- [x] This PR contains no AI generated code
- [ ] book chapter included (if relevant)
- [ ] design document included (if relevant)

PR #4454, opened 2026-07-09
20260701 application password cli

Firstyear:
This adds the application password management CLI. There are still a lot of rough edges here but it is enough that it can be used. I think certainly UX improvements will come in future. One major one is that in the server side we need to do some SCIM value resolution to prevent the need for UUID's to be passed about. 

Relates #3700 

Checklist


CEbbinghaus (approved):
Sweet. That looks good. I presume the server code has existed for a while?

Firstyear:
Yeah it has :)

PR #4471, opened 2026-07-18
Show sccache stats

Firstyear:
# Change summary

- Refactor CI to make better use of build stages to prevent wasteful build cycles.
- Remove beta as a test target, no one cares about that noise - nightly will catch any upcoming issues anyway
- Make run release script part of the rust_build profile since the build is already done, so it doesn't have to build anything at all!
- Clean up env vars in some places to prevent rebuilds

Checklist


.github/workflows/rust_build.yml
```diff
@@ -79,50 +38,17 @@ jobs:
...
-            libudev-dev \
-            libssl-dev \
-            libsystemd-dev \
-            ripgrep
+      - name: "Build the workspace"
+        run: cargo build --locked --workspace
+
       - name: "Run the release build test script"
-        env:
-          BUILD_MODE: --release
```
yaleman, line 125: The *point* of running this in `--release` was that it previously found some issues where the difference between `--debug` and `--release` builds were failing...

Firstyear: It's also consuming a huge amount of CI time every PR. I also don't remember a failure because of debug vs release? Like unless there was a debug assertion involved then it will be the same.

yaleman: Yeah there was a few situations where it was caught, but if you're not worried then 🤷🏻

Firstyear:
Also not done yet, I have more improvements.

scripts/test_run_release_server.sh
```diff
@@ -9,32 +9,20 @@ set -e
...
 if [ -z "${BUILD_MODE}" ]; then
-    BUILD_MODE="--release"
+    BUILD_MODE="--profile dev"
 fi
 
+export KANIDM_CONFIG="./server/daemon/insecure_server.toml"
+
+mkdir -p /tmp/kanidm/client_ca
+
 echo "Building release binaries..."
```
yaleman:
```suggestion
echo "Building $BUILD_MODE binaries..."
```

.github/workflows/rust_build.yml
```diff
@@ -79,50 +38,17 @@ jobs:
...
-          sudo apt-get install -y \
-            libpam0g-dev \
-            libudev-dev \
-            libssl-dev \
-            libsystemd-dev \
-            ripgrep
+      - name: "Build the workspace"
+        run: cargo build --locked --workspace
+
       - name: "Run the release build test script"
```
yaleman, line 50:
```suggestion
      - name: "Run the integration build/test script"
```

scripts/test_run_release_server.sh
```diff
@@ -9,32 +9,20 @@ set -e
...
 # shellcheck disable=SC2086
-cargo build --locked $BUILD_MODE --bin kanidm --bin kanidmd --quiet || {
+cargo build --locked $BUILD_MODE || {
     echo "Failed to build release binaries, please check the output above."
```
yaleman, line 23:
```suggestion
    echo "Failed to build $BUILD_MODE binaries, please check the output above."
```

Firstyear:
This also now makes it so we don't cd into server/daemon during run server, which prevents another round of rebuilds that could occur.

Firstyear:
Okay, this resolves the issues with path handling, and most importantly keeps all the run-release tests working as expected but they now run in the context of the build/test cycle to reduce compiler/worker times and improve cache hits.

yaleman (approved):

PR #4478, opened 2026-07-21
Update crypto glue

### LINKED ISSUE #4477: database restore cannot restore a backup from a replication-enabled server (ReplicationKey keyhandle → DB0002MismatchedRestoreVersion)
### Summary

`kanidmd database restore` cannot restore a backup (either `online_backup` output or `database backup` output) that was produced by a **replication-enabled** server. The restore aborts with:

```
ERROR 🚨 [error]: The provided backup data is from an older server version and is unable to be restored.
ERROR 🚨 [error]: Failed to restore database: Err(DB0002MismatchedRestoreVersion)
```

The error message is misleading — it is **not** a version mismatch. The backup and the restoring binary are the exact same build. The real cause is that the `keyhandles.ReplicationKey` (`KeyHandle::X509Key`) entry, which is only present once replication has been configured, fails to deserialize, and the `#[serde(untagged)]` `DbBackup` enum then silently falls back to an older variant that has no `version` field — which is what trips the `DB0002` path.

Single-node backups have an empty `keyhandles` map and restore without any problem, so this only bites in an HA/replicated deployment — i.e. exactly the disaster-recovery case where you most need the restore to work.

This appears to contradict the design guarantee stated by @Firstyear in #3941:

> The current backup json files already include all replication metadata required for a restore. It was designed so that if you have two nodes A, and B, and you restore B, even though you restored it will automatically "catch up" to A.

In practice the restore never completes, so the catch-up never happens.

### Version

`kanidmd 1.10.0` (also reproduced with the released 1.10.0 binary). The backup was produced by the same 1.10.0 binary.

### Steps to reproduce

1. Stand up a replication-enabled deployment (two nodes, mutual-pull) so that a `ReplicationKey` keyhandle is generated.
2. Take a backup of a node — either the scheduled `online_backup` output, or `kanidmd database backup <path>`.
3. On the same version, attempt `kanidmd database restore <path>` into a fresh data directory.

Result: restore aborts with `DB0002MismatchedRestoreVersion` / "from an older server version".

The backup JSON is structurally a valid `V5` (it has `version`, `db_s_uuid`, `db_d_uuid`, `db_ts_max`, `keyhandles`, `repl_meta`, `entries`), e.g.:

```json
{"version":"1.10","db_s_uuid":"…","db_d_uuid":"…","db_ts_max":{…},
 "keyhandles":{"ReplicationKey":{"X509Key":{"private":"MIGH…","x509":"MIIB…"}}},
 "repl_meta":{"V1":{"ruv":[…]}},
 "entries":[…]}
```

### Isolation (what actually fails)

Starting from a single-node backup that restores cleanly and mutating one field at a time:

- Inject the replicated backup's `keyhandles` object → restore fails with `DB0002`.
- Inject the replicated backup's `repl_meta` (RUV) instead → restore **succeeds**.
- Set `keyhandles = {}` on the full replicated backup → the entire backup (all entries + RUV) **restores cleanly**, `database verify` passes, and the server boots and serves.
- Re-encoding the `private` / `x509` base64 between url-safe / standard / padded variants does **not** fix it — so this is not a base64-alphabet mismatch; the `KeyHandle::X509Key` value itself does not round-trip through serialize → deserialize.

So the failing field is unambiguously `keyhandles` (the `KeyHandle::X509Key` / `ReplicationKey` entry), not `repl_meta` or `entries`.

### Root cause (two parts)

**1. `KeyHandle::X509Key` does not round-trip.** In `server/lib/src/be/keystorage.rs`:

```rust
pub enum KeyHandle {
    X509Key {
        #[serde(with = "pkeyb64")]
        private: SecretDocument,
        #[serde(with = "x509b64")]
        x509: Certificate,
    },
}
```

The value written into the backup by the same binary cannot be read back by that binary's restore path.

**2. The `#[serde(untagged)]` `DbBackup` enum masks the real error and misreports it as a version mismatch.** In `server/lib/src/be/dbentry.rs`, `DbBackup` is `#[serde(untagged)]` with `V5 { version, …, keyhandles, repl_meta, entries }` first, then `V4 { …, keyhandles, repl_meta, entries }` (identical to V5 minus `version`), then `V3` (no `repl_meta`), etc. None of the older variants use `deny_unknown_fields`, so when V5 fails to deserialize on `keyhandles`, serde silently falls through to a variant with no `version`. In `server/lib/src/be/mod.rs::restore`, `maybe_version` is then `None`, which returns:

```rust
} else {
    error!("The provided backup data is from an older server version …");
    return Err(OperationError::DB0002MismatchedRestoreVersion);
};
```

So a `KeyHandle` deserialization failure is reported as an "older server version" — which sent us a long way down the wrong debugging path.

### Impact

An HA/replicated deployment cannot restore its own backups with the documented `database restore` procedure. This defeats disaster recovery for exactly the topology that most needs it.

### Related

- #3941 (states the design intent that replicated backups are restorable + auto-catch-up — which this bug contradicts).

Firstyear:
# Change summary

- Update crypto-glue to resolve a deserialisation error. 

Fixes #4477 

Checklist

- [ ] This PR contains no AI generated content (code, docs, etc)


CEbbinghaus (dismissed):
Lgtm

CEbbinghaus (approved):
Lgtm

Issue #4467, opened 2026-07-16
renew-replication-certificate times out before new certificate is generated


fischer-felix:
### I did this
As part of our ansible bootstrapping process for kanidm, we first run `kanidmd renew-replication-certificate` on each node, then get the certificate using `kanidmd show-replication-certificate` to automatically distribute them to the partner nodes (ansible managed).

### I expected the following
The `renew-replication-certificate` command should run and exit after the certificate has been renewed (or fail if that is somehow impossible). Afterwards `show-replication-certificate` should also run and show the cert.

### This happened instead
`renew-replication-certificate` exits with `INFO     ｉ [info]: Error - you should inspect the logs.`, immediately running `show-replication-certificate` yields the same result.
Upon checking the logs of kanidmd with log level set to DEBUG it seems the client times out before the new certificate is generated, as some time after the renewal finishes independently. Of course if you try to show the certificate in this time frame it will not work either.

Relevant part of the logs: 
```
Jul 16 23:32:32 kanidm-test-1 kanidmd[39777]: 00000000-0000-0000-0000-000000000000 INFO     ｉ [info]: Allowing admin socket access | pid: Some(39887)
Jul 16 23:32:32 kanidm-test-1 kanidmd[39777]: 00000000-0000-0000-0000-000000000000 DEBUG    🐛 [debug]: Accepted admin socket connection
Jul 16 23:32:32 kanidm-test-1 kanidmd[39777]: c6a5e3b9-1ef8-4837-a893-675ae225cbb6 DEBUG    supplier_accept_loop [ 3.15ms | 1.57% / 100.00% ]
Jul 16 23:32:32 kanidm-test-1 kanidmd[39777]: c6a5e3b9-1ef8-4837-a893-675ae225cbb6 DEBUG    ┝━ 🐛 [debug]: renewing replication certificate ...
Jul 16 23:32:32 kanidm-test-1 kanidmd[39777]: c6a5e3b9-1ef8-4837-a893-675ae225cbb6 DEBUG    ┝━ proxy_write [ 274µs | 8.44% / 8.69% ]
Jul 16 23:32:32 kanidm-test-1 kanidmd[39777]: c6a5e3b9-1ef8-4837-a893-675ae225cbb6 DEBUG    │  ┕━ write_acquire_ticket [ 7.81µs | 0.25% ]
[...]

PR #4481, opened 2026-07-22
20260721 actor framework

Firstyear:
# Change summary

Throughout Kanidm and our related projects we use tokio/async tasks heavily. This has led to a lot of duplicated code such as signal handling, task management, shutdown coordination and more. 

And in many cases those features are sometimes spotty. For example our current connection handling code doesn't actually block and wait for client operations to complete cleanly. Our replication task manager has issues with reloads of child tasks. There are lots of inconsistencies as we have been using the raw tokio primitives and missed out on things.

To resolve this I have made an erlang-esque actor library. It is not as full fledged wrt things like process restarting, but it does allow supervision, task cancelation, supervisor hosted tasks to be cleanly stopped, and more. 

The goal is to introduce this first as a library, then to convert over unix_integration, unixd tasks, and kanidm to use this. Rather than do it all at once, I've just added the library first a long with a number of tests to show how this works. 

Certainly this won't be perfect and I'll find edge cases I'm sure, but this makes it easier to review 😅

I did look for existing libraries in the space, but nothing fit what we needed, handled signals, or had other issues. So NIH YOLO FAFO lets go. 

Relates #4467

Checklist


CEbbinghaus (approved):
Alright. Ill bite. No idea what the reasoning is behind building this actor framework ourselves by i'd hazard a guess that there is a pretty important reason.

Looks fine albeit incomplete and I cannot tell where this would slot into kani as a whole.

PR #4500, opened 2026-08-03
Container build cleanup part uno

yaleman:
# Change summary

- start building the tools container automagically so one doesn't have to remember to keep publishing dev builds on dockerhub
- create a reusable container build workflow that we can use to drop all the repeat code in the other container build workflows
- moves the fido-mds tool install into its own builder step so it stops polluting the build cache in general

Checklist

### DIFF
--- a/tools/Dockerfile
+++ b/tools/Dockerfile
@@ -18,18 +18,18 @@ ENV RUSTFLAGS="-Clinker=clang"
 RUN \
     --mount=type=cache,id=zypp,target=/var/cache/zypp \
     zypper install -y --no-recommends \
-    sccache \
-    cargo \
-    cargo-auditable \
-    clang \
-    gcc \
-    make \
-    automake \
-    autoconf \
-    pam-devel \
-    libudev-devel \
-    sqlite3-devel \
-    rsync
+        sccache \
+        cargo \
+        cargo-auditable \
+        clang \
+        gcc \
+        make \
+        automake \
+        autoconf \
+        pam-devel \
+        libudev-devel \
+        sqlite3-devel \
+        rsync
 
 
 FROM build_base AS fido_mds_builder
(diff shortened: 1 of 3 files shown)


tools/Dockerfile
```diff
@@ -18,18 +18,37 @@ ENV RUSTFLAGS="-Clinker=clang"
...
-        clang \
-        gcc \
-        make \
-        automake \
-        autoconf \
-        pam-devel \
-        libudev-devel \
-        sqlite3-devel \
-        rsync
+    sccache \
```
Firstyear: Weird formatting change?

tools/Dockerfile
```diff
@@ -18,18 +18,37 @@ ENV RUSTFLAGS="-Clinker=clang"
...
+    gcc \
+    make \
+    automake \
+    autoconf \
+    pam-devel \
+    libudev-devel \
+    sqlite3-devel \
+    rsync
+
+FROM build_base AS fido_mds_builder
```
Firstyear, line 35: Why make a second builder step here?

tools/Dockerfile
```diff
@@ -40,30 +59,25 @@ RUN \
...
     export CARGO_HOME=/cargo; \
+    export RUSTC_WRAPPER="$(which sccache)"; \
     export SCCACHE_DIR=/sccache; \
-    export RUSTC_WRAPPER=/usr/bin/sccache; \
     export CC="/usr/bin/clang"; \
     cargo auditable build --locked -p kanidm_tools ${KANIDM_BUILD_OPTIONS} \
-        --target-dir="/usr/src/kanidm/target/" \
-        --features="${KANIDM_FEATURES}" \
-        --release && \
+    --target-dir="/usr/src/kanidm/target/" \
```
Firstyear: Again, weird formatting change ....

Firstyear (commented):
What does this change do? I'm not sure I get what it's trying to achieve? 

EDIT: Apparently I can't read the PR description 🤣

tools/Dockerfile
yaleman: because it means the docker cache doesn't get polluted by the install of a thing being pulled from another repo, reducing build time on multiple runs

Firstyear: Ohhhhh I see it now. Maybe we should have that as a comment in the dockerfile then?

tools/Dockerfile
yaleman: it's just the autoformatter fighting me as usual

tools/Dockerfile
yaleman: done

Firstyear (approved):
I fixed the formatting for you.

PR #4510, opened 2026-08-06
20260724 user sign up

Firstyear:
# Change summary

- This adds the foundations for user self-signup. The feature is NOT exposed currently, it's only used in tests.
- Trying to do this in smaller steps to avoid the classique huge PR's. 

Relates #2287 

Checklist


janhaa:
Very interesting! Do you need help with the UI / views / etc.?

Firstyear:
@janhaa Sure, that'd be great. Once I have the underlying parts in place and the apis exposed into the server for consumption I can ping you?

server/lib/src/idm/account_signup.rs
```diff
@@ -0,0 +1,152 @@
+use crate::idm::server::IdmServerProxyWriteTransaction;
+use crate::prelude::*;
+// use kanidm_proto::scim_v1::client::
```
CEbbinghaus: 🎗️ unused

book/src/developers/designs/account_signup_2026.md
```diff
@@ -0,0 +1,74 @@
...
+
+Kanidm is going to be used in applications in future where public users need to be able to self-signup
+without the intervention of an admin. We have already had multiple requests for this feature, and there
+have been multiple community-indepedent implementations of this.
+
+This now makes accounts signup an internal component of Kanidm.
+
+## Considerations
+
+Account signup must be opt-in for deployments.
```
CEbbinghaus, line 11: Honestly, I'd love to see this extended so its not just enabled / disabled but also partially enabled, which is to say that an administrator can generate a signup url / key (similar to how we do the credential reset right now) and that url/key can be used to self serve sign up. So without a key signup isn't possible.

This allows for small private instances to have the administrator send a invite allowing a friend or family member to sign up without needing to do the data entry themselves.

book/src/developers/designs/account_signup_2026.md
```diff
@@ -0,0 +1,74 @@
...
+
+If the account signup feature flag is enabled, the signup form will require a user to enter the
+minimal details for a new person which includes:
+
+* username
+* displayname
+* email address
+
+While email address is not a mandatory attribute in the schema today, it is required for sending
+the initial credential reset request to the user account.
```
CEbbinghaus, line 35: This poses a problem for instances hosting their own mail servers utilizing the email attribute as the definition of the user's mail account.

If this account were to be set to a gmail address for example, the mail server would be trying to handle all outbound gmail destined emails, thus all email addresses must match the SPN of the user.

This may preclude instances hosting mail servers from using the self signup feature entirely as I cannot see an easy way to fix this. But I wanted to have mentioned it

CEbbinghaus: Perhaps this could be solved by only using the email for the credential reset and having an option to disable writing of the email to the DB?

server/lib/src/migration_data/dl_1_12/system_config.rs
```diff
@@ -65,6 +65,20 @@ pub fn e_hmac_name_history_feature() -> EntryInitNew {
     ])
 }
 
+// Waiting on DL16
+pub fn e_account_signup_feature() -> EntryInitNew {
+    entry_init_fn([
+        (Attribute::Class, EntryClass::Object.to_value()),
+        (Attribute::Class, EntryClass::Feature.to_value()),
```
CEbbinghaus, line 72: haven't seen "Feature" EntryClass before :3

server/lib/src/server/mod.rs
```diff
@@ -148,9 +148,15 @@ pub struct HmacNameHistoryConfig {
     pub(crate) key: HmacSha256Key,
 }
 
+#[derive(Clone, Default)]
+pub struct AccountSignupConfig {
+    pub(crate) enabled: bool,
```
CEbbinghaus, line 153: Maybe an enum of:
* Enabled
* Limited
* Disabled

With Limited being only with code? Otherwise two bools could do the trick.

CEbbinghaus (dismissed):
Fine from a code perspective

server/lib/src/server/mod.rs
Firstyear: It comes directly from a bool, so that can't be done. Also what would "limited" even mean?

server/lib/src/migration_data/dl_1_12/system_config.rs
Firstyear: Yeah, it's to try to cut down on bool flags littered through out the place,

book/src/developers/designs/account_signup_2026.md
Firstyear: @CEbbinghaus We already kind of have this, the admin makes the account, then can trigger an account credential reset link to be sent.

book/src/developers/designs/account_signup_2026.md
Firstyear: I see what you're getting at, but we also want on public instances a way for the user to recover their account post-signup so storing the mail seems pretty important?

CEbbinghaus: I wish that stalwart would be a little more accommodating re mail attribute contents, or maybe kani allow custom attributes. But for now it does end in a bit of a stalemate. Public instances would just keep the "disable writing email to db" option disabled.

server/lib/src/server/mod.rs
CEbbinghaus: Limited meaning the feature is enabled (so the page can be loaded) but it requires an extra layer to access, namely a key similar to the credential reset one.

If this is written to the database then perhaps sticking to a bool is the right way to go? Maybe we allow the "authenticated" (where a pre-generated secret allows signup) by default. So the feature is never completely disabled (such as disallowing the form page) but rather just gated behind a secret.

book/src/developers/designs/account_signup_2026.md
CEbbinghaus: Sure, but to create an account you already need the 2/3 parts of information this signup is collecting. It would be nice not to need to sit there trying to tickle a username out of a friend but instead just send them the link to set it themselves

Firstyear: Then just send them the link to the signup form?

Firstyear: Like I get the idea of the admin has to authenticate to send the signup link, that can be a later feature I think.

book/src/developers/designs/account_signup_2026.md
Firstyear: This is intended today to be for like ... proper anonymous signup to a service. What you're talking about is a lot more similar to an enterprise where you have more interconnected parts and some kind of identity / hr system that pushes in accounts to a service and controls all the data.

book/src/developers/designs/account_signup_2026.md
CEbbinghaus: Not saying it can't be added later. Just something I wanted to mention 😊

CEbbinghaus (approved):

PR #4516, opened 2026-08-08
Enforce consent on localhost redirections to protect against malicious clients

### LINKED ISSUE #4514: Force consent on every OAuth2 authorization request for public clients with localhost redirect URL

Currently, consent is only asked on the first OAuth2 authorization request of a public client with a specific set of scopes (unless the consent prompt is disabled). This causes all subsequent authorization requests to succeed without user interaction, as long as the Kanidm session in the browser is still valid.

This could allow an attacker to obtain an authorization token without user interaction if localhost redirect URLs are allowed, by creating a new authorization request using the same (public) client and scopes and its own localhost redirect URL. A requirement for this would be that the attacker can listen on any port on localhost and can open arbitrary URLs in a browser. I think it is quite likely that those are available on typical desktop environments (as they are also requirements for legitimate native OAuth2 clients that do not use their own browser). While users *can* notice this, it usually happens quite fast and only leaves mitigation (like logging out). If an attacker times it right, it could even be done when the user is away and doesn’t notice a browser tab opening for 2 seconds.

Ideally, Kanidm should prompt for interactive consent on every authorization request of a public client with a localhost redirect URL (possibly in a simpler variant than the initial consent, similar to “Grant access to a service running on http​://localhost:12345? \[Grant access\]”). This should be the default. It could be made optional as an opt-out (like other security-related settings) for administrators that can exclude this attack vector.

If I understand correctly, this is also what [Section 8.6 of RFC 8252](https://datatracker.ietf.org/doc/html/rfc8252#section-8.6) requires, as it allows silent authorization requests only when the client’s identity can be proven (typically not the case for public clients with localhost redirect URLs).

RFC 8252 Section 8.6 describes mechanisms for how public clients can prove their identity: Claimed https redirect URLs and platform-specific identity features (like schemes registered by an application with the operating system). Both of which are not directly applicable to usage on Linux desktop systems (at least I do not know of widespread usage of them).

I considered a workaround using a “redirector” endpoint instead of the actual localhost endpoint, which then would display a confirmation button before redirect. However, this would require support from clients, as they might choose different ports with different authorization requests, and seems tacked on instead of a proper secure solution.

I also considered using confidential clients, but that would then require distributing a secret to only somewhat trusted clients. This either means that it increases friction by requiring some sort of authentication for retrieving the client secret (from a secret manager, password manager or similar), or if the secret is available without requiring user interaction, it would likely also be readable by the hypothetical attacker.

While https://github.com/kanidm/kanidm/issues/4274 is related, I think it would not actually fix the issue, as malicious clients could simply not pass `prompt=consent`. It would work if the consent is tied to the value of `prompt`, but is still easily overlooked by users (as clients likely need to be configured specifically). Since Kanidm already enforces (or strongly encourages) the most secure option (like TLS, PKCE, legacy cryptography), I think it should not be that easy to miss a security-relevant setting.

### Additional context

As a practical example, this would currently happen with the configuration of opkssh given in the [Example Configurations section of the documentation](https://kanidm.github.io/kanidm/stable/integrations/oauth2/examples.html#opkssh), which uses a public client. While the official implementation of opkssh leaves the tab open after authorization (making detection by a user easier), a malicious client could auto-close the tab (like [oauth2c](https://github.com/SecureAuthCorp/oauth2c) does) and then exfiltrate the SSH private key it generated using the ID token from Kanidm. If a forced consent on every request does not get implemented, I think this scenario should be noted somewhere in the documentation, as it is not obvious (at least it wasn’t for me).

Firstyear:
# Change summary

Fixes #4514

Checklist

### DIFF
--- a/server/lib/src/idm/oauth2.rs
+++ b/server/lib/src/idm/oauth2.rs
@@ -2509,7 +2509,23 @@ impl IdmServerProxyReadTransaction<'_> {
 
         let session_id = ident.get_session_id();
 
-        if consent_previously_granted || !o2rs.enable_consent_prompt() {
+        let consent_required = (
+                // consent was NOT previously granted, we SHOULD request it.
+                !consent_previously_granted
+                // OR
+                ||
+                // client is NOT basic AND is localhost
+                // - This prevents a malicious localhost client hijacking an existing localhost
+                // - consent that was granted.
+                (!o2rs.is_basic() && loopback_uri_matched)
+                // Future - if we add DCR then we will always force TRUE here.
+            )
+            // AND
+            &&
+            // consent prompt is enabled. This can only be false on confidential clients.
+            o2rs.enable_consent_prompt();
+
+        if !consent_required {
             if event_enabled!(tracing::Level::DEBUG) {
                 let pretty_scopes: Vec<String> =
                     granted_scopes.iter().map(|s| s.to_owned()).collect();


sbruder (approved):
Thank you for the fast fix! This resolves the problem I reported (tested the patch by applying it on top of my instance running 1.11.0).

server/lib/src/idm/oauth2.rs
```diff
@@ -2509,7 +2509,23 @@ impl IdmServerProxyReadTransaction<'_> {
...
+                // OR
+                ||
+                // client is NOT basic AND is localhost
+                // - This prevents a malicious localhost client hijacking an existing localhost
+                // - consent that was granted.
+                (!o2rs.is_basic() && loopback_uri_matched)
+                // Future - if we add DCR then we will always force TRUE here.
+            )
+            // AND
+            &&
```
yaleman, line 2524: and and and and and

ATM machine 😄

yaleman (approved):

server/lib/src/idm/oauth2.rs
Firstyear: It's actually so that the formatting doesn't change to help keep the logic clearer in the comments.

Issue #4460, opened 2026-07-12
kanidm-mail-sender broken


andreas-bulling:
I have been [fighting with this for several days now](https://github.com/kanidm/kanidm/discussions/4450) and I am pretty sure that the problem is not my particular docker setup but either a bug in the lettre library or in how lettre is being using within kanidm-mail-sender.

I am running the latest devel version of the kanidm-tools (1.11.0-dev).

I have tried various variants of specifying the SMTP host - including schema and without. Including a port and without. The error messages change but the result is always the same: No connection can be made and the mail queue is not processed.

I tried for `mail_relay`

1) Just the hostname. Error

`Unable to parse config from '/data/mail-sender' error: Error { message: "relative URL without a base: \"REDACTED\"", input: Some("# The token used to identify and authenticate to Kanidm to read the mail queue. This\n# token *must* have read-write privileges in order to ensure [...]`

That is, it prints the whole config file. This was not the case with the latest stable version.

2) Hostname + port (25 in my case). Error

`2026-07-12T20:14:20.904439Z ERROR kanidm_mail_sender: Unable to build mail transport err=lettre::transport::smtp::Error { kind: Connection, source: "smtp host undefined" }`

3) `smtp://` + hostname + port. Error

```
2026-07-12T20:15:05.635862Z ERROR kanidm_mail_sender: Unable to send message err=lettre::transport::smtp::Error { kind: Client, source: "No compatible authentication mechanism wasfound" }
2026-07-12T20:15:05.635912Z ERROR kanidm_mail_sender: err=Transport
```

4) `smtps://` + hostname + port. Error

```
2026-07-12T20:15:50.398637Z ERROR kanidm_mail_sender: Unable to send message err=lettre::transport::smtp::Error { kind: Connection, source: lettre::transport::smtp::Error { kind: Connection, source: Custom { kind: InvalidData, error: InvalidMessage(InvalidContentType) } } }
2026-07-12T20:15:50.398712Z ERROR kanidm_mail_sender: err=Transport
```

5) Same as 4) but with port 587 or 465. Error

```
2026-07-12T20:17:09.819235Z ERROR kanidm_mail_sender: Unable to send message err=lettre::transport::smtp::Error { kind: Connection, source: Os { code: 101, kind: NetworkUnreachable, message: "Network is unreachable" } }
2026-07-12T20:17:09.819298Z ERROR kanidm_mail_sender: err=Transport
```

yaleman:
[...]

PR #4532, opened 2026-08-16
Avoid logging secrets when mail-sender config parsing fails

booxter:
# Change summary

TOML parser errors include the original input, causing mail-sender to
log tokens and credentials when its configuration is invalid.

Only log the parser message to avoid secret leaks through logs.

See also: https://github.com/kanidm/kanidm/issues/4460

Checklist

### DIFF
--- a/tools/mail_sender/src/main.rs
+++ b/tools/mail_sender/src/main.rs
@@ -390,6 +390,10 @@ fn build_mailer(
     Ok(mailer)
 }
 
+fn parse_config(contents: &str) -> Result<Config, String> {
+    toml::from_str(contents).map_err(|e| e.message().to_owned())
+}
+
 async fn driver_main(opt: Opt) -> Result<(), ()> {
     let mut f = match File::open(&opt.mail_sender_config) {
         Ok(f) => f,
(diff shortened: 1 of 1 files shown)


booxter:
Note: I have not exhaustively covered the codebase, only the case that I personally hit on my setup. There may be more of similar issues in the tree.

Firstyear:
Yeah I think this won't be trivial - perhaps the best option is to just avoid secrets in the config and have them as paths-to-secrets as we have started to do elsewhere in the project.

booxter:
I think systemd EnvironmentFile directive to pass secrets-from-a-file as environment variables on service startup is a possible model to use. Seems to be a common pattern.

Paths to secrets would also do. Probably preferred?

If kanidm just splits secrets into a separate TOML file, it would still need a parser to meld the two, with - potentially - same leakage-through-parser-logging issues.

---

I also think that once the service gains a safer means to inject secrets into its environment, it should nevertheless deal with this current leakage: either by revoking the unsafe mechanism (implies an incompatible change) - or fixing the leakage (this patch).

Firstyear:
@yaleman In response to your deleted comment, that won't work here as this is before parsing completes, not after so SecretString is not applicable.

@booxter Yes, that's the model I'm proposing given how we've done it elsewhere. I'll look into it a bit more soon.

yaleman:
Hence the delete 😝

PR #4543, opened 2026-08-21
Allow setting the provided http protocol versions

### LINKED ISSUE #4542: http/2 causing misdirected requests
- Please review our projects [AI usage policy](https://github.com/kanidm/kanidm/blob/master/book/src/developers/ai_llm_policy.md) and avoid LLM outputs in issues.

### I did this
With kanidm configured through a kgateway TLSRoute, connections can end up misdirected due to http/2.
Hitting reload on a protected page (httpbin behind oauth2-proxy) is sent to the wrong listener and results in a 404.

### I expected the following
SNI to work, routing to kanidm

### This happened instead
404 from the default listener, nothing sent to kanidm

### Kanidm version details

- Output of `kanidm(d) version`: kanidmd 1.11.1
- Are you running it in a container? If so, which image/tag?: kanidm/server:1.11.1@sha256:7c3d7ed868e91f78c24a7fb9c548876563b375a4203021b730d58369b97ad154
- Operating System / Version (On Unix please post the output of `uname -a`): `Linux hz-worker-tiny-8675309 6.8.0-117-generic #117-Ubuntu SMP PREEMPT_DYNAMIC Tue May  5 19:26:24 UTC 2026 x86_64 Linux`

### Any other comments
The root cause looks like a known upstream problem (https://gateway-api.sigs.k8s.io/geps/gep-3567/). The other listener uses a wildcard certificate, and both offer http/2, so the browser is free to reuse connections. (According to the [kgateway docs](https://kgateway.dev/docs/envoy/latest/setup/listeners/tls-settings/), it is possible to globally disable http/2 but I would prefer to avoid that.)

I think an option to disable http/2 on kanidm would solve this, but I wasn't able to find any info on whether that is already possible. (GH search is .. bad.)
Fresh connections work, and force-reloading works. "Try again" generally doesn't change the result until the browser decides to open a new connection.

Configs:
kgateway listener:
```yaml
  listeners:
    - name: tls
      protocol: TLS
      port: 443
      hostname: kidm.example.com
      tls:
        mode: Passthrough
      allowedRoutes:
        namespaces:
          from: Selector
          selector:
            matchExpressions:
              - key: kubernetes.io/metadata.name
                operator: In
                values:
                  - kanidm
```
kanidm tlsroute:
```yaml
spec:
  hostnames:
  - kidm.example.com
  parentRefs:
  - group: gateway.networking.k8s.io
    kind: Gateway
    name: http
    namespace: kgateway-system
    sectionName: tls
  rules:
  - backendRefs:
    - group: ""
      kind: Service
      name: kanidm-main
      namespace: kanidm
      port: 8443
      weight: 1
```

Firstyear:
# Change summary

- Allows setting the provided http protocol versions 

Fixes #4542

Checklist

### DIFF
--- a/examples/server.toml
+++ b/examples/server.toml
@@ -94,6 +94,12 @@ domain = "idm.example.com"
 origin = "https://idm.example.com"
 #
 
+#   Select the HTTP protocol versions offered by the server.
+#   Defaults to "all".
+# http_server_versions = "1"
+#   OR
+# http_server_versions = "2"
+
 #   HTTPS requests can be reverse proxied by a loadbalancer.
 #   To preserve the original IP of the caller, these systems
 #   will often add a header such as "Forwarded" or
--- a/server/core/src/config.rs
+++ b/server/core/src/config.rs
@@ -17,7 +17,6 @@ use std::io::Read;
 use std::net::IpAddr;
 use std::path::{Path, PathBuf};
 use std::str::FromStr;
-use std::sync::Arc;
 use url::Url;
 
 use crate::repl::config::ReplicationConfiguration;
@@ -132,12 +131,12 @@ pub enum LdapAddressInfo {
 }
 
 impl LdapAddressInfo {
-    pub fn trusted_tcp_info(&self) -> Arc<TcpAddressInfo> {
-        Arc::new(match self {
+    pub fn trusted_tcp_info(&self) -> TcpAddressInfo {
+        match self {
             LdapAddressInfo::None => TcpAddressInfo::None,
             LdapAddressInfo::ProxyV2(trusted) => TcpAddressInfo::ProxyV2(trusted.clone()),
             LdapAddressInfo::ProxyV1(trusted) => TcpAddressInfo::ProxyV1(trusted.clone()),
-        })
+        }
     }
 }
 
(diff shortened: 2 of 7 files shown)


examples/server.toml
```diff
@@ -94,6 +94,12 @@ domain = "idm.example.com"
 origin = "https://idm.example.com"
 #
 
+#   Select the HTTP protocol versions offered by the server.
+#   Defaults to "all".
```
yaleman, line 98: defaults to "all" but the listed options only include "1" or "2" - this should include the available options

server/core/src/config.rs
```diff
@@ -577,6 +601,7 @@ impl fmt::Display for Configuration {
             None => write!(f, "arcsize: AUTO, "),
         }?;
         write!(f, "max request size: {}b, ", self.maximum_request)?;
+        write!(f, "http server versions: {}, ", self.http_server_versions)?;
```
yaleman, line 604:
```suggestion
        write!(f, "HTTP server versions: {}, ", self.http_server_versions)?;
```

server/core/src/config.rs
```diff
@@ -279,6 +278,27 @@ impl FromStr for ServerRole {
...
+    #[serde(rename = "1")]
+    V1_1,
+}
+
+impl Display for HttpVersions {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Self::All => f.write_str("http/2 + http/1"),
+            Self::V2_0 => f.write_str("http/2"),
+            Self::V1_1 => f.write_str("http/1"),
```
yaleman, line 297:
```suggestion
            Self::All => f.write_str("HTTP/2 + HTTP/1"),
            Self::V2_0 => f.write_str("HTTP/2"),
            Self::V1_1 => f.write_str("HTTP/1"),
```

yaleman (changes requested):
UX nits

PR #4549, opened 2026-08-25
20260825 make prompt none error more clearly

### LINKED ISSUE #4172: Improvement Web-UI: Redirect on expired session to login screen
### I did this
Having the web UI open in a browser tab and being logged in to the web UI until the login session expires. (Maybe unloading the tab and then reloading it.)

### I expected the following
Being redirected on to the login page, expecting a banner stating that my session expired and I have to log in again.

### This happened instead
I get something like a plain HTML site with the only text on it being `"sessionexpired"`. Reloading does not redirect me to the login page, so that I have to manually remove the path part of the URL to get to the login page.

### Kanidm version details

- Output of `kanidm(d) version`: 1.8
- If not a container, how'd you install it: NixOS

### Any other comments
Just the wish to smoothen user experience.

Firstyear:
# Change summary

- `prompt=none` when set by a client can lead to unintuitive error states. Make it clear that it's the client who is causing this.

Fixes #4172 

Checklist


yaleman:
This isn't spec-compliant per OIDC Core 3.1.2.6 - if the `redirect_uri` is valid, the authorisation server returns the user agent to the client's redirect URI with the appropriate error and state parameters. If it is invalid then we can just throw a "nope something is wrong" error.

Firstyear:
> This isn't spec-compliant per OIDC Core 3.1.2.6 - if the `redirect_uri` is valid, the authorisation server returns the user agent to the client's redirect URI with the appropriate error and state parameters. If it is invalid then we can just throw a "nope something is wrong" error.

Yes, I have just fixed that, I'm getting there :)

Firstyear:
@yaleman It now does the proper redirection for those errors, which should fix the issue with Forgejo.

CEbbinghaus (approved):
Oh damn, Yeah that one is 100% on me not understanding the OAuth2 Spec well enough / the Kani internals to fulfill it 😅

Luckily this seems to fix the last of the problems I introduced so 👍🏻 from me.

Surprising I never encountered the Forgejo problem that caused this PR but glad its getting fixed sooner rather than later

Firstyear:
Don't worry man, I completely missed this one too. Depending on which part of the OAuth2 rfc you read, sometimes it's a 302 sometimes its 400, so it's easy to overlook.

PR #4560, opened 2026-09-02
Respect output_mode (-o flag) in all kanidm (client tools) commands

### LINKED ISSUE #4559: JSON output for all client tool commands (-o json)
### I did this
Run many (not all) kanidm (client tools) commands with the `-o json` option, e.g.
```
kanidm system sync get foo -o json --name admin
---
class: object
class: sync_account
name: foo
sync_token_session: a55ed6b5-c857-4bfe-b12b-4ec9848df856: ApiToken { label: "baz", expiry: None, issued_at: 2026-09-02 15:34:58.591936161 +00:00:00, issued_by: User(00000000-0000-0000-0000-000000000000), scope: Synchronise }
uuid: a845928f-0316-4b48-a9ef-0144a829b9dc
```

### I expected the following
Json output that could be easily used to automate processes.

### This happened instead
Regular text output.

### Kanidm version details

- Output of `kanidm(d) version`: 1.10.4
- Are you running it in a container? If so, which image/tag?: N/A
- If not a container, how'd you install it: AUR package
- Operating System / Version (On Unix please post the output of `uname -a`): `Linux ryzen5 7.2.2-zen1-1-zen #1 ZEN SMP PREEMPT_DYNAMIC Fri, 28 Aug 2026 16:40:27 +0000 x86_64 GNU/Linux`

### Any other comments
We are currently in the process of automating the setup for `kanidm-ldap-sync` amongst other things using ansible. For this, it would be very useful to be able to parse command outputs as json.
I know about the api and that it should be used instead of the cli, however to use it, the cli is also required to generate service-account and bearer token. As everything should be automatd in ansible this leaves us with the same issue.

I have gone ahead and tried implementing the parts I need most myself (see PR #4560), but wanted to also check in here before investing a lot of time.

fischer-felix:
# Change summary

- implement new function `print_vec(input vec<T>)` for `OutputMode` enum to either print json output (same logic as `print_struct()` or print each element by iterating through the vector
- Change the printing logic of shell commands to use the `OutputMode` `print_xxx()` functions instead of `println!()`

Fixes #4559 
As mentioned in my issue, these are some relatively quick-n-dirty fixes for my immediate needs. That's why I have not continued modifying any of the other commands in case you prefer a different approach.

Checklist

### DIFF
--- a/tools/cli/src/cli/serviceaccount.rs
+++ b/tools/cli/src/cli/serviceaccount.rs
@@ -24,7 +24,7 @@ impl ServiceAccountOpt {
                         .await
                     {
                         Ok(cstatus) => {
-                            println!("{cstatus}");
+                            opt.output_mode.print_message(cstatus);
                         }
                         Err(e) => {
                             error!("Error getting credential status -> {:?}", e);
@@ -38,7 +38,7 @@ impl ServiceAccountOpt {
                         .await
                     {
                         Ok(new_pw) => {
-                            println!("Success: {new_pw}");
+                            opt.output_mode.print_message(format!("Success: {new_pw}"));
                         }
                         Err(e) => {
                             error!("Error generating service account credential -> {:?}", e);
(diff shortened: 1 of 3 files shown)


Firstyear:
I'm okay to merge this under the caveat that our cli is *not* an api and so we may change command names/locations in future.

PR #4568, opened 2026-09-11
Fix pam.h constants for FreeBSD/MacOS

Firstyear:
We previously used a single set of pam constants for Linux and the BSDs, however FreeBSD/MacOS use openpam which has a small number of differences. This caused a bug in FreeBSD 16-CURRENT when logging in via the TTY as a local user.

Checklist


unix_integration/pam_sparkle_common/src/pam/constants_openpam.rs
```diff
@@ -0,0 +1,88 @@
+use libc::{c_int, c_uint};
+
+pub type PamFlag = c_uint;
+pub type PamItemType = c_int;
+pub type PamMessageStyle = c_int;
+pub type AlwaysZero = c_int;
+
+// See /usr/include/security/pam_constants.h
+
```
yaleman, line 10: Might be worth linking to https://github.com/freebsd/freebsd-src/blob/stable/15/contrib/openpam/include/security/pam_constants.h or similar so it's more easily findable in future.

Any reason they're const's instead of making PamFlag an enum like the original source? I guess just aligning with the current implementation but it seems like a more Rust-y thing. 
```suggestion
// <https://github.com/freebsd/freebsd-src/blob/stable/15/contrib/openpam/include/security/pam_constants.h>

```

yaleman (dismissed):

Firstyear:
>  I guess just aligning with the current implementation but it seems like a more Rust-y thing.

Yeah, it's to keep it compatible with the existing impl, but I suspect it's because they're flags they need have bitwise ops applied that you can't do with an enum.

yaleman:
> because they're flags they need have bitwise ops applied that you can't do with an enum.

the values would have to be in an as_ref or something, but yeah that's just extra work

yaleman (approved):

PR #4581, opened 2026-09-17
Update account signup requests to expire

Firstyear:
# Change summary

Checklist


Firstyear:
In the spirit of keeping reviews smaller, this just improves the account signup feature to have the entry deleteAfter, an the access controls to allow this feature to be activated.

server/lib/src/server/access/mod.rs
```diff
@@ -684,7 +684,10 @@ pub trait AccessControlsTransaction<'a> {
         let sync_agmts = self.get_sync_agreements();
 
         match apply_modify_access(ident, related_acp, sync_agmts, entry) {
-            ModifyResult::Deny => false,
+            ModifyResult::Deny => {
+                security_error!("modify access denied.");
```
yaleman, line 688: This seems like an event for the sake of it, wouldn't it be better to surface it elsewhere?

server/lib/src/server/access/modify.rs
```diff
@@ -248,14 +248,15 @@ fn modify_ident_test(ident: &Identity) -> AccessBasicResult {
             return AccessBasicResult::Grant;
         }
         IdentType::Internal(InternalRole::Migration) => {
-            return AccessBasicResult::Grant;
+            return AccessBasicResult::Ignore;
```
yaleman, line 251: That's quite a change, from grant to ignore?

yaleman (changes requested):

server/lib/src/server/access/mod.rs
Firstyear: It was needed to work out *what* module was triggering the denial state.

server/lib/src/server/access/modify.rs
Firstyear: Yes, it was over-granting here but then constrained later where the real interaction here should be to ignore this stack in the process.

